PR #213, opened 2014-12-19
better handling of input format change for non-adaptive codecs

martinbonnin:
* this fixes a bug when switching from HE-AAC 22050Hz to AAC 44100Hz (the AudioTrack was not reset and we were trying to send a bad number of bytes, triggering a "AudioTrack.write() called with invalid size" error)
* this also improves quality switches, making it almost seamless


martinbonnin:
Any news about this ?

ojw28:
Sorry for the delay. I think https://github.com/google/ExoPlayer/pull/221 is the one that goes into the correct branch, so I'll comment there. Can we close this one?

martinbonnin:
yep, sure. closing it

PR #220, opened 2014-12-23
log all AudioTrack errors

martinbonnin:


library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
```diff
@@ -405,8 +405,6 @@ public int handleBuffer(ByteBuffer buffer, int offset, int size, long presentati
         bytesToWrite = Math.min(temporaryBufferSize, bytesToWrite);
         bytesWritten = audioTrack.write(temporaryBuffer, temporaryBufferOffset, bytesToWrite);
         if (bytesWritten < 0) {
```
ojw28: I think you meant to swap the sign around here (should be >=).

library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
```diff
@@ -414,10 +412,15 @@ public int handleBuffer(ByteBuffer buffer, int offset, int size, long presentati
       bytesWritten = writeNonBlockingV21(audioTrack, buffer, temporaryBufferSize);
     }
 
-    temporaryBufferSize -= bytesWritten;
-    submittedBytes += bytesWritten;
-    if (temporaryBufferSize == 0) {
-      result |= RESULT_BUFFER_CONSUMED;
+    if (bytesWritten < 0) {
+        Log.w(TAG, "AudioTrack.write returned error code: " + bytesWritten);
```
ojw28: No need to change as part of this request, since it's not a regression, but I wonder about the logic of just printing a warning here. The AudioTrack documentation suggests failures will be permanent (particularly ERROR_INVALID_OPERATION and ERROR_DEAD_OBJECT), in which case just claiming the buffer was consumed and trying to continue isn't going to help.

Did you actually see any instances where you were able to survive one of these errors successfully? If not, I'll follow this change with another one that changes this code to propagate the error and fail playback instead.

library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
martinbonnin: Ooooops... Will change this

library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
martinbonnin: The idea was to at least get one warning. The previous code did not log errors in the 'SDK >= 21' case. So it took me some time to find the root cause. But the failing was permanent indeed, I have never been able to recover such an error so failing playback would be good as well

library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
martinbonnin: see commit 75b5a59f597973d2cae77bbe83c826ee9eb6a344

library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
ojw28: Yeah. I'll merge this and follow up with a change that properly propagates the error and fails playback.

ojw28:
Thanks!

PR #221, opened 2014-12-23
better handling of input format change for non-adaptive codecs

martinbonnin:
* this fixes a bug when switching from HE-AAC 22050Hz to AAC 44100Hz (the AudioTrack was not reset and we were trying to send a bad number of bytes, triggering a "AudioTrack.write() called with invalid size" error)
* this also improves quality switches, making it almost seamless

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
@@ -493,9 +493,10 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
     decodeOnlyPresentationTimestamps.clear();
     // Workaround for framework bugs.
     // See [Internal: b/8347958], [Internal: b/8578467], [Internal: b/8543366].
-    if (Util.SDK_INT >= 18) {
+    if (Util.SDK_INT >= 18 && codecReinitState == REINIT_STATE_NONE) {
       codec.flush();
     } else {
+      codecReinitState = REINIT_STATE_NONE;
       releaseCodec();
       maybeInitCodec();
     }
@@ -504,6 +505,8 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
       // avoid this issue by sending reconfiguration data following every flush.
       codecReconfigurationState = RECONFIGURATION_STATE_WRITE_PENDING;
     }
+
+    hasQueuedOneInputBuffer = false;
   }
 
   /**
--- a/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
@@ -148,11 +148,6 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
    * The end of stream has been sent, wait for the codec to acknowledge it.
    */
   private static final int REINIT_STATE_WAIT_END_OF_STREAM = 2;
-  /**
-   * The last sample has been processed, we can safely reinit now.
-   */
-  private static final int REINIT_STATE_DO_REINIT_NOW = 3;
-
 
   public final CodecCounters codecCounters;
 
@@ -521,12 +516,6 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
       return false;
     }
 
-    if (codecReinitState == REINIT_STATE_DO_REINIT_NOW) {
-      releaseCodec();
-      maybeInitCodec();
-      return false;
-    }
-
     if (inputIndex < 0) {
       inputIndex = codec.dequeueInputBuffer(0);
       if (inputIndex < 0) {
@@ -691,7 +680,8 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
     } else {
       if (!hasQueuedOneInputBuffer) {
         // no need to signal end of stream if nothing has been queued, we can just reinit asap
-        codecReinitState = REINIT_STATE_DO_REINIT_NOW;
+        releaseCodec();
+        maybeInitCodec();
       } else {
         codecReinitState = REINIT_STATE_SIGNAL_END_OF_STREAM;
       }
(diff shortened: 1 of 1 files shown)


ojw28:
This looks pretty good; thanks! I think you're missing some handling in the flushCodec method though. It looks like you probably need to do the following in that method:

- Set hasQueuedOneInputBuffer to false
- If the state is something other than REINIT_STATE_NONE, release and re-init the decoder
- Set the state to REINIT_STATE_NONE

What do you think?

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
```diff
@@ -135,6 +135,25 @@ private static String getDiagnosticInfoV21(Throwable cause) {
...
+   */
+  private static final int REINIT_STATE_SIGNAL_END_OF_STREAM = 1;
+  /**
+   * The end of stream has been sent, wait for the codec to acknowledge it.
+   */
+  private static final int REINIT_STATE_WAIT_END_OF_STREAM = 2;
+  /**
+   * The last sample has been processed, we can safely reinit now.
+   */
+  private static final int REINIT_STATE_DO_REINIT_NOW = 3;
```
ojw28: I don't really understand  the need for this state. Why don't you just do the reinit directly at the points where you set the state to this value, and set back to REINIT_STATE_NONE?

martinbonnin: not sure why. I liked the idea of doing most of the work from feedInputBuffer() as I was not sure what other code paths could lead to the other places. It's certainly not needed though.

martinbonnin:
You're right about the flushCodec() thing, just made a new commit. It's quite difficult to reach this state though so I haven't been able to verify the code is actually doing what it should but it should be safe.

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
ojw28: It seems cleaner to do it immediately, rather than to have an additional state. So I'd suggest the other way.

ojw28: This aside, everything looks good. Thanks!

martinbonnin: additional state removed.

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
```diff
@@ -502,6 +525,16 @@ private boolean feedInputBuffer(boolean firstFeed) throws IOException, ExoPlayba
       sampleHolder.data.clear();
     }
 
+    if (codecReinitState == REINIT_STATE_SIGNAL_END_OF_STREAM) {
+      codec.queueInputBuffer(inputIndex, 0, 0, 0, MediaCodec.BUFFER_FLAG_END_OF_STREAM);
+      inputIndex = -1;
+      codecReinitState = REINIT_STATE_WAIT_END_OF_STREAM;
+      return false;
+    } else if (codecReinitState != REINIT_STATE_NONE) {
```
ojw28: Small thing: IMO this is clearer:

```
codecReinitState == REINIT_STATE_WAIT_END_OF_STREAM
```

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
```diff
@@ -469,9 +488,10 @@ private void flushCodec() throws ExoPlaybackException {
     decodeOnlyPresentationTimestamps.clear();
     // Workaround for framework bugs.
     // See [Internal: b/8347958], [Internal: b/8578467], [Internal: b/8543366].
-    if (Util.SDK_INT >= 18) {
+    if (Util.SDK_INT >= 18 && codecReinitState == REINIT_STATE_NONE) {
```
ojw28: Taking a final pass at this, I'm not sure I follow the logic here. Shouldn't this be left as it was, and then codecReinitState should be set to REINIT_STATE_NONE at the bottom of this method, where you set hasQueuedOneInputBuffer to false?

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
```diff
@@ -166,6 +181,7 @@ private static String getDiagnosticInfoV21(Throwable cause) {
   private boolean outputStreamEnded;
   private boolean waitingForKeys;
   private boolean waitingForFirstSyncFrame;
+  private boolean hasQueuedOneInputBuffer;
```
ojw28: Small thing: Can we call it hasQueuedInputBuffer? The current name kind of suggests it might be false if we've queued more than one.

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
martinbonnin: no because just calling flush() is not enough if we have a reinit pending due to an input format change, we really need to release and reinit the codec.

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
martinbonnin: done

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
ojw28: Ah, yes :). Thanks.

library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
martinbonnin: yup, done

martinbonnin:
any news on this one ? I fear we wont be able to merge it if we wait too long.

ojw28:
I still don't see a signed CLA on my side.

martinbonnin:
gasp. Ok, I will double check

martinbonnin:
The CLA should be ok now :)

ojw28:
Yes, it is. Thanks. I'll be sure to merge this soon.

ojw28:
And. Finally. It is done! Thanks for your patience.

martinbonnin:
Great ! Thanks :)

PR #241, opened 2015-01-18
Fix infinite loop in the case where an output buffer of size 0 is return...

jafletch:
There exists a case where dequeueOutputBuffer will return an outputIndex >= 0 with an outputBufferInfo.size == 0. When this happens, processOutputBuffer will return false so the method will exist without reseting outputIndex to -1. This effectively causes an infinite loop because subsequent calls to drainOutputBuffer will never attempt to dequeueOutputBuffer so the stream will not progress.

Not sure if my fix is the correct one but it fixes the problem for me.

Repros for me attempting to stream:
http://feeds.thisamericanlife.org/~r/talpodcast/~5/3LflCztSswE/544.mp3

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
@@ -737,6 +737,11 @@ public abstract class MediaCodecTrackRenderer extends TrackRenderer {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


ojw28:
I'm pretty sure this is already fixed on the "dev" branch. See: https://github.com/google/ExoPlayer/pull/208.

I'm going to close this based on the above, but if you can still reproduce the issue on the dev branch then please comment/re-open.

Thanks!

Issue #246, opened 2015-01-20
FrameworkSampleSource with FileDescriptor


Nezz:
Currently FrameworkSampleSource is designed to work with URIs. This makes it rather inconvenient to play media from other sources, like the APK (Assets for example). It would be a [one line change](https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/FrameworkSampleSource.java#L68) to make it work with with a FileDescriptor, since MediaExtractor has a [setDataSource overload](http://developer.android.com/reference/android/media/MediaExtractor.html#setDataSource%28java.io.FileDescriptor,%20long,%20long%29) for that.

My questions are:
Is there a way to play files from the Assets folder?
Are there any plans to add support for it?
Would it make sense to add a new SampleSource that is the same as FrameworkSampleSource just with that single line changed?

spark85:
I changed the implementation with a custom SampleSource to use a FileDescriptor.  I obtained the FileDescriptor by creating an InputStream wrapped around a File object with the correct file name.  That fixed the issue with the MediaExtractor; and when I examined the MediaFormat it had all of the correct attributes.  (* I will post them by tomorrow morning). 

However, that still didn't solve the problem.  The player crashes when it tries to play the file, with a Signal 6 error; and FILE_DATA_SOURCE_END. 

I could use the asset folder; however, for security reasons we prefer to keep the media in internal storage.

ojw28:
We can add an additional FrameworkSampleSource constructor that takes a FileDescriptor, if that helps. I'd like to understand the "still doesn't solve the problem" part though.

Nezz:
Me too. I would have done it already, but I was too lazy to set up the environment to build it :)

Nezz:
Works flawlessly for me.

ojw28:
Closing this as the change is merged. It's likely the "still doesn't solve the problem" issue is something else. Feel free to open a separate issue to track that, if needed.

PR #250, opened 2015-01-21
FrameworkSampleSource with FileDescriptor

Nezz:
As discussed in #246

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/FrameworkSampleSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/FrameworkSampleSource.java
@@ -23,6 +23,7 @@ import android.content.Context;
 import android.media.MediaExtractor;
 import android.net.Uri;
 
+import java.io.FileDescriptor;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


Nezz:
Should I submit this to master or dev?

ojw28:
It'll need submitting into dev. You'll also need to sign the appropriate CLA before we can merge the change. See:

Thanks!

CCorrado:
Was this PR not included in r.2.4.1 ? Looking for the proper way to use FileDescriptor (instead of a URI) with ExoPlayer, this is a perfect solution. Not seeing "FrameworkSampleSource" or another MediaSource implementation that takes in a FD instead of URI.

ojw28:
`FrameworkSampleSource` is not part of ExoPlayer V2, since we removed all our dependencies on Android's `MediaExtractor` API. It's probably best if you file a new issue/question explaining what you're actually trying to do (in terms of what the use case is, rather than "I want to use a FileDescriptor"). There may well be an easy alternative or better way of doing whatever you're trying to do.

PR #256, opened 2015-01-22
Upgrade ExoPlayer with controllable speed

mikumi:
Goal of this pull request

- Update to Google's most recent ExoPlayer version
- Port previous speed change modifications to new version

There have been some modifications compared to Hooloop's earlier modifcation:
- Now support for x86 as well (e.g. Genymotion simulator)
- This one is using a different approach than previously. Here we preserve the partial buffer writing to audio track so we don't lose performance. Instead, the speed is changed directly before initializing the temporary buffer. Additionally, the channel configuration for libsonic has also been fixed.
This approach works with all audio file types including vorbis.

Other notes:
- I am not sure if it is still necessary to set ```AudioTrack``` type to ```AudioManager.STREAM_VOICE_CALL``` instead of ```AudioManager.STREAM_MUSIC```. This has been directly ported from Hooloop and didn't hurt, so I just left it like that 90f86c6
- The same goes for changing the ```DEFAULT_MIN_BUFFER_MULTIPLICATION_FACTOR```. This has also been directly ported from Hooloop and didn't hurt, so I just left it like that dd28f61
- Also I disabled the previous codec fix for Xiaomi as it forced the AAC codec. We need to see if this problem still applies and fix it properly if necessary. 77fd609
- In case there are any problems with the new ```MediaCodecAudioTrackRenderer``` implementation, I added the previous one as an alternative renderer as ```MediaCodecHooloopAudioTrackRenderer``` adc6aa5
- The newest ExoPlayer implements a feature called non-blocking write to AudioTrack. This will only work with Android 5.0+. I have disabled that for our renderer as it requires a different approach for libsonic. It won't be a big change, but it is not required now. Just keep it in mind, I have added todo+notes in code for that.


mikumi:
Shit, pull request on wrong repository. I am sorry :-D

PR #260, opened 2015-01-23
Fix a signed right shift issue in BitArray.readUnsignedByte()

mine260309:
To reproduce the issue, you can use the data: `0x00 0x01 0xe1 0x90` to create the BitArray, then:
```java
buffer.skipBits(19);
int value = buffer.readBits(13);
```
The expected value is `400 (0x0190)`, but the returned value is `8080 (0x1F90)`

The root cause is in `BitArray.readUnsignedByte()`, it uses signed right shift, so the high bits are all 1 if the byte is a negative value.
The fix is simply to make it unsigned right shift.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/util/BitArray.java
+++ b/library/src/main/java/com/google/android/exoplayer/util/BitArray.java
@@ -151,7 +151,7 @@ public final class BitArray {
     byte b;
     if (bitOffset != 0) {
       b = (byte) ((data[byteOffset] << bitOffset)
-          | (data[byteOffset + 1] >> (8 - bitOffset)));
+          | ((data[byteOffset + 1] & 0xFF) >> (8 - bitOffset)));
     } else {
       b = data[byteOffset];
     }
--- a/library/src/main/java/com/google/android/exoplayer/util/BitArray.java
+++ b/library/src/main/java/com/google/android/exoplayer/util/BitArray.java
@@ -148,16 +148,15 @@ public final class BitArray {
    * @return The value of the parsed byte.
    */
   public int readUnsignedByte() {
-    byte b;
+    int value;
     if (bitOffset != 0) {
-      b = (byte) ((data[byteOffset] << bitOffset)
-          | ((data[byteOffset + 1] & 0xFF) >> (8 - bitOffset)));
+      value = (data[byteOffset] << bitOffset)
+          | ((data[byteOffset + 1] & 0xFF) >>> (8 - bitOffset));
     } else {
-      b = data[byteOffset];
+      value = data[byteOffset];
     }
     byteOffset++;
-    // Converting a signed byte into unsigned.
-    return b & 0xFF;
+    return value & 0xFF;
   }
 
   /**


ojw28:
Thanks for this! You'll need to sign the appropriate CLA before we can merge the change. See:

mine260309:
CLA signed. Thanks.

library/src/main/java/com/google/android/exoplayer/util/BitArray.java
```diff
@@ -151,7 +151,7 @@ public int readUnsignedByte() {
     byte b;
```
ojw28: By the way, the problem isn't that the shift is signed. It's that data[byteOffset+1] is cast from byte to int before the shift happens. If the byte is negative, the cast does more than just adding zeros to the front.

Your fix is correct, but I feel the method could be written more clearly (and without an unnecessary int->byte->int conversion), like:

```
  public int readUnsignedByte() {
    int value;
    if (bitOffset != 0) {
      value = ((data[byteOffset] & 0xFF) << bitOffset)
          | ((data[byteOffset + 1] & 0xFF) >>> (8 - bitOffset));
    } else {
      value = data[byteOffset] & 0xFF;
    }
    byteOffset++;
    return value;
  }
```

Thoughts? Happy for you to amend your change to do this, or I can submit it separately.

mine260309: Yes, it looks a bit more clear.
I'll test it and update the pull request

mine260309: Your code's return value is not correct, it actually needs to `return value & 0xFF`

ojw28:
Oh yeah, you're right. It should be:

```
public int readUnsignedByte() {
    int value;
    if (bitOffset != 0) {
      value = ((data[byteOffset] & 0xFF) << bitOffset)
          | ((data[byteOffset + 1] & 0xFF) >>> (8 - bitOffset));
    } else {
      value = data[byteOffset];
    }
    byteOffset++;
    return value & 0xFF;
  }
```

By the way - We'll be pushing some tests to Github soon. So when that happens, we should make sure this class is unit tested.

library/src/main/java/com/google/android/exoplayer/util/BitArray.java
ojw28: Oh yeah, you're right. It should be:

```
public int readUnsignedByte() {
    int value;
    if (bitOffset != 0) {
      value = ((data[byteOffset] & 0xFF) << bitOffset)
          | ((data[byteOffset + 1] & 0xFF) >>> (8 - bitOffset));
    } else {
      value = data[byteOffset];
    }
    byteOffset++;
    return value & 0xFF;
  }
```

By the way - We'll be pushing some tests to Github soon. So when that happens, we should make sure this class is unit tested.

mine260309:
Updated the pull request, note some of `& 0xFF` in your proposed code is unnecessary, so they're not included in the patch.

mine260309:
Good to know it will be tested :)
The pull request is now updated.

ojw28:
Thanks!

ojw28:
By the way - Did you see any bad playback behavior as a result of this bug, or did you just spot it by reading code? Just curious.

mine260309:
It's noticed that some stream's pmt pid becomes incorrect, that's why my test case is skipping 19bits and reading 13bits, which is to read pmt pid :)

PR #304, opened 2015-02-18
Fixing issue with getSampleQueue

jeoliva:
As far as I see, doesn’t make sense that track parameter received by
getSampleQueue method is always set to 0. It could vary depending on
the number of tracks defined in each TS segment and typically has at
least 2 (one for video, one for audio).

This is breaking HLS playback for any stream containing more than 1
track so I am commenting out this method (this line should be removed).


ojw28:
Thanks. You're right, however I beat you to it and merged a fix a few minutes ago :)!

jeoliva:
Ouuugh!! You are really fast! :)

PR #305, opened 2015-02-18
Codecs parsing issue

jeoliva:
Previous regular expression for extracting codec information was wrong. Given a line that defines a variant it added information from “CODEC=“
text to the end of the line (including also information about
RESOLUTION or alternate rendition groups as part of the CODEC field).

So, previous regular expression is parsing CODEC field well just when CODEC is the last parameter of a variant definition line.

This is not causing a functional problem (at least known by me)
although is making codecs field storing information that is not related
with the codec.


ojw28:
Thanks! This should merge into dev, not master. Note that you also need to submit a CLA. See here for details: https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md

PR #307, opened 2015-02-19
Issue with regular expression for CODEC attribute

jeoliva:
Previous regular expression for extracting codec information was wrong,
given a line that defines a variant it added information from “CODEC=“
text to the end of the line (including also information about
RESOLUTION or alternate rendition groups as part of the CODEC field).

This is not causing a functional problem (at least known by me)
although is making codecs field storing information that is not related
with the codec.


ojw28:
Thanks! I'll merge this as soon as you complete a CLA (see https://github.com/google/ExoPlayer/blob/dev/CONTRIBUTING.md). If you've done so already then let me know, since I'm not seeing it on my side. Thanks!

jeoliva:
Done! I did that yesterday. Sorry for not saying that before.

Many thanks and congratulations for the great work you are doing with Exoplayer

ojw28:
Ah, I see it now. Thanks!

PR #314, opened 2015-02-21
Update build.gradle

pavelmartovskiyfilamon:
Add uploadArchives task for publish to maven repository


ojw28:
Ref'd issue is now fixed, so we don't need this. Thanks though!

PR #316, opened 2015-02-23
Better management of blacklisted playlists

jeoliva:
Added an expiration time field to playlists blacklisted to allow
Exoplayer to continue playback when playlists that failed were
recovered from a bad state.

In live environments, some times occur that primary encoder stop
working for a while. In that cases, HLS failover mechanism in the
player should detect the situation and “switch” to playlists served by
the backup encoder (in case a backup encoder exists). This was well
managed before these changes.

However, and to ensure a playback experience that can recover itself
from temporary issues, we cannot blacklist a playlist forever. When
streaming live events using HLS, it is quite typical that the player
needs to switch from primary to backup playlists, and from backup to
primary ones, from time to time to have playback working when temporary
issues in the network/encoder are happening. Most of the issues are
recoverable, so what I have implemented is a mechanism that makes
blacklisted playlists to be available again after a while (60 seconds).

Evaluation of this algorithm should happen just when something fails.
If player is working with a backup playlist, it shouldn’t switch to the
primary one at least something fail.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
@@ -258,7 +258,6 @@ public class HlsChunkSource {
     if (adaptiveMode == ADAPTIVE_MODE_NONE) {
       // Do nothing.
     } else {
-      clearStaleBlacklistedPlaylists();
       nextVariantIndex = getNextVariantIndex(previousTsChunk, playbackPositionUs);
       switchingVariant = nextVariantIndex != variantIndex;
       switchingVariantSpliced = switchingVariant && adaptiveMode == ADAPTIVE_MODE_SPLICE;
@@ -392,6 +391,7 @@ public class HlsChunkSource {
   }
 
   private int getNextVariantIndex(TsChunk previousTsChunk, long playbackPositionUs) {
+    clearStaleBlacklistedPlaylists();
     int idealVariantIndex = getVariantIndexForBandwdith(
         (int) (bandwidthMeter.getBitrateEstimate() * BANDWIDTH_FRACTION));
     if (idealVariantIndex == variantIndex) {
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
@@ -372,7 +372,7 @@ public class HlsChunkSource {
         MediaPlaylistChunk playlistChunk = (MediaPlaylistChunk) chunk;
         mediaPlaylistBlacklistFlags[playlistChunk.variantIndex] = true;
         mediaPlaylistBlacklistedTimeMs[playlistChunk.variantIndex] = SystemClock.elapsedRealtime();
-        evaluatePlaylistBlacklistedTimestamps();
+        clearStaleBlacklistedPlaylists();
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
```diff
@@ -541,6 +553,17 @@ private boolean allPlaylistsBlacklisted() {
     return true;
   }
 
+  private void evaluatePlaylistBlacklistedTimestamps()
```
ojw28: Maybe this should be called clearStaleBlacklistedPlaylists, or something?

library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
```diff
@@ -541,6 +553,17 @@ private boolean allPlaylistsBlacklisted() {
...
+  {
```
ojw28: nit - Please put { on previous line, to be consistent with style elsewhere.

library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
```diff
@@ -541,6 +553,17 @@ private boolean allPlaylistsBlacklisted() {
...
+    long currentTime = SystemClock.elapsedRealtime();
+    for (int i = 0; i < mediaPlaylistBlacklistFlags.length; i++) {
+      if (mediaPlaylistBlacklistFlags[i] && currentTime - mediaPlaylistBlacklistedTimeMs[i] > DEFAULT_MAX_TIME_MEDIA_PLAYLIST_BLACKLISTED_MS) {
```
ojw28: nit - Line too long. Please line break before the && to get it back to within 100 chars.

ojw28:
Looks good, thanks!

One question: Are you concerned at all about ending up in a situation where a high quality playlist is blacklisted, and playback continues fine using a lower quality stream, without ever switching back up to the higher quality again? If so it would seem beneficial to periodically check the blacklist even without failures, to ensure switching back up to the higher quality playlist can happen? On the flip side, if the playlist really is permanently broken, that'll just mean we make a request that fails every 60s. So I think it can be argued either way (and it's not clear to me whether one approach is better or worse than the other).

jeoliva:
Yeah, good point. I was more thinking on a failover scenario, but you are right, we should also take into account that case: not just thinking on providing a "player that always tries to continue playback" solution but also providing a solution that ensures player plays the highest available quality and then provides the best user experience. 

From where do you think we should call clearStaleBlacklistedPlaylists? My suggestion is calling it from getChunkOperation method, right before evaluating next variant index.

if (adaptiveMode == ADAPTIVE_MODE_NONE) {
      // Do nothing.
    } else {
      clearStaleBlacklistedPlaylists();
      nextVariantIndex = getNextVariantIndex(previousTsChunk, playbackPositionUs);
      switchingVariant = nextVariantIndex != variantIndex;
      switchingVariantSpliced = switchingVariant && adaptiveMode == ADAPTIVE_MODE_SPLICE;
    }

ojw28:
I think you've found the right point to call it, although I think it's marginally clearer if you make the call right at the top of the getNextVariantIndex method (which is exactly the same point; just inside the method as opposed to outside). How about we put it there?

Otherwise all looks great; thanks.

jeoliva:
Makes sense, done!

PR #324, opened 2015-02-26
Added support for new types of ID3 frames

jeoliva:
- Added support for ID3 frames of types GEOB and PRIV.
 - GEOB type is commonly used by dynamic ads providers to include in the
stream information about the ads to be played.
 - PRIV type is commonly used for time synchronization (example:
synchronizing playback of a live stream and its webvtt captions) and
also by analytics companies to include tracking information in the
stream.

- Added a sample stream from Apple that contains ID3 metadata.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
+++ b/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
@@ -67,6 +67,39 @@ public class Id3Parser implements MetadataParser<Map<String, Object>> {
         String value = new String(frame, valueStartIndex, valueEndIndex - valueStartIndex,
             charset);
         metadata.put(TxxxMetadata.TYPE, new TxxxMetadata(description, value));
+      } else if (frameId0 == 'P' && frameId1 == 'R' && frameId2 == 'I' && frameId3 == 'V') {
+        // Check frame ID == PRIV
+        byte[] frame = new byte[frameSize];
+        id3Data.readBytes(frame, 0, frameSize);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String owner = new String(frame, 0, firstZeroIndex);
+        byte[] privateData = new byte[frameSize - firstZeroIndex - 1];
+        System.arraycopy(frame, firstZeroIndex + 1, privateData, 0, frameSize - firstZeroIndex - 1);
+        metadata.put(PrivMetadata.TYPE, new PrivMetadata(owner, privateData));
+      } else if (frameId0 == 'G' && frameId1 == 'E' && frameId2 == 'O' && frameId3 == 'B') {
+        // Check frame ID == GEOB
+        int encoding = id3Data.readUnsignedByte();
+        String charset = getCharsetName(encoding);
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String mimeType = new String(frame, 0, firstZeroIndex);
+        int filenameStartIndex = firstZeroIndex + 1;
+        int filenameEndIndex = indexOf(frame, filenameStartIndex, (byte) 0);
+        String filename = new String(frame, filenameStartIndex,
+                                     filenameEndIndex - filenameStartIndex, charset);
+        int descriptionStartIndex = filenameEndIndex + 1;
+        int descriptionEndIndex = indexOf(frame, descriptionStartIndex, (byte) 0);
+        String description = new String(frame, descriptionStartIndex,
+                descriptionEndIndex - descriptionStartIndex, charset);
+
+        byte[] objectData = new byte[frameSize - descriptionEndIndex - 2];
+        System.arraycopy(frame, descriptionEndIndex + 1, objectData, 0,
+                         frameSize - descriptionEndIndex - 2);
+        metadata.put(GeobMetadata.TYPE, new GeobMetadata(mimeType, filename,
+                                                         description, objectData));
       } else {
         String type = String.format("%c%c%c%c", frameId0, frameId1, frameId2, frameId3);
         byte[] frame = new byte[frameSize];
--- a/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
+++ b/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
@@ -70,7 +70,7 @@ public class Id3Parser implements MetadataParser<Map<String, Object>> {
         int valueStartIndex = firstZeroIndex + delimiterLength(encoding);
         int valueEndIndex = indexOfEOS(frame, valueStartIndex, encoding);
         String value = new String(frame, valueStartIndex, valueEndIndex - valueStartIndex,
-            charset);
+                                  charset);
         metadata.put(TxxxMetadata.TYPE, new TxxxMetadata(description, value));
       } else if (frameId0 == 'P' && frameId1 == 'R' && frameId2 == 'I' && frameId3 == 'V') {
         // Check frame ID == PRIV
@@ -98,7 +98,7 @@ public class Id3Parser implements MetadataParser<Map<String, Object>> {
         int descriptionStartIndex = filenameEndIndex + delimiterLength(encoding);
         int descriptionEndIndex = indexOfEOS(frame, descriptionStartIndex, encoding);
         String description = new String(frame, descriptionStartIndex,
-                descriptionEndIndex - descriptionStartIndex, charset);
+                                        descriptionEndIndex - descriptionStartIndex, charset);
 
         byte[] objectData = new byte[frameSize - descriptionEndIndex - 2];
         System.arraycopy(frame, descriptionEndIndex + delimiterLength(encoding), objectData, 0,
@@ -121,15 +121,6 @@ public class Id3Parser implements MetadataParser<Map<String, Object>> {
   private static int indexOf(byte[] data, int fromIndex, byte key) {
     for (int i = fromIndex; i < data.length; i++) {
       if (data[i] == key) {
-        return i;
-      }
-    }
-    return data.length;
-  }
-
-  private static int indexOfNot(byte[] data, int fromIndex, byte key) {
-    for (int i = fromIndex; i < data.length; i++) {
-      if (data[i] != key) {
         return i;
       }
     }
--- a/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
+++ b/library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
@@ -70,7 +70,7 @@ public class Id3Parser implements MetadataParser<Map<String, Object>> {
         int valueStartIndex = firstZeroIndex + delimiterLength(encoding);
         int valueEndIndex = indexOfEOS(frame, valueStartIndex, encoding);
         String value = new String(frame, valueStartIndex, valueEndIndex - valueStartIndex,
-                                  charset);
+            charset);
         metadata.put(TxxxMetadata.TYPE, new TxxxMetadata(description, value));
       } else if (frameId0 == 'P' && frameId1 == 'R' && frameId2 == 'I' && frameId3 == 'V') {
         // Check frame ID == PRIV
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -67,6 +67,39 @@ public boolean canParse(String mimeType) {
...
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String mimeType = new String(frame, 0, firstZeroIndex);
+        int filenameStartIndex = firstZeroIndex + 1;
+        int filenameEndIndex = indexOf(frame, filenameStartIndex, (byte) 0);
+        String filename = new String(frame, filenameStartIndex,
+                                     filenameEndIndex - filenameStartIndex, charset);
+        int descriptionStartIndex = filenameEndIndex + 1;
```
andudo: Use indexOfNot(frame, filenameEndIndex, (byte) 0); In some encodings it could be a single 0 byte delimiter, in some it might be two 0 bytes. Same applies for the objectData start. I'd also use it for filenameStartIndex for consistency.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -67,6 +67,39 @@ public boolean canParse(String mimeType) {
...
+        metadata.put(PrivMetadata.TYPE, new PrivMetadata(owner, privateData));
+      } else if (frameId0 == 'G' && frameId1 == 'E' && frameId2 == 'O' && frameId3 == 'B') {
+        // Check frame ID == GEOB
+        int encoding = id3Data.readUnsignedByte();
+        String charset = getCharsetName(encoding);
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String mimeType = new String(frame, 0, firstZeroIndex);
```
andudo: Use ISO-8859-1 explicitly. Android platform's default is UTF-8.

andudo:
Thank you! Overall looks good, just a few small comments.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
jeoliva: You meant for creating owner and mimeType strings, right? 

Thanks

library/src/main/java/com/google/android/exoplayer/metadata/PrivMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+ * with time indices.
+ */
+public class PrivMetadata {
+
+  public static final String TYPE = "PRIV";
+
+  public final String owner;
+  public final byte[] privateData;
+
+  public PrivMetadata(String owner, byte [] privateData) {
```
andudo: Please remove a space between byte and [].

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
andudo: Yes.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
jeoliva: Sorry, not sure about this change. In some cases filename and/or description fields are empty fields and then just a byte 0 is found in their locations (from ID3 format spec: "The first two
   strings may be omitted, leaving only their terminations"). This means that, for example, in case filename is fullfilled (with value = "any", for example) but description is empty I will found the following in frame byte array:
// x == filenameStart
frame[x] = 'a'
frame[x+1] = 'n'
frame[x+2] = 'y'
frame[x+3] = 0 // End of filename field
frame[x+4] = '0' // End of description field, because it is empty
frame[x+5] = // Start of objectData field

This scenario is happening, for example, in the description field of GEOB frames included in Apple test stream. They are always set as empty.

In those cases if I use indexOfNot(frame, filenameEndIndex, (byte) 0); then I would skip the byte 0 indicating description field is empty and will take objectData field as the description one.

Am I wrong? Am I missing anything?

andudo: Good point. But then you'll have to skip one or two zero bytes based on the encoding. For ISO-8859-1 and UTF-8 it's a single 0 byte, for UTF-16 and UTF-16BE it's two.

jeoliva: Ok, right what I thought. I am going to take a look and propose something that works for all the available codifications.

Thanks for your comments

jeoliva:
Andudo, I have added a patch for UTF-16/UTF-16B encoding and the method for finding the termination string. I noticed there was something wrong in the way used for finding termination string. Although it was fine for single byte encoded strings, there was a mistake for UTF-16/UTF-16B in which we should look for a 2 consecutive zero bytes. Otherwise, as many characters encoded with UTF-16 or UTF-16BE has one of their 2 bytes set with the value zero, we will truncate text fields.

Note: I added a method for finding EOS character (indexOfEOS). In this method I have reused the already defined method (indexOf) just for making easier understanding the code, although it would work more efficiently if indexOfEOS is coded without using indexOf at all.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -98,6 +136,25 @@ private static int indexOfNot(byte[] data, int fromIndex, byte key) {
...
+
+    // For single byte encoding charsets, we are done
+    if(encodingByte == ID3_TEXT_ENCODING_ISO_8859_1 || encodingByte == ID3_TEXT_ENCODING_UTF_8) {
+      return terminationPos;
+    }
+
+    // Otherwise, look for a two zero bytes
+    while(terminationPos < data.length - 1) {
+      if(data[terminationPos + 1] == (byte) 0) {
+        return terminationPos + 1;
```
andudo: This returns not the index of the delimiter but the index of the second byte of the delimiter. When you use this it makes the first 0 byte become part of the string.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -60,13 +65,46 @@ public boolean canParse(String mimeType) {
         byte[] frame = new byte[frameSize - 1];
         id3Data.readBytes(frame, 0, frameSize - 1);
 
-        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        int firstZeroIndex = indexOfEOS(frame, 0, encoding);
         String description = new String(frame, 0, firstZeroIndex, charset);
-        int valueStartIndex = indexOfNot(frame, firstZeroIndex, (byte) 0);
-        int valueEndIndex = indexOf(frame, valueStartIndex, (byte) 0);
+        int valueStartIndex = firstZeroIndex + 1;
```
andudo: If firstZeroIndex is the actual index of the first zero then this should be int valueStartIndex = firstZeroIndex + delimiterLength(encoding);

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
jeoliva: In fact, I did that intentionally but you are right, I included the termination delimiter in the resulting string but that's wrong and will cause wrong results when converting to string. Thanks for the tip.

jeoliva:
Right added a few changes to take into account the length of delimiter for different encodings. I could test it successfully with single byte encoding although couldn't test it with UTF16 nor UTF16BE as id3taggenerator tool from Apple only generates UTF8 strings.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -60,13 +65,46 @@ public boolean canParse(String mimeType) {
...
-        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        int firstZeroIndex = indexOfEOS(frame, 0, encoding);
         String description = new String(frame, 0, firstZeroIndex, charset);
-        int valueStartIndex = indexOfNot(frame, firstZeroIndex, (byte) 0);
-        int valueEndIndex = indexOf(frame, valueStartIndex, (byte) 0);
+        int valueStartIndex = firstZeroIndex + delimiterLength(encoding);
+        int valueEndIndex = indexOfEOS(frame, valueStartIndex, encoding);
         String value = new String(frame, valueStartIndex, valueEndIndex - valueStartIndex,
-            charset);
+                                  charset);
```
ojw28: optional style nit - Can you put the indentation back? Our style is to indent 4 chars for a line continuation (also on L97/L101). Feel free to ignore and this will get cleaned up automagically at some point. Something to be aware of for future patch sets though :).

ojw28:
Had a quick scan (didn't look at the gritty details of around charsets!). Looks good to me, thanks! Will leave for @andudo to do a final pass / merge.

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
jeoliva: Thanks for your comments. I have fixed indentation issues although will wait for @andudo comments before committing. Just to minimize number of commits with very small changes.

@ojw28, just one question. I am wonder how you detect this kind of code formatting stuffs! :). Not this one, that is a bit easier (my fault introducing it...), but detecting an extra space in a very large sentence. Are you using any Android Studio plugin that does this format code checking or you just have that plugin in your eyes/brain? :)

ojw28: Hah. Yeah, for indentation you can normally setup your IDE to automatically fix it to the intended style (more or less). Andudo saw the extra space, not me. I think he probably has the eyes/brain plugin thing; guess you just get good at spotting that kinda thing once you've spent enough time staring at code!

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -60,13 +65,46 @@ public boolean canParse(String mimeType) {
...
+      } else if (frameId0 == 'G' && frameId1 == 'E' && frameId2 == 'O' && frameId3 == 'B') {
+        // Check frame ID == GEOB
+        int encoding = id3Data.readUnsignedByte();
+        String charset = getCharsetName(encoding);
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String mimeType = new String(frame, 0, firstZeroIndex, "ISO-8859-1");
+        int filenameStartIndex = firstZeroIndex + delimiterLength(encoding);
```
andudo: There is always a single 0 byte after mime type.
int filenameStartIndex = firstZeroIndex + 1;

library/src/main/java/com/google/android/exoplayer/metadata/Id3Parser.java
```diff
@@ -60,13 +65,46 @@ public boolean canParse(String mimeType) {
...
+        int filenameEndIndex = indexOfEOS(frame, filenameStartIndex, encoding);
+        String filename = new String(frame, filenameStartIndex,
+                                     filenameEndIndex - filenameStartIndex, charset);
+        int descriptionStartIndex = filenameEndIndex + delimiterLength(encoding);
+        int descriptionEndIndex = indexOfEOS(frame, descriptionStartIndex, encoding);
+        String description = new String(frame, descriptionStartIndex,
+                                        descriptionEndIndex - descriptionStartIndex, charset);
+
+        byte[] objectData = new byte[frameSize - descriptionEndIndex - 2];
```
andudo: You forgot to take the delimiterLength into the account here.
int objectDataSize = frameSize - 1 /* encoding byte */ - descriptionEndIndex - delimiterLength(encoding);
byte[] objectData = new byte[objectDataSize];
System.arraycopy(frame, descriptionEndIndex + delimiterLength(encoding), objectData, 0, objectDataSize);

jeoliva: Fixed. Thanks for your comment.

Andudo, just an additional comment. Following ID3 specifications, for most of ID3 tags (for example, for TXXX, PRIV or GEOB ones) you can find more than one tag of the same type in a single frame. However, given we are storing metadata in a Hashmap, and using type of the tag as the key, we are missing that possibility and we will save just the last tag of each type. 

To be honest I never see that (two ID3 tags of same type in same frame) in any HLS stream, although seems to be a possibility. A quick solution to this could be moving from a Hashmap to an ArrayList although we will lose the efficiency of looking for tags of a specific type that we have right now. Otherwise, we can set as the key a good combination of TYPE + a parameter of the ID3 tag that is unique (and different for each type of id3 tag) although this will make the Hashmap useless (difficult to know the combination of type + parameter) and at the end of the day, dev user will need to iterate Hashmap to find the metadata she is looking for that is more inefficiency that iterating over an ArrayList.

What do you think?

PR #325, opened 2015-02-26
Added MPEG audio support to TsExtractor

tresvecesseis:
This is the first try to support more formats in TsExtractor, it works properly with streams with only one audio track (same as current AAC support)


library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
```diff
@@ -0,0 +1,297 @@
...
+        if (!hasMediaFormat()) {
+            mpaScratch.skipBits(12);
+            int isLSF = (!mpaScratch.readBit()) ? 1 : 0;
+            int layer = mpaScratch.readBits(2) ^ 3;
+            mpaScratch.skipBits(1);
+            int audioObjectType = 32 + layer;
+            int bitRate = MPA_BITRATES[isLSF][layer][mpaScratch.readBits(4)];
+            int sampleRate = MPA_SAMPLING_RATES[3 - isLSF][mpaScratch.readBits(2)];
+            int sampleRateIndex = 0;
+            for (; sampleRateIndex < (CodecSpecificDataUtil.AUDIO_SPECIFIC_CONFIG_SAMPLING_RATE_TABLE).length; sampleRateIndex++) {
```
andudo: Please hide this logic inside CodecSpecificDataUtil, similar to the existing buildAudioSpecificConfig() method for AAC.
And keep AUDIO_SPECIFIC_CONFIG_SAMPLING_RATE_TABLE and AUDIO_SPECIFIC_CONFIG_CHANNEL_COUNT_TABLE private.

library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
```diff
@@ -49,13 +49,22 @@ private CodecSpecificDataUtil() {}
...
   public static Pair<Integer, Integer> parseAudioSpecificConfig(byte[] audioSpecificConfig) {
     int audioObjectType = (audioSpecificConfig[0] >> 3) & 0x1F;
-    int byteOffset = audioObjectType == 5 || audioObjectType == 29 ? 1 : 0;
-    int frequencyIndex = (audioSpecificConfig[byteOffset] & 0x7) << 1
-        | ((audioSpecificConfig[byteOffset + 1] >> 7) & 0x1);
-    Assertions.checkState(frequencyIndex < 13);
-    int sampleRate = AUDIO_SPECIFIC_CONFIG_SAMPLING_RATE_TABLE[frequencyIndex];
-    int channelCount = (audioSpecificConfig[byteOffset + 1] >> 3) & 0xF;
-    return Pair.create(sampleRate, channelCount);
+    if (audioObjectType != 31) {
```
andudo: Make it (audioObjectType < 31) just to be consistent with buildAudioSpecificConfig().

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
```diff
@@ -0,0 +1,297 @@
...
+                case STATE_READING_HEADER:
+                    int targetLength = hasCrc ? HEADER_SIZE + CRC_SIZE : HEADER_SIZE;
+                    if (continueRead(data, mpaScratch.getData(), targetLength)) {
+                        parseHeader();
+                        startSample(timeUs);
+                        /**
+                         * Reset buffer after Header data extraction in order to pass full
+                         *  mpeg audio frame to the mpeg decoder
+                         */
+                        data.setPosition(data.getPosition() - targetLength);
```
andudo: This looks hacky. Could you give some more details on why is this needed?

andudo:
Thank you! This is nice. Although some of this code is the same as in AdtsReader, maybe it makes sense to put it into a common class for example.

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
tresvecesseis: In the case of AAC, the header parsing is done in is own class, Im only reusing the tables in the CodecSpecificDataUtil but I could repeat these private tables here if this is pose a problem

library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
tresvecesseis: The Mpeg Audio specifies these object type extensions to the object type, I´m only completing the spec

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
tresvecesseis: The mpeg decoder needs the complete frame including the header to decode de mpeg audio, in the case of AAC it is wrapped in an adts wrapper that is what its removed

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
andudo: I was referring to this method https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java#L84
It's better to leave them private and expose a getSampleRateIndex(int sampleRate) function CodecSpecificDataUtil, and reuse it here an in the method I referred to. Same for channelConfig.

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
andudo: If you make parseHeader() handle appending header data to the sample, then parseHeader() will be the only method in this class that is different from AdtsReader.

library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
andudo: In this particular case it doesn't matter if this line says if (audioObjectType != 31) or if (audioObjectType < 31)

andudo:
@tresvecesseis can you also share the media (URL) that could me used to verify this?

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
tresvecesseis: Ok

library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
tresvecesseis: Ok, i didnt understand you, i will change this

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
tresvecesseis: OK, I will try to create a common class for the two audio readers

tresvecesseis:
Our URLs are private to our VDN and contain multiple audio tracks, I'll try to share with you a TS with only an audio track tomorrow.

library/src/main/java/com/google/android/exoplayer/hls/parser/MpaReader.java
ojw28: I'd rather we keep them separate for the initial push. It's not much duplication (simply because the classes are small) and we can establish a common class as a following change if necessary.

FYI - I'm working on some larger extractor/reader related changes, and keeping this separate for now will make it much easier come merge-conflict time.

tresvecesseis: My next work was going to reorganize the extractors and readers to be able to reuse them with other transports/sources, it seems that I´ll have to go towards my next work item, an RTP extractor

ojw28: I'm already working on an attempt to unify all of ExoPlayer's extractors under a single interface (which, if it works out, would let you do thinks like put fragmented mp4 inside HLS streams, if you really wanted to!). It's a pretty scary change though; it'll likely take a while. We're also working on supporting additional extractors (MP4, MP3 will be the first ones). Just a heads up in any case, to try and minimize duplication.

ojw28: By the way - this line looks like it'll crash if continueRead is actually continuing a read (i.e. the read was started from the previously consumed ParsableByteArray). In this case you'll end up setting the position to a negative value here, which is invalid.

tresvecesseis: in the case of MP3, if you attach the MpaReader to the fileDataSource or httpDataSource you  sould be able to reproduce audio files as the frame format of the mpeg audio is the same as in the TS

tresvecesseis:
During the testing of the path, it seems that there are sync problems between the Audio and Video tracks, in live channels the video seems to stutter but in carrousel channels where there are bigger clock differences between the tracks, the video is frozen until the video reach the same timestamp. I think it would more sensible to discard the audio frames with a timestamp lower than the timestamp of the first video frame

tresvecesseis:
My audio tracks are MPEG 1 Layer 2, 48khz, stereo. The problem is that the google decoder (OMX.google.mp3.decoder) seems to be working with these tracks in some boxes and failing in others, the failing ones seem to have other alternative decoders that are able to decode the audio tracks but I don´t know how I can test the available mpeg audio decoders without impacting in the video starting time.

library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
```diff
@@ -0,0 +1,293 @@
...
+
+            byte[] audioSpecificConfig = CodecSpecificDataUtil.buildAudioSpecificConfig(
+                    audioObjectType, sampleRateIndex, channelConfig);
+            Pair<Integer, Integer> audioParams = CodecSpecificDataUtil.parseAudioSpecificConfig(
+                    audioSpecificConfig);
+
+            // need to investigate how to detect if the mpeg decoder supports Layers other than Layer III
+            MediaFormat mediaFormat = MediaFormat.createAudioFormat(/*isLSF == 1 ?*/ MimeTypes.AUDIO_MPEG/* : MimeTypes.AUDIO_MP1L2*/,
+                    MediaFormat.NO_VALUE, audioParams.second, audioParams.first,
+                    Collections.singletonList(audioSpecificConfig));
```
optman: the code crash on my machine, and change this parameter to null, fix it!

tresvecesseis: could you post your mpeg audio track characteristics, I can´t reproduce your problem with my test streams

optman: I don't think it's about audio info, it works on some other phone. My phone is Nexus 5, with android version 5.1.1。I think set initialData parameter to null just be fine,  media codec can figure it out anyway.

optman: I have try the links you provided, it crash when initialData is not set to null.
http://www.endlesspools.com/iphone/iphone_streams/Fastlane/Fastlane_128k.m3u8
http://www2.lapresse.it/video/mp4/1.15753/1.15753_128k.m3u8

tresvecesseis: It must be a codec implementation issue, it works properly in all my boxes
(5 different boxes from 4 different vendors, 2 Android TVs and another 2
TVs with  Android 4)

Sergio

tresvecesseis: I have to double check the audio channels code with these test URLs, I will
keep you updated

tresvecesseis: i have a bug in the buildAudioSpecificConfig, the channel config is 4 bits, could you try the following patch?

```java
diff --git a/library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java b/library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
index 8e9bc9b..26094a0 100644
--- a/library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
+++ b/library/src/main/java/com/google/android/exoplayer/util/CodecSpecificDataUtil.java
@@ -77,8 +77,8 @@ public final class CodecSpecificDataUtil {
       int frequencyIndex = (audioSpecificConfig[1] & 0x1E) >> 1;
       Assertions.checkState(frequencyIndex < 13);
       int sampleRate = AUDIO_SPECIFIC_CONFIG_SAMPLING_RATE_TABLE[frequencyIndex];
-      int channelCount = (audioSpecificConfig[1] & 0x01) << 2 |
-              ((audioSpecificConfig[2] >> 6) & 0x03);
+      int channelCount = (audioSpecificConfig[1] & 0x01) << 3 |
+              ((audioSpecificConfig[2] >> 5) & 0x07);
       return Pair.create(sampleRate, channelCount);
     }
   }
@@ -108,8 +108,8 @@ public final class CodecSpecificDataUtil {
               (audioObjectTypeExt >> 3) & 0x07);
       audioSpecificConfig[1] = (byte) ((audioObjectTypeExt << 5) & 0xE0 |
               (sampleRateIndex    << 1) & 0x1E |
-              (channelConfig      >> 2) & 0x01);
-      audioSpecificConfig[2] = (byte) ((channelConfig << 6) & 0xC0);
+              (channelConfig      >> 3) & 0x01);
+      audioSpecificConfig[2] = (byte) ((channelConfig << 5) & 0xE0);
       return audioSpecificConfig;
     }
   }
```

tresvecesseis: there is a remaining bug in the processing of the audioSpecificConfig related to tracks with 7+1 audio channels but is present in original implementation so I will submit another patch separately

optman: apply the new patch, still crash.

tresvecesseis: when you say you are settin initialData to null, you are talking about the last parameter of createAudioFormat, aren´t you?

library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
```diff
@@ -0,0 +1,293 @@
...
+
+        if (!hasOutputFormat) {
+            mpaScratch.setPosition(0);
+            mpaScratch.skipBits(12);
+            int isLSF = (!mpaScratch.readBit()) ? 1 : 0;
+            int layer = mpaScratch.readBits(2) ^ 3;
+            mpaScratch.skipBits(1);
+            int audioObjectType = 32 + layer;
+            int bitRate = MPA_BITRATES[isLSF][layer][mpaScratch.readBits(4)];
+            int sampleRate = MPA_SAMPLING_RATES[3 - isLSF][mpaScratch.readBits(2)];
```
optman: when sample rate is 11025(using FME), it doesn't work(table lookup result is wrong). it seen that mpeg 2.5 can't be reach.

library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
optman: yes

library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
tresvecesseis: could you try this patch to add mpeg 2.5 support?

```java
diff --git a/library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java b/library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
index 596314f..0be3ea4 100644
--- a/library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/ts/MpaReader.java
@@ -269,13 +269,14 @@ import java.util.Collections;

         if (!hasOutputFormat) {
             mpaScratch.setPosition(0);
-            mpaScratch.skipBits(12);
-            int isLSF = (!mpaScratch.readBit()) ? 1 : 0;
+            mpaScratch.skipBits(11);
+            int MPAVersionID = mpaScratch.readBits(2);
+            int isLSF = (MPAVersionID & 0x01) ^ 0x01;
             int layer = mpaScratch.readBits(2) ^ 3;
             mpaScratch.skipBits(1);
             int audioObjectType = 32 + layer;
             int bitRate = MPA_BITRATES[isLSF][layer][mpaScratch.readBits(4)];
-            int sampleRate = MPA_SAMPLING_RATES[3 - isLSF][mpaScratch.readBits(2)];
+            int sampleRate = MPA_SAMPLING_RATES[MPAVersionID][mpaScratch.readBits(2)];
             int sampleRateIndex = CodecSpecificDataUtil.getSampleRateIndex(sampleRate);
             int paddingBit = (mpaScratch.readBit()) ? 1 : 0;
             mpaScratch.skipBits(1);
```

optman: it seem that FME encoded mp3 is incorrect, setting is:

channels = stereo,  
sample rate = 11025,
bit rate = 32 kbps  

the mp3 header parse result is :
MPAVersionID = 11
isLSF = 0
layer = 0
bitRate = 224
sampleRate = 44100

obviously mismatch.

tresvecesseis: >
> it seem that FME encoded mp3 is incorrect, setting is:
>
> channels = stereo,
>
> sample rate = 11025,
> bit rate = 32 kbps
>
> the mp3 header parse result is :
> MPAVersionID = 11
> isLSF = 0
> layer = 0
> bitRate = 224
> sampleRate = 44100
>
> obviously mismatch.
>
> —
> Reply to this email directly or view it on GitHub.

Could you send me the file?

tresvecesseis: the above patch was a quick hack to test mpeg 2.5 support, being an undocumented proprietary format it was bit reckless and as your data shows, it's not possible to reliably detect an mpeg  2.5 stream. MPEG 2.5 use a 11 bit sync word, stealing a bit from the normal 12 bit MPEG audio sync word to transform the LSF bit in a MPEG version ID, my patch was faulty because the needed adaption of the skipToNextSync method to cope with the new sync word length, but you got 12 bit set in a row that, obviously, is not a sync word so with the adapted logic in skipToNextSync we could randomly sync to incorrect frame header.

optman: because the audio source is from FME live stream, i can only provide fragment of a frame. 

ff e3 20 70 b8 0b e8 8b 67 1f ce 20 00 01 20 02 5c ad 80 00 00 e9 ff d1 69 f4 6d 7a 96 ee d6 ed 3c 6f ff ff c0 00 00 ad 34 49 35 55 5b 6a 5c db 96 c6 c0 07 2d 40 6f e8 c1 c7 a8 71 58 cf 31 86

it seem that the header is correct! 

http://www.mp3-tech.org/programmer/frame_header.html

tresvecesseis: yep, I used the same page, the problem is that to reliable detect the header i have to store a mask of the first candidate header skip the calculate frame length and check if the next header is equal to the candidate header, if not we have to move one bit forward and try again, that´s trivial in a file but I have to see how to do it with the buffer classes that we use.

ojw28:
FYI, there appears to be a suitable test stream in https://github.com/google/ExoPlayer/issues/578, which also crashes (native crash in the MP3 decoder) using this patch. I suspect the codec configuration data is being parsed incorrectly from the stream, although I didn't check.

ojw28:
Aha, this does very nearly work. I tried replacing some of the logic in parseHeader with usage of com.google.android.exoplayer.extractor.mp3.MpegAudioHeader instead, and was able to get the sample in #578 to play without much hassle.

@tresvecesseis - If you could sign the CLA that's required to accept this request, then I can merge it and make the changes on top to get it working nicely. Please see [here](https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md) for CLA information.

tresvecesseis:
Ok, I just signed the CLA. I have checked this code lots of times and it
should work with any mpeg 1/2 streams (the mpeg 2.5 is not working because
different syncword), the only remaining problem is with skipToNextSync(),
It seems that the sync word could be misdetected as twelve ones is valid in
the MPA frame, so to reliable detect the MPA header we have to store a mask
(int HEADER_SYNC_MASK = 0xFFFE0CCF) of the first candidate header, skip the
calculated frame length and check if the next masked header is equal to the
candidate header mask (this masked fields of the header cannot change in
the stream), if they are not the same we have to move one bit forward and
repeat the whole process again, normally your stream will start in a
syncword but in the generic streaming scenario (multicast) is possible to
join to an already started stream in an arbitrary position.

ojw28:
Did you sign it using your github name? I can't find a record of it in our system. Thanks!

tresvecesseis:
Sorry, I put an incomplete github name (tresvecesseis instead of
tresvecesseis@gmail.com), it should be ok now

ojw28:
Great.

I think the problem with the stream in #578 and the change as it is here is related to the data passed as the final argument of MediaFormat.createAudioFormat. It appears that the decoder doesn't actually expect anything, as in our [Mp3Extractor](https://github.com/google/ExoPlayer/blob/dev/library/src/main/java/com/google/android/exoplayer/extractor/mp3/Mp3Extractor.java#L258) that passes the empty list. It's also possible to remove quite a bit of mpeg header parsing code, since equivalent code already exists in [MpegAudioHeader](https://github.com/google/ExoPlayer/blob/dev/library/src/main/java/com/google/android/exoplayer/extractor/mp3/MpegAudioHeader.java).

In any case, I'll merge this tomorrow and push a change on top that removes some of the duplication, cleans a few things up and fixes the MediaFormat. It's possible that the latter change may break playback of your streams, but in the worst case it should be straightforward to get them working again.

tresvecesseis:
ok, I was using audioSpecificConfig because I thought  it would reduce the
initialization time of the decoder and because it was used in the
AdtsReader, I only extended it a bit to support the MPA format and It
didn't cause any problem  with my Android TVs (Philips and Sony) or with my
Android 4.4 boxes, but given that the code was wrong until the last
committed patch it seems that the initialization data was just being
ignored by the decoders.

PR #336, opened 2015-03-05
Allow SingleFetchHelper to cancel

gitanuj:
This exposes a way to cancel SingleFetchHelper. In scenarios, like a music streaming app, where rapid switching between media is required, we need better control over cancelling previous requests.


library/src/main/java/com/google/android/exoplayer/util/ManifestFetcher.java
```diff
@@ -143,11 +143,13 @@ public void updateManifestUrl(String manifestUrl) {
    * @param callbackLooper The looper associated with the thread on which the callback should be
    *     invoked.
    * @param callback The callback to receive the result.
+   * @return Instance of SingleFetchHelper which is loading the manifest.
    */
-  public void singleLoad(Looper callbackLooper, final ManifestCallback<T> callback) {
+  public SingleFetchHelper singleLoad(Looper callbackLooper, final ManifestCallback<T> callback) {
```
ojw28: I don't see how you can actually use this, because SingleFetchHelper is a private inner class. We don't want to make it public, either.

library/src/main/java/com/google/android/exoplayer/util/ManifestFetcher.java
```diff
@@ -333,7 +348,6 @@ public void onLoadCompleted(Loadable loadable) {
 
     @Override
     public void onLoadCanceled(Loadable loadable) {
-      // This shouldn't ever happen, but handle it anyway.
```
ojw28: I don't think onSingleManifestError should be invoked if cancellation is now expected, so this would need to do something else.

ojw28:
Could you clarify what you're actually trying to achieve with this change? Are you trying to optimize network usage through cancellation? It's likely that the underlying network stack will choose to fully read the network response regardless. This is a common optimization in the case that the response is known to be relatively small (which is typically true for manifests), because completing the request allows for the socket to be kept open for potential subsequent requests.

 So I don't think you'll actually see any benefit there. Given this, can you not just ignore the response in your callback?

gitanuj:
I guess you're right. It only makes sense to actually cancel a request if it hasn't really started. Also, now that ManifestFetcher uses a supplied HttpDataSource I can probably use that to have better control over manifest downloading.

PR #349, opened 2015-03-11

canatella:

where the IV attribute is not set. The fix is to reset the initial
vector value for each media sequence so that it is correctly set to the
IV attribute value or the current media sequence number.

This is commit  9a302d4.

Apprently I messed up this pull request (first time using pull requests), I did not use multiple branches so it also has a temptative fix for issue 73 in the other commits.


ojw28:
You've added a whole bunch of commits to this request that you didn't mean to add. Can you get rid of them? Thanks!

canatella:
Is there an easy way to remove the commits from the pull request ?

ojw28:
http://stackoverflow.com/questions/8939808/git-remove-unwanted-commits-from-a-pull-request

ojw28:
No worries, thanks for the fix. One other thing - You need to sign the Google CLA before we can merge your new request. See: https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md

PR #350, opened 2015-03-11

canatella:

where the IV attribute is not set. The fix is to use the provided IV if
given or use the current media sequence number.


library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
```diff
@@ -216,13 +213,15 @@ private static HlsMediaPlaylist parseMediaPlaylist(LineIterator iterator, Uri ba
       } else if (line.equals(DISCONTINUITY_TAG)) {
         segmentDiscontinuity = true;
       } else if (!line.startsWith("#")) {
+        String thisSegmentEncryptionIV =
```
ojw28: Looking more closely at this, presumably you should only generate the segmentMediaSequence derived encryptionIV if the content is actually encrypted as AES-128. In which case you need an extra condition. Something like:

```
String thisSegmentEncryptionIV;
if (segmentEncryptionIV != null) {
  thisSegmentEncryptionIV = segmentEncryptionIV;
} else if (HlsMediaPlaylist.ENCRYPTION_METHOD_AES_128.equals(segmentEncryptionMethod)) {
  thisSegmentEncryptionIV = Integer.toHexString(segmentMediaSequence);
} else {
  thisSegmentEncryptionIV = null;
}
```

Does that make sense? Sorry, should have seen that the first time around!

ojw28:
One other comment - This needs to be merged into the dev branch, not master.

PR #353, opened 2015-03-12
Hls alternate audio

canatella:
This is a tentative fix for the audio part of issue 73 . It is loosely based on the multi track support for Dash. It only support one audio group for all the streams and completely bypass the TS audio. I think this is the basic use case for alternative audio as mentioned by @perchrh and @johanlindquist.

It has problem with audio synchronisation. Any idea what's the problem? When switching audio tracks, it gets synchronised but it's quickly drifting after.

I updated the parser so that it parses all alternate media types and not only the subtitles.

Also there is no load control as mentioned in the comment by @ojw28.

I tried to split the change in a way that make sense so please do a per commit review. Also I'm new to ExoPlayer and media players so please forgive me if I made stupid assumptions ;-)

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/player/DemoPlayer.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/player/DemoPlayer.java
@@ -25,7 +25,7 @@ import com.google.android.exoplayer.MediaCodecVideoTrackRenderer;
 import com.google.android.exoplayer.TrackRenderer;
 import com.google.android.exoplayer.audio.AudioTrack;
 import com.google.android.exoplayer.chunk.ChunkSampleSource;
-import com.google.android.exoplayer.chunk.MultiTrackChunkSource;
+import com.google.android.exoplayer.MultiTrackSource;
 import com.google.android.exoplayer.drm.StreamingDrmSessionManager;
 import com.google.android.exoplayer.metadata.MetadataTrackRenderer;
 import com.google.android.exoplayer.text.TextRenderer;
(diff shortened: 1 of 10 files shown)


andreimarinescu:
This fixes my issue in #372. I did find some bugs that cause the player to crash (the biggest issues being that if a manifest contains a SUBTITLE type EXT-X-MEDIA directive, HlsPlaylistParser throws an exception, and that the regex for MEDIA_TAG should be #EXT-X-MEDIA: instead of #EXT-X-MEDIA).  I was able to work around these for testing purposes though and audio works perfectly.

Any plans of getting this merged, any way I can contribute? I'm quite new to ExoPlayer but I'd love to help.

ojw28:
This is roughly how this feature should be implemented, but there are some important caveats around handling of HLS Live streams, and around buffer management (not using too much memory, and also ensuring that the audio and video streams load at the same rate, rather than buffering way more audio than video).

We're working on some changes that better align the Chunk and HLS packages. These changes will allow components in the Chunk package, such as LoadControl (for proper buffer management) and hopefully MultiTrackChunkSource (to avoid forking) to be re-used. We'll likely push these changes towards the end of this week, at which point we'll be better placed to implement this feature.

RikHeijdens:
@ojw28 Any updates on aligning the Chunk and HLS packages? It would be nice if we could use setSelectedTrack() for Audio tracks just as we can with DASH in the future.

itaykinnrot:
@ojw28  Any timeline for this feature? Anything we can help with?

ojw28:
We support multi-audio for HLS now, so closing this.

PR #354, opened 2015-03-12

canatella:

where the IV attribute is not set. The fix is to use the provided IV if
given or use the current media sequence number.


ojw28:
Thanks!

ojw28:
FYI - I simplified this a little more in https://github.com/google/ExoPlayer/commit/fb36f85e0d5e4c9f7c5e0d92586b004b71f43496. Let me know if I broke something.

PR #365, opened 2015-03-20
The value of AUTOSELECT, DEFAULT attributes of subtitles in HLS playist doesn't include ' " ' character

kenial:
They are represented as these and takes only YES / NO values:
  AUTOSELECT=YES,DEFAULT=YES


ojw28:
Thanks for this. We've fixed this in a way that defines the allowable values in a more strict way, here: https://github.com/google/ExoPlayer/commit/5a5935cb72857580803573f0d44703516d332558

PR #381, opened 2015-04-03
Enabled Multi Adaption sets for Type in Manifest (Dash)

s3xy4ngyc:
This Applies to Manifests Having Multiple Audio Adaptions in Manifest (For Example on Multi Language Content)

Example for Multi Language Manifest File:

```
<MPD ... >
...
  <Period>
   ...
    <AdaptationSet
      group="1"
      lang="de"
      minBandwidth="1536000"
      maxBandwidth="1536000"
      segmentAlignment="true"
      audioSamplingRate="48000"
      mimeType="audio/mp4"
      codecs="mp4a.40.2">
      <AudioChannelConfiguration
        schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011"
        value="2">
      </AudioChannelConfiguration>
      ...
      <Representation
        id="audio_deu=1536000"
        bandwidth="1536000">
      </Representation>
    </AdaptationSet>
    <AdaptationSet
      group="1"
      lang="en"
      minBandwidth="1536000"
      maxBandwidth="1536000"
      segmentAlignment="true"
      audioSamplingRate="48000"
      mimeType="audio/mp4"
      codecs="mp4a.40.2">
      <AudioChannelConfiguration
        schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011"
        value="2">
      </AudioChannelConfiguration>
     ... 
      <Representation
        id="audio_eng=1536000"
        bandwidth="1536000">
      </Representation>
    </AdaptationSet>
   ...
  </Period>
</MPD
```

This Also Applies to Video Sources (For Example Different Aspect Ratios)

I only Extended the Dash Renderer for Multi Audio Files

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/player/DashRendererBuilder.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/player/DashRendererBuilder.java
@@ -70,7 +70,9 @@ import android.widget.TextView;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * A {@link RendererBuilder} for DASH.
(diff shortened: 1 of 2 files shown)


ojw28:
- The patch is valid, but there's no particular need for us to support multiple audio adaptation sets in the demo app (particularly since we don't have any sample content in the demo app that makes use of it). This is the kind of thing that developers should add to their own app integration if they need it.

- Re the Period.java change, you can already iterate over adaptationSets directly from app code, so there's no need for that method.

PR #382, opened 2015-04-03
Fix: made HttpDataSource compatible with servers that do not support range requests

nbruning:
When a http request is re-requested, a Range request is used. If the server returns a full response (200 OK), ExoPlayer would throw an UnexpectedLengthException exception.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java
@@ -56,7 +56,8 @@ public class DefaultHttpDataSource implements HttpDataSource {
   private HttpURLConnection connection;
   private InputStream inputStream;
   private boolean opened;
-  private long skipBytes;
+  private long bytesToSkipAtBegin;
+  private long stopReadAfter;
 
   private long dataLength;
   private long bytesRead;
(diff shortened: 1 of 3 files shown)


ojw28:
Could you clarify what use case you're finding this problematic? If you're hitting the problem for non-chunked media then I think it's right to treat it as an error, since serving non-chunked media and not supporting range requests is fundamentally flawed. For chunked media it's less flawed for the server to not support range requests, and I agree we should handle servers not doing so for that case.

Thanks!

library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java
```diff
@@ -165,8 +166,22 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
       throw new InvalidContentTypeException(contentType, dataSpec);
     }
 
+    // only used if gzip not used
+    long contentLength = getContentLength(connection);
+
+    if (responseCode == 200 && dataSpec.position != 0) {
```
ojw28: This doesn't consider range requests that start at offset 0 but impose an upper limit less than the length of the content. I think the condition should be:

if (responseCode == 200 && !(dataSpec.position == 0 && dataSpec.length != C.LENGTH_UNBOUNDED)) {
  ....
}

You then also need to impose the upper bound in the read method (that can happen unconditionally on all requests that impose such a limit).

nbruning:
I hit this problem using a chunked (DASH) source.
The Unified streaming server does not handle range requests, probably due to caching servers somewhere in the media path.
As supporting range requests is not mandated by any standard (right?), I believe we should have a workaround in the ExoPlayer. Appended is a second patch also supporting range requests with a bounded length.

ojw28:
If you're streaming non-chunked media then it's mandatory for a sensible user experience. But yes, we should have a solution for the chunked case at least, where actually it's kind of reasonable to not support range requests (as long as the chunks are not too large). I'll take a look at the updated patch.

In the meantime, note that we need a signed CLA to accept pull requests. See:

ojw28:
One other question - Do you have any test streams that reproduce this? The Unified Streaming test loop (http://live.unified-streaming.com/loop/loop.isml/loop.mpd) supports partial requests when I try. This is independent to the question of whether we should fix this (we should). I'm just interested in having a test stream available for this case.

ojw28:
Ignore last question - I've found an example (albeit not a DASH one).

ojw28:
Turns out it's a little bit more complicated to cover everything off correctly. I've submitted https://github.com/google/ExoPlayer/commit/362dc5f382e3a87f04fed017248e225eeb416396 to support servers that don't support partial requests.

PR #429, opened 2015-05-05
Fix DataSourceInputStream.read() for EOF cases

gitanuj:
Currently, read() returns garbage data when EOF is reached. Ideally it should return -1 if EOF is reached.


ojw28:
Thanks for this. You need to sign the CLA before we can merge. See: https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md

gitanuj:
Done

ojw28:
Thanks!

PR #465, opened 2015-05-18
support content uri (e.q. content://xxx)

Ood-Tsen:
In android, content-uri is widely used.
For example, "content://download/" , "content://media/external/video" and so on.

We can query the file descriptor from content resolver and create the input stream from it.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/upstream/ContentDataSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/upstream/ContentDataSource.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright (C) 2014 The Android Open Source Project
+ *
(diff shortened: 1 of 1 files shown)


ojw28:
Thanks. We'll push a change shortly that makes this usable via DefaultUriDataSource as well (DefaultUriDataSource will take http/https/asset/content Uris and delegate to the correct more-specific DataSource implementation).

Ood-Tsen:
It's great !!

Can we consider the empty scheme case ?(e.g. like #467)

> /storage/emulated/0/Android/data/com.getvictorious.staging.eatyourkimchi/files/DCIM/Video/Video-20150518_142803.mp4

ojw28:
Yeah, I think we should probably assume file:// if the scheme is empty.

PR #489, opened 2015-05-23
Reset positionUs, durationUs & bufferedPositionUs in prepare, stop & release

gitanuj:
If we use single ExoPlayer instance for playing media, the position parameters are not updated until the player has finished preparing and it would return parameters from last prepared media. This provides unreliable data for the UI. This case is encountered very often in a music application where the user can rapidly switch media. A workaround is to not rely on these parameters until the media has prepared.

Also, would it make sense to reset positionUs to TrackRenderer.UNKNOWN_TIME_US too?


ojw28:
There are valid reasons why these variables shouldn't be reset in these places. For example if you reset in prepare, you break the ability to seek to a non-zero position before prepare is called. If you reset in stop, you break the ability to query the final playback position if an error occurred, and so on.

You may as well just create a new ExoPlayer instance every time to achieve what you need; the overhead of doing so should be small. Alternatively you should wait until the player transitions to the prepared state before you read the values.

PR #498, opened 2015-05-27
Subrip (SRT) support

ened:
For details on SRT, see http://en.wikipedia.org/wiki/SubRip

Still missing in this PR is support for x1,x2,y1,y2 coordinate parsing.

@ojw28 seeking your advice on the required data model changes. `Cue` could probably have traits like `Positionable` (SRT) or `Alignable` (for WebVTT) or `ImageRepresentable` (for SDH, at a later stage?).

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripCue.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripCue.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2014 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.exoplayer.text.subrip;
+
+import com.google.android.exoplayer.text.Cue;
+
+/**
+ * A representation of a SubRip cue.
+ */
+/* package */ final class SubripCue extends Cue {
+
+  public final long startTime;
+  public final long endTime;
+
+  public SubripCue(CharSequence text) {
+    this(Cue.UNSET_VALUE, Cue.UNSET_VALUE, Cue.UNSET_VALUE,text);
+  }
+
+  public SubripCue(long startTime, long endTime, int position, CharSequence text) {
+    super(text, Cue.UNSET_VALUE, position, null, Cue.UNSET_VALUE);
+    this.startTime = startTime;
+    this.endTime = endTime;
+  }
+
+  /**
+   * Returns whether or not this cue should be placed in the default position and rolled-up with
+   * the other "normal" cues.
+   *
+   * @return True if this cue should be placed in the default position; false otherwise.
+   */
+  public boolean isNormalCue() {
+    return (line == UNSET_VALUE && position == UNSET_VALUE);
+  }
+
+}
--- a/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright (C) 2014 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.exoplayer.text.subrip;
+
+import android.text.Html;
+
+import com.google.android.exoplayer.C;
+import com.google.android.exoplayer.ParserException;
+import com.google.android.exoplayer.text.Cue;
+import com.google.android.exoplayer.text.SubtitleParser;
+import com.google.android.exoplayer.util.MimeTypes;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A simple SRT parser.
+ * <p/>
+ *
+ * @see <a href="https://en.wikipedia.org/wiki/SubRip">Wikipedia on SRT</a>
+ */
+public final class SubripParser implements SubtitleParser {
+
+  private static final String TAG = "SubRipParser";
+
+  private static final String SUBRIP_POSITION_STRING = "^(\\d)$";
+  private static final Pattern SUBRIP_POSITION = Pattern.compile(SUBRIP_POSITION_STRING);
+
+  private static final String SUBRIP_CUE_IDENTIFIER_STRING = "^(.*)\\s-->\\s(.*)$";
+  private static final Pattern SUBRIP_CUE_IDENTIFIER =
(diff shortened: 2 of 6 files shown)


ened:
SRT streams can be external (.srt-File) or embedded in the container, like MKV. I believe we would need to setup potentially multiple SampleSources for multiple .srt files, if needed (e.g. when the a player scans the directory and detects more files).

For embedded (e.g. mkv) containers, I'd hope we can do something like this:
``` java
TrackRenderer videoRenderer = new MediaCodecVideoTrackRenderer(sampleSource, MediaCodec.VIDEO_SCALING_MODE_SCALE_TO_FIT);
TrackRenderer audioRenderer = new MediaCodecAudioTrackRenderer(sampleSource);
TrackRenderer subtitleRenderer = new TextTrackRenderer(sampleSource, new TextRenderer() {
    @Override
    public void onCues(List<Cue> list) {

    }
}, Looper.myLooper(), new SubripParser());
```

Ood-Tsen:
Yes, @ened something like that,

android [mediaplayer](http://developer.android.com/reference/android/media/MediaPlayer.html) has
function "addTimedTextSource" for external text file,

Maybe ExoPlayer can do so.

ened:
Actually, snippet above is actual code. There are few things not yet clear, for example how to switch the tracks (which should be done without re-creating the player object).

dogo:
+1

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -0,0 +1,149 @@
...
+    textBuilder = new StringBuilder();
+  }
+
+  @Override
+  public SubripSubtitle parse(InputStream inputStream, String inputEncoding, long startTimeUs)
+          throws IOException {
+    ArrayList<SubripCue> subtitles = new ArrayList<>();
+
+    // file should not be empty
+    if (inputStream.available() == 0) {
```
ojw28: It's allowable for an InputStream to say it has nothing available when it actually has something available. IMO you can just not do this. It'll end up parsing an empty subtitle in this case, I think, which is fine?

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripCue.java
```diff
@@ -0,0 +1,66 @@
...
+  public SubripCue(CharSequence text) {
+    this(Cue.UNSET_VALUE, Cue.UNSET_VALUE, Cue.UNSET_VALUE,text);
+  }
+
+  public SubripCue(long startTime, long endTime, int position, CharSequence text) {
+    this(startTime, endTime, text, Cue.UNSET_VALUE, Cue.UNSET_VALUE, Cue.UNSET_VALUE,
+            Cue.UNSET_VALUE, Cue.UNSET_VALUE, position, null, Cue.UNSET_VALUE);
+  }
+
+  public SubripCue(long startTime, long endTime, CharSequence text, long x1, long x2, long y1,
```
ojw28: Can we get rid of this from this change? It doesn't appear to be used. Makes sense to bring it back in a subsequent request that does make use of it. Ditto for the x1/x2/y1/y2 member variables.

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A simple SRT parser.
+ * <p/>
+ *
+ * @see <a href="https://en.wikipedia.org/wiki/SubRip">Wikipedia on SRT</a>
+ */
+public class SubripParser implements SubtitleParser {
```
ojw28: final?

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -0,0 +1,149 @@
...
+    SubripSubtitle subtitle = new SubripSubtitle(subtitles, startTimeUs);
+    return subtitle;
+  }
+
+  @Override
+  public boolean canParse(String mimeType) {
+    return MimeTypes.APPLICATION_SUBRIP.equals(mimeType);
+  }
+
+  protected void handleNoncompliantLine(String line) throws ParserException {
```
ojw28: private?

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -0,0 +1,149 @@
...
+    if (strictParsing) {
+      throw new ParserException("Unexpected line: " + line);
+    }
+  }
+
+  private static long parseTimestampUs(String s) throws NumberFormatException {
+    if (!s.matches(SUBRIP_TIMESTAMP_STRING)) {
```
ojw28: Normally we'd use a Matcher for this, I think? Like you do for SUBRIP_POSITION. Any reason why not to?

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripSubtitle.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A representation of a SubRip subtitle.
+ */
+public class SubripSubtitle implements Subtitle {
```
ojw28: Can probably be

/* package */ final class

or at least final?

ojw28:
This looks good, thanks. Sorry for the delay in looking at it. Things like switching track + setting up the renderer + loading the subtitle should all be possible using the existing classes in the chunk package. I can make a nice util method for doing this once this is submitted.

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
ened: Actually the method is not used. Will check.

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
ened: How would this work with network-loaded SRTs? Say I'm loading a movie by http, and attach a SRT thats placed next to it?

ojw28: I'm not sure I understand the question. Once you get to this method, you don't know (or care) what the InputStream source is...

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripCue.java
ened: Hm, those fields and parameters should come back, because SubRip supports them.
This is a valid entry:

```
1
00:00:10,500 --> 00:00:13,000  X1:63 X2:223 Y1:43 Y2:58
<i>Elephant's Dream</i>
```

ojw28: They should eventually, yes. I'd rather defer them to a different pull request though. It's perfectly possible to implement srt support without supporting this feature, and then add it subsequently.

ened: OK got it.

library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -0,0 +1,149 @@
...
+
+      // parse text
+      textBuilder.setLength(0);
+      while (((line = subripData.readLine()) != null) && (!line.isEmpty())) {
+        if (textBuilder.length() > 0) {
+          textBuilder.append("<br>");
+        }
+        textBuilder.append(line.trim());
+      }
+      text = Html.fromHtml(textBuilder.toString());
```
ojw28: Is the only purpose of Html.fromHtml here to handle the <br> elements that you've inserted yourself? If so, please just append "\n" instead, so that you don't need to do this.

ojw28: Ignore this; worth keeping it for the "unofficial formatting support" mentioned in the wiki page.

ojw28:
I'm planning to pull this as-is and clean it up a bit after the merge. Unless you're actively working on it still.

ojw28:
I simplified this significantly in https://github.com/google/ExoPlayer/commit/fbbf3f27fdce46950f5cbe5e57fcc07325eb52f3. It looks a bit like you started off by copying the WebVTT implementation, but there was actually quite a lot of complexity there that's unnecessary for SRT.

ened:
Thanks @ojw28 for getting a much cleaner version of this into Exo. Sorry as I just had zero time left to work on it after submitting this.

And yes you were right that I started based on WebVTT :)

PR #509, opened 2015-06-01
parse mp4 tx3g

Ood-Tsen:
Atom "tx3g" is for text-based subtitle. Android MediaPlayer has support this format.

We can use [Mp4box](http://gpac.wp.mines-telecom.fr/mp4box/) of [QuickTimePro](https://developer.apple.com/library/mac/documentation/QuickTime/QT6_3/Chap1/QT6WhatsNew.html) to encode the SRT into the a MP4 file.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2014 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.exoplayer.text.tx3g;
+
+import com.google.android.exoplayer.text.Subtitle;
+import com.google.android.exoplayer.text.SubtitleParser;
+import com.google.android.exoplayer.util.MimeTypes;
+
+import android.util.Log;
+
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * A simple Text parser that supports Tx3g presentation profile.
+ * <p>
+ * Supported features in this parser are:
+ * <ul>
+ *   <li>content
+ *   <li>core
+ *   <li>presentation
+ *   <li>profile
+ *   <li>structure
+ *   <li>time-offset
+ *   <li>timing
+ *   <li>tickRate
+ *   <li>time-clock-with-frames
+ *   <li>time-clock
+ *   <li>time-offset-with-frames
+ *   <li>time-offset-with-ticks
+ * </ul>
+ * </p>
+ * @see <a href="http://www.w3.org/TR/ttaf1-dfxp/">TTML specification</a>
+ */
+public class TextParser implements SubtitleParser {
+  private static final String TAG = "TextParser";
+
+
+  private final List<SubtitleData> mSubtitleList;
+
+  /**
+   * Equivalent to {@code TtmlParser(true)}.
+   */
+  public TextParser() {
+    Log.i(TAG,"TextParser ");
+    mSubtitleList = new LinkedList<SubtitleData>();
+  }
+
+
+  @Override
+  public Subtitle parse(InputStream inputStream, String inputEncoding, long startTimeUs)
+  throws IOException {
+
+    DataInputStream in  = new DataInputStream(inputStream);
+    String text = in.readUTF();
+    text = (text == null) ? "" : text;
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/google/android/exoplayer/util/MimeTypes.java
```diff
@@ -53,6 +53,7 @@
   public static final String AUDIO_OPUS = BASE_TYPE_AUDIO + "/opus";
 
   public static final String TEXT_VTT = BASE_TYPE_TEXT + "/vtt";
+  public static final String TEXT_TX3G = BASE_TYPE_TEXT + "/tx3g";
```
ojw28: I think this should be:
public static final String APPLICATION_TX3G = BASE_TYPE_APPLICATION + "/x-quicktime-tx3g";

library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Comparator;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * A simple Text parser that supports Tx3g presentation profile.
+ * <p>
+ * Supported features in this parser are:
+ * <ul>
+ *   <li>content
```
ojw28: Please fix the Javadoc to be accurate.

library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
```diff
@@ -0,0 +1,102 @@
...
+ * @see <a href="http://www.w3.org/TR/ttaf1-dfxp/">TTML specification</a>
+ */
+public class TextParser implements SubtitleParser {
+  private static final String TAG = "TextParser";
+
+
+  private final List<SubtitleData> mSubtitleList;
+
+  /**
+   * Equivalent to {@code TtmlParser(true)}.
```
ojw28: Ditto

library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
```diff
@@ -0,0 +1,102 @@
...
+   */
+  public TextParser() {
+    Log.i(TAG,"TextParser ");
+    mSubtitleList = new LinkedList<SubtitleData>();
+  }
+
+
+  @Override
+  public Subtitle parse(InputStream inputStream, String inputEncoding, long startTimeUs)
```
ojw28: I'm confused about this method. It should be stateless and read the whole InputStream, returning the parsed result. So specifically, mSubtitleList should not exist (and should not accumulate entries across multiple calls to parse(). 

The SubtitleData class probably isn't needed at all (if it is, it should be completely immutable and not contain anything unnecessary, for example the end time - which doesn't appear to be used).

library/src/main/java/com/google/android/exoplayer/text/tx3g/TextSubtitle.java
```diff
@@ -0,0 +1,97 @@
...
+package com.google.android.exoplayer.text.tx3g;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.android.exoplayer.text.Cue;
+import com.google.android.exoplayer.text.Subtitle;
+
+/**
+ * A representation of a TTML subtitle.
```
ojw28: Javadoc needs fixing.

library/src/main/java/com/google/android/exoplayer/util/MimeTypes.java
Ood-Tsen: Yes, from this document, you are correct.
http://www.iss.net/security_center/reference/vuln/XML_QuickTime_3GPP_Detected.htm

library/src/main/java/com/google/android/exoplayer/text/tx3g/TextParser.java
Ood-Tsen: The "parse" function may be called for many times. and each time contains a single cue.

And ent time is not necessary, remove it.

ojw28: Right, but each call to parse is supposed to return a Subtitle containing just the single cue parsed during that call, not the accumulated result of all calls to parse so far.

The current approach can end up adding an arbitrarily large number of cues to the object, and will add duplicates if the user ever seeks backwards.

Ood-Tsen:
We don't need to keep the entire list.
As long as keep last few.

ojw28:
I don't understand why you need to keep any previous ones at all. Have you actually tried not doing so? It should work correctly.

Ood-Tsen:
Yes, we have to keep the cues in a list.

Not always to present the latest cue on screen, sometime is the second-last or the third-last.

There is a latency between "the time parse the cue" and "the time to present".

ojw28:
I don't really understand what you're saying. Provided the cues are in the correct order and do not overlay, you shouldn't need to keep more than one. Are they? Do you have some sample content handy that I can look at?

TextTrackRenderer already handles the difference between parsing and presentation times.

Ood-Tsen:
this is a sample of the tx3g file.
https://drive.google.com/open?id=0B7AjFwY8h0HrUmgweU5aSGlwcGc&authuser=0

ojw28:
Please share access to the link; thanks,

ojw28:
Ok, you can definitely do this without maintaining a list, however some changes in TextTrackRenderer are needed to make it do something more sensible. We'll pull this as it is, and make some modifications on top once it's merged. Thanks!

ojw28:
I've submitted https://github.com/google/ExoPlayer/commit/01affbb93ef685e91ab38b085dfaccb8bb011406 which simplifies this change. It should give the same end result, but please give it a try. Thanks!

PR #610, opened 2015-07-08
Dev

somsubhojit1988:


ojw28:
Not sure if this is an accident, but we only merge to master when we do actual releases.

Issue #426, opened 2015-05-03
Getting 'sending message to a Handler on a dead thread' (dev branch)


dnutcracker:
When performing the following call from the UI main thread: 
exoPlayer.blockingSendMessage(videoRenderer, MediaCodecVideoTrackRenderer.MSG_SET_SURFACE, null);

I randomly get the following exception:
05-03 13:41:39.468  23310-25973/me.samba.app W/MessageQueue﹕ Handler (com.google.android.exoplayer.upstream.Loader$LoadTask) {44fe3a10} sending message to a Handler on a dead thread
    java.lang.RuntimeException: Handler (com.google.android.exoplayer.upstream.Loader$LoadTask) {44fe3a10} sending message to a Handler on a dead thread
            at android.os.MessageQueue.enqueueMessage(MessageQueue.java:320)
            at android.os.Handler.enqueueMessage(Handler.java:626)
            at android.os.Handler.sendMessageAtTime(Handler.java:595)
            at android.os.Handler.sendMessageDelayed(Handler.java:566)
            at android.os.Handler.sendEmptyMessageDelayed(Handler.java:530)
            at android.os.Handler.sendEmptyMessage(Handler.java:515)
            at com.google.android.exoplayer.upstream.Loader$LoadTask.run(Loader.java:249)
            at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
            at java.util.concurrent.FutureTask.run(FutureTask.java:237)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
            at java.lang.Thread.run(Thread.java:841)

What am I doing wrong?

ojw28:
It's only a warning in the logs, right? What's actually happening here is that one of ExoPlayer's loading threads is notifying the playback thread after the playback thread has exited (because you called release() on the player). This is working as intended, and the warning can be safely ignored.

It's slightly unfortunate that the platform chooses to print this warning at all, since there are cases such as this one where the behavior is intentional and expected.

dnutcracker:
Yep - it is a little bit annoying.

Can the release() ask  the loading threads to stop?
Can the loading threads check if release() was called before trying to notify the playback thread?

ojw28:
[...]

PR #624, opened 2015-07-15
Fix issue 426 - send messages to dead thread. 

MaTriXy:
issue https://github.com/google/ExoPlayer/issues/426


ojw28:
Please style this code in the same way as the rest of the code base. It's very difficult to review it when virtually every line has changed due to application of a different code style.

MaTriXy:
Sorry - i thought apply the code formatting fixed it.. 
Doing a revert now.

ojw28:
Hm. This doesn't prevent the message in all cases, although probably catches most occurrences. I'll have a think.

MaTriXy:
Hi, 
I can't kill the handler from where we are but i can avoid sending messages when the tread is dead. 
This way messages are not sent when not needed.

thomas-mann:
What about this pull request? I have also issues with sending messages to a dead thread. 
It happens when you're starting multiple streams in a short period of time. For example selecting the same video multiple times in a listview. 
For clearance: Every click releases the current exo player instance and creates a new one with the specific streaming url.

ojw28:
The only issue this causes is a warning in the logs. There are no user facing issues related to this warning. For now, I think we're ok doing nothing until we have time to think about a complete fix.

ojw28:
This is fixed in a more correct way in the [2.x.x experimental branch](https://github.com/google/ExoPlayer/tree/dev-2.X.X-experimental). We're not planning to take a fix for 1.x, since the issue is largely benign.

PR #661, opened 2015-07-27
Upgrade Mockito and dexmaker versions.

### LINKED ISSUE #651: Upgrade mockito version to avoid getting warning for difference in hamcrest-core versions
Since the latest update, I keep getting a warning message "Conflict with dependency 'org.hamcrest:hamcrest-core'. Resolved versions for app (1.1) and test app (1.3) differ."

I noticed that ExoPlayer uses older version of mockito, which uses 1.1 of hamcrest-core where as my app uses the latest version of junit for testing, which uses 1.3 of hamcrest-core.

ened:
Fixes #651.
The third_party directory still exists and should probably be removed (the only reference to it is in a Eclipse .project file - which can be removed as well. ;-)).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,3 +16,8 @@
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
+
+
+JUNIT_VERSION = 4.12
+DEXMAKER_VERSION = 1.4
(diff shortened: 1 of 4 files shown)


ojw28:
See the referenced issue. Upgrading these to match the versions you happen to use will presumably fix things for you and break things for someone else. The dependencies just shouldn't exist at all, I think.

ojw28:
I don't think we have a need to update these versions. The libs are I think still required for people using Eclipse, as is the project file. Please comment if you disagree :). Thanks!

PR #682, opened 2015-08-03
Correctly identify gzipped response in DefaultHttpDataSource

gitanuj:
There's a comment added in DefaultHttpDataSource regarding identification of gzipped response:
"there isn't a reliable way to determine whether the gzip was used or not".

However according to HTTP/1.1 specs, (http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11):
"If the content-coding of an entity is not "identity", then the response MUST include a Content-Encoding entity-header (section 14.11) that lists the non-identity content-coding(s) used."

This change uses 'Content-Encoding' header to determine if content is compressed using gzip.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java
@@ -223,16 +223,15 @@ public class DefaultHttpDataSource implements HttpDataSource {
     bytesToSkip = responseCode == 200 && dataSpec.position != 0 ? dataSpec.position : 0;
 
     // Determine the length of the data to be read, after skipping.
-    if ((dataSpec.flags & DataSpec.FLAG_ALLOW_GZIP) == 0) {
+    if (!"gzip".equals(connection.getContentEncoding())) {
       long contentLength = getContentLength(connection);
       bytesToRead = dataSpec.length != C.LENGTH_UNBOUNDED ? dataSpec.length
           : contentLength != C.LENGTH_UNBOUNDED ? contentLength - bytesToSkip
           : C.LENGTH_UNBOUNDED;
     } else {
       // Gzip is enabled. If the server opts to use gzip then the content length in the response
-      // will be that of the compressed data, which isn't what we want. Furthermore, there isn't a
-      // reliable way to determine whether the gzip was used or not. Always use the dataSpec length
-      // in this case.
+      // will be that of the compressed data, which isn't what we want. Always use the dataSpec
+      // length in this case.
       bytesToRead = dataSpec.length;
     }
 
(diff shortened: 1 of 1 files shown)


ojw28:
I'm pretty sure you can't do this, because okhttp (which is used under HttpUrlConnection on more recent Android versions) will strip the Content-Encoding header when it decompresses the response. See the source code (here)[https://github.com/square/okhttp/blob/master/okhttp%2Fsrc%2Fmain%2Fjava%2Fcom%2Fsquareup%2Fokhttp%2Finternal%2Fhttp%2FHttpEngine.java#L627].

ojw28:
If this is necessary, one option is to explicitly add the Accept-Encoding gzip header ourselves, in which case OkHttp wont transparently decompress the response or strip the header. DefaultHttpDataSource would then be responsible for handling the decompression itself, which would add additional complexity. What problem is this actually fixing, if any? I think we'd rather not do that unless there's an actual issue that requires us to do so.

gitanuj:
In scenarios where DataSpec.FLAG_ALLOW_GZIP is used but the response is not gzipped, we do not read the content length from headers, assuming that the content must be gzipped. This doesn't provide length of the content to the caller, even if we can get actual length. 

In my particular scenario, I have a custom FileDataSink which relies on length provided by open() to run some logic. Since HlsChunkSource.newMediaPlaylistChunk(int) enables gzip by default, I had to write a wrapper around DefaultHttpDataSource to override DataSpec flags to disable gzip (knowing that my server doesn't support gzip).

Also, I can see in the documentation of HttpEngine.unzip(the link that you provided) that they remove 'Content-Length' header as well. So, this change shouldn't provide any false positives.

Although after looking at the implementation of HttpEngine.unzip(Response), I'm thinking that iterating through the list of encodings in 'Content-Encoding' header might be an overkill to determine if the content is gzip encoded.

ojw28:
On that version of okhttp isContentGzipped will always return false. You will then get false positives because the Content-Range header isn't stripped, which getContentLength uses if Content-Length is missing (this is required, I think due to proxies that some mobile carriers use that strip Content-Length). If I remember correctly, there are also some versions of the Android network stack that don't strip Content-Length for this case.

Why don't you modify your custom component to not rely on the content length being provided? In general components should be able to cope with this case anyway (although admittedly, if you know your server's behavior then you can get away without). Also, playlist files are quite amenable to gzip, I think, so I'd guess that you're losing out by not compressing them.

gitanuj:
If we can get false positives by this, it makes sense to not use the length.

I want to write complete file for corresponding DataSpec. I'm using TeeDataSource and using length is one way to ensure that complete data has been written. Is there any other way we can detect if complete data has been written in a DataSink when closing it?

gitanuj:
@ojw28 I have one question regarding DataSource's read method. If the method returns less number of bytes than requested is it safe to assume that EOF has been reached? Or should we only rely on encountering C.RESULT_END_OF_INPUT on subsequent reads?

ojw28:
It's common for DataSource implementations to read fewer bytes than are requested (but each call will always read at least 1 byte, to guarantee progress is made). See the Javadoc for that method for details. So you need to wait until you see END_OF_INPUT. This is the same as how Java InputStream objects work.

Issue #524, opened 2015-06-07
Provide OkHttp variant of HttpDataSource


b95505017:
Hi,
As title, I create this small class. Not sure could merge this into demo module?
https://gist.github.com/b95505017/09b8da221af7f14740b7

Thanks!

b95505017:
Look like this when trying to play HLS sample in demo app
![2015-06-07 5 53 36](https://cloud.githubusercontent.com/assets/320815/8023535/11d6d4da-0d42-11e5-8e02-cdbb85276147.png)

Ood-Tsen:
Just want to know what the tool is ?

b95505017:
You mean Chrome dev tools or Stetho?
Stetho is developed by Facebook
http://facebook.github.io/stetho/
It could easily integrated with OkHttp to monitor all request in Chrome dev tools.

Ood-Tsen:
@b95505017 Cool, thanks for the information.

jpshelley:
+1 this would be really awesome to have.

ojw28:
Is it easy to use stetho with DefaultHttpDataSource as well? It looks like stetho supports urlconnection in some form at least (I didn't look in detail).

b95505017:
Not as easy as using OkHttp
> If you are using HttpURLConnection, you can use StethoURLConnectionManager to assist with integration though you should be aware that there are some caveats with this approach. In particular, you must explicitly add Accept-Encoding: gzip to the request headers and manually handle compressed responses in order for Stetho to report compressed payload sizes.

ojw28:
I don't think we want this in the core library or demo app, since we don't want to mandate that developers include the okhttp network stack in their applications. This is definitely useful as an optional extension though, particularly for developers who already include okhttp in their applications anyway.

If you were to package it up as a standalone extensions project in the [extensions](https://github.com/google/ExoPlayer/tree/dev/extensions) directory, then we'd happily merge it there so that developers who do want it can get access to it.

ojw28:
Note - I've updated the title of this issue to correspond more closely to what it's tracking. This is really tracking providing an okhttp variant of HttpDataSource, and the primary use would be to allow applications to use okhttp as the network stack. The fact that you can debug using Stetho when you're using it is more or less just a (very nice) added bonus!

ojw28:
This is merged in dev. Thanks!

Issue #420, opened 2015-04-27
Generalize/enhance persistent caching functionality


ojw28:
ExoCache is designed specifically for DASH implementations where every request can be in the form of a bounded range request (which in practice means single-segment on-demand streams containing sidx boxes). It doesn't work for anything else (DASH where streams are segmented in the manifest, SmoothStreaming, HLS, ExtractorSampleSource). There are three things to fix:
1. We don't support unbounded range requests. This is because the cache currently has no concept of the length of a piece of data. If we have a file in the cache with byte range [0-1000], and make a request for range [0-*], we currently don't know what to do after we've return the first 1000 bytes from the cache. We don't want to make an unconditional request to the network starting at offset=1000 because it's inefficient, wont work in the offline case, and because the server may return a 416 (unsatisfiable byte-range) in the case that the content really is 1000 bytes long, which would be awkward to handle.
2. It's probably more common for servers that are configured to serve every chunk from a separate URL to not support range requests. We may need the cache to support a "request the whole chunk or nothing from the upstream source" mode for this use case. I'm not sure how we'd decide when to turn it on. Awkward.
3. The cache currently indexes: contentId->Tree[byte-range->file]. For chunks that are requested from different URLs, we need the Tree to be indexed by chunk-index+byte-range, or perhaps just chunk-index depending on the answer to (2).

nzer:
It also should be able to cache mp3, mp4 files distributed through plain http. In my use case app plays media file from server, but URL and CDN server may change everytime file is requested. I think we should use key based caching.

nzer:
I'm making my own cache for now, should I modify DataSource interface or create CacheableDataSource instead?

nzer:
I did some work, please check it. https://github.com/nzer/ExoPlayer/commit/6c3c89073e25b7af97fc7c55e6a7d15d308a9fbc
Idea is when media is being downloaded (MP3, AAC, whatever) its written to disk at same time. If download succeeds to the end, we mark file finished and offer it to datasource to instead of loading it from network.
Key should be designed by app developer. MD5 of artist+title for music files for example.

Arqu:
One question about the indexes (key), why not just use a hashing function of the DataSpec info, like uri+position+length? Would be unique enough for the caching structure and would also avoid having another segments chunk index overlap with some other.

ojw28:
[...]

PR #735, opened 2015-08-23
Add an extension that provide DataSource using OkHttp

b95505017:
#524


b95505017:
Not sure where should I write a demo for it, cause I need to rewrite all render builder in demo app.

ojw28:
I don't think it needs a demo app. It would be trivial for someone to swap it into the normal demo app in a couple of minutes, if they want to experiment with it, so I don't think it needs a demo of its own.

talklittle:
:+1:

Tried this, swapping into the `ExtractorRendererBuilder` in r1.4.2 demo, works great with `ExtractorSampleSource` and mp4 files.

I love that it can integrate with the OkHttp's `Cache` to avoid redownloading the mp4 file when looping. So for me, this serves as a pretty good solution to #420 

Thank you @b95505017 !

b95505017:
@talklittle You could also integrate [Stetho](https://github.com/facebook/stetho) via OkHttp's Interceptor mechanism, then all request could monitor via chrome dev tool.
@ojw28 any review progress update?

ojw28:
We're finalizing a 1.5.x release at the moment. We'll be merging it after that's done. Sorry for the long delay; perhaps we need more branches :). Thanks!

ojw28:
1.5.0 is out, so we're pretty much ready to merge it. Our plan is to merge it as-is, since it looks in pretty good shape already and works well. Just FYI, we then plan to make some follow-up modifications, including:

- Fixing code style to be consistent with the rest of the project.
- Force application to inject OkHttpClient instance through all constructors, and remove allowCrossProtocolRedirect/connectTimeout/readTimeoutMillis parameters. Right now if you inject a client, the OkHttpDataSource will change the default properties of that client, which is confusing given it might be used elsewhere. It's probably simplest just to require the application to instantiate, configure and inject a client as appropriate.

ojw28:
This is now merged, with the modifications described above applied. Please take a look and check that it still needs the original needs. Thanks!

b95505017:
LGTM :+1:  :)

TommyVisic:
@b95505017, thanks for this OkHttp version. @talklittle can you share how you set this up to get caching working with mp4s? Are you working with partial responses at all?

talklittle:
@TommyVisic Nothing fancy at all, just setting a Cache on the OkHttpClient.

b95505017:
@TommyVisic take a look with [OkHttp recipes](https://github.com/square/okhttp/wiki/Recipes#response-caching)

kwelsh:
The ReadMe for the OkHttp Extension doesn't have any build instructions. I'm assuming this is similar to how other extensions are built? Anyone with more experience care to comment on the process of including it in a project? Any info would greatly be appreciated!

Would be nice to have some way of just adding a single `compile` line to your Gradle file add these extensions...

b95505017:
Yes just using `compile ':extension-okhttp'` in your submodule or some another path if you have another settings.gradle.

Just like the usage of [DefaultHttpDataSource.java](https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/upstream/DefaultHttpDataSource.java), if you really need a example them I could provide it. :)

brianchu:
Adding a line to the settings.gradle will get it compile into the build, but we still need to program the app to use okhttp instead default one. There is no magic (like reflection) to pick up the okhttp.

b95505017:
OK I'll try to write an example AOAP.

econnelly:
@talklittle OkHttp doesn't cache partial requests. The OkHttp cach may work on a fast connection where the entire file can be downloaded in one shot, but it's not reliable. I just went through the exact scenario where a looping mp4 was redownloading every time it started over.

There are two ways to prevent this.

1) Download the file somewhere else using OkHttp directly and use `FileDataSource` for playback
2) Modify `OkHttpDataSource` and `CacheDataSource` to produce and accept the final file size before starting playback

talklittle:
@econnelly Hey thanks for pointing that out.

Another way is by forcing OkHttp to request the full file each time using a network interceptor and removing the `Range` header that ExoPlayer uses. Then you'll get 200 responses instead of 206's, which OkHttp's Cache will handle.

I'll update my other comment for posterity.

kwelsh:
Thanks, @talklittle, @econnelly, @b95505017, and @brianchu for the extra info. It's been very insightful.

I feel like I'm getting tripped up in the set up here, as my IDE (Android Studio) doesn't seem to auto-complete OkHttpDataSource. I added `compile 'com.google.android.exoplayer:exoplayer:extension-okhttp'` in my dependencies and received no errors/warnings, but this doesn't seem to work.

Does someone mind sharing their Gradle?

talklittle:
@kwelsh Copy the file into your project. It's not in the Maven artifact on JCenter.

You can verify this in Android Studio in the Project view: External Libraries -> exoplayer-r1.5.2 -> classes.jar. The extensions are not included. The Maven artifact corresponds to the `library` module of this repo only.

b95505017:
Here is my example which replace demo data source with OkHttp, default is no cache.
https://github.com/b95505017/ExoPlayer/tree/okhttp_http_data_source

kwelsh:
@talklittle Ah I see. Thanks for clarifying this.

@b95505017 That is awesome! It seems a lot simpler than I envisioned in my head. I'm gonna give this another go.

kwelsh:
I got this thing up and running. Just need to configure it to cache now. Thanks again, everyone.

laurencedawson:
@kwelsh just a heads up, partial requests are not cached. 

If you modify the method makeRequest in OkHttpDataSource you can get caching to work correctly.

welshk91:
@laurencedawson Thanks for the heads up.

I'm still struggling with getting caching to work on the demo for ExoPlayer. I've added in OkHttp to handle the networking requests, but it still seems to re-download the video when you rotate the screen or power off (anything that recreates the fragment again).

I'd be extremely interested in your solution, if just to see a different perspective on the matter! :)

b95505017:
You could inject your own CacheControl (or just give it `null`) into OkHttpDataSource, or using Interceptor skill to monitor/override each request & response.

b95505017:
I'v upgrade the extension to supporting okhttp3 #1119 
One thing should notice that OkHttp now using its own `CookieJar` to handle cookie stuff, checkout [JavaNetCookieJar](https://github.com/square/okhttp/blob/master/okhttp-urlconnection/src/main/java/okhttp3/JavaNetCookieJar.java) to pass original `java.net.CookieHandler` into it.

laminsk:
Thank you b95505017 and everyone for valuable inputs. I've been trying to download audio stream from server to disk using okhttpdatasource and still no luck so far. Has anyone manage to do this yet? Any help would be greatly appreciated. Thanks in advance.

b95505017:
What do you mean about "no luck"?

ceetah:
Hi

I want to cache an mp4.
I also wrote about it here: 
https://github.com/google/ExoPlayer/issues/420#issuecomment-185734470
it seems that in OkHttpDataSource on makeRequest the DataSpec has an unbounded length and this makes it to download the file every time because it cannot cache it.
How can I cache it ? Is there a way?

ceetah:
@talklittle so you made cache work for mp4s ?

PR #776, opened 2015-09-09
Ignore extra returns in Subrip parsing

IanDBird:
I was testing SRT support when I found that a number of my test files didn't parse. It turned out the reason was that they contained additional empty lines before the next numeric counter. I don't think this is uncommon, and fairly easy to handle.

Here's an example of an SRT which includes this:

```
1
00:00:34,160 --> 00:00:35,366
Start simple.


2
00:00:35,440 --> 00:00:36,965
Start with what you know is true.
```

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
@@ -58,6 +58,10 @@ public final class SubripParser implements SubtitleParser {
     String currentLine;
 
     while ((currentLine = reader.readLine()) != null) {
+      // Skip blank lines.
+      if (currentLine.length() == 0)
+        continue;
+
       // Parse the numeric counter as a sanity check.
       try {
         Integer.parseInt(currentLine);


library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java
```diff
@@ -58,6 +58,10 @@ public SubripSubtitle parse(InputStream inputStream) throws IOException {
     String currentLine;
 
     while ((currentLine = reader.readLine()) != null) {
+      // Skip blank lines.
+      if (currentLine.length() == 0)
```
IanDBird: This could have used `TextUtils.isEmpty` but since we already know it's not null, I thought simplest to just check it's length directly.

ojw28:
Change looks fine, thanks! You need to sign the CLA before we can merge it though (see https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md).

IanDBird:
@ojw28 - I've signed that before, since i've already contributed: https://github.com/google/ExoPlayer/commit/646b39622fbfa4fb9308c0094658113e808f8fc0

ojw28:
Ah, thanks! The source I have doesn't have your GitHub name (only email address) so it wasn't coming up in my search :).

PR #781, opened 2015-09-10
Explicitly state 16 bits per sample for media format

IanDBird:
I acknowledge this might be a hard sell, but... :)

I have been testing the E/AC3 decoding of audio out on number of devices. One of these is a OnePlus One running CM12 (I know). The resulting audio was static. After a bit of digging, I found that CM12 supports 24bit audio which it attempts to default to. I'm not suggesting at all that ExoPlayer might want to support this. However, this change makes it explicit what the expected MediaFormat should be and therefore "fixes" the compatibility issue here. Arguably, I did have to read through the CM12 source code to work this out... but if there was ever a "fix" for this, this one seems the most sensible.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
@@ -258,6 +258,7 @@ public final class MediaFormat {
       maybeSetIntegerV16(format, android.media.MediaFormat.KEY_MAX_HEIGHT, maxHeight);
       maybeSetIntegerV16(format, android.media.MediaFormat.KEY_CHANNEL_COUNT, channelCount);
       maybeSetIntegerV16(format, android.media.MediaFormat.KEY_SAMPLE_RATE, sampleRate);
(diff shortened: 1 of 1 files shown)


IanDBird:
@ojw28 - Any thoughts on this PR? Is it under consideration or been rejected?

ojw28:
Why does CM12 try and "default" to 24-bit audio? That's just wrong/broken, and they'll need to fix it if they expect applications using AudioTrack to work properly? I'd suggest filing a bug on CM12 instead?

IanDBird:
It's a good point, and I can try to get more information from them directly...

However, I don't think it's feasible to expect a fix from CM, and to then be merged to all the other variants based off it, and then updated on the significant amount of devices released out in the field which use it. This (what I thought) simple change makes things compatible rather than waiting the (potentially) months before users would start getting any updates.

As I mentioned in my first comment, I thought it would be a hard sell. So am happy to just keep this change in our own application fork of ExoPlayer.

PR #786, opened 2015-09-11
Fixed issue with WebVTT percentage values

jeoliva:
Current WebVTT parser is assuming percentages are always integer values.
However, following WebVTT specification percentage values format is as
follows (http://dev.w3.org/html5/webvtt/#webvtt-file-structure):

“
A WebVTT percentage consists of the following components:

1.- One or more ASCII digits.
2.- Optionally:
1.- A U+002E DOT character (.).
2.- One or more ASCII digits.
3.- A U+0025 PERCENT SIGN character (%).
”

This change modifies the way the percentage values are checked to ensure
parser takes into account the format defined in the specification.

Note: While checking format I am keeping check of existence of “%”
character and check of number format separated (instead of using the regex for checking all) just for keeping as
detailed logs as we have today (log message advising about lack of “%”
character; log message advising about wrong number format).

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
@@ -60,6 +60,8 @@ public final class WebvttParser implements SubtitleParser {
   private static final String WEBVTT_CUE_SETTING_STRING = "\\S*:\\S*";
   private static final Pattern WEBVTT_CUE_SETTING = Pattern.compile(WEBVTT_CUE_SETTING_STRING);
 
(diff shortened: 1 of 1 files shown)


ojw28:
Hm, I think there are slightly bigger problems here, because "line" and "position" can both have an optional alignment suffix.

jeoliva:
Yeah, you are right. In fact, that's not the only "parsing issue". There are other "params" that are more complex than what current webvtt parser is assuming.  I am going to take a look at this.

jeoliva:
@ojw28, closing this PR as I am preparing a more complete one with the WebVTT "features" I have seen frequently used.

PR #795, opened 2015-09-15
Improve WebVTT parser according to WebVTT spec

jeoliva:
- Line parameter
  - Added support for value and line alignment attributes.
  - Support negative numbers when line is an absolute number (not a
percentage).
- Position parameter
  - Added support for value and position alignment attributes
- Added support for WebVTT comment blocks
- Percentage values now accept decimal numbers (as webvtt spec states)
- Added new WebVTT tests for testing all new implemented features

This PR is focused on improving the parser. Later work is needed on improving WebVTT cue rendering process to use new parsed parameters (line alignment, position alignment, line as percentage, etc)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
@@ -24,6 +24,7 @@ import com.google.android.exoplayer.util.MimeTypes;
 import android.text.Html;
 import android.text.Layout.Alignment;
 import android.util.Log;
+import android.util.Pair;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -42,9 +43,13 @@ public final class WebvttParser implements SubtitleParser {
 
   private static final String TAG = "WebvttParser";
 
-  private static final String WEBVTT_FILE_HEADER_STRING = "^\uFEFF?WEBVTT((\\u0020|\u0009).*)?$";
+  private static final String WEBVTT_FILE_HEADER_STRING = "^\uFEFF?WEBVTT((\u0020|\u0009).*)?$";
   private static final Pattern WEBVTT_FILE_HEADER =
       Pattern.compile(WEBVTT_FILE_HEADER_STRING);
+
+  private static final String WEBVTT_COMMENT_BLOCK_STRING = "^NOTE((\u0020|\u0009).*)?$";
+  private static final Pattern WEBVTT_COMMENT_BLOCK =
+      Pattern.compile(WEBVTT_COMMENT_BLOCK_STRING);
 
   private static final String WEBVTT_METADATA_HEADER_STRING = "\\S*[:=]\\S*";
   private static final Pattern WEBVTT_METADATA_HEADER =
@@ -59,6 +64,8 @@ public final class WebvttParser implements SubtitleParser {
 
   private static final String WEBVTT_CUE_SETTING_STRING = "\\S*:\\S*";
   private static final Pattern WEBVTT_CUE_SETTING = Pattern.compile(WEBVTT_CUE_SETTING_STRING);
+
+  private static final String WEBVTT_PERCENTAGE_NUMBER_STRING = "^([0-9]+|[0-9]+\\.[0-9]+)$";
 
   private static final String NON_NUMERIC_STRING = ".*[^0-9].*";
 
@@ -118,9 +125,18 @@ public final class WebvttParser implements SubtitleParser {
 
     // process the cues and text
     while ((line = webvttData.readLine()) != null) {
+      // parse webvtt comment block in case it is present
+      Matcher matcher = WEBVTT_COMMENT_BLOCK.matcher(line);
+      if(matcher.find()) {
+        // read lines until finding an empty one (webvtt line terminator: CRLF, or LF or CR)
+        while (((line = webvttData.readLine()) != null) && (!line.isEmpty())) {
+          // just ignoring comment text
+        }
+        continue;
+      }
 
       // parse the cue identifier (if present) {
-      Matcher matcher = WEBVTT_CUE_IDENTIFIER.matcher(line);
+      matcher = WEBVTT_CUE_IDENTIFIER.matcher(line);
       if (matcher.find()) {
         // ignore the identifier (we currently don't use it) and read the next line
         line = webvttData.readLine();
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttCue.java
```diff
@@ -26,20 +26,33 @@
 
   public final long startTime;
   public final long endTime;
+  public final Alignment lineAlignment;
```
ojw28: Can we please leave this class alone for the purposes of this change? It's not clear that adding more things here is the right thing to do with the new parameters that you're parsing. It's quite likely they'll be used to modify the existing line/position/alignment values that get passed to the super class).

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
```diff
@@ -46,6 +47,10 @@
   private static final Pattern WEBVTT_FILE_HEADER =
       Pattern.compile(WEBVTT_FILE_HEADER_STRING);
 
+  private static final String WEBVTT_COMMENT_BLOCK_STRING = "^NOTE((\\u0020|\u0009).*)?$";
```
ojw28: "^NOTE\\s" should be sufficient? In practice "^NOTE" is also fine, and probably preferable in case anyone forgets the space/tab/newline.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
```diff
@@ -60,6 +65,8 @@
   private static final String WEBVTT_CUE_SETTING_STRING = "\\S*:\\S*";
   private static final Pattern WEBVTT_CUE_SETTING = Pattern.compile(WEBVTT_CUE_SETTING_STRING);
 
+  private static final String WEBVTT_PERCENTAGE_NUMBER_STRING = "^([0-9]+|[0-9]+\\.[0-9]+)$";
```
ojw28: You can use ```\\d+``` instead of ```[0-9]+```, I think?

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
```diff
@@ -164,29 +182,32 @@ public final WebvttSubtitle parse(InputStream inputStream) throws IOException {
...
+            value = lineMetadata.first;
+            lineAlignment = lineMetadata.second;
             if (value.endsWith("%")) {
               lineNum = parseIntPercentage(value);
-            } else if (value.matches(NON_NUMERIC_STRING)) {
-              Log.w(TAG, "Invalid line value: " + value);
             } else {
-              lineNum = Integer.parseInt(value);
+              // Following WebVTT spec, line number can be a negative number
+              int sign = 1;
```
ojw28: Integer.parseInt can parse the string including the negative sign, so you don't need to do this manually. You just need to make sure the ```value.matches(NON_NUMERIC_STRING)``` line doesn't trigger for negative numeric values.

jeoliva: Yeah, you are right, Integer.parseInt can parse the negative sign. However, doing the check using the regular expression add extra complexity to the regular expression (create a separated regex for line numbers apart from the one used for checking position numbers; the new regex should check string doesn't include numbers nor an initial hyphen character). 

Personally I thought using a new and modified regex for this could make code looks less clean.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
jeoliva: Sure. [0-9] was used for consistency with all the other regex of this same file but of course it can be replaced by \\d

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
jeoliva: Accordingly to WebVTT spec:

A WebVTT comment block consists of the following components, in the given order:
 - The string "NOTE".
- Optionally, the following components, in the given order:
  Either:
   - A U+0020 SPACE character or U+0009 CHARACTER TABULATION (tab) character.
   - A WebVTT line terminator.

So:
 1.- "^NOTE" wouldn't be valid because a separation character (space, tab or new line) is required between NOTE and the comment.
 2.- "^NOTE\\s" strictly speaking is not valid because "\s" includes characters out of space and tab characters although, in the practice, I think it could be valid. It would be weird having someone using a vertical tab or a form feed (included by "\s") wrongly and unintentionally for separating NOTE keyword from the comment text.

Regarding point 2, and for consistency, we can do the same for WEBVTT regex:
 - From: "^\uFEFF?WEBVTT((\u0020|\u0009).*)?$"
 - To: "^\uFEFF?WEBVTT\\s"

ojw28: Heh, sorry, guess it's fine to have it how you have it already (given it's consistent with the existing WEBVTT regex). 

One question: Do you understand why The u0020 has two backslashes before it where-as the u0009 one only has one? I think I wrote that code, but it looks like a typo? Should both have only a single backslash?

jeoliva: In fact, even without talking strictly, "^NOTE\s" is not valid. 

For example, given the following text block:

NOTE
This is a sample comment

As the string returned by readLine method of BufferedReader doesn't include the newline sequence, "^NOTE\s" wouldn't be correct to validate the previous example comment.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
ojw28: Fine to leave this one too, although wouldn't it be clearer and match the spec more directly if it were like: ```"^([0-9]+(\\.[0-9]+))$```

ojw28: (NB - Edited the comment above to add a question mark!)

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
jeoliva: You are right, I think it was a typo + copy&paste error in my side. Both should have just one single backslash. I am applying the fix in my branch and will add it to the PR.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
ojw28: Makes sense.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttCue.java
jeoliva: Totally agree

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttCue.java
```diff
@@ -36,7 +36,7 @@ public WebvttCue(long startTime, long endTime, CharSequence text) {
   }
 
   public WebvttCue(long startTime, long endTime, CharSequence text, int line, int position,
-      Alignment alignment, int size) {
+                   Alignment alignment, int size) {
```
ojw28: Fully revert this file to avoid git history pollution :)? (see also rm'd newline at bottom).

jeoliva: Sure, done! :)

ojw28:
Thanks!

ojw28:
I'm taking a quick look at further cleaning this up. Since you've been looking at the spec, I was wondering if I could pick your brains about this section:

<i>
This example shows how to position cues at explicit positions in the video viewport.

WEBVTT

00:00:00.000 --> 00:00:04.000 position:10%,start align:start size:35%
Where did he go?

00:00:03.000 --> 00:00:06.500 position:90% align:end size:35%
I think he went down this lane.

00:00:04.000 --> 00:00:06.500 position:45%,end align:middle size:35%
What are you waiting for?
The cues cover only 35% of the video viewport's width. The first cue has its cue box left aligned at the 10% mark of the video viewport width and the text is left aligned within that box - probably underneath a speaker on the left of the video image. "start" alignment of the cue box is the default for start aligned text, so does not need to be specified in "position". <b>The second cue has its cue box right aligned at the 90% mark of the video viewport width. The same effect can be achieved with "position:55%,start", which explicitly positions the cue box</b>. The third cue has middle aligned text within the same type of cue box as the first cue.
</i>

The text in bold looks wrong to me. Doesn't "align" apply within the cue box? For the bold text to be correct, shouldn't the example be:

<i>
00:00:03.000 --> 00:00:06.500 position:90%,end size:35%
</i>

instead?

ojw28:
Ignore me. The "computed position alignment" inherits from "align", so this does actually make sense.

jeoliva:
:). Coincidentally I had the same "internal" discussion yesterday. Example is not very clear for point 2.

ojw28:
FYI - I have a change that (hopefully) further cleans up the parser, propagates the anchors and applies them along with the width in CuePainter.. Will merge next week.

jeoliva:
Good to know as I am also working on parser and CuePainter adding support for vertical alignment, line position alignment, position alignment and some other rendering changes. 

I am going to put my changes on hold until you apply the merge.

Thanks!

ojw28:
Draft change is [here](https://github.com/google/ExoPlayer/commit/6a9c88358c54afe56847290e25e16b81e10d0bd3). Any comments appreciated (just dump them directly on the change, if you have any!). Thanks.

Issue #748, opened 2015-08-31
I found a strange bug may be due to proguard.


AlanCheen:
I found my app can't play some videos well when i proguarded my app but play the same video well when debug.

At first i think it's Color OS's bug,cause it happens on ColorOS only but when i build a debug app it plays well.

Then, i tried the exoplayer demo ,but nothing changed.

So i think may be it's the proguard affect the exoplayer? I can't find any other reason.  

e.g. URL : `http://s.5dktv.com/hls/f4/30100127_72_1.m3u8`

The log when can't play:
```
08-31 10:07:24.297  23357-23357/? D/EventLogger﹕ start [0]
08-31 10:07:24.297  23357-23357/? D/EventLogger﹕ state [0.00, false, P]
08-31 10:07:24.307  23357-23357/? D/EventLogger﹕ state [0.01, true, P]
08-31 10:07:27.841  23357-23357/? D/EventLogger﹕ state [3.54, true, B]
08-31 10:07:27.851  23357-23357/? D/EventLogger﹕ videoFormat [3.55, 0, 0]
08-31 10:07:28.001  23357-23357/? D/EventLogger﹕ decoderInitialized [3.70, OMX.qcom.video.decoder.avc]
08-31 10:07:28.262  23357-23357/? D/EventLogger﹕ videoSizeChanged [320, 568, 1.0]
08-31 10:07:29.673  23357-23357/? D/EventLogger﹕ state [5.38, true, R]
08-31 10:07:29.893  23357-23357/? E/EventLogger﹕ internalError [5.59, loadError]
    com.google.android.exoplayer.i.u$d: Unexpected ArrayIndexOutOfBoundsException: length=9; index=9
            at com.google.android.exoplayer.i.u$b.run(Loader.java:222)
            at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:390)
            at java.util.concurrent.FutureTask.run(FutureTask.java:234)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
            at java.lang.Thread.run(Thread.java:856)
     Caused by: java.lang.ArrayIndexOutOfBoundsException: length=9; index=9
            at com.google.android.exoplayer.j.n.c(ParsableBitArray.java:147)
            at com.google.android.exoplayer.e.c.k$b.b(TsExtractor.java:552)
[...]

PR #796, opened 2015-09-16
Prevent ArrayIndexOutOfBoundsException when proguard is enabled

souji1103:
When I proguraded my app which play HLS movies, surely I encountered ArrayIndexOutOfBoundsException #748 by using 303SH(4.2.2), SH-01F(4.2.2), F-01F(4.2.2) and other 4.2.2 devices.

I logged ParsableBitArray.readBits as below:
```
  public int readBits(int n) {

    Log.d("point1", String.valueOf(n) + " : " + String.valueOf(byteOffset));
    
    if (n == 0) {
      return 0;
    }

    int returnValue = 0;

    // While n >= 8, read whole bytes.
    while (n >= 8) {
      int byteValue;
      if (bitOffset != 0) {
        byteValue = ((data[byteOffset] & 0xFF) << bitOffset)
            | ((data[byteOffset + 1] & 0xFF) >>> (8 - bitOffset));
      } else {
        byteValue = data[byteOffset];
      }
      n -= 8;
      returnValue |= (byteValue & 0xFF) << n;
      byteOffset++;
    }

    Log.d("point2", String.valueOf(n) + " : " + String.valueOf(byteOffset));
```
At the crash case, unexpected values were logged.

| point | n | byteOffset |
|---|---|---|
| 1 | 16 | 4 |
| 2 | -8 | 7 |

As the result, TSExtractor read the 10th byte of PES header only holding 9 bytes.

Though I don't accurately know what happens, as perhaps the variable "n" used as loop condition is updated within the loop, some runtime environments execute this loop in parallel at the code optimized by proguard.

So I define the number of loops before that loop, and the exception isn't thrown when proguard is enabled.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/util/ParsableBitArray.java
+++ b/library/src/main/java/com/google/android/exoplayer/util/ParsableBitArray.java
@@ -138,7 +138,8 @@ public final class ParsableBitArray {
     int returnValue = 0;
 
     // While n >= 8, read whole bytes.
-    while (n >= 8) {
+    int numBytes = (n / 8);
+    for (int i = 0; i < numBytes; i ++) {
       int byteValue;
       if (bitOffset != 0) {
         byteValue = ((data[byteOffset] & 0xFF) << bitOffset)


ojw28:
Thanks for investigating this! My guess would be that proguard is making a valid optimisation, but that the resulting optimised code is then erroneously further optimised by Dalvik on 4.2.2 (I think Dalvik must be involved, otherwise I don't think the issue would be platform version specific).

Please could you sign the CLA so that we can merge this request. See [here](https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md) for details.

souji1103:
@ojw28 Thanks for your comment. I also decompile apk for investigating that codes, but don't know why only 4.2.2 (4.3 too?) mistake.
So as you indicated, further optimization would cause that error.

Submitted the CLA, please check that. (thanks for pointing to the link)

ojw28:
Hi. Apologies, you need to select "Any other Google project" rather than "Android OS" when submitting the CLA, since ExoPlayer is not part of the OS. Could you submit again? Thanks!

We should probably clarify this in the Contributing file :).

souji1103:
@ojw28 Sorry for my mistake... I agreed the CLA. Please check it again.

ojw28:
Thanks!

PR #797, opened 2015-09-16
Allow demo app to respond to ACTION_VIEW intents

ened:
That way, the app can respond to any sort of playback request (especially for offline files or gallery intents).

### DIFF
--- a/demo/src/main/AndroidManifest.xml
+++ b/demo/src/main/AndroidManifest.xml
@@ -43,7 +43,42 @@
     <activity android:name="com.google.android.exoplayer.demo.PlayerActivity"
         android:configChanges="keyboardHidden|orientation|screenSize"
         android:label="@string/application_name"
-        android:theme="@style/PlayerTheme"/>
+        android:theme="@style/PlayerTheme">
+
+        <intent-filter>
+            <action android:name="android.intent.action.VIEW"/>
+
+            <category android:name="android.intent.category.DEFAULT"/>
+            <category android:name="android.intent.category.BROWSABLE"/>
+
+            <data android:mimeType="video/*"/>
+        </intent-filter>
+
+        <intent-filter>
+            <action android:name="android.intent.action.VIEW"/>
+
+            <category android:name="android.intent.category.DEFAULT"/>
+            <category android:name="android.intent.category.BROWSABLE"/>
+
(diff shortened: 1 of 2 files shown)


demo/src/main/AndroidManifest.xml
```diff
@@ -43,7 +43,42 @@
...
+            <data android:mimeType="video/*"/>
+        </intent-filter>
+
+        <intent-filter>
+            <action android:name="android.intent.action.VIEW"/>
+
+            <category android:name="android.intent.category.DEFAULT"/>
+            <category android:name="android.intent.category.BROWSABLE"/>
+
+            <data android:scheme="rtsp"/>
```
ojw28: ExoPlayer doesn't support rtsp.

demo/src/main/AndroidManifest.xml
```diff
@@ -43,7 +43,42 @@
...
+        </intent-filter>
+
+        <intent-filter>
+            <!-- HTTP live support -->
```
ojw28: Given TYPE_OTHER is being used, would this actually work? You'd need the type to end up being TYPE_HLS. Also, why not support asset, file and content as well as http and https?

demo/src/main/AndroidManifest.xml
```diff
@@ -43,7 +43,42 @@
...
-        android:theme="@style/PlayerTheme"/>
+        android:theme="@style/PlayerTheme">
+
+        <intent-filter>
+            <action android:name="android.intent.action.VIEW"/>
+
+            <category android:name="android.intent.category.DEFAULT"/>
+            <category android:name="android.intent.category.BROWSABLE"/>
+
+            <data android:mimeType="video/*"/>
```
ojw28: - Why not also audio/*
- Should this be restricted to schemes http/https/file/asset/content?

ojw28:
We added some intent support in https://github.com/google/ExoPlayer/commit/bcb4ea4f703e667e910d8b4016ddab67d11af2eb. Note that we require a custom action, so this is aimed toward being able to easily start a playback with an adb command, as opposed to having us actually appear as a playback option in the disambiguate UI when, for example, selecting a video from a gallery. We believe that this is suitable for most developer needs, and so I'll close this request.

PR #798, opened 2015-09-16
Fixed crash when having different startNumber values in a DASH manifest

miquelbeltran:
The crash happens when we have a dash stream that has different startNumbers on each Representation SegmentTemplate as this example shows.

The sequenceNumber - startNumber is a negative number that causes a crash.

Setting the sequenceNumber - startNumber to 0 (by setting the sequenceNumber = startNumber) avoids the crash and the playback continues uninterrupted.

```xml
<AdaptationSet mimeType="video/mp4" segmentAlignment="true" subsegmentAlignment="true" startWithSAP="1" subsegmentStartsWithSAP="1" bitstreamSwitching="true">
      <Representation id="1" width="960" height="540" frameRate="30/1" bandwidth="1499968" codecs="avc1.4D401F">
        <SegmentTemplate timescale="90000" media="_____" initialization="_____" startNumber="48785" presentationTimeOffset="774314720">
          <SegmentTimeline>
            <S t="774314720" d="900000" r="125"/>
            <S t="887714720" d="204000"/>
          </SegmentTimeline>
        </SegmentTemplate>
      </Representation>
      <Representation id="2" width="640" height="360" frameRate="30/1" bandwidth="899968" codecs="avc1.4D401F">
        <SegmentTemplate timescale="90000" media="_____" initialization="_____" startNumber="48780" presentationTimeOffset="774314720">
          <SegmentTimeline>
            <S t="774314720" d="900000" r="125"/>
            <S t="887714720" d="204000"/>
          </SegmentTimeline>
        </SegmentTemplate>
      </Representation>
      <Representation id="3" width="480" height="270" frameRate="30/1" bandwidth="499968" codecs="avc1.4D401F">
        <SegmentTemplate timescale="90000" media="_____" initialization="____" startNumber="48782" presentationTimeOffset="774314720">
          <SegmentTimeline>
            <S t="774314720" d="900000" r="125"/>
            <S t="887714720" d="180000"/>
          </SegmentTimeline>
        </SegmentTemplate>
      </Representation>
      <Representation id="4" width="320" height="180" frameRate="30/1" bandwidth="299968" codecs="avc1.4D400D">
        <SegmentTemplate timescale="90000" media="_____" initialization="____" startNumber="48782" presentationTimeOffset="774314720">
          <SegmentTimeline>
            <S t="774314720" d="900000" r="125"/>
            <S t="887714720" d="180000"/>
          </SegmentTimeline>
        </SegmentTemplate>
      </Representation>
    </AdaptationSet>
    <AdaptationSet mimeType="audio/mp4" segmentAlignment="0">
      <Representation id="5" bandwidth="64034" audioSamplingRate="48000" codecs="mp4a.40.2">
        <SegmentTemplate timescale="90000" media="_____" initialization="_____" startNumber="48785" presentationTimeOffset="774315680">
          <SegmentTimeline>
```

(filenames removed for confidentiality)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/dash/mpd/SegmentBase.java
+++ b/library/src/main/java/com/google/android/exoplayer/dash/mpd/SegmentBase.java
@@ -175,6 +175,7 @@ public abstract class SegmentBase {
      */
     public final long getSegmentDurationUs(int sequenceNumber, long periodDurationUs) {
       if (segmentTimeline != null) {
+        sequenceNumber = sequenceNumber < startNumber ? startNumber : sequenceNumber;
         long duration = segmentTimeline.get(sequenceNumber - startNumber).duration;
         return (duration * C.MICROS_PER_SECOND) / timescale;
       } else {
@@ -188,6 +189,7 @@ public abstract class SegmentBase {
      * @see DashSegmentIndex#getTimeUs(int)
      */
     public final long getSegmentTimeUs(int sequenceNumber) {
+      sequenceNumber = sequenceNumber < startNumber ? startNumber : sequenceNumber;
       long unscaledSegmentTime;
       if (segmentTimeline != null) {
         unscaledSegmentTime = segmentTimeline.get(sequenceNumber - startNumber).startTime
@@ -328,6 +330,7 @@ public abstract class SegmentBase {
     @Override
     public RangedUri getSegmentUrl(Representation representation, int sequenceNumber) {
       long time = 0;
+      sequenceNumber = sequenceNumber < startNumber ? startNumber : sequenceNumber;
       if (segmentTimeline != null) {
         time = segmentTimeline.get(sequenceNumber - startNumber).startTime;
       } else {


ojw28:
You might be avoiding crashes, but that doesn't look right to me. ExoPlayer relies on segment indices being aligned across different representations within each adaptation set. Specifically, we calculate the next segment index as the previous segment index plus one, so if they're not aligned then you'll end up skipping or repeating media whenever a quality change occurs.

Why are the start numbers different between the different representations? That seems very odd to me.

miquelbeltran:
Hi ojw28, I see your point. This seems to be part of a live stream if that makes more sense. I'm asking for more information to the content provider.

A different way to support this would be that ExoPlayer looks for the greatest startNumber within the SegmentTemplates and the sequenceNumber starts with it (in this case, it would be 48785).

A second option would be that each representation keeps its own sequenceNumber, instead of using a common sequenceNumber across representations which is what I think is happening now.

Seems that there's still some issues when playing the content, so feel free to close the pull request, but let me know what you think would be the correct solution here.

miquelbeltran:
Hi ojw28, I consulted with my colleagues and also think that having different startNumbers is spec compliant.

    Start Number: specifies the number of the first Media Segment in this Representation in the Period. 

So they can be different and not aligned.

However the proposed solution in this pull request is incorrect, so I'll close it.

Debugging a bit more, I've found that in

    public final void getChunkOperation(List<? extends MediaChunk> queue, long seekPositionUs,

To calculate the segmentNumber is using

    queue.get(out.queueSize - 1).chunkIndex + 1;

And this queue corresponds to the representation with startNumber="48782" (format id = 3), but the current representationHolder belongs to the representation with startNumber="48785" (format id = 1). Which causes the issue.

This would be wrong when the different representations follow a different enumeration.

I've found a better solution which is checking if the format changed, and then "normalizing" the segmentNum. This avoids any dropped frames. I'm creating a new pull request with that.

```
if (queue.get(out.queueSize - 1).format.id != selectedFormat.id) {
          int queueStartNum = periodHolder.representationHolders.get(queue.get(out.queueSize -1).format.id).getFirstAvailableSegmentNum();
          int formatStartNum = periodHolder.representationHolders.get(selectedFormat.id).getFirstAvailableSegmentNum();
          segmentNum += formatStartNum - queueStartNum;
        }
```

ojw28:
It might be spec compliant, but given every other DASH stream I've ever seen ensures consistent numbering across representations, I doubt we'll be supporting it. There's no benefit to a server having mismatched sequence numbers across representations that I can think of, so it doesn't make sense to make clients more complicated to handle that possibility.

The spec in general is not specific enough, in my opinion (hence the existence of things like the DASH interop recommendation docs, that try to further constrain what's actually allowed).

ojw28:
Actually, the sample manifest is in violation of the base DASH spec. It indicates ```segmentAlignment="true"```. The specification says:

```
when not set to ‘false’, this specifies that for any two Representations, X and Y, within the same Adaptation Set, the m-th Segment of X and the n-th Segment of Y are non-overlapping (as defined in 4.5.2) whenever m is not equal to n. 
```

which is violated unless the indices are properly aligned. This means that you'd need to set `````segmentAlignment="false"``` to allow misaligned indices. We don't support this case, because playback where segments are not aligned is in the general case significantly more complicated and switching is less efficient. Also, the DASH-IF Interoperability points require that this parameter is true, and the base profiles in the DASH spec indicate that non-aligned adaptation sets may be ignored.

miquelbeltran:
Thanks you very much for the detailed explanation.

I will keep support for this on my codebase because I need it for a project but I understand you won't like it.

An alternative I know other players are doing is starting by the first common Segment number (in this case 48785).

ojw28:
Is it not possible to fix whatever source is generating these manifests, given they're invalid ;)?

miquelbeltran:
We still have a discussion internally, we believe the content is correct and ExoPlayer is not.

We think that segmentAlignment does not mean that the $Number$ need to be the same.

Following your example, when `segmentAlignment=true` m and n are aligned and cover the same time span. But it does not mean that they have to share the same segment number.

The SegmentList@startNumber is just a way to generate the URLs and has nothing to do with segmentAlignment.

Checking the chapter 5.3.9.5.3 Media Segment information

The problem is that ExoPlayer is calculating the $Number$ in a different way that the specification. 

The specification says to use (k + 1) + startNumber to generate the $Number$. It does not matter if the startNumbers are different, what matters is that the position k is the same for each representation.

In ExoPlayer the formula is different. ExoPlayer is using the "old Number" + 1 - startNumber to get the representation URL in an array that goes from 0 to N. The problem is that this "old number" comes from a Representation which has a different startNumber. So it only works when all the startNumber are the same.

ojw28:
In the part of the specification I quoted, m and n are segment numbers. It's quite a stretch to argue that startNumber values are completely unrelated to segment numbers, and that startNumber values are used only for generation or URLs. In fact, this claim is disproven in A.3.3 of the specification, which clearly shows startNumber as being used for indexing in addition to URL generation.

Also, from a "what would be useful" point of view, requiring proper alignment is very useful for efficient playback, where-as not requiring it is not only not useful, but also has no other obvious benefit. It should be trivial for a server to properly align the indices across different representations. Given this, why would it make sense for a specification to define the latter? And why have all other DASH serving infrastructures that I'm aware of chosen to align them properly ;)? To ask the same question again, why don't you just fix the source that's generating the manifests?

ojw28:
Note also that startNumber is defined as:

<i>specifies the number of the first Media Segment in this Representation in the Period.</i>

So you're basically stating that you believe m and n to not be segment numbers, or that they're somehow different numbers to the numbers that this part of the specification refers to. Seems highly unlikely.

ojw28:
Annoyingly A.3.3 has been completely replaced in the 2014 spec (I was looking at 2012).

PR #803, opened 2015-09-17
Normalize the segmentNum based on the different startNumber between the representations

miquelbeltran:
Based on the discussion in https://github.com/google/ExoPlayer/pull/798

Different representations can have different startNumber values.

Using the segmentNum from a representation to get the next segments from another representation may be wrong when they have different startNumbers.

This commit fixes the issue by normalizing the segmentNum by adding the difference between the new and the previous representations. This will only be done when switching from one representation to another.


ojw28:
Please see  #798.

PR #807, opened 2015-09-18
Parse audioChannels only when on mpeg-dash

miquelbeltran:
Only parse the `value` as Int when the `schemeIdUri` is `urn:mpeg:dash:23003:3:audio_channel_configuration:2011`

For example this stream contains Dolby audio tracks which follow a different scheme: http://demo.cf.castlabs.com/media/TOS/abr/Manifest.mpd

Example:

`<AudioChannelConfiguration schemeIdUri="urn:dolby:dash:audio_channel_configuration:2011" value="A000"/>`

For that case the `audioChannel` is set to -1, as the Format class expects when it is not applicable.


ojw28:
Thanks!

PR #814, opened 2015-09-23
Stability when playing HLS

SebastienChauvin:
Hi,

This is our first pull request for the project. We use the ExoPlayer in production for our apps : ch.rts.player , ch.srf.mobile.srfplayer and others
The CLA should be signed ( Radio Télévision Suisse , rts-developers@googlegroups.com )

We encountered a few issues while using the player:
- some AAC decoding issues in our stream where causing the player to crash without reporting the exception devices < API 21. This is solved by the IllegalStateException commit ( e77a2a1 )
- A rare codec null race condition ( 3921919 )
- One big issue that happens very rarely ( about every 20 minutes of playing an HLS stream ) where the SampleHolder has a mismatch between declared size and data field actual size. The only safe option I could think of was to be "more OO" and encapsulate the size and data fields. SampleHolder is now fully responsible for reallocs ( 482b045 ) 

Let me know if you have any feedback on any of those. They seem to work fine for us in our HLS streams.

For your information, we are also working on HLS DVR, but we can discuss this later!

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/EventLogger.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/EventLogger.java
@@ -150,6 +150,11 @@ public class EventLogger implements DemoPlayer.Listener, DemoPlayer.InfoListener
   }
 
   @Override
+  public void onDecoderError(IllegalStateException e) {
+    printInternalError("decoderError", e);
+  }
+
+  @Override
   public void onAudioTrackInitializationError(AudioTrack.InitializationException e) {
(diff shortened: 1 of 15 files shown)


ojw28:
Hi, I've started looking at these. I'm commenting on the individual commits. So far:
https://github.com/rtsmb/ExoPlayer/commit/99167da07bfb2597c351614a2fe84a14474712dd
https://github.com/rtsmb/ExoPlayer/commit/3921919115899cc096e90023db19e600507d7715

PR #818, opened 2015-09-24
Add onDrmSessionReady callback

pakerfeldt:
This adds a callback for when the DRM session is ready.
I.e. when keys required for decryption has been acquired.

-----

We have the need of getting notified when DRM session has been successfully loaded and ready to be used. Therefore, I suggest an addition to StreamingDrmSessionManager's EventListener. Do you think it makes sense?

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/drm/StreamingDrmSessionManager.java
+++ b/library/src/main/java/com/google/android/exoplayer/drm/StreamingDrmSessionManager.java
@@ -51,6 +51,12 @@ public class StreamingDrmSessionManager implements DrmSessionManager {
   public interface EventListener {
 
     /**
+     * Invoked when DRM keys have been loaded. Depending on license setup, this might occur multiple
+     * times during playback.
+     */
+    void onKeysLoaded();
+
+    /**
      * Invoked when a drm error occurs.
      *
      * @param e The corresponding exception.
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/android/exoplayer/drm/StreamingDrmSessionManager.java
```diff
@@ -386,6 +391,14 @@ private void onKeyResponse(Object response) {
     try {
       mediaDrm.provideKeyResponse(sessionId, (byte[]) response);
       state = STATE_OPENED_WITH_KEYS;
+      if (eventHandler != null && eventListener != null) {
```
ojw28: Depending on your license setup, this might be called many times during playback of a single video. For example if keys are valid for X minutes and fresh keys are loaded every Y minutes, where Y < X. So I don't think onDrmSessionReady is a great name. Maybe the listener method should be called onKeysLoaded, which accurately reflects what just happened?

library/src/main/java/com/google/android/exoplayer/drm/StreamingDrmSessionManager.java
```diff
@@ -51,6 +51,11 @@
   public interface EventListener {
 
     /**
+     * Invoked when the session is open and the keys required for decryption has been acquired.
```
ojw28: NB - Comment will need adjusting if you rename this method as below.

ojw28:
I have no particular objection, but I can't think of a use case either. Can you say what you actually need this for? There may be another way of achieving what you need.

library/src/main/java/com/google/android/exoplayer/drm/StreamingDrmSessionManager.java
pakerfeldt: That's a very valid point.

pakerfeldt:
The use case is collecting start up metrics. We want to know how long it took to initiate drm for a specific playback.

ojw28:
Ok, that's fine :).

pakerfeldt:
The commit has been updated. Also, only pull this in if you feel it adds value. I don't mean to pollute the project in any way.

ojw28:
Note - Tweaked slightly in https://github.com/google/ExoPlayer/commit/e652019bb3eb2041218ecf928c10b3166f728f78

PR #827, opened 2015-09-29
Master (Sony Bravia 4K TV)

piterwilson:
Simple fork that permits 4K playback on Sony Bravia 4K TV sets.


ojw28:
Thanks for this. I think @andrewlewis is planning a similar but more targeted change.

piterwilson:
That's great to hear. Look forward to the update.

PR #828, opened 2015-09-29
FLV (H.264 + AAC) support

jeoliva:
FLV is not one of the most used containers right now, and of course not a priority for any player, but I though it could be useful adding it to Exoplayer for 2 reasons:
1. FLV was pretty used not too much time ago and still there are a lot of legacy video content libraries that are using this container. I have even seen a couple of reported issues asking for FLV support so there are some interest on it. 
2. And second, and most important, a share knowledge reason. In other words, I think this is really useful to allow other developers to better understand how to build their own extractors for Exoplayer.  FLV is one of the simplest and best documented video/audio container formats so it is easy going through the code, looking at the parser, flv extractor, FLV specification, and see what is the architecture behind Exoplayer extractors. MP4 type of containers are much more complex, and are a so hard way of starting coding or trying to understand extractors :). 

Regarding the implementation, as title stated, this pull request is adding support for FLV container, and more specifically, when content within FLV is encoded using H.264 and AAC.


ojw28:
In general I'm not a fan of adding support for legacy container formats. If we take it, it's something we have to commit to maintaining. However, as you say, this does look pretty simple, and so it seems like a reasonable request!

I'll post a few initial comments now, and try and have a more in-depth look later this week. Thanks!

library/src/main/java/com/google/android/exoplayer/extractor/ExtractorSampleSource.java
```diff
@@ -146,6 +146,13 @@ public UnrecognizedInputFormatException(Extractor[] extractors) {
     } catch (ClassNotFoundException e) {
       // Extractor not found.
     }
+    try {
```
ojw28: Please update the list in the javadoc at the top of this class to say that FLV is supported. Also, if seeking is not supported (see comment below) then update the sentence that says seeking in AAC and MPEG-TS is not supported, to include FLV too.

library/src/main/java/com/google/android/exoplayer/extractor/flv/TagPayloadReader.java
```diff
@@ -0,0 +1,92 @@
...
+   * @param data Buffer where tag payload is stored
+   * @param timeUs Time position of the frame
+   */
+  protected abstract void parsePayload(ParsableByteArray data, long timeUs);
+
+  /**
+   * Evaluate if for the current tag, payload should be parsed
+   * @return
+   */
+  protected abstract boolean shouldParsePayload();
```
ojw28: It looks like you could have parseHeader return whether the payload should be parsed, rather than having a separate method. Current logic in the implementations is to stash information in member variables during the parseHeader call, which is then used during this call. Having parseHeader return the boolean directly means you don't need these member variables.

library/src/main/java/com/google/android/exoplayer/extractor/flv/TagPayloadReader.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Consumes (possibly partial) payload data.
```
ojw28: Does this doc need updating? I'm not sure, but a brief scan of FlvExtractor suggests this is consuming tag data, which is complete?

library/src/main/java/com/google/android/exoplayer/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Parses audio tags of from an FLV stream and extracts AAC frames.
+ */
+final class AudioTagPayloadReader extends TagPayloadReader {
+
+  private static final String TAG = "AudioTagPayloadReader";
+
+  // Sound format
+  private static final int AUDIO_FORMAT_LINEAR_PCM_PLATFORM_ENDIAN = 0;
```
ojw28: A whole bunch of these aren't used anywhere (applies to TAG above also). Can we delete all of the unused ones? Not sure what IDE you're using, but I always turn on a warning that highlights unused stuff, which makes them easy to spot and delete.

library/src/main/java/com/google/android/exoplayer/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -0,0 +1,156 @@
...
+  private static final int AAC_PACKET_TYPE_AAC_RAW = 1;
+
+  // SAMPLING RATES
+  private static final int[] AUDIO_SAMPLING_RATE_TABLE = new int[] {
+      5500, 11000, 22000, 44000
+  };
+
+  // Variables to hold audio configuration.
+  private int format;
+  private int sampleRate;
```
ojw28: Unused (ditto x2 below). format would also be used if you merge shouldParsePayload into parseHeader.

library/src/main/java/com/google/android/exoplayer/extractor/flv/FlvExtractor.java
```diff
@@ -0,0 +1,310 @@
...
+      tagData.reset();
+    }
+
+    parserState = STATE_READING_TAG_HEADER;
+
+    return RESULT_CONTINUE;
+  }
+
+  // SeekMap implementation.
+  // TODO: Add seeking support
```
ojw28: It would be nice if this could be done in the initial commit, assuming the container provides sufficient information to allow for seeking. It's a pretty glaring absence :).

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -0,0 +1,258 @@
...
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Parses Script Data tags from an FLV stream and extracts metadata information.
+ */
+final class ScriptTagPayloadReader extends TagPayloadReader {
```
ojw28: It seems preferable to cut this right down to something that just parses the duration for the first commit. The rest doesn't actually do anything, at least for now.

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -0,0 +1,258 @@
...
+  public boolean canSeekOnTime;
+  public String httpHostHeader;
+
+  /**
+   * @param output A {@link TrackOutput} to which samples should be written.
+   */
+  public ScriptTagPayloadReader(TrackOutput output) {
+    super(output);
+
+    output.format(MediaFormat.createTextFormat(MimeTypes.APPLICATION_AMF, MediaFormat.NO_VALUE,
```
ojw28: Don't do this if you're never going to send any sample data.

library/src/main/java/com/google/android/exoplayer/extractor/flv/FlvExtractor.java
```diff
@@ -0,0 +1,310 @@
...
+    // Pass payload to the right payload reader.
+    if (currentTagHeader.type == TAG_TYPE_AUDIO && audioReader != null) {
+      audioReader.consume(tagData, currentTagHeader.timestamp);
+    } else if (currentTagHeader.type == TAG_TYPE_VIDEO && videoReader != null) {
+      videoReader.consume(tagData, currentTagHeader.timestamp);
+    } else if (currentTagHeader.type == TAG_TYPE_SCRIPT_DATA && metadataReader != null) {
+      metadataReader.consume(tagData, currentTagHeader.timestamp);
+      if (metadataReader.durationUs != C.UNKNOWN_TIME_US) {
+        if (audioReader != null) {
+          audioReader.durationUs = metadataReader.durationUs;
```
ojw28: - If the audio reader has already output its format, this happens too late. Is there a guarantee that the duration metadata will arrive before this happens? If not, do you need a guard somewhere that stops the audio reader from outputting the format until the duration has been set? Ditto applies for video below.

- IMO a setDurationUs method would be more appropriate than directly setting the variable for this case. durationUs should probably be private to each reader.

jeoliva:
Taking a look to all these things.

library/src/main/java/com/google/android/exoplayer/extractor/flv/FlvExtractor.java
jeoliva: In fact, FLV container doesn't include any kind of information to help on implementing seeking. The most we can do is building some kind of timestamps-byte range map while we are downloading the stream and using that table for implementing seeking. FLV doesn't define that kind of mapping, as you can find in MP3, MP4, etc

ojw28: Thanks for the info! Let's leave it as not implemented in that case.

Out of interest, when playing a really long FLV video (e.g. in a browser), how would a player typically handle seeking to near the end? Would it just guess an offset (e.g. assuming constant bit-rate) and then try and re-synchronize to the container?

jeoliva: In fact, when playing an FLV file in a browser (and then, using a Flash based player) you will notice that player is only able to seek to the portion of the file that was already downloaded. There is no way of seeking to a position that was not previously downloaded as there is no way of calculating bytes given a timestamp. 

As a solution to that, most of CDN's provide some kind of "virtual seeking" (probably not the best name). This is stream url accepting an optional query string parameter with the seeking offset (Ex: http://domain/file.flv?time_offset=xxxx) and then, CDN analyzing the file and returning FLV tags from the position specified in the query string parameter.

Assuming constant bitrate could be enough when working with only audio (I implemented something like this in the CDN side when working with MP3 CBR streams without access to MP3 "seeking headers") but it accuracy would be really poor when working with video tracks. With H.264 baseline profile you can do something, but not accurate at all, if you even use more complicated stuffs like main profile and bidirectional frames, any assumption you can take regarding time <-> byte will be wrong. 

Anyway, in the specific case of FLV, there is not a sync sequence as you can find in other formats, so there is not a very good way of re-sync once you jump into the middle of the stream.

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -0,0 +1,258 @@
...
+      Map<String, Object> extractedMetadata = (Map<String, Object>) obj;
+      for (Map.Entry<String, Object> entry : extractedMetadata.entrySet()) {
+        if (entry.getValue() == null) {
+          continue;
+        }
+
+        switch (entry.getKey()) {
+          case "duration":
+            this.durationUs = (long)(C.MICROS_PER_SECOND * (Double)(entry.getValue()));
+            break;
```
mondain: Duration in an FLV's metadata, is in seconds. http://help.adobe.com/en_US/AS2LCR/Flash_10.0/help.html?content=00002413.html

library/src/main/java/com/google/android/exoplayer/extractor/flv/AudioTagPayloadReader.java
jeoliva: Yep, I preferred to keep all the types as a way of declaring all the possible values but you are right, as they are never used doesn't make sense keeping all that code there. Thanks!

library/src/main/java/com/google/android/exoplayer/extractor/flv/FlvExtractor.java
jeoliva: - That's a difficult decision. The problem with waiting for the duration to arrive is there is no any guarantee that duration field is going to arrive at any time. Although that metadata field is typically added at the very beginning of the file, before any other FLV tag (and in fact some CDN's require onMetadata with duration data to be at the very beginning of each FLV file), that's not a must. The only sure way of calculating duration would be reading the last 4 bytes of the FLV file (length of the last FLV tag) and then reading that tag and its PTS value. However, this would require doing an HTTP range request before playback starts that I think modify a bit how Exoplayer HTTP request workflow works. 

For your info, what Flash players are doing is just assuming duration is 0 when metadata duration field is not there (Flash player can't do http byte range requests).

- Ok, makes sense.

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
jeoliva: Ok, makes sense.

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
jeoliva: Ok, makes sense

library/src/main/java/com/google/android/exoplayer/extractor/flv/TagPayloadReader.java
jeoliva: In fact, shouldParsePayload and parseHeader are separated right now because I had in mind implementing some extra features that were going to make use of that. Following with all the modifications I am applying, doesn't make sense making code a more complex for something that still was not done. As soon as new features are implemented I will add the extra complexity in case that's needed.

jeoliva:
Just incorporated suggested changes. Thanks @ojw28 for all your comments!

ojw28:
Sorry for the delay looking at this. Will try and get to it tomorrow!

library/src/main/java/com/google/android/exoplayer/extractor/ExtractorSampleSource.java
```diff
@@ -60,7 +60,7 @@
  * <li>MPEG TS ({@link com.google.android.exoplayer.extractor.ts.TsExtractor}</li>
```
ojw28: This list should contain FLV.

library/src/main/java/com/google/android/exoplayer/extractor/mp4/Mp4Extractor.java
```diff
@@ -125,7 +125,6 @@ public int read(ExtractorInput input, PositionHolder seekPosition)
   }
 
   // SeekMap implementation.
```
ojw28: Revert this file?

library/src/main/java/com/google/android/exoplayer/util/MimeTypes.java
```diff
@@ -60,6 +60,7 @@
   public static final String APPLICATION_TTML = BASE_TYPE_APPLICATION + "/ttml+xml";
   public static final String APPLICATION_M3U8 = BASE_TYPE_APPLICATION + "/x-mpegURL";
   public static final String APPLICATION_TX3G = BASE_TYPE_APPLICATION + "/x-quicktime-tx3g";
+  public static final String APPLICATION_AMF = BASE_TYPE_APPLICATION + "/x-amf";
```
ojw28: Is this used anywhere? If not, can we revert for now?

library/src/main/java/com/google/android/exoplayer/extractor/flv/FlvExtractor.java
ojw28: setDurationUs method?

library/src/main/java/com/google/android/exoplayer/extractor/flv/ScriptTagPayloadReader.java
ojw28: Is there a cut down version of this? I'm not sure whether this is still pending? Thanks!

jeoliva: Yes, this was totally reduced to just the needed for parsing duration and also removed the creation of the text track.

library/src/main/java/com/google/android/exoplayer/extractor/mp4/Mp4Extractor.java
jeoliva: Ouugh, sorry...

jeoliva:
Changes applied

winlinvip:
FLV is useful in live streaming with low-latency, similar to RTMP for flash.
For example, most live streaming server supports RTMP/HLS, but HLS is large latency 10s+ while RTMP 1s+. Actually the RTMP is similar to HTTP-FLV and some server(for instance, [srs](https://github.com/winlinvip/simple-rtmp-server/issues/293)) already supported the HTTP-FLV/RTMP.

winlinvip:
@jeoliva Thanks for your work, I merge this pull request to my fork, :+1:

library/src/main/java/com/google/android/exoplayer/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -0,0 +1,125 @@
...
+      int audioObjectType = adtsScratch.readBits(5);
+      int sampleRateIndex = adtsScratch.readBits(4);
+      int channelConfig = adtsScratch.readBits(4);
+
+      byte[] audioSpecificConfig = CodecSpecificDataUtil.buildAacAudioSpecificConfig(
+          audioObjectType, sampleRateIndex, channelConfig);
+      Pair<Integer, Integer> audioParams = CodecSpecificDataUtil.parseAacAudioSpecificConfig(
+          audioSpecificConfig);
+
+      MediaFormat mediaFormat = MediaFormat.createAudioFormat(MimeTypes.AUDIO_AAC,
```
winlinvip: Compile failed when merge to dev-1.5.1-rc：

git/ExoPlayer/library/src/main/java/com/google/android/exoplayer/extractor/flv/AudioTagPayloadReader.java
Error:(111, 44) error: method createAudioFormat in class MediaFormat cannot be applied to given types;
required: int,String,int,int,long,int,int,List<byte[]>,String
found: String,int,int,long,Integer,Integer,List<byte[]>,<null>
reason: actual and formal argument lists differ in length

winlinvip: Seems change to the bellow is ok:
```
MediaFormat.createAudioFormat(MediaFormat.NO_VALUE, MimeTypes.AUDIO_AAC,
```

winlinvip: I fix it at: https://github.com/winlinvip/ExoPlayer/commit/115aee8984ca298d072b4469d362e55708226fba

ojw28: Yes. I've merged this and fixed the break. Thanks @jeoliva / @winlinvip !

ojw28:
I've also submitted some stylistic tweaks just to bring the code in line with the rest of the source: https://github.com/google/ExoPlayer/commit/950cc70003d7633d69afb2e2d4d0859c9b1fa0c6

ojw28:
And some further changes. I was able to simplify ScriptTagPayloadReader somewhat. Please take a look, and let me know if I accidentally broke anything!

ojw28:
One final question: Where did the test link you've added come from? I want to know if it's fine to copy it across to our own hosting. Obviously the underlying content (big buck bunny) is fine for us to copy. Depending on who packaged it, we might need to re-package it ourselves to copy it over. Please advise. Thanks!

winlinvip:
I test the flv, seems ok for vod stream, for example, it's ok for http://www.ossrs.net/mp4/source.200kbps.768x320.flv。
But failed to play a flv live stream, which is send by chunked encoding by SRS server, the demo stream is http://ossrs.net:8081/live/livestream.flv or [link](http://www.ossrs.net/players/srs_player.html?vhost=ossrs.net&stream=livestream.flv&server=ossrs.net&port=8081&schema=http)。
I will find out why, maybe ExoPlayer depends on the file Content-Length? So not support chunked http stream?

winlinvip:
BTW, I want to introduce more about the HTTP-FLV low latency live streaming for Android.
I have start a Android encoder, which fetch the image from camera and encode by MediaCodec and send to SRS over HTTP POST, that is [srs-sea](https://github.com/simple-rtmp-server/srs-sea).
I plan to start a player named srs-spa, but I found ExoPlayer and this pull request, thanks very much, I have canceled the srs-spa project and linked to ExoPlayer.

It's very very strong feature for low latency live streaming on Android, user can use [srs-sea](https://github.com/simple-rtmp-server/srs-sea) to publish stream and  play with ExoPlayer, the latency is 1s+, it's awesome~

Thanks~

winlinvip:
Seems HTTP-TS is ok~
I use ExoPlayer to play the http://ossrs.net:8081/live/livestream.ts, it works~

I test the latency, use srs-sea to publish and exoplayer to play, the latency is 1s - 3s:
sea exoplayer

ojw28:
The live stream you link to doesn't set the TypeFlagsAudio or TypeFlagsVideo flags in the FLV header, which are supposed to be set to indicate audio and video are present in the stream. So I think it's an issue with the stream rather than the player. If I hardcode hasAudio and hasVideo to true in FlvExtractor, then the stream (sort of - albeit with a lot of apparent distortion) plays ok.

jeoliva:
I am taking a look. 

@winlinvip, I am also very interested on low latency streaming so thank you very much for sharing the link to your SEA project.

jeoliva:
Regarding FLV sample file, I got it from the following demo site: http://demo.leasewebcdn.com. I was looking for something not right protected so chose the typical Big Bug Bunny video.

ojw28:
Thanks. for the info. I made some further changes to the extractor in 4422e8a0155642a4708e1f50b1fb2c340c78f56d based on an internal review. I've put a couple of comments in the old version of the code in that change that explain why the new state "skipping to next tag header" state is necessary :).

winlinvip:
@ojw28 Thanks~ I will try to add hasVideo and hasAudio in header in SRS server.

winlinvip:
Fixed in https://github.com/simple-rtmp-server/srs/commit/50a7b9c1ff2124017abc6a3c6fdd2337d11046d3, set hasVideo and hasAudio tag in flv header.
It works! And the ExoPlayer can play the HTTP-FLV and HTTP-TS stream, both live and vod:

* HTTP-FLV: http://ossrs.net:8081/live/livestream.flv
* HTTP-TS: http://ossrs.net:8081/live/livestream.ts

chodison:
But live time is not moving

winlinvip:
@chodison Not understand you, what's `time moving` means? The timestamp in FLV stream is increasing, maybe the player not display it?

winlinvip:
I see, you means the timeline of ExoPlayer, that's caused by the flv onMetaData packet:
```
String onMetaData
Object (20 items)
    Property 'duration' Number 210.7
```

Because the flv stream is generated by ffmpeg to publish a flv file, which contains the duration, I will remove this field for live stream.

chodison:
The player‘s position is always zero. And Cycle two times, it don't work.

winlinvip:
Fixed in https://github.com/simple-rtmp-server/srs/commit/ef00005ab4d0456925e5f65e51a02f7100b4bf7f, remove the duration of flv for live streaming.

chodison:
@ojw28 For flv vod video, can not ExoPlayer to support seeking?

ojw28:
See @jeoliva's note on this thread above seeking: *In fact, FLV container doesn't include any kind of information to help on implementing seeking*

If you want your media to be seekable you should use a more appropriate container format, like MP4.

winlinvip:
@ojw28 @jeoliva @chodison There is some workaround for flv to support seek, the service which produce the flv file must inject all keyframe offset to the onMetaData script tag, then player can got the offset to seek to specified time. However, it's not the standard of flv, it's possible to implements by this workaround.

winlinvip:
Please refer to https://github.com/simple-rtmp-server/srs/wiki/v2_EN_FlvVodStream.

chodison:
@ojw28 why the ffplay player can seek fly vod video?

winlinvip:
@chodison The ffplay generally play a local flv vod file, it can skip one tag by tag, it's ok for disk but not good for http flv vod stream. The network file always need a `header` which store the keyframe and offset map, similar to mp4 container.

chodison:
@winlinvip My tested url is http flv vod stream.

winlinvip:
@chodison Http flv vod stream is ok to seek, to skip one tag by tag, it's not efficient way however. The acceptable way for a big flv vod file to seek, is inject the keyframe offset to the onMetaData.

ojw28:
There's an alternative "accepted" way where the server will allow you to specify the seek position as, for example, a query parameter. In that case the server does the brute force tag-by-tag search (it's probably implemented to have built an index already) and serve you the media from the specified seek position.

The takeaway here, however is that (a) FLV does not support seek properly, (b) we don't intend to support any hacks to make seeking in FLV possible in ExoPlayer, unless they're external contributions and do not extend anywhere outside of the FLV extractor package, and (c) if you're knowingly packaging your own media as FLV, you should stop and choose a more appropriate container like MP4.

winlinvip:
@ojw28 Aggree to use MP4 instead for http vod stream. :smile:

chodison:
@winlinvip @ojw28 flv vod stream server supports http start or http range to seek.

winlinvip:
@chodison Yep, but need the player to finger it out where to seek, for example:
```
player.seek(10s)
```

The player must conver the `10s` to file offset, for instance, `3421bytes`, then player must request from the offset:
```
http://server/f.flv?start=3421
``
Or request by range whatever.

The question is: how to know the offset and seconds relationship? The flv container never describe it.

There is a workaround, to save the offset-seconds map to the `onMetadata`(the scirpt tag) of flv, which need the player to parse this tag -- this is exactly what @ojw28 said about the standard flv not support seek.

I aggree, should never use none-standard, especially the player.

chodison:
@winlinvip http://server/f.flv?start=time or request by range, file offset is from he proportion of the position  to the duration.
seekTime/flvDuration == position/flvFileLength.

winlinvip:
@chodison seekTime/flvDuration != position/flvFileLength.

mondain:
The offset and time relationships are often sent in the metadata response; depending upon your server's implementation of course.

ojw28:
We have no plans to support seeking in FLV. All of the approaches mentioned would add complexity to the upstream package, and would not work in a consistent way across different `DataSource` implementations. If you want seeking, you should use a container format the properly supports it (or DASH, SmoothStreaming or HLS).

chodison:
@winlinvip Estimation calculation. The flv vod stream server will return the key frame for closest position

winlinvip:
@chodison It is possible to implement the flv stream seek, but not official standard for it, so I think @ojw28 is right to not support this feature and use HLS or DASH.

For the server to seek the closest position, not all server support it, that is why not official standard.

PR #832, opened 2015-09-30
WebVTT - Cue positioning fixes

jeoliva:
- Fix: fixed positioning when line is defined as a negative number.
 - Fix: added sanity checks to adjust positioning when it is out of the
screen.


library/src/main/java/com/google/android/exoplayer/text/CuePainter.java
```diff
@@ -350,4 +362,164 @@ private static boolean areCharSequencesEqual(CharSequence first, CharSequence se
     return first == second || (first != null && first.equals(second));
   }
 
+  /**
+   * Gets position anchor given an anchor and an alignment value. This method takes into
+   * account that the parameter position anchor could be not set and then anchor "inherits" its
```
ojw28: The inheritance behavior is a property of WebVTT and so should be implemented in WebvttParser rather than here. It's actually implemented there already (see the alignmentToAnchor method).

library/src/androidTest/java/com/google/android/exoplayer/text/webvtt/WebvttParserTest.java
```diff
@@ -35,6 +35,7 @@
   private static final String WITH_POSITIONING_FILE = "webvtt/with_positioning";
   private static final String WITH_TAGS_FILE = "webvtt/with_tags";
   private static final String EMPTY_FILE = "webvtt/empty";
+  private static final String WITH_WRITING_DIRECTION = "webvtt/with_writing_direction";
```
ojw28: I think you forgot to add this file to the patch?

ojw28:
As a general point, does anyone actually use these vertical modes? I'd quite like us to not support things that no-one uses, because there's a non-zero maintenance cost associated with every feature we take.

library/src/androidTest/java/com/google/android/exoplayer/text/webvtt/WebvttParserTest.java
jeoliva: ooops.... my fault. Sorry for that. I just added it.

Thanks!

library/src/main/java/com/google/android/exoplayer/text/CuePainter.java
jeoliva: Sounds good. I think this was wrong in the previous version so this failed in some of my tests and I added this behavior here. Anyway, as you suggest, WebVttParser is a better way for implementing that, as that's very specifc of WebVTT.

Just added two changes:
 1.- What you suggest. Removing inheritance behavior from CuePainter.
 2.- Modified WebVTTParser so whenever an alignment is not defined, alignmentToAnchor method returns MIDDLE alignment (that's what WebVTT spec says).

jeoliva:
Honestly, not very common, although I have worked with some Asian based companies that needed it (Japanese and Chinese customers). So, not the most used feature but somehow needed.

ojw28:
To ask a (possibly stupid) question: Is rotation the correct way to achieve vertical flow for such cases, given that this approach will also rotate all of the characters? I guess I kind of assumed that something fundamentally different would need to happen during text rendering, that involves writing vertically but not doing any character rotation. Please educate :). Thanks!

Will take a more detailed look at this on Monday.

jeoliva:
Hmmm.... Let me feel stupid myself for a while....... I was pretty used to see players rotating captions, but knowing that webvtt vertical mode is primarily thought to be used with Asian languages that are written from top to bottom, doesn't make sense at all implementing vertical mode using rotation. I skipped that obvious point.

Let me remove vertical mode implementation from this PR as this would need a complete redesign and just include the fixes I prepared for the horizontal mode. I will do that on Monday.

ojw28:
Sounds good; thanks!

jeoliva:
Done. Thanks for your comments!

library/src/main/java/com/google/android/exoplayer/text/Cue.java
```diff
@@ -126,7 +126,7 @@ public Cue(CharSequence text) {
   }
 
   public Cue(CharSequence text, Alignment textAlignment, float line, int lineType,
-      int lineAnchor, float position, int positionAnchor, float size) {
+             int lineAnchor, float position, int positionAnchor, float size) {
```
ojw28: Completely revert this file.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttCue.java
```diff
@@ -37,7 +37,7 @@ public WebvttCue(long startTime, long endTime, CharSequence text) {
   }
 
   public WebvttCue(long startTime, long endTime, CharSequence text, Alignment textAlignment,
-      float line, int lineType, int lineAnchor, float position, int positionAnchor, float width) {
+                   float line, int lineType, int lineAnchor, float position, int positionAnchor, float width) {
```
ojw28: Completely revert file.

library/src/main/java/com/google/android/exoplayer/text/webvtt/WebvttParser.java
```diff
@@ -329,4 +329,4 @@ private static int alignmentToAnchor(Alignment alignment) {
 
   }
 
-}
+}
```
ojw28: Completely revert file.

library/src/main/java/com/google/android/exoplayer/text/CuePainter.java
```diff
@@ -220,48 +220,20 @@ public void draw(Cue cue, boolean applyEmbeddedStyles, CaptionStyleCompat style,
...
-      textRight = textLeft + textWidth;
-    }
+
+    // Position.
+    textLeft = calculateCueboxLeftPosition(parentWidth, cuePosition, cuePositionAnchor,
+        parentRight, parentLeft, textWidth);
+    textRight = Math.min(parentRight, textLeft + textWidth);
 
     int textTop;
     int textBottom;
```
ojw28: This isn't used anywhere (it's only written, never read).

library/src/main/java/com/google/android/exoplayer/text/CuePainter.java
```diff
@@ -350,4 +322,96 @@ private static boolean areCharSequencesEqual(CharSequence first, CharSequence se
...
+   *
+   * @param parentWidth Width of parent enclosing the cuebox.
+   * @param position Position value as it comes from the cue object.
+   * @param positionAnchor Position anchor as it comes from the cue object.
+   * @param parentRight Parent right.
+   * @param parentLeft Parent left
+   * @param textWidth Text width
+   * @return Left position of the cue box within parent layout.
+   */
+  private static int calculateCueboxLeftPosition(int parentWidth, float position,
```
ojw28: Hm. The logic in this method seems more confusing than the previous logic. Are you actually fixing anything here? I'm finding it hard to parse what's changed.

library/src/main/java/com/google/android/exoplayer/text/CuePainter.java
```diff
@@ -350,4 +322,96 @@ private static boolean areCharSequencesEqual(CharSequence first, CharSequence se
...
+   * @param lineAnchor Line anchor as it comes from the cue object.
+   * @param firstLineHeight Height of the first line of the text in the cuebox.
+   * @param parentTop Parent top position.
+   * @param parentBottom Parent bottom position.
+   * @param textHeight Height of the text enclosed by the cuebox.
+   * @param bottomPadding The bottom padding fraction to apply when {@link Cue#line} is
+   *     {@link Cue#DIMEN_UNSET}, as a fraction of the viewport height.
+   * @return Top position of the cue box.
+   */
+  private static int calculateCueBoxTopPosition(int parentHeight, float line, int cueLineType,
```
ojw28: Ditto. Although I think you are fixing something here.

ojw28:
Hm. I'm not sure about this. I think the (now backed out) vertical text changes have ended up moving things around a little that wouldn't have been moved around otherwise. If you were to make the minimal change necessary to fix the two things that this change is now fixing, what would that look like (NB - It'll be much easier to review if the code doesn't move into separate helper methods; if that's really needed/preferable, it would be good to do it as a subsequent change that changes only that and nothing else).

jeoliva:
Applied changes to previous method. Thanks for your comments. Just for your info, what I see was wrong was:
1.- When line is provided as percentage, line position seems to be wrong. 
line_30_percentage

2.- When end alignment is used for position parameter, position is not correct.
position_20_end

3.- Line position is not correct when using negative values. Example:
line_minus_5_start

ojw28:
Thanks for the screenshots! I'm not convinced the previous behavior was incorrect though. Looking at the spec:

- For the first screenshot, I think a line percentage is a percentage from the top of the viewport. From the spec: *A WebVTT line cue setting configures the offset of the cue box from the video viewport’s edge in the direction opposite to the writing direction. For horizontal cues, this is the vertical offset from the top of the video viewport.*

- For the second screenshot, I think `position:20%,end` means you should position the right hand side of the cue box 20% from the left of the viewport. There's an example in the VTT spec illustrating this, that states `position:10%,start size:35%` and `position:45%,end size:35%` are equivalent.

- For the third screenshot, I think negative line numbers are supposed to count from the bottom of the viewport, where-as positive line numbers count from the top.

Thoughts?

jeoliva:
You are right, after taking a look to WebVTT spec seems there was a misunderstood in my side. From the point of view of the user, specially the specification related with position and end alignment looks a bit counter intuitive for me. 

Thanks for your comments. I am going to close this one as fault was on my side, positioning is already fine in the current implementation and I was misunderstanding webvtt spec.

ojw28:
Doh :). Oh well; it was an interesting verification exercise at least! The positioning defined by the WebVTT spec is indeed. . .complicated. It's somewhat unclear to me whether it's somewhat over-engineered, or whether everything in the spec is actually useful!

Issue #87, opened 2014-10-22
Implement seeking within HLS moving live window in ExoPlayer and its demo app


perchrh:
The server may keep old chunks available to the streaming clients so that they can rewind the video, for example, to 10 minutes before the live position. Imagine you missed the start of a live tv-show or you were interrupted and missed a part, so you want to rewind it. There are two main, common cases here, the live playlist received by the clients can be of a short or a long duration. 

1) Often CDNs will give you a short playlist with 3-10 chunks, for a playlist duration of 2-3 minutes or less. In this case seeking inside the playlist is not very useful. Instead you would make a request to the CDN for a new playlist at a different position, for example live head - 20 minutes. 
The NASA TV public live stream is an example of this approach, http://www.nasa.gov/multimedia/nasatv/NTV-Public-IPS.m3u8

2) The second case, that this issue covers, is where you get a long playlist, for example 120 chunks, or 1 hour, from the server. In this case seeking within the playlist is desirable. Because the server is continuously deleting old chunks (and writing new ones), you have a moving live window of a certain fixed size in seconds. 
The player/library must not request chunks outside of the window, as that will give an HTTP error (404). The seekbar must take the moving live window into consideration. While watching the stream, the seekbar should be in the same position, e.g. 10 minutes before live head. The playlist must be refetched when reaching the end of the playlist, or when the user seeks to the end of the playlist. The player UI could for example display "live" as the duration and "-10:00" for the current position when the user has rewinded the live video 10 minutes. 

Please let me know if you would like a public sample stream for case (2).

fougere-mike:
Another +1 for seeking in live HLS content

zikavele:
Yet another +1 for seeking in live HLS content

ldimitroff:
Is there an ETA on when is this going to be tackled down? Is there anybody working on it now or has an ExoPlayer workaround to support DVR?

andreimarinescu:
+1 as well

zsmatyas:
+1 :)

ojw28:
No promises, but we're aiming to add DVR window support for DASH by the end of this month. As we do that, we'll be looking at how the same approach can be used for HLS too, specifically for case (2) in the original post. Hopefully this wont be too hard, since they're roughly analogous to one another. I don't think we have near-term plans to support case (1) in the original post.

Arqu:
+1 for case (2)

Guro:
+1

fougere-mike:
[...]

PR #848, opened 2015-10-07
HLS Support for Live Events

thasso:
This pull requests adds support for HLS Live Event Streams where `EXT-X-PLAYLIST-TYPE` is set to `EVENT` and the player should allow seeking within the currently available range.

The seek range change is exposed through the EventListener and a delegation constructor was added so the only API change is the change of the EventListener interface.

(FYI Company Contributor Agreement should be on the way)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
@@ -17,6 +17,7 @@ package com.google.android.exoplayer.hls;
 
 import com.google.android.exoplayer.C;
 import com.google.android.exoplayer.MediaFormat;
+import com.google.android.exoplayer.TimeRange;
 import com.google.android.exoplayer.chunk.BaseChunkSampleSourceEventListener;
 import com.google.android.exoplayer.chunk.Chunk;
 import com.google.android.exoplayer.chunk.ChunkOperationHolder;
@@ -36,6 +37,7 @@ import com.google.android.exoplayer.util.UriUtil;
 import com.google.android.exoplayer.util.Util;
 
 import android.net.Uri;
+import android.os.Handler;
 import android.os.SystemClock;
 import android.text.TextUtils;
 import android.util.Log;
@@ -60,7 +62,14 @@ public class HlsChunkSource {
   /**
    * Interface definition for a callback to be notified of {@link HlsChunkSource} events.
    */
-  public interface EventListener extends BaseChunkSampleSourceEventListener {}
+  public interface EventListener extends BaseChunkSampleSourceEventListener {
+    /**
+     * Invoked when the available seek range of the stream has changed.
+     *
+     * @param availableRange The range which specifies available content that can be seeked to.
+     */
+    public void onAvailableRangeChanged(TimeRange availableRange);
+  }
 
   /**
    * Adaptive switching is disabled.
@@ -141,6 +150,8 @@ public class HlsChunkSource {
 
   private byte[] scratchSpace;
   private boolean live;
+  private boolean liveEvent;
+  private boolean firstLiveEventChunkSelected;
   private long durationUs;
   private IOException fatalError;
   private PtsTimestampAdjuster ptsTimestampAdjuster;
@@ -149,11 +160,15 @@ public class HlsChunkSource {
   private byte[] encryptionKey;
   private String encryptionIvString;
   private byte[] encryptionIv;
+  private TimeRange availableRange;
+
+  private Handler eventHandler;
+  private EventListener eventListener;
 
   public HlsChunkSource(DataSource dataSource, String playlistUrl, HlsPlaylist playlist,
       BandwidthMeter bandwidthMeter, int[] variantIndices, int adaptiveMode) {
     this(dataSource, playlistUrl, playlist, bandwidthMeter, variantIndices, adaptiveMode,
-        DEFAULT_MIN_BUFFER_TO_SWITCH_UP_MS, DEFAULT_MAX_BUFFER_TO_SWITCH_DOWN_MS);
+        DEFAULT_MIN_BUFFER_TO_SWITCH_UP_MS, DEFAULT_MAX_BUFFER_TO_SWITCH_DOWN_MS, null, null);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


thasso:
@ojw28 Any feedback on this would be much appreciated. Do you you think this can be merged to dev? I think I could also add another commit that tackles #87 and support seeking in general live streams.

ojw28:
Hi. I'll get to this soon (sorry!).

pajatopmr:
Would it be possible to add a sample to the HLS demo project that shows off this code?  It is not clear if this PR fixes HLS live seeking.  Does it?

ojw28:
Note that we have issues elsewhere saying "make the demo app simpler". So it's pretty unclear whether that's a good idea :)...

thasso:
I added two commits to enable event support in the demo application. I think c64a1cf is okay, but I'm not so sure about 36f7f7e. Passing the listener information (the seek range info) into ``PlayerControl`` doesn't look very clean to me, but it might help to test this and I couldn't find a way to pass this information without more major changes. 

For testing purposes, I only have a dummy stream at http://demo.castlabs.com/media/BBB-HLS/BBB-Event.m3u8 . It's an ``EVENT`` type stream that is not finished yet (no end tag), but please note that it does **not update**. Unfortunately I currently do not have access to a _real_ live event stream.

pajatopmr:
@thasso thanks for the testing information.  I tried your BBB stream and an iTV24 live stream to test your code (from your repo) but get no joy.  The player controls start and end time are both 0.  Is there more that I need to do to test this PR?

Here's what I added to Samples.java: 
    new Sample("Live/DVR Fixed", "http://bcoveliveios-i.akamaihd.net/hls/live/215102/master_english/398/master.m3u8",
        PlayerActivity.TYPE_HLS),
    new Sample("Live/DVR Event", "http://demo.castlabs.com/media/BBB-HLS/BBB-Event.m3u8", PlayerActivity.TYPE_HLS),

thasso:
@pajatopmr I have a hard time reproducing this for BBB. Here, the playback for BBB starts at 7:40 and the duration is displayed as 8:10 correctly (this is not a _real_ event, so the duration will not change). There are not further changes required to make this work. Do you see output like ``availableRange [true, 0, 490000000]`` in the logs when you start playback?

For the iTV24 stream, this is expected behaviour for this PR. The stream playlists are not marked as "Event" so this PR has no effect and the player as well as the controls behave as before.

pajatopmr:
@thasso I do not see the available range in the logcat.  Also, should not a
fixed size live window also report an available seek range change every
time a new "chunk" arrives?  If yes, then your change should also apply to
live.

On Wed, Oct 28, 2015 at 9:45 AM Thasso Griebel <notifications@github.com>
wrote:

> @pajatopmr <https://github.com/pajatopmr> I have a hard time reproducing
> this for BBB. Here, the playback for BBB starts at 7:40 and the duration is
> displayed as 8:10 correctly (this is not a *real* event, so the duration
> will not change). There are not further changes required to make this work.
> Do you see output like availableRange [true, 0, 490000000] in the logs
> when you start playback?
>
> For the iTV24 stream, this is expected behaviour for this PR. The stream
> playlists are not marked as "Event" so this PR has no effect and the player
> as well as the controls behave as before.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/ExoPlayer/pull/848#issuecomment-151848648>.
>
-- 
-pmr

thasso:
@pajatopmr Can you please double check that you are using the correct branch to test this, or maybe add some logging statements in the ``HlsPlaylistParser`` to verify that the event tag is parsed correctly? If you do not see the range change log messages, something is wrong but I can not reproduce the issue. Here its working.

You might be right about the general range changes, but this PR was intended to add support for Events only, where the seeking window is clearly defined. But I agree that this could rather easily be extended to also send seek range change events for general live streams, which would put this more in the context of  #83 and I'm not sure if that is a good idea for this PR ?!?

pajatopmr:
@thasso Sorry about the confusion.  Yes indeed, I did use the wrong branch initially but quickly realized and fixed my mistake to find your changes working as advertised.  This PR thus provides me with the first ExoPlayer experience I have encountered where some Live behavior works from the player controls perspective, both from showing current time and duration as well as having seek behavior that works.

Now I would like to use your changes (in the lack of anything as good or better) to get Live working with a fixed size moving window as illustrated by the iTV24 stream I referenced above.  Fwiw, the approach I am using is to generate an available range change event on top of this PR for every new chunk loaded.

library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java
```diff
@@ -573,7 +624,16 @@ private void setMediaPlaylist(int variantIndex, HlsMediaPlaylist mediaPlaylist)
     variantLastPlaylistLoadTimesMs[variantIndex] = SystemClock.elapsedRealtime();
     variantPlaylists[variantIndex] = mediaPlaylist;
     live |= mediaPlaylist.live;
-    durationUs = live ? C.UNKNOWN_TIME_US : mediaPlaylist.durationUs;
+    liveEvent |= mediaPlaylist.liveEvent;
+    durationUs = live || liveEvent ? C.UNKNOWN_TIME_US : mediaPlaylist.durationUs;
+
+    if(!live || liveEvent) {
```
pajatopmr: Why is "live" being filtered out here.  I would think that both a sliding fixed size live window and a growing sized live window (is this the same as an "Event" type?) would want to notify listeners about the available seek range change.

pajatopmr: @ojw28 @thasso When I remove the conditional at line 630 and then test against the iTV24 live stream, it sort of works correctly.  There is an excessive delay in forward seeking but backward seeking works fine.  Can either of you make a suggestion where I can find and attempt to fix this seek bug.  I have a hunch it is a known, existing issue.

ckarthikv:
@pajatopmr - Needed the HLS live seeking feature in ExoPlayer, safe to cherry pick these 4 commits now?

pajatopmr:
@ckarthikv No, it is not really safe to take these changes yet.  The problem is that seeking forward still does not work correctly for the live stream mentioned in an earlier comment on this PR.  There is an unacceptably long delay (sometimes measured in minutes) before playback continues from the desired seek position.  I am currently working on solving this problem and will report any progress both here and in a separate issue.

ckarthikv:
@pajatopmr  - Thanks. My requirement is to get the HLS video to always start from the start of the LiveStream rather than the live-edge, which essentially means just one-time-backward-seeking to time 0L. Would these 4 commits help achieve the same?

thasso:
@ckarthikv The PR as it is right now will not help with that as it only adds support for "Live Event" streams, not for general live streams and seeking support general live streams. I have another patch locally to enable that, but I am waiting for some feedback from @ojw28 in the PR first, before making it more complicated and tackle multiple problems at the same time.

ckarthikv:
@thasso - Is there any way I could access the patch implemented locally? Would be of great help.

pajatopmr:
@thasso @ckarthikv If I read the HLS spec correctly (which matches some comments I have read by @ojw28) seeking to 0 should cause a seek to the live edge.  This PR handles the seek to live edge for EVENT type accordingly.  Seek to the live edge via 0L does not work for live (no EVENT playlist type) but I am working on making this change.  @thasso if you have a fix for this I would love to see it.

thasso:
The patch is internal only at this moment, but I can try to get it public during this week maybe. Not sure though if I should merge it with the PR or keep it separate, cause it might indeed solve both live and event-live.

General note the seek(0) questions: I don't think this is spec specific, but simply how seeking in HLS Live in master/dev is currently implemented. Take a look here 

https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java#L281

and you will see that in case of a _live_ stream, _any_ seek operation will jump to the current live edge. NOTE that this _did change_ in this PR and you will need to seek to the end of the seek range as reported by the events.

jonasevcik:
Hi @ojw28  I've written this for dev-v2 in much simpler fashion - no new entities, nor listeners. You can see the changes here: https://github.com/google/ExoPlayer/compare/dev-v2...jonasevcik:startover?expand=1
If it's ok with you, I can create a pull request.

jglanfield:
No updates on this?

PR #884, opened 2015-10-20
Added id property to the MediaFormat class

RikHeijdens:
In our application we give people control of the "selected" track on the player for DASH streams however there was no way for people to identify the track, unless if you knew the bitrate or width/height properties. This change adds the `id` property to the MediaFormat class which is read from the DASH Media Presentation Description. This way it's easier for users to identify tracks on the player.

CLA should be signed: https://groups.google.com/forum/#!forum/jwplayer-google-contributors

Feedback is appreciated, feel free to ask questions.
Thanks.


ojw28:
Hey. Thanks for the contribution. I'm a little bit confused by this comment:

*We saw a case where a user would load a stream with various video representations that had the same height and width but a different bitrate. In our UI there was no way to differentiate the streams from each other*

It's unclear why you need the identifier to resolve this issue. The getTrackFormat method gives you the MediaFormat of the track, which includes the bitrate. So why not just look at the bitrate field to solve this case?

The above question aside, I do think it's a good idea to propagate the identifier, and can see cases where it would be useful.

We'll try and get this merged soon, or an equivalent change in place. One thing we're mulling over is whether a separate field is the best option, or whether we should just convert the existing field to a string. I think we're thinking that converting the existing field is probably the best option.

library/src/main/java/com/google/android/exoplayer/MediaFormat.java
```diff
@@ -38,7 +38,10 @@
    * the timestamps of their parent samples.
    */
   public static final long OFFSET_SAMPLE_RELATIVE = Long.MAX_VALUE;
-
+  /**
+   * An identifier for the format.
+   */
+  public final String id;
```
ojw28: As per comment, it seems better just to convert trackId into a string?

library/src/main/java/com/google/android/exoplayer/MediaFormat.java
```diff
@@ -206,31 +210,31 @@ public static MediaFormat createFormatForMimeType(int trackId, String mimeType,
...
   public MediaFormat copyWithDurationUs(long durationUs) {
-    return new MediaFormat(trackId, mimeType, bitrate, maxInputSize, durationUs, width, height,
+    return new MediaFormat(id, trackId, mimeType, bitrate, maxInputSize, durationUs, width, height,
         rotationDegrees, pixelWidthHeightRatio, channelCount, sampleRate, language,
         subsampleOffsetUs, initializationData, adaptive, maxWidth, maxHeight);
   }
 
   public MediaFormat copyAsAdaptive() {
-    return new MediaFormat(trackId, mimeType, NO_VALUE, NO_VALUE, durationUs, NO_VALUE, NO_VALUE,
+    return new MediaFormat(id, trackId, mimeType, NO_VALUE, NO_VALUE, durationUs, NO_VALUE, NO_VALUE,
```
ojw28: This is problematic, because you're creating an adaptive MediaFormat (i.e. one that really spans multiple individual formats) from a non-adaptive one (i.e. an arbitrary one of the individual formats), and the adaptive MediaFormat is getting the id from the arbitrary individual format that you use.

You should probably take a look at the places that call this method, and either decide to set the id to null here or allow a different id to be passed in as an argument, so you can set it to some other value, depending on what you think makes most sense.

ojw28:
Looks pretty good. If you convert the existing id field to a string and fix the id when creating adaptive formats to do something sensible, then we'll merge this. Thanks!

RikHeijdens:
> It's unclear why you need the identifier to resolve this issue. The ExoPlayer.getTrackFormat method gives you the MediaFormat of the track, which includes the bitrate. So why not just look at the bitrate field to solve this case?

We would like to give users the possibility to use a 'custom'  name for their tracks instead of using a predfined format. We figured out we could make this possible by using the `id` field in DASH streams and the `NAME` field in HLS streams. See https://github.com/google/ExoPlayer/pull/885 for the HLS variant of our implementation for this.

I will convert the existing field and fix the id when creating adaptive formats.

By the way: I was just told we had some technical problems signing the corporate CLA, one of my supervisors was not able to sign it through the webpage. He contacted Google and he's working on it.

library/src/androidTest/java/com/google/android/exoplayer/MediaFormatTest.java
```diff
@@ -45,20 +45,20 @@ public void testConversionToFrameworkFormat() {
     initData.add(initData2);
 
     testConversionToFrameworkFormatV16(MediaFormat.createVideoFormat(
-        MediaFormat.NO_VALUE, "video/xyz", 5000, 102400, 1000L, 1280, 720, initData));
+            null, MediaFormat.NO_VALUE, "video/xyz", 5000, 102400, 1000L, 1280, 720, initData));
```
ojw28: Please put the indentation back to what it used to be like (4 chars) (ditto x2 below).

library/src/main/java/com/google/android/exoplayer/MediaFormat.java
```diff
@@ -38,12 +38,11 @@
    * the timestamps of their parent samples.
    */
   public static final long OFFSET_SAMPLE_RELATIVE = Long.MAX_VALUE;
-
```
ojw28: Add blank line back again?

library/src/main/java/com/google/android/exoplayer/dash/DashChunkSource.java
```diff
@@ -846,6 +846,7 @@ public void run() {
 
     public ExposedTrack(MediaFormat trackFormat, int adaptationSetIndex, Format fixedFormat) {
       this.trackFormat = trackFormat;
+
```
ojw28: Rm additional blank line.

library/src/main/java/com/google/android/exoplayer/dash/DashChunkSource.java
```diff
@@ -865,9 +866,9 @@ public ExposedTrack(MediaFormat trackFormat, int adaptationSetIndex, Format[] ad
 
     public boolean isAdaptive() {
       return adaptiveFormats != null;
-    }
 
-  }
+    }
```
ojw28: Revert bad style change.

ojw28:
Happy to merge this once stylistic alterations are reverted. I don't think the CLA has been signed yet. Any update? Thanks!

RikHeijdens:
I have no idea why Android Studio screwed up the code identation, it should detect existing file identation and use that for editing.

The person responsible for signing the CLA is out of office today. I'll try to update you on the CLA tomorrow.

Thanks.

RikHeijdens:
@ojw28 The CLA has been signed last Friday.

ojw28:
I don't see the CLA on this side. Do you know what name it was signed under? Has your GitHub username been added to it?

ojw28:
Note - We've added a googlebot hook to the repository that will reply to pull requests stating whether it found a signed CLA. If you open a dummy pull request you should receive an update from the googlebot there stating whether it found a CLA.

johnluther:
@ojw28, I signed the CLA under my name and email address (jluther@jwplayer.com). I listed the corporation as "Longtail Video DBA JW Player."

JL

ojw28:
I don't see a corresponding entry on my side. Are you sure the corporate CLA is active? I can't actually see the link at the top of this post (because I'm not a member of the group), so I don't know if there's anything useful there. I think we probably also need the email addresses of everyone who will be sending pull requests to be added to the CLA, so that there's something linking them to it.

ojw28:
CLA looks good now. Feel free to resolve the merge conflict and update if you have a chance. Otherwise I can probably do this myself tomorrow. Thanks!

PR #885, opened 2015-10-20
Parse HLS NAME label from the MasterPlaylist

RikHeijdens:
For the same usecase as mentioned in: https://github.com/google/ExoPlayer/pull/884 we needed to parse the `NAME` label from the HlsMasterPlaylist and have it available in the `variants` array.

CLA should be signed: https://groups.google.com/forum/#!forum/jwplayer-google-contributors

Feel free to ask questions, feedback is appreciated.
Thanks.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
@@ -153,13 +153,12 @@ public final class HlsPlaylistParser implements UriLoadable.Parser<HlsPlaylist> 
         String type = HlsParserUtil.parseStringAttr(line, TYPE_ATTR_REGEX, TYPE_ATTR);
         if (SUBTITLES_TYPE.equals(type)) {
           // We assume all subtitles belong to the same group.
-          name = HlsParserUtil.parseStringAttr(line, NAME_ATTR_REGEX, NAME_ATTR);
+          String subtitleName = HlsParserUtil.parseStringAttr(line, NAME_ATTR_REGEX, NAME_ATTR);
           String uri = HlsParserUtil.parseStringAttr(line, URI_ATTR_REGEX, URI_ATTR);
           String language = HlsParserUtil.parseOptionalStringAttr(line, LANGUAGE_ATTR_REGEX);
           boolean isDefault = HlsParserUtil.parseOptionalBooleanAttr(line, DEFAULT_ATTR_REGEX);
           boolean autoSelect = HlsParserUtil.parseOptionalBooleanAttr(line, AUTOSELECT_ATTR_REGEX);
-          subtitles.add(new Subtitle(name, uri, language, isDefault, autoSelect));
-          name = null;
+          subtitles.add(new Subtitle(subtitleName, uri, language, isDefault, autoSelect));
         } else {
           // TODO: Support other types of media tag.
         }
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
```diff
@@ -152,18 +153,20 @@ private static HlsMasterPlaylist parseMasterPlaylist(LineIterator iterator, Stri
         String type = HlsParserUtil.parseStringAttr(line, TYPE_ATTR_REGEX, TYPE_ATTR);
         if (SUBTITLES_TYPE.equals(type)) {
           // We assume all subtitles belong to the same group.
-          String name = HlsParserUtil.parseStringAttr(line, NAME_ATTR_REGEX, NAME_ATTR);
+          name = HlsParserUtil.parseStringAttr(line, NAME_ATTR_REGEX, NAME_ATTR);
```
ojw28: Can we just use a local variable for this (could probably be called subtitleName)? Then get rid of L162 (no longer necessary).

library/src/main/java/com/google/android/exoplayer/hls/Variant.java
```diff
@@ -27,10 +27,15 @@
   public final String url;
   public final Format format;
 
-  public Variant(int index, String url, int bitrate, String codecs, int width, int height) {
+  public Variant(int index, String name, String url, int bitrate, String codecs, int width, int height) {
     this.url = url;
-    format = new Format(Integer.toString(index), MimeTypes.APPLICATION_M3U8, width, height, -1, -1,
-        -1, bitrate, null, codecs);
+    if (name != null) {
```
ojw28: Hm. I guess it would be a weird case (some variants with names, others without in the master playlist), but it would be strange if we ended up exposing some tracks with indices for names and other tracks with actual names. How would you feel about:

No name: Integer.toString(index)
With name: Integer.toString(index) + ':' + name

The same will apply for the TODO in the WebmExtractor, where I would guess some tracks having names and others not having them may be more likely.

RikHeijdens: I don't really feel like mixing up NAME labels and track identifiers, maybe we should store the trackId (or the NAME label) in another variable? What do you think about that?

library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
RikHeijdens: Yes I will change this tomorrow, the main reason I reused this variable was because it already had the identifier `name`.

library/src/main/java/com/google/android/exoplayer/hls/Variant.java
ojw28: Meh. I don't really want to start adding extra fields, unless there's some kind of dire need for them. In the worst case we might end up four ids (index-in-manifest, index-in-container, name-in-manifest, name-in-container), and it all starts getting a bit out of hand :). Let's just leave it how you have it for now and see if works out ok. Although as a nit, can we change this to be a bit more concise, like:

String formatName = name != null ? name : Integer.toString(index);
format = new Format(formatName, MimeTypes.APPLICATION_M3U8, width, height, -1, -1,
    -1, bitrate, null, codecs);

Thanks!

ojw28:
Unclear whether we both commented on this at the same time, or if you just replied really fast :). See inline comment I just posted in any case; I think it's probably the simplest thing to be going on with.

RikHeijdens:
It seems like we commented on the same time, I didn't see your comment before committing. Well I think we're on the same page here :)

ojw28:
Ack. If you put it back to how it was with the simplification suggestion then I'll merge in. Thanks!

RikHeijdens:
Done, see: https://github.com/google/ExoPlayer/commit/d49d3e2cd3f92d87737eed9b5c4a35224a0bca6f

Issue #151, opened 2014-11-19
Implement WebVTT support for HLS


andudo:

ojw28:
Question for anyone following this issue - Do we envisage a case where we need to support both VTT and embedded EIA608 simultaneously (in a single playback)? Or if VTT is present, is it always safe to assume that we can just rely on VTT tracks? I'm hoping we can just rely on VTT if they're present.

jeoliva:
From my experience, people is using EIA608 or VTT, never both at the same time.

johanlindquist:
Yes, same thing from my experience

Ood-Tsen:
We may download vtt files separately in the class 'HlsChunkSource'.

perchrh:
@ojw28 Yes, that's a good simplification, I don't see a case where you would need both at the same time. Subtitles declared in the playlist as separate downloads (VTT) supersedes subtitles embedded in the TS (EIA?08 and TTML), as I see it.

caleb-allen:
+1 for WebVTT support for HLS

nigelmegitt:
@perchrh @ojw28 It seems like an over-simplification to me. The VTT track may not be subtitles but instead could be chapterisation or other metadata, intended to be additional to the embedded subtitles/captions rather than a replacement. Even if they are intended to be the same subtitle/caption content, there are likely to be differences in styling and positioning between the two available tracks. At least give the user the opportunity to select from the multiple tracks available even if only presenting one of them, rather than hard coding player prioritisation of available tracks.

jlacivita:
it might be true that 80% of the time you'd want WebVTT over CEA-608/708, it's not alway true.

Generally, you want to expose a list of TextTracks that are available, and allow each one to be activated/deactivated by the API. It's up to the App using ExoPlayer to decide whether there's hard coded logic to pick CEA-608/708 over WebVTT or whether more than one track can be rendered at once.

perchrh:
@jlacivita Yes, and the API could expose mimetype or subtitle type together with the other TextTrack information. This way the API user can implement his or her own logic for subtitle preferences. 

The implementation could still be done in two steps: first support both subtitle sources and prefer one of them, then make both available. The last step will be quick to implement, I think.

jlacivita:
Hi guys.  Any update on the priority of HLS support with WebVTT?

ojw28:
First pieces that we need for this were:
[...]

Issue #514, opened 2015-06-02
Ability to switch audio track during playback


IanDBird:
I'm attempting to add support to our video player that allows individual audio tracks to be selected during playback. Initially i'm focussing on my ExtractorRendererBuilder (with MP4s) but this really affects any media which can contain multiple different audio tracks.

#### First Issue

`MediaCodecAudioTrackRenderer` by default will select, and use, the first audio track provided by the `SampleSource`. It therefore seemed appropriate to extend this class and override the `handlesMimeType` used during `onPrepare` in order to try and "select" the audio stream that you want. A problem then arrises that "mimeType" is not a unique identifier within a video. For example, you could easily have an MP4 which has 2 AAC tracks, the main feature vs the directors commentary. In order to make it easier to identify individual tracks, it might be worth considering including the `trackIndex` into the method, for example:

```
protected boolean handlesMimeType(int trackIndex, String mimeType)
```

With this change in place, it should then be possible to create N individual audio track renderers (one for each known audio track), and to disable / enable them on demand.

#### Second Issue

There are two possible (related) issues with what I described above:
 - Are there any overheads associated with having lots of separate instances of `MediaCodecAudioTrackRenderer`?
 - What if you don't know up front how many audio tracks are available?

The audio tracks are available via the `SampleSource` but this is prepared after the `MediaCodecAudioTrackRenderer`'s have been created. I therefore looked at what it would take to have a single `MediaCodecAudioTrackRenderer` that when we needed to "change" it (because we needed to select a new track index) we would attempt to re-use the original `SampleSource` but create a new instance of the audio track renderer. However, there doesn't appear to be any way to replace/substitute a new track renderer like this. As mentioned in my approach above, the `handlesMimeType` is called when being prepared. However, if you call the `ExoPlayer.prepare` a second time it will dispose of all the original renderers and eventually their `SampleSource` (that we're trying to re-use).

Is there something that i'm missing on how best to implement this?

Ood-Tsen:
It seems only SS and Dash are support switch between multiple audio tracks.

Those have implement [MultiTrackChunkSource](https://github.com/google/ExoPlayer/blob/dev/library/src/main/java/com/google/android/exoplayer/chunk/MultiTrackChunkSource.java).

ojw28:
Correct. Also, when we implement this for HLS, we'll use MultiTrackChunkSource there as well.

[...]

PR #891, opened 2015-10-22
Implement Webvtt support for HLS

ferrytanu:
Addded WebVttChunkSource which can be used with ChunkSampleSource to
manage downloading subtitle media playlist, and subtitle chunk. For live
streaming subtitle, it refreshes the media playlist every segment
duration, and half of segment duration when the playlist is stale.

Added TimestampOffsetListener file to synch between Audio/Video frame
with subtitle. The value is used as an offset to subtitle timestamp,
this is required esp in live streaming subtitle.

Bringing back X-TIMESTAMP-MAP to handle VOD with initial PTS > 0

Issue: #151


ojw28:
[Ignore this comment. Just trying to kick googlebot] I signed it!

ojw28:
Note that it's necessary to sign the Google CLA to contribute to this repository, as described [here](https://github.com/google/ExoPlayer/blob/master/CONTRIBUTING.md). I'll take a look at this this week. From a brief initial scan, I'm not sure I like the extent to which timestamps are being propagated between various components. However it may be possible to clean it up without too much effort.

demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
```diff
@@ -125,6 +125,8 @@ public Sample(String name, String contentId, String uri, int type) {
         + "prog_index.m3u8", PlayerActivity.TYPE_HLS),
     new Sample("Apple ID3 metadata", "http://devimages.apple.com/samplecode/adDemo/ad.m3u8",
         PlayerActivity.TYPE_HLS),
+    new Sample("Apple WebVtt", "https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8",
```
ojw28: This exact example is already added at L117, so you can revert changes to this file.

library/src/main/java/com/google/android/exoplayer/hls/WebvttChunkSource.java
```diff
@@ -0,0 +1,342 @@
...
+            for(Subtitle subtitle: subtitles) {
+                Format format = new Format(subtitle.name, MimeTypes.TEXT_VTT, MediaFormat.NO_VALUE,
+                        MediaFormat.NO_VALUE, MediaFormat.NO_VALUE, MediaFormat.NO_VALUE,
+                        MediaFormat.NO_VALUE, MediaFormat.NO_VALUE, subtitle.language);
+
+                // timestampOffset value is unknown at this time, set 0 as default
+                // subsampleOffset must be set to 0, if not it will be set to OFFSET_SAMPLE_RELATIVE
+                // which will treated as relative time, as a result it won't work for WebVtt
+                // see SubtitleParserHelper for details
+                MediaFormat trackFormat = MediaFormat.createTextFormat(tracks.size(), format.mimeType,
```
ojw28: First argument to createTextFormat should be format.id

library/src/main/java/com/google/android/exoplayer/text/SubtitleParserHelper.java
```diff
@@ -179,8 +179,13 @@ private void handleSample(long sampleTimeUs, SampleHolder holder) {
       if (sampleHolder != holder) {
         // A flush has occurred since this holder was posted. Do nothing.
       } else {
-        this.result = new PlayableSubtitle(parsedSubtitle, subtitlesAreRelative, sampleTimeUs,
-            subtitleOffsetUs);
+        // it's more accurate to read startTime from eventTime compare to sampleTime
```
ojw28: I'm pretty sure you shouldn't need to be changing this class. The sample works perfectly without this. Can we revert it?

ojw28: Oh right, I see what this is for. Although maybe we could clean this up a bit by doing the following:

- If there aren't any events in the subtitle, this.result may as well just be set to null, I think?
- There's no need for the PlayableSubtitle constructor to take startTimeUs. It should be able to calculate it internally.

ojw28: Actually, you still need startTimeUs for the absolute timestamp case. But you could do the following here:

```
        this.result = parsedSubtitle.getEventTimeCount() == 0 ? null : new PlayableSubtitle(
            parsedSubtitle, subtitlesAreRelative, sampleTimeUs, subtitleOffsetUs);
```

And then implement the PlayableSubtitle constructor as:

```
    this.subtitle = subtitle;
    this.offsetUs = (isRelative ? startTimeUs : 0) + offsetUs;
    this.firstEventTimeUs = getEventTime(0);
```

Where `firstEventTimeUs` replaces `startTimeUs` as the member variable.

demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
ferrytanu: @ojw28 thank you for looking into this. CLA has been signed.

You are right about duplicate sample. I'll remove it.

library/src/main/java/com/google/android/exoplayer/text/SubtitleParserHelper.java
ferrytanu: Gotcha, let's revisit live subtitles on different PR. I'll revert the class for the time being

ojw28:
One other thing to note with this change is that I don't think it'll handle either TS timestamp rollover or discontinuities properly (and it seems very non-trivial to do so, also). Any idea how that might be done within this model?

ferrytanu:
The changes covers that actually.
In WebvttParser, we parse the X-TIMESTAMP-MAP=LOCAL, to get the timestamp, and use that to offset the cue timestamp. As such, the cue timestamp will not be greater than 26.5 hours (which is the limitation of the 33 bits PTS). It serves well so far.

ojw28:
I don't think that's the correct behaviour, unfortunately. Other HLS components use PtsTimestampAdjuster to un-wrap the timestamps. The timestamps parsed from X-TIMESTAMP-MAP are not subject to this un-wrapping logic, and so do wrap. So you'll end up with the WebVTT timestamps wrapping but the timestamps for other media streams not wrapping, I think?

I have a feeling it'll fall out nicer if you parse just X-TIMESTAMP-MAP from the sample somewhere in the hls package, where the timestamp adjustment can be applied.

ojw28:
Discontinuities are even more problematic. HlsChunkSource necessarily instantiates a new PtsTimestampAdjuster when it sees a discontinuity. Matching discontinuities should be present in the WebVTT playlist too, but it's necessary to make sure the right PtsTimestampAdjuster is applied before and after each discontinuity, which is not currently enforced.

ojw28:
I think things actually fall out more nicely using the existing HlsChunkSource implementation in conjuction with a lightweight implementation of the `Extractor` interface specific for HLS+WebVtt.

The `WebVttExtractor` would scan just as much of each sample as is required to establish the TS->LOCAL mapping and the first cue timestamp. This would allow the correct timestamp to be calculated and pushed through `PtsTimestampAdjuster`, which is in-line with what `TsExtractor` does for the other media streams, and ensures that rollover is handled the same in both cases. Handling discontinuities would require extending `PtsTimestampAdjuster` to be discontinuity aware, rather than replacing it with a new instance at each discontinuity point as is currently the case.

Note that with the above approach `WebvttParser` does not need to parse `X-TIMESTAMP-MAP`, which would be done by `WebVttExtractor`. This is actually quite a nice property, because it keeps the HLS specific logic isolated within the hls package.

I intend to give this approach a try over the next couple of days, to see if it's viable.

ferrytanu:
Sounds like that's the more right approach. Wondering how we can avoid calling the webvtt parsing twice, as it will be in WebVttExtractor and SubtitleParserHelper.
Thanks for looking into this, can't wait to see how it being refactor-ed. :smile:

ojw28:
We wont really be parsing the VTT twice. The extractor part will only be parsing as far as the first cue header; it wont need to look any further in the data. I have a hacked together version sort of working, and will look more next week.

amed500:
Hey @ojw28  i have just question about this Pull, will it be merged in the new version of ExoPlayer ?, i need WebVtt Subtitle in My application, for now the version 1.5.2 for Hls it shows only Closed captions subtitle. thank you

ojw28:
This change will not be merged, but we are actively working on a different approach to enabling WebVTT + HLS. Some of the first changes were pushed yesterday (https://github.com/google/ExoPlayer/commit/4bb8bea9528a803a201b809e8bd4abe3f0b460e8, https://github.com/google/ExoPlayer/commit/72f093c4f6d66cea021756e1cbe931b52f55a79e) and more will follow shortly.

amed500:
@ojw28  thank you for your fast answer, will be waiting this feature then, but the multi Audio language Hls will be supported too ?

ojw28:
It's a little further out. There's only so much we can do at once ;). No ETA as of this moment.

amed500:
@ojw28  i really need that feature also, is there any work around you could suggest me to implement this feature ? i saw some workings from this issue #514, thanks

tommymc:
@ojw28 i've seen, you've written WebVttExtractor in 1.5.3rc. Can you add add sample how we can use webvtt + hls? thx

ojw28:
It's not yet possible to fully wire everything up. I guess it's a little misleading that the class was part of the release, given it's still work in progress. Sorry about that!

tommymc:
@ojw28 the version of @ferrytanu use webvtt internal, i means a lots of small webvtt files in playlist m3u8, does your refactor version support webvtt extenal (only one vtt file for one language of the video)?
Thx

ojw28:
Closing this because support should be landing next week. Much thanks for the effort though; it was definitely a helpful reference!

PR #892, opened 2015-10-23
[RollingSampleBuffer] prevent internal runtime error (NoSuchElementException)

pipelka:
Hi,

I'm developing a TVInputService for Android TV using a custom Extractor class.
Unfortunately I'm sometimes facing a "NoSuchElementException" in 
"RollingSampleBuffer.dropDownstreamTo(RollingSampleBuffer.java:321".

This patch helped to fix the problem for me.

Regards,
Alex

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/RollingSampleBuffer.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/RollingSampleBuffer.java
@@ -317,7 +317,7 @@ import java.util.concurrent.LinkedBlockingDeque;
   private void dropDownstreamTo(long absolutePosition) {
     int relativePosition = (int) (absolutePosition - totalBytesDropped);
     int allocationIndex = relativePosition / allocationLength;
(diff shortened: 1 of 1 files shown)


ojw28:
There isn't enough information provided to determine why you're encountering the issue, but it's likely to be a bug in your extractor rather than an issue with RollingSampleBuffer. It doesn't look like the check being added should be necessary.

pipelka:
I'm quite sure my extractor still has a lot of problems ;-)
I opened the PR because this exception should not be possible but it happens, ...

I think it may be some threading issue in my code. I'll recheck that.

ojw28:
Well, the exception likely only occurs because your extractor code is broken. You should fix the root cause, not add a check in the library to hide the issue :).

dnutcracker:
What do you mean by 'extractor code broken'? I'm using the library's
ExtractorSampleSource out of the box.

pipelka:
It's my custom extractor code that's broken.

PR #897, opened 2015-10-25
Add error messages to assertion checks

noamtamim:
Finding the inconsistency in my test stream took a lot of debugging time. An error message would lead me directly to the offending element.


ojw28:
I don't understand how this actually saves you any time. As soon as you have a stack trace (for the assertion failure), it should only takes a few seconds to look up the corresponding line of source code and see exactly what condition has failed.

By convention, we're trying to avoid adding strings like this, for the most part.

PR #915, opened 2015-11-02
Adds mpeg2 video support in MPEG−TS extractor

sureshcnvidia:
Final changes to support mpeg2 video in MPEG-TS streams.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/ts/H262Reader.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/ts/H262Reader.java
@@ -1,0 +1,296 @@
+package com.google.android.exoplayer.extractor.ts;
+
+import com.google.android.exoplayer.C;
+import com.google.android.exoplayer.MediaFormat;
+import com.google.android.exoplayer.extractor.TrackOutput;
+import com.google.android.exoplayer.smoothstreaming.SmoothStreamingManifest;
+import com.google.android.exoplayer.util.Assertions;
+import com.google.android.exoplayer.util.MimeTypes;
+import com.google.android.exoplayer.util.ParsableByteArray;
+
+import android.util.Log;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Parses a continuous mpeg2 byte stream and extracts individual frames.
+ */
+/* package */ final class H262Reader extends ElementaryStreamReader {
+    private static final String TAG = "H262Reader";
+
+    private final int H262_FRAME_TYPE_IFRAME = 0;
+    private final int H262_FRAME_TYPE_AUD = 1;
+
+    private boolean hasSequenceCompleted;
+    private boolean foundFirstSample;
+
+    // State that should not be reset on seek.
+    private boolean hasOutputFormat;
+
+    // State that should be reset on seek.
+    private long totalBytesWritten;
+    private final boolean[] prefixFlags;
+
+    // Per sample state that gets reset at the start of each sample.
+    private boolean isKeyframe;
+    private long samplePosition;
+    private long sampleTimeUs;
+
+    public H262Reader(TrackOutput output) {
+        super(output);
+        prefixFlags = new boolean[4];
+    }
+
+    @Override
+    public void seek() {
+        clearPrefixFlags(prefixFlags);
+        hasSequenceCompleted = false;
+        samplePosition = 0;
+        totalBytesWritten = 0;
+    }
+
+    @Override
+    public void consume(ParsableByteArray data, long pesTimeUs, boolean startOfPacket) {
+        while (data.bytesLeft() > 0) {
+            int offset = data.getPosition();
+            int limit = data.limit();
+            byte[] dataArray = data.data;
+            // Append the data to the buffer.
+            totalBytesWritten += data.bytesLeft();
+            output.sampleData(data, data.bytesLeft());
+
+            if (!hasSequenceCompleted) {
+               hasSequenceCompleted = checkForCompleteSequence(dataArray, offset, limit, prefixFlags);
+               if (hasSequenceCompleted && !hasOutputFormat)
+                   output.format(parseMediaFormatFromSeqHeader(dataArray, offset, limit, prefixFlags));
+            }
+
+            while (hasOutputFormat && offset < limit) {
+                int nextframeAtOffset = findValidStartCode(dataArray, offset, limit, prefixFlags);
+                if (nextframeAtOffset < limit) {
+                    int frameType = getH262FrameType(dataArray, nextframeAtOffset);
+                    int bytesWrittenPast = limit - nextframeAtOffset;
+                    switch (frameType) {
+                        case H262_FRAME_TYPE_IFRAME:
+                            isKeyframe = true;
+                            break;
+                        case H262_FRAME_TYPE_AUD:
+                            if (foundFirstSample) {
+                                int flags = isKeyframe ? C.SAMPLE_FLAG_SYNC : 0;
+                                int size = (int) (totalBytesWritten - samplePosition) - bytesWrittenPast;
+                                output.sampleMetadata(sampleTimeUs, flags, size, bytesWrittenPast, null);
+                            }
+                            foundFirstSample = true;
+                            samplePosition = totalBytesWritten - bytesWrittenPast;
+                            sampleTimeUs = pesTimeUs;
+                            isKeyframe = false;
+                            break;
+                    }
+                    offset = nextframeAtOffset + 4;
+                } else {
+                    offset = limit;
+                }
+            }
+        }
+    }
+
+    @Override
+    public void packetFinished() {
+        // Do nothing.
+    }
+
+    /**
+     * Locates the next frame start code (0x00, 0x00, 0x01, 0xXX)
+     * where XX can be anything lesser than 0x01 or more than 0xaf
+     * @param data The buffer whose position should be advanced.
+     * @param startOffset The offset of the data in {@code data}.
+     * @param endOffset The limit (exclusive) of the data in {@code data}.
+     * @param prefixFlags A boolean array whose first four elements are used to store the state
+     *     required to detect H262 frame units where the frame unit prefix spans array boundaries. The array
+     *     must be at least 4 elements long.
+     * @return The offset into the buffer where a new frame starts from.
+     */
+    private int findValidStartCode(byte[] data, int startOffset, int endOffset,
+                                   boolean[] prefixFlags ) {
+        int length = endOffset - startOffset;
+
+        Assertions.checkState(length >= 0);
+        if (length == 0) {
+            return endOffset;
+        }
+
+        if (prefixFlags != null) {
+            if (prefixFlags[0]) {
+                clearPrefixFlags(prefixFlags);
+                return startOffset - 4;
+            } else if (length > 1 && prefixFlags[1] && checkMpegVideoFrame(data[startOffset])) {
+                clearPrefixFlags(prefixFlags);
+                return startOffset - 3;
+            } else if (length > 2 && prefixFlags[2] && data[startOffset] == 1
+                    && checkMpegVideoFrame(data[startOffset + 1])) {
+                clearPrefixFlags(prefixFlags);
+                return startOffset - 2;
+            }  else if (length > 3 && prefixFlags[3] && data[startOffset] == 0
+                    && data[startOffset + 1] == 1 && checkMpegVideoFrame(data[startOffset + 1])) {
+                clearPrefixFlags(prefixFlags);
+                return startOffset - 1;
+            }
+        }
+
+        int limit = endOffset - 1;
+        for (int i = startOffset + 3; i < limit; i += 4) {
+            if ((data[i - 1] & 0xFE) != 0) {
+                // There isn't a frame prefix here, or at the next two positions. Do nothing and let the
+                // loop advance the index by four.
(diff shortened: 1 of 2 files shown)


sureshcnvidia:
I signed it!

sureshcnvidia:
I signed it!

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+                i -= 3;
+            }
+        }
+        return endoffset;
+    }
+
+    public  boolean checkMpegVideoFrame (int data) {
+        int byte1 = data & 0xFF;
+        if (byte1 < 0x01 || byte1 > 0xaf){
+            isSync = (byte1 == 0xb3) ? true : false;
```
ojw28: Are you sure this is sufficient information to determine whether it's a keyframe? 6.1.1.6 of 13818-2:2013 indicates that there could be an I-frame or a P-frame following a sequence header. But presumably only I-frames should be marked as sync frames.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Parses the sample size from sequence header and prepares media format
+     * @param data The buffer whose position should be advanced.
+     * @param offset The offset of the data in {@code data}.
+     * @return None
+     */
+    private void findSampleSizeAndPrepareMediaFormat(byte[] data, int offset) {
+        if (seqHeaderFound) {
```
ojw28: seqHeaderFound is always true here, so you don't need this check.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Locates the next sync word, advancing the position to the byte that immediately follows it.
+     * If a sync word was not located, the position is advanced to the limit.
+     * @param data The buffer whose position should be advanced.
+     * @param startoffset The offset of the data in {@code data}.
+     * @param endoffset The limit (exclusive) of the data in {@code data}.
+     * @return True if a sync word position was found. False otherwise.
+     */
+
+    private boolean skipToNextSync(byte[] data, int startoffset, int endoffset) {
```
ojw28: I don't think the way this class looks for start codes is robust against the code being split half in one byte[] and half in the next. NalUnitUtil.findNalUnit with prefixFlags non-null will do this for you (see the H264Reader as an example).

Note - If you do use this, which I think you should do, then we'll rename it to something without "nal" in the name. Don't do this in this change though.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+            int frameAtOffset = 0;
+            int offset = data.getPosition();
+            int limit = data.limit();
+            byte[] dataArray = data.data;
+            // Append the data to the buffer.
+            totalBytesWritten += data.bytesLeft();
+            output.sampleData(data, data.bytesLeft());
+
+            if (!seqHeaderFound && skipToNextSync(dataArray, offset, limit)) {
+                if (seqHeaderFound) {
```
ojw28: seqHeaderFound is always true here?

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+ */
+/* package */ final class Mpeg2Reader extends ElementaryStreamReader {
+    private static final String TAG = "Mpeg2Reader";
+
+    private int flags;
+    private boolean seqHeaderFound;
+    private boolean frameCollected;
+    private boolean isSync;
+
+    private int width;
```
ojw28: These don't need to be member variables. They can be local in the one method that uses them.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Parses a continuous mpeg2 byte stream and extracts individual frames.
+ */
+/* package */ final class Mpeg2Reader extends ElementaryStreamReader {
+    private static final String TAG = "Mpeg2Reader";
+
+    private int flags;
```
ojw28: This doesn't need to be a member variable. It can be local to the one method that uses it.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+            int offset = data.getPosition();
+            int limit = data.limit();
+            byte[] dataArray = data.data;
+            // Append the data to the buffer.
+            totalBytesWritten += data.bytesLeft();
+            output.sampleData(data, data.bytesLeft());
+
+            if (!seqHeaderFound && skipToNextSync(dataArray, offset, limit)) {
+                if (seqHeaderFound) {
+                    flags = C.SAMPLE_FLAG_SYNC;
```
ojw28: You should IMO set isSync to true here rather than touching flags...

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+
+            if (startOfPacket && seqHeaderFound) {
+
+                frameAtOffset = findValidStartCode(dataArray, offset, limit);
+
+                if (frameAtOffset < limit) {
+                    frameCollected = true;
+                    if (!foundFirstSample)
+                        flags = (isSync == true) ? C.SAMPLE_FLAG_SYNC : 0;
```
ojw28: Move this line to just before the output.sampleMetadata call, and make flags a local variable there.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+                    flags = C.SAMPLE_FLAG_SYNC;
+                    foundFirstSample = true;
+                    timeUs = pesTimeUs;
+                    findSampleSizeAndPrepareMediaFormat(dataArray, offset); // Seq header has been already been found
+                }
+            }
+
+            if (startOfPacket && seqHeaderFound) {
+
+                frameAtOffset = findValidStartCode(dataArray, offset, limit);
```
ojw28: I'm quite unclear in general what an h262 decoder expects to be fed. Could you clarify exactly what the decoder is expecting to receive, and how it's supposed to be segmented into samples? Thanks!

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+
+import android.util.Log;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Parses a continuous mpeg2 byte stream and extracts individual frames.
+ */
+/* package */ final class Mpeg2Reader extends ElementaryStreamReader {
```
ojw28: IMO call this H262Reader, for consistency with the H264/H265 readers.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+            // Append the data to the buffer.
+            totalBytesWritten += data.bytesLeft();
+            output.sampleData(data, data.bytesLeft());
+
+            if (!seqHeaderFound && skipToNextSync(dataArray, offset, limit)) {
+                if (seqHeaderFound) {
+                    flags = C.SAMPLE_FLAG_SYNC;
+                    foundFirstSample = true;
+                    timeUs = pesTimeUs;
+                    findSampleSizeAndPrepareMediaFormat(dataArray, offset); // Seq header has been already been found
```
ojw28: It's probably clearer if you do this here:

```
          if (!hasOutputFormat) {
            output.format(parseMediaFormatFromSeqHeader(dataArray, offset));
            hasOutputFormat = true;
          }
```

Where findSampleSizeAndPrepareMediaFormat has been renamed to parseMediaFormatFromSeqHeader and has been made to return the format.

Note that a reader should call output.format only once, even if seek is called. So you need a hasOutputFormat boolean to ensure this.

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+        int byte1 = data & 0xFF;
+        if (byte1 < 0x01 || byte1 > 0xaf){
+            isSync = (byte1 == 0xb3) ? true : false;
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Parses the sample size from sequence header and prepares media format
```
ojw28: Javadoc is incorrect (sample size isn't parsed).

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
ojw28: I guess it depends on what the underlying decoder expects (which I'm not sure about).

library/src/main/java/com/google/android/exoplayer/extractor/ts/TsExtractor.java
```diff
@@ -47,6 +47,7 @@
   private static final int TS_STREAM_TYPE_ATSC_E_AC3 = 0x87;
   private static final int TS_STREAM_TYPE_H264 = 0x1B;
   private static final int TS_STREAM_TYPE_H265 = 0x24;
+  private static final int TS_STREAM_TYPE_MPEG2 = 0x02;
```
ojw28: TS_STREAM_TYPE_H262

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+
+    private int flags;
+    private boolean seqHeaderFound;
+    private boolean frameCollected;
+    private boolean isSync;
+
+    private int width;
+    private int height;
+
+    private long timeUs;
```
ojw28: Can we use names consistent with the H264 reader? So specifically `sampleTimeUs` and `isKeyframe` rather than `timeUs` and `isSync`. Some grouping would also be nice, as in the H264 case, like:

```
  // State that should not be reset on seek.
  private boolean hasOutputFormat;

  // State that should be reset on seek.
  private boolean foundFirstSample;
  private long totalBytesWritten;

  // Per sample state that gets reset at the start of each sample.
  private boolean isKeyframe;
  private long samplePosition;
  private long sampleTimeUs;
```

library/src/main/java/com/google/android/exoplayer/extractor/ts/Mpeg2Reader.java
```diff
@@ -0,0 +1,189 @@
...
+            // Width and Height
+            int b1 = data[4 + offset] & 0xFF;
+            int b2 = data[5 + offset] & 0xFF;
+            int b3 = data[6 + offset] & 0xFF;
+            width = (b1 << 4) | (b2 & 0xF0);
+            height = (b2 & 0x0F) << 8 | b3;
+            Log.i(TAG, "Width is " + width + " Height is " + height);
+        }
+        List<byte[]> initializationData = new ArrayList<>();
+        output.format(MediaFormat.createVideoFormat(MediaFormat.NO_VALUE, MimeTypes.VIDEO_MPEG2,
```
jacktech24: pixelWidthHeightRatio is not determined and passed, which causes wrong rendered video

library/src/main/java/com/google/android/exoplayer/extractor/ts/H262Reader.java
```diff
@@ -0,0 +1,296 @@
...
+        int b1 = data[4 + offset] & 0xFF;
+        int b2 = data[5 + offset] & 0xFF;
+        int b3 = data[6 + offset] & 0xFF;
+        width = (b1 << 4) | (b2 & 0xF0);
+        height = (b2 & 0x0F) << 8 | b3;
+        aspectRatioCode = (data[7 + offset] & 0xF0) >> 4;
+        float pixelWidthAspectRatio = 1; //default
+        switch(aspectRatioCode) {
+            case 2:
+                pixelWidthAspectRatio = 4 / 3;
```
jacktech24: I think it should be like this, from what I have been testing with the video, because you need to calculate the pixel width scaling ratio
```java
        switch(aspectRatioCode) {
            case 2:
                pixelWidthHeightRatio = (4f / 3) / videoWidthHeightRatio;
                break;
            case 3:
                pixelWidthHeightRatio = (16f / 9) / videoWidthHeightRatio;
                break;
            case 4:
                pixelWidthHeightRatio = (1.21f / 1f) / videoWidthHeightRatio;
                break;
            default:
                // Do nothing.
                break;
        }
```

PR #919, opened 2015-11-03
Added missing MimeTypes to Smooth Streaming manifest parser

miquelbeltran:
This is necessary to avoid parsing errors when playing Smooth Streaming content with DTS or Dolby audio tracks.


ojw28:
Thanks!

PR #934, opened 2015-11-06
Dev

peddisri:
Hi,

These are some of the patches created to work around issues in Fire TV Gen1 Box  & Fire TV Stick,  and Fire OS 4 based Kindle Tablets. 

Let me know your opinion.

Regards,
Srikanth

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/C.java
+++ b/library/src/main/java/com/google/android/exoplayer/C.java
@@ -41,6 +41,11 @@ public final class C {
   public static final long MICROS_PER_SECOND = 1000000L;
 
   /**
+   * The number of milliseconds in one second.
+   */
+  public static final long MILLIS_PER_SECOND = 1000L;
+
+  /**
    * Represents an unbounded length of data.
    */
   public static final int LENGTH_UNBOUNDED = -1;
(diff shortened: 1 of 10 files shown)


peddisri:
Hi,

Amazon has signed CLA as a corporation.

Regards,
Srikanth

ojw28:
a2e2626 - This doesn't look like a good idea. It'll generate a significant amount of garbage because it's creating String objects every time a video frame is rendered. This overhead is incurred even if logging is disabled for the module. We already expose sufficient information for playback analytics via the EventLogger interfaces that many components expose. So I don't think we want to merge this.

245c745 - Happy to take this as an individual pull request, since it looks pretty isolated. Not keen on having an AmazonQuirks class though. Let's just put the special case directly into MediaCodecUtil. We already have other special cases in that class (e.g. blacklisting of known broken decoders).

5d52107, 45d0eef - These look too involved for it to be worth taking them, given we'll have to maintain the code going forward. Amazon can upgrade these devices to resolve the underlying issues.

ojw28:
Amazon probably have a group that you need to become a member of for you to be linked against the corporate CLA.

peddisri:
I am already part of that group. Please verify at your end.

ojw28:
I think your Amazon address is part of that group, but you're using a Lab126 address that isn't. You need to get the Lab126 address added to the group, or use your Amazon address on GitHub. The bot does verify corp CLA signers, so you should be able to reply `I signed it!` once this is resolved.

peddisri:
Thanks. Will fix the email id in the next pull request containing only https://github.com/google/ExoPlayer/commit/245c7457f8dfa95d1f23c72cda323c5a4f9ba74f

ojw28:
Sounds good to me. I'll close this one, in that case!

Issue #909, opened 2015-10-30
Consider allowing alternative MimeTypes to handle device differences


IanDBird:
One thing i've seen is that some devices use slightly different MimeTypes for a number of audio codecs (mainly DTS/DTS-HD/TrueHD). I think this might be because as far as i'm aware, this isn't strictly specified anywhere and is more of an adopted standard.

Of course, one stance would be to just not support these devices which are "wrong". I did however have a go to see what it would take to allow alternative MimeType lookups to allow their MediaCodec's to be used. I found the following changes were required:
 - Modify `MediaCodecUtil.getMediaCodecInfo` method to consider the specified `mimeType` but also any other alternatives known for that type.
 - Modify the `MediaFormat.getFrameworkMediaFormatV16` method to translate the `mimeType` into one specific for the device.
   - The way I did this was to simply add another method into the `MediaCodecUtil` to perform the lookup.

I actually have a patch that I could submit as a Pull Request for this, but thought it was unclear as to whether you consider this an ExoPlayer issue to be resolved, or whether these devices should simply not be supported. The device I was testing around with was the "Minix Neo", but I believe that the Sony Android TVs also use a different MimeType for DTS than that defined in ExoPlayer (which maybe they've fixed already in a recent FW update).

Thoughts?

ojw28:
I'm unsure as to why the second bullet point would be necessary, if you were to do the first one. I can't think of a reason why we couldn't do the first one; that seems like a fairly sensible thing to do.

IanDBird:
The reason I thought (think) the second point was required is that when initially testing, I found that when initialising the codec (in `MediaCodecTrackRenderer.configureCodec`) it resulted in a failure because an unknown or unexpected mime type was being specified in the provided `MediaFormat`. I thought it made sense to provide a mime type that the codec was expecting, and registered against, rather than what ExoPlayer was using internally.

IanDBird:
If you're happy with the explanation for the second point, i'm happy to submit a PR with the suggested changes for further discussion?

ojw28:
A PR would be the best way to proceed, yes.

ojw28:
Decided not to do this for the time-being, but will reconsider in the future if it becomes a more widespread issue.

PR #941, opened 2015-11-09
Provide mechanism to include alternative mime types for MediaCodec lookup

IanDBird:
Context: #909 

This provides a simple mechanism to include alternative mime types to be considered when attempting to identify a devices associated MediaCodec.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
@@ -29,6 +29,8 @@ import android.util.Log;
 import android.util.Pair;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 
 /**
@@ -54,6 +56,14 @@ public final class MediaCodecUtil {
   private static final String TAG = "MediaCodecUtil";
 
   private static final HashMap<CodecKey, Pair<String, CodecCapabilities>> codecs = new HashMap<>();
+
+  private static final HashMap<String, ArrayList<String>> mimeTypeAlternatives = new HashMap<>();
+
+  static {
+    // It's possible that some devices use alternative MimeTypes for codecs to those used internally. We
+    // therefore should consider them when attempting to locate the associated MediaCodec.
+    mimeTypeAlternatives.put(MimeTypes.AUDIO_DTS_HD, new ArrayList<>(Arrays.asList(new String[]{"audio/dtshd"})));
+  }
 
   private MediaCodecUtil() {}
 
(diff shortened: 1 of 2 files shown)


IanDBird:
I signed it!

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
```diff
@@ -95,34 +105,72 @@ public static synchronized void warmCodec(String mimeType, boolean secure) {
...
+  /**
+   * Returns the actual mime type known by this device for the one provided. This allows the library
+   *     to use a potentially different mime type internally, but translate back to the device
+   *     specific one when required.
+   *
+   * @param originalMimeType The mime type.
+   * @return The devices specific mime type for the one specified.
+   */
+  public static synchronized String getDeviceMimeType(String originalMimeType) {
+    ArrayList<String> alternatives = mimeTypeAlternatives.get(originalMimeType);
```
IanDBird: This code has a dependency on the above code being used first. I wasn't 100% comfortable with this but as it works in practice, I wanted to get feedback before refactoring. Essentially, the MediaCodec identification cleans up any alternatives so that they are no longer considered in the future.

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
```diff
@@ -55,6 +57,14 @@ private DecoderQueryException(Throwable cause) {
 
   private static final HashMap<CodecKey, Pair<String, CodecCapabilities>> codecs = new HashMap<>();
 
+  private static final HashMap<String, ArrayList<String>> mimeTypeAlternatives = new HashMap<>();
+
+  static {
+    // It's possible that some devices use alternative MimeTypes for codecs to those used internally. We
+    // therefore should consider them when attempting to locate the associated MediaCodec.
+    mimeTypeAlternatives.put(MimeTypes.AUDIO_DTS_HD, new ArrayList<>(Arrays.asList(new String[]{"audio/dtshd"})));
```
IanDBird: It's likely that other alternatives will be added but I wanted to test on these devices before suggesting the changes. This specific alternative came from the Minix Neo device. Interestingly, it only provides a DTS-HD MediaCodec but not a DTS one (strange).

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
ojw28: Heh. I think this is too hacky to be classed as maintainable code ;)...

If we were to make this change why not just include the device mimeType in DecoderInfo, which would remove the need for this method?

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
```diff
@@ -95,34 +105,72 @@ public static synchronized void warmCodec(String mimeType, boolean secure) {
...
-      return codecs.get(key);
-    }
-    MediaCodecListCompat mediaCodecList = Util.SDK_INT >= 21
-        ? new MediaCodecListCompatV21(secure) : new MediaCodecListCompatV16();
-    Pair<String, CodecCapabilities> codecInfo = getMediaCodecInfo(key, mediaCodecList);
-    // TODO: Verify this cannot occur on v22, and change >= to == [Internal: b/18678462].
-    if (secure && codecInfo == null && Util.SDK_INT >= 21) {
-      // Some devices don't list secure decoders on API level 21. Try the legacy path.
-      mediaCodecList = new MediaCodecListCompatV16();
+      String originalMimeType, boolean secure) throws DecoderQueryException {
```
ojw28: This is quite a bit more invasive than I had envisioned. Are there any devices other than Minix Neo that are using the wrong mimeTypes, or is this *just* for that device?

IanDBird: The Sony Android TVs registered MediaCodecs against a different MimeType, at least at one point. I've received a few firmware updates, so it's possible that they have changed that. I think this issue is likely to happen on a percentage of devices which bundle their own DTS/DTS-HD/TrueHD decoder, as there isn't any concrete specifications to define exactly what is expected for these codecs. Native decoding for these codecs is still fairly rare, but new devices are coming on to the market which include these. I would imagine any future Android TV based TV devices would also fall into this set, unless these MimeTypes are being checked during CTP?

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
IanDBird: That would be a much better idea :) I'm happy to change, but let me know if you think it's worth refactoring these code into something more acceptable / maintainable, or if you are having second thoughts on adding this feature.

IanDBird:
@ojw28 - Although I have signed the agreement previously, and had some PRs accepted, do I need to redo it again in order for the googlebot to detect it?

library/src/main/java/com/google/android/exoplayer/MediaCodecUtil.java
ojw28: Meh. It's quite a lot more complicated than I thought it was going to be. CodecKey makes things more complicated still, if we want to handle everything in a nice way.

I think the Sony TVs use the same MimeTypes as are currently in place in ExoPlayer these days, although @andrewlewis can confirm/deny. If that's the case, it might be worth holding the thought for a while to see if other mainstream devices do actually ship with the need for this kind of hack. If they do, maybe we should think again at that point?

IanDBird: I would completely understand if you decided not to accept these changes. I can try to identify additional devices, but I would guess they might also fall into the "not so mainstream" category. Unfortunately for me though, users still purchase these devices with the expectation that they will work well with our application and therefore we are still required to try and best utilise their capabilities.

ojw28: I think we should not take it for now. If it becomes a problem with more mainstream devices then we should reconsider.

PR #948, opened 2015-11-11
Add HLS Live/DVR support

pajatopmr:
<h1>Rationale:</h1>

Support in ExoPlayer for HLS live streaming is not as polished as for non-live streaming.  A number of issues exist against seeking (which implies DVR) in particular.  This commit builds on the good work done in PR#848 (and others) to address these issues in the following way:

* A live stream is added to the Demo samples for HLS.

* The Demo player uses the available seek range to bracket the sliding DVR window.

* Chunk processing takes into account the live edge on start.

* A bug in live seeking is fixed.

<h1>Changes to be committed:</h1>

demo/src/main/java/com/google/android/exoplayer/demo/Samples.java

- Add a live stream (the Israeli i24 TV news feed)

demo/src/main/java/com/google/android/exoplayer/demo/player/DemoPlayer.java (From PR#848)

- Add an event listener for Chunk events.

- Update the player on changes to the available seek range.

demo/src/main/java/com/google/android/exoplayer/demo/player/HlsRendererBuilder.java (From PR#848)

- Inject the player and main handler into the chunk source object (using a new HlsChunkSource constructor).

library/src/main/java/com/google/android/exoplayer/hls/HlsChunkSource.java (Mostly from PR#848)

- Extend the EventListener interface to handle changes to the available seek range.

- Provide a constructor that injects the player and a handler to deal with available seek range changes.

In getChunkOperation():

- Obtain the live start chunk media sequence only when seeking to playback position 0.

- Fix a bug in live sseking that causes all chunks preceding the one containing the desired seek position to be loaded (which causes a huge delay in seeking towards the live edge).

In setMediaPlaylist():

- Process changes to the available seek range.

library/src/main/java/com/google/android/exoplayer/util/PlayerControl.java (From PR#848)

- Provide support for handling changes to the available seek range.

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
@@ -132,8 +132,6 @@ import java.util.Locale;
         + "prog_index.m3u8", PlayerActivity.TYPE_HLS),
     new Sample("Apple ID3 metadata", "http://devimages.apple.com/samplecode/adDemo/ad.m3u8",
         PlayerActivity.TYPE_HLS),
-    new Sample("Live/DVR Fixed", "http://bcoveliveios-i.akamaihd.net/hls/live/215102/master_english/398/master.m3u8",
-        PlayerActivity.TYPE_HLS),
   };
 
   public static final Sample[] MISC = new Sample[] {
--- a/demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/Samples.java
@@ -132,6 +132,8 @@ import java.util.Locale;
         + "prog_index.m3u8", PlayerActivity.TYPE_HLS),
     new Sample("Apple ID3 metadata", "http://devimages.apple.com/samplecode/adDemo/ad.m3u8",
         PlayerActivity.TYPE_HLS),
+    new Sample("Live/DVR Fixed", "http://bcoveliveios-i.akamaihd.net/hls/live/215102/master_english/398/master.m3u8",
+        PlayerActivity.TYPE_HLS),
   };
 
   public static final Sample[] MISC = new Sample[] {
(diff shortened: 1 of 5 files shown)


pajatopmr:
Fwiw, I have signed CLAs under both email addresses I typically use when committing changes to Github.  But I may have to add another commit to trigger a check to catch the CLA I signed this morning.

fredx21:
@pajatopmr, I played with your patched files and it's a very good start to live HLS. I have a few suggestions to improve the result in the demo app:
- In your implementation of "getChunkOperation" when the seekPositionUs != 0, you should prevent seeking beyond the "3 frame threshold" from stream end, as it is already done in "getLiveStartChunkMediaSequence".
- The player position in the seek-bar should never move for live streams because it repensent a fixed reading position inside the server's DVR buffer, as opposed to the non-live streams where the time in the seek-bar represents a time-span since the beggining of the stream.

pajatopmr:
Thanks @fredx21 (and @thasso).
* Ensuring that the seek does not go beyond the "live edge" is a good point.  I will fix that as it feels consistent with the HLS spec.
* As for your second point, I think there is some room for disagreement.  My model is that live/EVENT and VOD both should use the seekbar to represent the player position from a start value of 0.  For live with a fixed sliding window I think the current position can be interpreted/represented either from the live edge or the start of the window (my choice).  I have seen both styles employed.  I'd love to get the "official" ExoPlayer preference or hear about an existing, dominant player model.  That said, I think app developers should be able to apply their own style where a spec does not clearly define behavior, which is the case here with the demo player.  Using the available seek range gives the app developer wiggle room.

pajatopmr:
@fredx21 In testing my fix for your first point, your second point became clearer,especially to the extent that my representation of the stream showed one end as the current playhead position and the other end of the seekbar as the fixed window size.  I am now inclined to change that to show playhead positions on both ends.  I think this makes more sense.

TBoshoven:
When using this patch, the playlists don't seem to be updated frequently. This makes it possible to request very old segments by seeking after pausing. Furthermore, it makes it difficult to seek to "live".
Building on top of this patch, what would be a good approach to making sure that the playlists keep updating in a DVR stream?

programkom:
when will this be merged to master branch? I've been using Pajatopmr's branch because on that variant LIVE streams can be rewinded, which is not true for current master branch

PR #950, opened 2015-11-11
loosen overly restrictive regex to allow whitespace before numbers, etc.

machenmusik:
Out in the wild, some manifests have spaces before numbers, e.g. EXT-X-TARGETDURATION: 10
but are otherwise valid and playable; this change loosens ExoPlayer's strict compliance with https://tools.ietf.org/html/draft-pantos-http-live-streaming-07#page-7 to allow playback of these streams.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsPlaylistParser.java
@@ -68,21 +68,21 @@ public final class HlsPlaylistParser implements UriLoadable.Parser<HlsPlaylist> 
   private static final String METHOD_AES128 = "AES-128";
 
   private static final Pattern BANDWIDTH_ATTR_REGEX =
-      Pattern.compile(BANDWIDTH_ATTR + "=(\\d+)\\b");
+      Pattern.compile(BANDWIDTH_ATTR + "=[ \\t]*(\\d+)\\b");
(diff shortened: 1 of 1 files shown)


ojw28:
Whose manifests are these? Can you provide some examples?

machenmusik:
unfortunately I can't provide examples, but the exception ExoPlayer was
throwing looked like this:

Couldn't match #EXT-X-TARGETDURATION tag in #EXT-X-TARGETDURATION: 10
which made no sense to me until I saw it with monospaced font and realized
there was a space in front.  If you'd like to stick with stricter parsing
policy, I understand. (since technically exoplayer was right in rejecting,
but in practice other players work)

ojw28:
I don't agree with the argument that we should handle things because other players do. If every player were to do that then you end up with all players relaxing specifications in the same ill defined ways, which then encourages the generation of bad media, undermines the existence of specifications in the first place, and makes in much harder for someone to write a new player because any new implementations will be expected to handle the same unspecified behaviors.

We could handle it if it's a widespread problem (e.g. if some popular HLS serving component is known to behave this way). Otherwise I think it's better to request the provider to fix their content.

machenmusik:
Right, that's why I understand if you choose to maintain strict compliance,
as I am not sure how prevalent this issue is.  Feel free to close/reject
this pull request

ojw28:
Ok, I think we'll close for now.

PR #957, opened 2015-11-13
 Check for valid periods in DASH MPD after parsing.

### LINKED ISSUE #952: Handling of invalid periods in DASH MPDs
In `MediaPresentationDescriptionParser` in method `parseMediaPresentationDescription`, it is possible that one or more periods are discarded/ignored. This can lead to an empty list of Periods for the MPD. Example: live profile, and only one period element, which lacks the start attribute (I know this is an erroneous MPD).

In this case, the whole app will crash due to an uncaught `IndexOutOfBoundsException`, when requesting the first period in `DashRendererBuilder.buildRenderers()`, line `Period period = manifest.getPeriod(0);`.

As there has to be at least one Period according to the spec anyway, I would suggest doing something along the following lines at the end of `parseMediaPresentationDescription`, in order to have a graceful shutdown of the player.

```
if (periods.isEmpty()) {
  throw new ParseException(
      "no valid periods present in media presentation description");
}
```

Bastian35022:
Fixes #952

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/dash/mpd/MediaPresentationDescriptionParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/dash/mpd/MediaPresentationDescriptionParser.java
@@ -159,6 +159,11 @@ public class MediaPresentationDescriptionParser extends DefaultHandler
       }
     }
 
+    if (periods.isEmpty()) {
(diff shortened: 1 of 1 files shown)


ojw28:
Thanks!

PR #969, opened 2015-11-19
Implement multiple audio tracks support for HLS

tipoc123:
1) Returns a correct number of audio streams for HLS (streamTypes -> streamPids)
2) Reading audio stream language from PMT (for AAC)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/ts/AdtsReader.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/ts/AdtsReader.java
@@ -56,6 +56,8 @@ import java.util.Collections;
   // Used when reading the samples.
   private long timeUs;
 
+  private String language;
+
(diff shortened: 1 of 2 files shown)


ojw28:
If you want multi-track support in HLS you should be using renditions rather than bundling multiple audio tracks in a single TS stream (I don't think the latter approach was ever an official part of the HLS spec). See:

https://tools.ietf.org/html/draft-pantos-http-live-streaming-17#section-4.3.4.2.1

Whilst we don't yet support renditions either, we do intend to do so in the future, and attempting to support both approaches simultaneously is likely going to get messy. So unless you can convince me otherwise, I don't think this is something we really want to be merging.

tipoc123:
I signed it!

ojw28:
We support multi-audio using renditions, so closing this.

PR #1007, opened 2015-11-30
Fixed large PMT packet parsing

alex3d:
Fixed large (>ts packet) PMT packet parsing in TsExtractor.


ojw28:
Do you have some sample media you can provide that's broken without this and fixed with it? Thanks!

ojw28:
Merging. Although we'd still really appreciate some test content if you're able to provide it. Thanks!

PR #1035, opened 2015-12-08
Release 1.5.3

ojw28:


ojw28:
Note - Googlebot is unhappy because it was only turned on part way through the 1.5.3 dev cycle. CLAs have been signed in all cases.

TagLoomis:
I have been following your FLV implementation on ExoPlayer and I am using ExoPlayer-dev-1.5.3-rc with the latest changes  ( checked out an hour ago )

Looking in “ScriptTagPayloadReader.java”  where do you callback any metadata that is coming in or is that not complete yet?

If it is complete, if you wouldn’t mind, how do I create a listener for FLV  metadata messages?

Also,  I have added temporary MP3 support to “AudioTagPayloadReader.java”  would you be interested in my additions?

PR #1054, opened 2015-12-12
Allow multiple identical but intermittent spans

aptly-io:
Multiple identical TTML fontStyles or fontWeights or textDecorations 
on the content for a specific moment were rendered as if there's only one decoration.
That's because SpannableStringBuilder.setSpan(Object span, start, end, flag) found
the earlier set span (the static allocated span's reference
is each time the same) and only refreshed this first set span's start and end values
instead of adding a new span at its (new) different start and end values.
This patch removes the static data members; the newly allocated span objects
are distinguishable.
A correct implementation is favoured over worries about memory consumption.


ojw28:
Thanks for this. The change needs to go into the `dev` branch though, not `master`. Please send a pull request to that branch.

PR #1070, opened 2015-12-16
[WIP] Add SequenceSampleSource for playing multiple sources back-to-back

hermanbanken:
Seamless playback is the goal of this pull request, without even the slightest gap.

Multiple people are trying to achieve a playlist like functionality, with either ExoPlayer or MediaPlayer, see for example:
- http://stackoverflow.com/questions/28277091/exoplayer-play-10-files-one-after-another
- http://stackoverflow.com/questions/33362339/does-exoplayer-support-mp4-playlist
- http://stackoverflow.com/questions/3877119/android-mediaplayer-gapless-or-seamless-video-playing
- http://stackoverflow.com/questions/2030487/play-playlist-with-mediaplayer (MediaPlayer)
- http://stackoverflow.com/questions/22743693/android-media-player-playlist-and-reset (MediaPlayer)
- http://stackoverflow.com/questions/30681002/play-list-of-mp3-file-with-mediaplayer-in-android (MediaPlayer)

Some of these questions are about simple audio playlists, some for video playlists. What I'm trying to do is create an application like (a very lightweight variant of) Apple's iMovie. So basically play 2 or more video files (stored locally) with 0-frames gap. On iOS the AVFoundation framework provides these features and it works easily and performs well.

On Android, implementations that rely on substituting the ExoPlayer or it's data sources do not perform well in our tests (either release/prepare-cycles, or having a spare player stand ready). Even on modern devices (tested old & new devices) there exists a noticeable freeze.

The SequenceSampleSource is an effort to solve this issue. It wraps multiple SampleSources so that buffering is continued cross video and the video timeline is seekable as a whole. The inner clip durations would be available for keeping track of which video is played currently, or another listener could be implemented, both options are fine by me.

I am completely open for feedback and would love to solve this 'problem' for the Android platform. The ultimate goal would be to have something like this in the MediaPlayer interface, something as simple as this:

```
mediaPlayer.setDataSources(new String[] { "http://video.url" });
mediaPlayer.prepare();
mediaPlayer.start();
```

For now I'm using the following setup myself:

```
String[] path = new String[] { "A.mp4", "B.mp4" };
ArrayList<SampleSource> playlist = new ArrayList<>();
for(String path : paths) {
  playlist.add(new ExtractorSampleSource(Uri.parse(path), new FileDataSource(), new DefaultAllocator(256 * 1024), 1024 * 1024));
}

SampleSource sampleSource = new SequenceSampleSource(playlist, null);

Handler mainHandler = player.getMainHandler();
MediaCodecVideoTrackRenderer videoRenderer = new MediaCodecVideoTrackRenderer(context, sampleSource, MediaCodec.VIDEO_SCALING_MODE_SCALE_TO_FIT, 5000, mainHandler, player, 50);
MediaCodecAudioTrackRenderer audioRenderer = new MediaCodecAudioTrackRenderer(sampleSource, null, true, mainHandler, player, AudioCapabilities.getCapabilities(context));

TrackRenderer[] renderers = new TrackRenderer[Player.RENDERER_COUNT];
renderers[Player.TYPE_VIDEO] = videoRenderer;
renderers[Player.TYPE_AUDIO] = audioRenderer;

player.onRenderers(renderers, new DefaultBandwidthMeter());
```

The code in this pull request is not finished. It is inspired by the HLS playback, but my knowlegde of video data formats and the internal defacto ExoPlayer lifecycle/ways is lacking. Known bugs:

- [ ] Although first playback works fine, after seeking the video tends to jump upon clip transition
- [ ] After first seek, some clip transitions seem to lack a keyframe, we see typical behaviour like when normal frame differences are applied to the wrong keyframe. Smudgy motions in regions that are not reset by the keyframe.

Any help would be wonderful!

### DIFF
--- a/library/src/androidTest/java/com/google/android/exoplayer/playlist/SequenceSampleSourceTest.java
+++ b/library/src/androidTest/java/com/google/android/exoplayer/playlist/SequenceSampleSourceTest.java
@@ -1,0 +1,114 @@
+package com.google.android.exoplayer.playlist;
+
+import com.google.android.exoplayer.MediaFormat;
+import com.google.android.exoplayer.MediaFormatHolder;
+import com.google.android.exoplayer.SampleHolder;
+import com.google.android.exoplayer.SampleSource;
+import com.google.android.exoplayer.util.MimeTypes;
+
+import junit.framework.TestCase;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Test for {@link SequenceSampleSource}
+ */
+public class SequenceSampleSourceTest extends TestCase {
+
+    public void testSequencePlaylist() {
+        final long[] duration = {0};
+
+        SequenceSampleSource.SequenceInfoListener listener = new SequenceSampleSource.SequenceInfoListener() {
+            @Override
+            public void onTimesKnown(long[] durationsUs, long[] startTimesUs) {
+                duration[0] = durationsUs[0];
+            }
+        };
+
+        SequenceSampleSource sequenceSampleSource = new SequenceSampleSource(buildPlaylist(), listener);
+        SampleSource.SampleSourceReader sourceReader = sequenceSampleSource.register();
+        sourceReader.prepare(0);
+
+        assertEquals("Check that duration is extracted from sub sources", (long) 10e6, duration[0]);
+    }
+
+
+    private Collection<SampleSource> buildPlaylist() {
+        ArrayList<SampleSource> list = new ArrayList<>();
+
+        list.add(new DummySource());
+        list.add(new DummySource());
+
+        return list;
+    }
+
+    class DummySource implements SampleSource {
+
+        @Override
+        public SampleSourceReader register() {
+            return new DummyReader();
+        }
+    }
+
+    class DummyReader implements SampleSource.SampleSourceReader {
+        @Override
+        public void maybeThrowError() throws IOException {
+
+        }
+
+        @Override
+        public boolean prepare(long positionUs) {
+            return true;
+        }
+
+        @Override
+        public int getTrackCount() {
+            return 2;
+        }
+
+        @Override
+        public MediaFormat getFormat(int track) {
+            if(track == 0)
+                return MediaFormat.createFormatForMimeType(track, MimeTypes.VIDEO_MP4, MediaFormat.NO_VALUE, (long) 10e6);
+            return MediaFormat.createFormatForMimeType(track, MimeTypes.AUDIO_AAC, MediaFormat.NO_VALUE, (long) 10e6);
+        }
+
+        @Override
+        public void enable(int track, long positionUs) {
+
+        }
+
+        @Override
+        public boolean continueBuffering(int track, long positionUs) {
+            return true;
+        }
+
+        @Override
+        public int readData(int track, long positionUs, MediaFormatHolder formatHolder, SampleHolder sampleHolder, boolean onlyReadDiscontinuity) {
+            return 0;
+        }
+
+        @Override
+        public void seekToUs(long positionUs) {
+
+        }
+
+        @Override
+        public long getBufferedPositionUs() {
+            return 0;
+        }
+
+        @Override
+        public void disable(int track) {
+
+        }
+
+        @Override
+        public void release() {
+
+        }
+    }
+
+}
--- a/library/src/main/java/com/google/android/exoplayer/playlist/SequenceSampleSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/playlist/SequenceSampleSource.java
@@ -1,0 +1,296 @@
+package com.google.android.exoplayer.playlist;
+
+import android.util.SparseArray;
+
+import com.google.android.exoplayer.MediaFormat;
+import com.google.android.exoplayer.MediaFormatHolder;
+import com.google.android.exoplayer.SampleHolder;
+import com.google.android.exoplayer.SampleSource;
+import com.google.android.exoplayer.TrackRenderer;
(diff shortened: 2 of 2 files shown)


ojw28:
Thanks for sending this! It looks like an interesting approach :). Playlists are definitely something that we want to do a better job of handling. As you may have observed, there are a few things that make this difficult currently. The main issue is that ExoPlayer doesn't support changes to the available tracks and track formats during a playback. For general purpose playlist support as well as for DASH multi-period manifest support (which is conceptually similar), this is a must. We don't want to disallow playlists from containing both MP3 and MP4 files, for example.

I think the first step toward supporting general purpose playlists will be to allow dynamic binding of SampleSource to TrackRenderers. Moving from one piece of media to the next would then involve (approximately speaking) rebinding the TrackRenderers to a new SampleSource, and at this point we'd permit the set of tracks to change. On top of that, for a general purpose solution, we'd also like to properly manage buffering when transitioning from one piece of media to the next (e.g. if the policy is to keep 30 seconds of media buffered and we're 10 seconds from the end of one piece of media, we should have buffered 20 seconds of the next piece).

We're hoping to get to looking at this pretty soon now, so stay tuned. As a final note, in general I don't think it will be viable/sensible to prepare all of the sources up front. That would be a significant amount of work for long playlists. This would imply that the application would need to obtain media durations from elsewhere (or alternatively accept that the durations are unknown). I think this is reasonable though; something like a music application most likely already has a database containing this kind of information.

hermanbanken:
General purpose playlists would be nice indeed. I think there are two kinds of multi-source playback: 
- those where you are interested in playing full clips - one after the other - where progress bars and UI change upon the next clip (for example album cover when playing music, current time, time remaining), having possibly very different formats (video/audio mixed)
- and secondly playback where the source is subdivided (like HLS) but where chunks have some meaning (are clips) and possibly have a slightly different MediaFormat (max buffer size or codec initialisation data).

In the first case 'some' delay would be acceptable and a simple `player.stop(); rendererBuilder.buildRenderers(newClipData);` would suffice. It's the latter case I'm interested in, but I'm not sure if that requires changing TrackRenderers. It would of course be nice if ExoPlayer would support a dynamic playlist where you can start multiple audio sources with different offsets / independently from the video... that way it would be more efficient to make TrackRenderers dynamically bound.

In our case we have a single timeline, showing all chunks/clips and you can seek through the whole thing. This is the latter case, since although we need to know which clip is playing (to apply selected GPU Fx for example) we do not swap UI's and would like to prevent any delay incurred by flushing the codec, sample sources and/or renderers.

Currently our UI needs to know the complete duration to show progress. We retrieve the durations from the sample sources. That's why we are preparing all sources right now. This would be solvable if we had a database of this information on before hand. That would be fine too. If I think about it more, the clips currently are represented as fixed width (horizontal-)list items, so if the player would know which clip is playing we would only need to know the duration of the current clip. For example `getCurrentPosition()` could simply return a tuple of the clip index and the position (s/ms/us) in that clip. This would be a breaking change though. Better to handle that outside ExoPlayer using that database of durations...

Buffering should indeed we handled the way you propose. Our situation only deals with clips of approximately 10 seconds and a total duration of like a minute, but for a general solution this should be looked into.

I'm looking forward to the outcome of you guys looking into this!

Some extra food for thought: say we want to create transitions between video's. Thats a thing that would need to be handled a layer above the codec, at SurfaceTexture level. Would it be possible to have multiple (2) textures (GL_TEXTURE_EXTERNAL_OES if I'm correct), render those simultaneously and blend them in the GPU? Or can't those textures co-exist?

Orbyt:
So was a solution for gap-less looping audio ever achieved here? Its ridiculous this basic functionality can't be found across any of Android's media apis.

ojw28:
There are multiple APIs that you can use to achieve gap-less looping audio:

- For small sound effects (<1MB), `SoundPool` can do this.
- You can use two `MediaPlayer` instances and its `setNextMediaPlayer` API.
- You can use `AudioTrack` directly and feed in the looping audio yourself.

Flyview:
Good point, those should work for audio. Unfortunately that doesn't help much for gapless video.

Orbyt:
@ojw28 My use case is for sounds that are a bit to long for `SoundPool`. I've always tried using the two MediaPlayer instances, but the gap is still there.

Have yet to use AudioTrack, mainly due to lack of examples on how to get that functionality.

ojw28:
Two MediaPlayer instances should work. Are you sure the media you're playing contains gapless metadata information? If the media isn't designed to be looped in a gapless way, it's not going to be gapless.

Orbyt:
@ojw28 I'm not the one generated the audio, but he has told me that when looking in his editing program that the audio bars "go all the way to the end". Is there some kind of metadata he needs to add to make an audio track loop without a gap?

ojw28:
That doesn't sound particularly scientific ;). The metadata that needs to be included is format dependent. You'll need to do some research I think. A good place to start would be [here](https://en.wikipedia.org/wiki/Gapless_playback#Format_support).

Orbyt:
@ojw28 Yes, was just reading that, will look into it, thanks. Will end this here as I don't want to clutter this post with off-topic discussion.

ojw28:
Support for playing sources back-to-back is landing in 2.x. Many of the required parts are already in place. For example [SampleSourceProvider](https://github.com/google/ExoPlayer/blob/c2b89d62859c67e4a589de1f6b6acfd605d5099f/library/src/main/java/com/google/android/exoplayer/SampleSourceProvider.java). We don't intend to add support to 1.x, therefore closing this pull request.

hermanbanken:
Great to hear you're doing back-to-back in 2.0! What is the planning of that release, becoming less experimental anytime soon?

PR #1072, opened 2015-12-17
The status is already "Ready" status. If  we call onPlayerStateChange…

chirs1985:
…d again, it will cause top layer execute the On Ready twice.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/ExoPlayerImpl.java
+++ b/library/src/main/java/com/google/android/exoplayer/ExoPlayerImpl.java
@@ -143,9 +143,6 @@ import java.util.concurrent.CopyOnWriteArraySet;
       this.playWhenReady = playWhenReady;
       pendingPlayWhenReadyAcks++;
       internalPlayer.setPlayWhenReady(playWhenReady);
-      for (Listener listener : listeners) {
-        listener.onPlayerStateChanged(playWhenReady, playbackState);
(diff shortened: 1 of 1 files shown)


ojw28:
I'm not sure what the intention is here, but this doesn't look right.

chirs1985:
I signed it!

ojw28:
Please see comment above.

chirs1985:
Sorry, I don't understand the sentence meaning. ---> "Please see comment above."

ojw28:
It's working as intended that the listener is invoked when playWhenReady changes. It's even explicitly documented that way [here](https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer/ExoPlayer.Listener.html#onPlayerStateChanged(boolean,%20int)). So I don't understand what this change is actually trying to achieve. It looks like it's removing correct behavior.

chirs1985:
I think whether it need to add one "running" status? Otherwise the status is always ready, not onPlayerStateChanged

ojw28:
Closing this because I still don't understand what it is. I'm almost certain this isn't necessary, and no-one else has reported any issues with how the listener works.

chirs1985:
Because our player meet this kind issue, we already add a check function to prevent execute some important function twice. I don't think it is a good method that add a check function at top layer program. Architecture should be reliability.

Issue #90, opened 2014-10-24
H.264/MP2T over HTTP 


juanpef:
I have the next code:

        Uri uri = Uri.parse("http://192.166.103.140:8000/udp/239.192.0.49:2121");
		SampleSource sampleSource = new FrameworkSampleSource(mContext, uri, /* headers */ null , RENDERER_COUNT);
		MediaCodecVideoTrackRenderer videoRenderer = new MediaCodecVideoTrackRenderer(sampleSource,
     		        MediaCodec.VIDEO_SCALING_MODE_SCALE_TO_FIT, 0, mainHandler,
     		        null, 50); 
		MediaCodecAudioTrackRenderer audioRenderer = new MediaCodecAudioTrackRenderer(sampleSource);
    		ExoPlayer exoPlayer = ExoPlayer.Factory.newInstance(RENDERER_COUNT);
    		exoPlayer.prepare(videoRenderer, audioRenderer);
    		exoPlayer.sendMessage(videoRenderer, MediaCodecVideoTrackRenderer.MSG_SET_SURFACE, surface);
    		exoPlayer.setPlayWhenReady(true);

It works with a closed file like http://html5demos.com/assets/dizzy.mp4 but it does not work with a live streaming. What is the problem???

Schizo:
This should be an out of the box feature.

ojw28:
- We're working on a new extractor model that should make this easier to support (ETA 1-2 weeks).
- Is anyone able to provide a working test stream to help us validate support?
- Your evaluation of what should and should not be an out of the box feature is dubious. On what basis did you reach your conclusion, other than it's a feature you personally want?

Schizo:
@ojw28  wheni was writing my comment, i asked my selfe the same question. But to answer your question, h264 streaming over http/udp is one of the most used cases. Nearly every surveillance system i came across was using h264 over http. 

For the test-case: 
-) What do you need exactly? Can you not test it with an ffmpeg stream easily sending an elementary H264 Stream?

ojw28:
Ideally just a test URL, that works, without us having to spend any time setting one up ourselves.

ojw28:
One other question - Am I correct in thinking that you shouldn't expect seeking to work for this use case (since there's no index, either in the form of an external manifest or in the media stream itself)?

ojw28:
[...]

PR #1083, opened 2015-12-22
Restore the flow of Transport Stream packets if synchronisation is lost

algrid:
For some reason TsExtractor can get a TS packet with wrong length and all the following packets become shifted by several bytes which completely breaks the video playback. This code restores the synchronisation if it's possible.

See: https://github.com/google/ExoPlayer/issues/90


algrid:
I signed it!

sreenadhkondru:
For me this issue surfaced only when using UDP as the transport. When a TS packet sent by the server is spread across two UDP packets, and one of the UDP packets is lost (which happens quite often since UDP is a lightweight unreliable transport) the TSExtractor goes out of sync. An incomplete TS packet rather than being discarded is fed with wrong bytes from another TS packet.

On the server side the fix is to set the UDP payload packet size to a multiple of 188 bytes ( I set it to 7*188 = 1316 bytes). That way even if a UDP packet gets lost, we do not end up with an incomplete TS packet. This worked for me and I never got the issue with TSExtractor again.

On the client side I tested the fix by Algrid, and it works even If the server is not configured to send UDP packets of consistent size. But I would prefer the UDPDatasource code to be refactored to catch such packet loss exceptions and throw the exception to the extractor which in turn would take the necessary steps in discarding  incomplete data.

algrid:
In my case it fixed problems with MPEG2 TS over HTTP.

ojw28:
I believe we've fixed this in `dev-v2`, [here](https://github.com/google/ExoPlayer/blob/dev-v2/library/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java#L180), so closing this.

PR #1113, opened 2016-01-05
Add smooth upper limit of getCurrentPositionUs in AudioTrack.java

vipjml:
this prevent getCurrentPositionUs from smooth too much.This is useful when debugging app,when you run into a break point, then wait for a while,for example, one minute , then you resume the app,the smoothed currentPositionUs will be much large than the real player play position,the video play will have problem.

I also suggest  add a upper limit advance of StandaloneMediaClock.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
+++ b/library/src/main/java/com/google/android/exoplayer/audio/AudioTrack.java
@@ -298,8 +298,13 @@ public final class AudioTrack {
       maybeSampleSyncParams();
     }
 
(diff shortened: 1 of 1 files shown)


vipjml:
I signed it!

vipjml:
I signed it!

vipjml:
I signed it!

vipjml:
I signed it!

vipjml:
I signed it!

vipjml:
I signed it!

vipjml:
I signed it!

PR #1128, opened 2016-01-11
Delete public from ManifestFetcher interface

suzukaze:
It is not necceary that interface has public attribute.


ojw28:
Happy to take this change, but it would need to go into the google:dev branch, not google:master. You'll also need to sign a CLA as above. Closing this request since it's into the wrong branch, but please do open a new one into google:dev. Thanks!

suzukaze:
Thank you for advising.

PR #1146, opened 2016-01-14
Fix RGBA color tuple parsing

aptly-io:
The TTML 1 spec. defines an exact RGBA color tuple as #rrggbbaa
See https://www.w3.org/TR/ttml1/#style-value-color

Android's internal representation is ARGB.
The correct parsing therefore requires a bit of extra byte shuffling ...

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/ttml/TtmlColorParser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/ttml/TtmlColorParser.java
@@ -94,7 +94,11 @@ import java.util.regex.Pattern;
       if (colorExpression.length() == 7) {
         // Set the alpha value
         color |= 0x00000000ff000000;
-      } else if (colorExpression.length() != 9) {
+      } else if (colorExpression.length() == 9) {
+        int alpha = (int) color & 0x00000000000000ff; // get the transparency
+        color >>= 8;                                  // put rgb bytes in their rightful place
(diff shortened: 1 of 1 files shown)


ojw28:
You'll need to update TtmlColorParserTest as well (`testHexCodeParsing` will fail as a result of this change, I think, because the test is also wrong)?

aptly-io:
That's right, sorry ... something for tomorrow. I'll investigate and add another commit to the branch, then it gets automatically into pull request as wel I guess?

ojw28 <notifications@github.com> schreef op 14 januari 2016 23:27:14 CET:
>You'll need to update TtmlColorParserTest as well (`testHexCodeParsing`
>will fail as a result of this change, I think, because the test is also
>wrong)?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/google/ExoPlayer/pull/1146#issuecomment-171801785

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ojw28:
I'm not sure; try and see. You can always send a new pull request if necessary. Thanks!

ojw28:
I merged this as it is because I'd like to land this change in front or another one that's coming. I'll fix the test shortly.

ojw28:
Tests fixed in https://github.com/google/ExoPlayer/commit/7974a61476e9f494076fad1f8ee8f6df0a87c83c. Thanks again for the fix!

PR #1155, opened 2016-01-16
link track language attribute in Smoothstream parser

### LINKED ISSUE #1095: SmoothStream audio track language metadata is not made available
When we request the Audio tracks on a SmoothStream stream the language attribute on the manifest xml is not made available on the _MediaFormat_ language property return from the _getTrackFormat_ method:
https://github.com/google/ExoPlayer/blob/0c060f429f459f1bac2ffa0963cadc45ce35b73a/library/src/main/java/com/google/android/exoplayer/ExoPlayer.java#L327

It seems the language attribute is parsed and passed down to the _StreamElement_
https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/smoothstreaming/SmoothStreamingManifestParser.java#L580

But when the MediaFormat instance is created the language property is linked to the _TrackElement_ (not to the StreamElement):
https://github.com/google/ExoPlayer/blob/c9caaad50c4148a8c369816c100c170436a3d4d6/library/src/main/java/com/google/android/exoplayer/smoothstreaming/SmoothStreamingChunkSource.java#L417

The _TrackElement_ does not contain the language flag:
https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/smoothstreaming/SmoothStreamingManifest.java#L134

As a result the _getTrackFormat_ will always return a _MediaFormat_ with a language set to _null_, even if the metadata is available on the manifest.

bartsidee:
resolve #1095

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/smoothstreaming/SmoothStreamingManifest.java
+++ b/library/src/main/java/com/google/android/exoplayer/smoothstreaming/SmoothStreamingManifest.java
@@ -132,10 +132,10 @@ public class SmoothStreamingManifest {
     public final byte[][] csd;
 
     public TrackElement(int index, int bitrate, String mimeType, byte[][] csd, int maxWidth,
-        int maxHeight, int sampleRate, int numChannels) {
+        int maxHeight, int sampleRate, int numChannels, String language) {
       this.csd = csd;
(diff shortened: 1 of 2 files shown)


bartsidee:
signed the cla

ojw28:
Thanks!

PR #1198, opened 2016-01-27
Implemented fade in and fade out volume animations.

Eyliss:
Implemented fade in and fade out volume animations using logarithmic formulas.

## Fade In Formula
image

## Fade Out Formula
image

Where x is time and v velocity.

# Example
`player.fadeIn(0, 15, 10, 2, fadeCallback)`

Where:
* 0 is the start volume
* 15 is the maximum device value
* 10 is the duration of the fade
* 2 the velocity of the fade
* fadeCallback is the completion callback (can be null)

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/player/DemoPlayer.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/player/DemoPlayer.java
@@ -174,7 +174,8 @@ public class DemoPlayer implements ExoPlayer.Listener, ChunkSampleSource.EventLi
 
   private Surface surface;
   private TrackRenderer videoRenderer;
-  private CodecCounters codecCounters;
+    private TrackRenderer audioRenderer;
+    private CodecCounters codecCounters;
   private Format videoFormat;
   private int videoTrackToRestore;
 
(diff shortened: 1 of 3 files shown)


Eyliss:
I signed it!

AquilesCanta:
Hello Eyliss, thank you very much for your contribution. I don't have the last work in this matter so you will have to wait. My impression, however, is that there are a few things that require reviewing. For example, indentation is not consistent with the rest of the library. This is the unimportant part. The fading feature, in my opinion, is something that escapes the purpose of the library: It is application specific, and seems to be implementable on application level. Finally, most importantly, the changes affect the very core of the library. It is unlikely that such a feature should take place in the interface of the library. One of the core design concepts is to keep the Player track-agnostic (roughly), while this codes assumes the existance of an audio track renderer. I do really appreciate your contribution though. It is @ojw28 who has the final call. To sum up, I think this code belongs in the application.

ojw28:
Agreed. It looks like you could take what you've implemented in ExoPlayerImpl and move it to application code. If your application happened to be the demo app, it looks like you could fairly trivially move it to the DemoPlayer class instead, where it's more appropriate to assume the existence of an audio renderer.

Eyliss:
Thank you so much for your opinion. You may be right and this feature could be more appropriate for the application side than for the library itself. If there is no objection I would like to include it in the demo app.

AquilesCanta:
Eyliss, some people have suggested that the DemoApp was already too bloat and complex. And since there is no "playlist" feature in the DemoApp, perhaps there is not much sense in fading in and out. That would be forcing videos to have a certain "fading" component which would be part of the video already, if the author intended it to be that way.

If you really want to contribute to the project, help is always welcome. You can check the open issues for feature requests or bug fixes.

PR #1201, opened 2016-01-27
Ps extractor

Narflex:
This is an extractor that supports MPEG2 Program Streams. It follows similar logic to the TsExtractor and makes use of many of the same classes as it.


Narflex:
Sorry...looks like I did this wrong...arrggg

PR #1202, opened 2016-01-27
Added extractor for MPEG2-Program Stream sources

Narflex:
This is based off the TS extractor and shares some code with it (PES parsing is greatly simplified though since you never have fragmented packets with PS, so I didn't try to promote & reuse the PesReader inner class).

I work for Google..so I think that covers the CLA. :)


ojw28:
Please see your corp inbox. Thanks!

PR #1223, opened 2016-02-02
Filtered out repeated control messages.

TBoshoven:
ANSI CEA-608-E (Annex D.2) specifies that control messages for captions / text with codes of which the first character is in the range 0x10 - 0x1f should be transmitted twice in successive frames.
In streams that follow this recommendation, there are issues where captions disappear one frame after appearing.
This pull request adds functionality for ignoring these duplicate control messages.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
@@ -156,11 +156,16 @@ public final class Eia608Parser {
         continue;
       }
 
-      // Ignore repeated control characters.
+      // Ignore repeated control characters between 0x10 and 0x1f.
       if (ccData1 == previousCtrl1 && ccData2 == previousCtrl2) {
         continue;
       }
       previousCtrl1 = previousCtrl2 = 0;
+      
+      if (0x10 <= ccData1 && ccData1 < 0x20) {
+        previousCtrl1 = ccData1;
+        previousCtrl2 = ccData2;
+      }
 
       // Special North American character set.
       // ccData2 - P|0|1|1|X|X|X|X
@@ -191,8 +196,6 @@ public final class Eia608Parser {
       // Control character.
       if (ccData1 < 0x20) {
         addCtrl(ccData1, ccData2);
-        previousCtrl1 = ccData1;
-        previousCtrl2 = ccData2;
         continue;
       }
 
--- a/library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
@@ -106,6 +106,9 @@ public final class Eia608Parser {
   private final StringBuilder stringBuilder;
   private final ArrayList<ClosedCaption> captions;
 
+  // Previous control message, used for identifying repeated ones.
+  private byte previousCtrl1 = 0, previousCtrl2 = 0;
+
   /* package */ Eia608Parser() {
     seiBuffer = new ParsableBitArray();
     stringBuilder = new StringBuilder();
@@ -153,6 +156,12 @@ public final class Eia608Parser {
         continue;
       }
 
+      // Ignore repeated control characters.
+      if (ccData1 == previousCtrl1 && ccData2 == previousCtrl2) {
+        continue;
+      }
+      previousCtrl1 = previousCtrl2 = 0;
+
       // Special North American character set.
       // ccData2 - P|0|1|1|X|X|X|X
       if ((ccData1 == 0x11 || ccData1 == 0x19)
@@ -182,6 +191,8 @@ public final class Eia608Parser {
       // Control character.
       if (ccData1 < 0x20) {
         addCtrl(ccData1, ccData2);
+        previousCtrl1 = ccData1;
+        previousCtrl2 = ccData2;
         continue;
       }
 


TBoshoven:
I think the CLA should be okay now.

ojw28:
Please reply with "I signed it!" to have the CLA bot perform the check again. Thanks.

Also, thanks for this; we had been trying (and failing) to figure that one out; so it's great that you've solved the problem for us!

TBoshoven:
I signed it!

ojw28:
The account/email you're using needs to match the one for which there's a CLA in place. Please could you make it so; either by adding the address you're using to the CLA, or adding the email address for which there's a CLA in place to your GitHub account, or similar. Thanks!

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
```diff
@@ -153,6 +156,12 @@
         continue;
       }
 
+      // Ignore repeated control characters.
+      if (ccData1 == previousCtrl1 && ccData2 == previousCtrl2) {
```
ojw28: Isn't this check too broad? Per the spec, you should only ignore repeated control characters if ccData1 is in the range [0x10, 0x1F]. You should also not ignore a second repeat (e.g. if you see the same control characters three times in a row).

ojw28:
Do you understand exactly what the specification means by "successive frames"? Video frames are quite often re-ordered during decode, and it's unclear whether successive is defined before or after this re-ordering occurs.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
zsmatyas: I tried to read everything publicly available about EIA608. Correct me if I am wrong, but as far as I understand, repeated **Preamble Address Code**s will not affect the output, any formatting/styling/positioning that is higher priority than the ones before should simply override the previous setting.

Repeated **Mid Row Command**s similarly override previous lower priority commands, but they also should move the cursor (effectively printing a transparent space character to the output). So we might miss space characters when dropping something intentional. (Duplication of commands seemed to be optional, isn't it? If it ever makes sense and want to repeat a command, do they have to send it 4 times?)

If we repeatedly change a state based on an incoming **Misc Code** we might incorrectly handle line breaks, but maybeAppendNewline() probably handles most of the possible situations correctly by dropping a second new line character. 

If we do not care about white space errors, than we might be good, the only thing that should be handled carefully seems to be the **Backspace** command. Unfortunately, I do not have sample streams to check how they usually send them.

ojw28: I'm not up to speed to the level of detail required to understand your comment yet (sorry :)), but to ask the my question another way, if the spec says to de-duplicate only if ccData1 is in the range [0x10, 0x1F], why would you implement logic to de-duplicate in cases outside of this range?

I'm not sure if you're arguing that (a) you should de-dupe outside of this range too, or (b) it doesn't make a difference if you de-dupe outside of this range too because duplicate operations wont actually do anything. If (b) then it seems best not to de-dupe; it's easier to understand and match onto the spec with it done that way.

Please clarify. Thanks!

zsmatyas: I am trying to implement the missing part, that is why I read about the commands, but I have not dealt with duplicates yet. I was arguing for (b), as in the renderer the second incoming command seems simply overriding the first (I modified the pop-up type). But your notes are valid, TBoshoven might update the change set.

First:
It seems that "previousCtrl1" is only set when "ccData1 < 0x20", but lower values can still be present and dropped.

Second:
When we exit the loop because of the condition "ccData1 == previousCtrl1 && ccData2 == previousCtrl2", we might want to reset previousCtrl1 and previousCtrl2 to 0, so the 3rd exact same incoming command message wont be dropped (but the 4th again etc.)

TBoshoven:
I signed it!

TBoshoven:
Unfortunately it's not clear to me which notion of successive frames is meant in the spec.
Because the purpose of the repeated messages seems to be dealing with lossy transfers, I would expect it to be before decoding. However, I could not find a reference either way (maybe zsmatyas can answer this question?). So far, I have just been digging through the spec to find an explanation of the repeated messages we found in a number of our streams.

For what it's worth, in all cases we have available the presented patch helps deal with the captions that immediately disappear.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java
TBoshoven: I completely agree about the first point, zsmatyas. I somehow expected all values before 0x10 to be covered, but they don't seem to be. I will update the change set to fix this.
I'm not sure about the second point. While I have a reference recommending the repeated messages, I have not been able to find any information on how to deal with these repeated messages. Can we expect a maximum of two or are more allowed?

zsmatyas:
These are again just my impressions, might be incorrect:
- When 608 was first used, it was for analog transmission, so frames should be transmitted in order.
- The channel is very limited, it is 16 bits per frame, 2 more bits used for parity, so at most we have 2 useful bytes in each frame. They expected the transmission still being lossy, so at least the control codes might be duplicated. (nothing else as the bandwidth is too small for more complex redundancy)
- A single Control Code is 2 byte, so each Control Code eats up an entire frame.
- If we receive a duplicate control code and we did not reorder the frames because the transmission was analog, they must have been transmitted in consecutive frames. (I am not sure where would the digital transmission reorder captions, it should be a separate text stream without necessary reordering for compression or faster encoding/decoding.)

TBoshoven:
Is it not possible for a frame to have nothing in this channel?
Or would there always be two null bytes?
Note: the parser skips if ccType != 0, but it could also be 1, 2, or 3.

ojw28:
As a heads up, we're exploring a way of fixing this in Eia608TrackRenderer, which looks like it might fall out a little nicer. Will provide another update in due course.

zsmatyas:
@TBoshoven 
"NUL characters (00H) should be sent when no display or control characters are being transmitted."
So we probably should not expect any lower values.

The First Byte of the control codes seems to be: 
7 useful bits + 1 parity.
the first (MSB) bits must be "001", 4th bit is the channel.
So without the parity bit the value must be on the range from 0b0010000 to 0b0011111, that is 0x10 to 0x1F. 

We might be closer to the ideal logic if we only checked duplicates for the Control Code handling after the condition is satisfied at line 197:

 if (ccData1 < 0x20) {

The condition might also be added to the lower side of the range as "0xF < ccData1". If we move every duplicate checking and loop exiting logic here, we might have a slightly more redundant and less efficient, but more logical code. But we might also wait to see the implementation in Eia608TrackRenderer ojw28 mentioned.

ojw28:
Hi. We've pushed https://github.com/google/ExoPlayer/commit/b3c954968f1fb94f5d8daf81b6994e7fa0e42e38, which we think addresses this issue. Closing the pull request, but please verify that the issue is resolved for you on the dev branch. Thanks for your help tracking this down!

TBoshoven:
I can verify that this b3c9549 fixes the repeated characters in our test streams and after looking at it I agree that the TrackRenderer is probably the better place for this.

PR #1250, opened 2016-02-10
Add storage of multiple baseUrls in RangedUri class.

Bastian35022:
Adds support for multiple locations defined in DASH MPDs using `baseURL` elements. The respective locations exposed in the `RangedUri` class, as a list of `baseUri`s. The first found location in the MPD is always used currently. Passing the list of baseUrls through varies calls (instead of a single String) had to be implemented to support this.


Bastian35022:
Needs more work.

PR #1257, opened 2016-02-11
Implemented fade in and fade out volume animations in demo application.

Eyliss:
Implemented fade in and fade out volume animations using logarithmic formulas.

## Fade In Formula
image

## Fade Out Formula
image

Where x is time and v velocity.

# Example
`player.fadeIn(0, 15, 10, 2, fadeCallback)`

Where:
* 0 is the start volume
* 15 is the maximum device value
* 10 is the duration of the fade
* 2 the velocity of the fade
* fadeCallback is the completion callback (can be null)


ojw28:
This feature is not important enough to be in the demo app, sorry.

PR #1259, opened 2016-02-11
Support multiple BaseURL elements in MPD

### LINKED ISSUE #771: DASH: Make use of multiple BaseURL elements, where more than one is defined.
From what i saw, the BaseURL element in the DASH MPD is generally supported by the ExoPlayer. However, it does not seem to support multiple/alternative segment locations, as described in ISO/IEC 23009-1:2014 5.6.5. Do you plan on adding this?

pakerfeldt:
All BaseURL elements defined in the MPD is now stored as a List from which a BaseUrlSelector is used to select one of the available BaseURLs on requests.

URLs at each level of the MPD are resolved with respect to the BaseURL elements specified at that level. Relative BaseURL elements are themselves resolved using the BaseURL elements of the level above. If no BaseURL elements are defined at a certain level, BaseURL elements from the level above are inherited.

Resolves #771


abhiguru:
@pakerfeldt had a couple of questions about your implementation while I work on some tweaks to suit my use case. 

Correct me if I am wrong I just went thru most of the Exoplayer code for the first time today, In your implementation It seems if there is a load error on a segment an alternate BaseURL is not attempted, 
Per spec it seems the BaseURL tag is supposed to have an attribute serviceLocation which is to kept  track off in order to blacklist BaseURLs which have failed. 

Here is one very crude approach I came up with : 
when OnLoadError fire in ChunkSampleSource the call it makes to  updateLoadControl is passed a bool tryAlternateBaseUrl which then is passed into the doChunkOperation() and eventually into the newMediaChunk() which in turn would ask segmentUri.getUriAlternate method to get the next baseUri instead of the first one

something like 
"return UriUtil.resolveToUri((baseUris != null) ? baseUrlSelector.getBaseUrl(baseUris, uriIndex) : null, referenceUri);" ...

One issue with this approach is the getChunkOperation does not need to execute when fetching alternateBaseUrl.. If you can think of a way to bypass the whole getChunkOperation and just get the newMediaChunk with alternative BaseURL directly pumped into the stream that would be great...

Below an MPD portion with all possible <BaseUrl> attributes :
<BaseURL dvb:priority="1" dvb:weight="9" serviceLocation="cdn1">http://cdn1.example.com/</BaseURL>
<BaseURL dvb:priority="1" dvb:weight="9" serviceLocation="cdn2">http://cdn2.example.com/</BaseURL>
<BaseURL dvb:priority="1" dvb:weight="2" serviceLocation="cdn3">http://cdn3.example.com/</BaseURL>
<BaseURL dvb:priority="2" dvb:weight="1" serviceLocation="cdn1">http://cdn1.example.com/</BaseURL>
<BaseURL dvb:priority="3" dvb:weight="1" serviceLocation="cdn2">http://cdn2.example.com/</BaseURL>

pakerfeldt:
@abhiguru It was never my intention to change the default behavior of ExoPlayer with this PR. That's why we have the `FirstBaseUrlSelector` implementation. If you want failover you would have to implement `BaseUrlSelector` on your own and pass it to `DashChunkSource`. At least that was the idea. I for instance, (locally) have an implementation of `BaseUrlSelector` that also implements `DashChunkSource.EventListener` to get notified about the success/failure of chunk loads.

pakerfeldt:
I removed the commit with updates on `DashChunkSource.EventListener` since I realized that already exists in `ChunkSampleSource`.

pakerfeldt:
@ojw28 What are your thoughts on this? Is it in line with how you would envision it?

abhiguru:
+1

Bastian35022:
+1 i agree, this is definitely useful for a number of use cases, and general enough for a common basis.

abhiguru:
@pakerfeldt and @Bastian35022,  For the times when <BaseURL> is not present in the mpd should we continue with the default behaviour of using the connectionUrl as the BaseURL example https://gist.github.com/abhiguru/37b4690fe6aaf435dea8 line 45 to 47 ?

Bastian35022:
Definitely! On the other levels, base URLs are inherited nicely, but it's missing on MPD level.

pakerfeldt:
@abhiguru Definitely, good catch. I modified it slightly to be more coherent with the other checks I had:
https://github.com/pakerfeldt/ExoPlayer/commit/4f0eb21cb89350f6a925d659289bfb24e30f2a6c#diff-823909097e504fe52cb2e72cd3664f56R135

pakerfeldt:
I have been looking more closely to this, this morning and realized that, even if it seems to work during my tests, it might not be optimal. A specific `Uri` is decided on during creation of the `Chunk`. And if a `Chunk` load fails it looks like ExoPlayer will, in some cases, try to re-load the same `Chunk` which would re-try the same `Uri` obviously.

It could be preferable to pass the list of available base urls down to `DataSpec` (and let it be decided upon making connections), however that would break the idea of `DataSpec` as a pure data holder. I also realize that I have the need of knowing the available base urls during requests (which I'm not with the current state of this PR unless going for some semi ugly hack). Would love to get @ojw28 involved in the discussions as well.

abhiguru:
Ideally, I think the MediaChunk Class should hold an attribute like triedAndFailedDueTo = 404 or something similar which would allow us to skip adding the additional guard condition and instead the getChunkOperation can make the decision to fetch alternateURI if the last one 404 or timeout etc...

pakerfeldt:
I prefer to not have any such logic in any other place but an implementation of `BaseUrlSelector`. What if all you want is some round robin logic that picks any base url regardless of any previous error?

abhiguru:
Yup correct that logic will be in BaseUrlSelector, the change in the Base Media Chunk is just to let the process of letting the Dashchunksource getChunkOperation() method know that the current chunk failed to load and it needs to be discarded and it should try to get another chunk with the different baseUrl instead of loading the same one again and when it does the BaseUrlSelector will make the decision based on whatever logic you have setup... round-robin, blacklisting etc...  That is how I have it done currently and seems to be pretty solid... 

btw props to the @ojw28  and the team who built this because the current mechanism makes such reloading of chunks and maintaining stream continuity simple

pakerfeldt:
But then it would require a failure before `BaseUrlSelector` would even be called? What if someone wants round-robin regardless of any errors?

abhiguru:
Good point, That would probably require BaseUrlSelector to hold all the BaseUrls in cache and some logic to figure out which one was last used ?

pakerfeldt:
@ojw28 Is this nothing that ExoPlayer plans to support or is there another commit that resolves the underlying issue?

ojw28:
This pull request was closed automatically because the branch it was targeting has been deleted. We continue to use issue #771 to track this feature request, but it remains relatively low priority for us, and we haven't seen much demand for it.

PR #1277, opened 2016-02-17
Un-final doSomeWork in SampleSourceTrackRenderer

pakerfeldt:


pakerfeldt:
Does it make sense to un-final this method again? We use it in our GL renderer extension and I had to un-final it after a recent exoplayer bump.

ojw28:
That method calls another `doSomeWork` method that you can override. Is there a reason why you can't override that one instead?

pakerfeldt:
I'm not entirely sure but I can try to verify it. I just assumed what happens in `doSomeWork` is vital. Like `boolean sourceIsReady = enabledSource.continueBuffering(enabledSourceTrackIndex, positionUs);`, which is later passed to this other `doSomeWork`.

ojw28:
It is vital. That's a good reason to not allow someone to override it (because they might forget to call through to super.doSomeWork). You should implement the other doSomeWork instead.

pakerfeldt:
But the final doSomeWork already calls through to the other doSomeWork? So there's nothing to forget. Or am I misunderstanding?

ojw28:
What I'm saying is:

- The work done in `final void doSomeWork(long, long)` is vital. If you remove final from that method, as proposed here, it would allow someone to make the mistake of overriding it and failing to call `super.doSomeWork(long, long)` from their override. This is why it's final.
- You should just be able to implement `doSomeWork(long, long, boolean)` to do what you need. You shouldn't need to override `doSomeWork(long, long)`.

PR #1312, opened 2016-02-25
EIA608 Styling

zsmatyas:
First patches for the missing parts of the EIA608 specification.
**Italic, Color, Underline** is supported. 
Row and Column **positioning** is not implemented yet.

Tested with
https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3bipbop_4x3_variant.m3u8
and a few other (not public) samples.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
@@ -70,6 +70,8 @@ import android.graphics.Color;
   public static final byte TAB_OFFSET_CHAN_1 = 0x17;
   public static final byte TAB_OFFSET_CHAN_2 = 0x1F;
 
+  public static final int COLUMN_INDEX_BASE = 1; // standard uses 1 based index for positioning
+
   private static final int[] COLOR_VALUES = {
           Color.WHITE,
           Color.GREEN,
@@ -172,17 +174,20 @@ import android.graphics.Color;
     return (cc1 > 0x17) ? 1 : 0;
   }
 
-  // for debug purposes
-  public String getPreambleAddressCodeMeaning() {
+  // returns a 1 based index!
+  public int getPreambleAddressCodePositionRow() {
     int cc1Bits = (cc1 & 0x7);
     int cc2bit = (cc2 & 0x20) > 0 ? 1 : 0;
 
     int index = (cc1Bits << 1) + cc2bit;
 
-    int rowAddress = ROW_INDICES[ index ];
+    return ROW_INDICES[ index ];
+  }
 
-    return "Row:"+rowAddress
-            + "; Col:" + getIndentation()
+  // for debug purposes
+  public String getPreambleAddressCodeMeaning() {
+    return "Row:"+getPreambleAddressCodePositionRow()
+            + "; Col:" + getPreambleAddressCodePositionColumn()
             + "; Color:" + getPreambleColorName()
             + "; italic:" + isPreambleItalic()
             + "; underline:" + isUnderline();
@@ -238,7 +243,7 @@ import android.graphics.Color;
   // the aspect ratio of the video content. What if the content is letterboxed? Than aspect ration
   // checks will return incorrect results. I do not introduce this dependency at this time.
   // See Federal Communications Commission §79.102
-  private int getIndentation() {
+  public int getPreambleAddressCodePositionColumn() {
     // The bits of the two bytes of the Preamble Are:
     // CC1: 0bP001CRRR CC2: 0b1RSSSSU
     // P is the parity bit for
@@ -250,7 +255,7 @@ import android.graphics.Color;
     // we are again interested in the 4 bits marked with S. If the most significant one is set to 1,
     // than the bits mean indentation, color otherwise.
     if ((cc2 & 0x10) == 0) {
-      return 0; // the S bits mean a color value.
+      return COLUMN_INDEX_BASE; // the S bits mean a color value.
     }
 
     // If we are here, the S bits mean indentation.
@@ -260,7 +265,7 @@ import android.graphics.Color;
     // the required indentation is 4 times of this value represented by the bits.
     // Note: Indentation is between 0 and 28 columns this way, and there can be 0-3 tabs also added
     // by following TAB offset codes.
-    return indentValue * 4;
+    return indentValue * 4 + COLUMN_INDEX_BASE;
   }
 
   // last bit of the styling commands (Preamble Address Code or Mid Row CODE) is "Underline Flag".
@@ -269,7 +274,7 @@ import android.graphics.Color;
   }
 
   // for debug purposes
-  private String getControlCodeMeaning() {
+  public String getMiscControlCodeMeaning() {
     switch ( cc2 ) {
       case RESUME_CAPTION_LOADING:     return "RESUME_CAPTION_LOADING";
       case ROLL_UP_CAPTIONS_2_ROWS:    return "ROLL_UP_CAPTIONS_2_ROWS";
@@ -285,19 +290,20 @@ import android.graphics.Color;
     return "UNKNOWN";
   }
 
-  // for debug purposes
+  // for debug purposes, useful as the IDE shows this string value of the commands while debugging
   @Override
   public String toString() {
-    String value = "CC1:" + Integer.toHexString(cc1) + "; CC2:" + Integer.toHexString(cc2) + " ";
 
     if (isPreambleAddressCode()) {
-      value += getPreambleAddressCodeMeaning();
+      return "PAC: " + getPreambleAddressCodeMeaning();
     } else if (isMidRowCode()) {
-      value += getMidRowCodeMeaning();
-    } else {
-      value += getControlCodeMeaning();
+      return "MRC: " + getMidRowCodeMeaning();
+    } else if (isTabOffsetCode()) {
+      return "TAB: " + (cc2 - 0x20) ;
+    } else if (isMiscCode()){
+      return "MISC: " + getMiscControlCodeMeaning();
     }
 
-    return value;
+    return "UNKNOWN - CC1:" + Integer.toHexString(cc1) + "; CC2:" + Integer.toHexString(cc2);
   }
 }
--- a/library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
+++ b/library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
@@ -15,6 +15,8 @@
  */
 package com.google.android.exoplayer.text.eia608;
 
+import android.graphics.Color;
+
 /* package */ final class ClosedCaptionCtrl extends ClosedCaption {
 
   /**
@@ -59,7 +61,6 @@ package com.google.android.exoplayer.text.eia608;
 
   public static final byte BACKSPACE = 0x21;
 
-
   public static final byte MID_ROW_CHAN_1 = 0x11;
   public static final byte MID_ROW_CHAN_2 = 0x19;
 
@@ -69,6 +70,43 @@ package com.google.android.exoplayer.text.eia608;
   public static final byte TAB_OFFSET_CHAN_1 = 0x17;
   public static final byte TAB_OFFSET_CHAN_2 = 0x1F;
 
+  private static final int[] COLOR_VALUES = {
+          Color.WHITE,
+          Color.GREEN,
+          Color.BLUE,
+          Color.CYAN,
+          Color.RED,
+          Color.YELLOW,
+          Color.MAGENTA,
+          Color.TRANSPARENT // Last color setting should be kept
+  };
+
+  // for debug purposes
+  private static final String[] COLOR_NAMES = {
+          "WHITE",
+          "GREEN",
+          "BLUE",
+          "CYAN",
+          "RED",
+          "YELLOW",
+          "MAGENTA",
+          "KEEP PREVIOUS COLOR" // Used to turn on Italics and UnderLine.
+  };
+
+  // The final position of the text should be one of the predefined ROWs of the display.
+  // The target row is selected by 3 bits of CC1 and 1 bit of CC2. So let's have an array of 16
+  // indices that can be addressed by the incoming 4 bits.
+  // There is one duplication, row 11 is the default, where the bit in CC2 is not used for row
+  // selection, so the first 2 value will be 11: if the CC1 has the value of 0x10 or 0x18 than the
+  // target row is '11' irrespective of CC2.
+  final static short[] ROW_INDICES = { 11, 11, 1, 2, 3, 4, 12, 13, 14, 15, 5, 6, 7, 8, 9, 10 };
+
+  // 608 parser drops the first bits of the incoming data (parity bit), so both bytes has the max
+  // value of 0x7F.
+  // For Closed Captioning, cc1 values in the range 0 - 0xF are not defined, they are used as XDS
+  // control values (independent from closed caption control codes), so probably the parser should
+  // drop such values: the following functions do not really expect cc1 values
+  // outside the 0x10 - 0x7F range.
   public final byte cc1;
   public final byte cc2;
 
@@ -91,6 +129,12 @@ package com.google.android.exoplayer.text.eia608;
   }
 
   public boolean isPreambleAddressCode() {
+    // Note: Current implementation might throw compile warnings as cc2 cannot be bigger than
+    // 0x7F when we call this function.
+    // Preamble Code could also be checked as
+    // cc1 & 0x70 == 0x10 // bits required: 0bx001xxxx
+    // and
+    // cc2 & 0x40 == 0x40 // bits required: 0bx1xxxxxx
     return (cc1 >= 0x10 && cc1 <= 0x1F) && (cc2 >= 0x40 && cc2 <= 0x7F);
   }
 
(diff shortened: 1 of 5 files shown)


ojw28:
Thanks for this. It might be a couple of weeks before we get around to looking at this properly and getting it merged (holidays etc), but from a quick scan it looks very helpful!

zsmatyas:
Thanks! I would appreciate any quick feedback as well, as the next change sets are heavily depending on these ones. :)

library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
```diff
@@ -69,6 +71,45 @@
...
+          Color.GREEN,
+          Color.BLUE,
+          Color.CYAN,
+          Color.RED,
+          Color.YELLOW,
+          Color.MAGENTA,
+          Color.TRANSPARENT // Last color setting should be kept
+  };
+
+  // for debug purposes
```
ojw28: Can we remove all the debug stuff from committed code? Feel free to keep it locally, but it's probably >50% of the whole change to this file! Note that getIndentation is also not used anywhere, so that method can go or be deferred until a subsequent change also.

library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
```diff
@@ -15,6 +15,9 @@
  */
 package com.google.android.exoplayer.text.eia608;
 
+import android.graphics.Color;
+import android.util.Log;
```
ojw28: Unused, as is TAG below. There's normally a setting you can enable in your IDE to highlight unused stuff, so that you can get rid of it.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -58,16 +67,40 @@
...
-  private final StringBuilder captionStringBuilder;
+  private final SpannableStringBuilder captionStringBuilder;
   private final TreeSet<ClosedCaptionList> pendingCaptionLists;
 
   private boolean inputStreamEnded;
   private int captionMode;
   private int captionRowCount;
-  private String caption;
-  private String lastRenderedCaption;
+  private SpannableStringBuilder caption;
```
ojw28: It looks like caption and lastRenderedCaption can just be of type CharSequence (see related comment below).

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -334,21 +477,32 @@ private String getDisplayCaption() {
       return null;
     }
 
+    // Every styling span should be closed at end of line
+    closeOpenSpans( STYLE_PRIORITY.STYLE_PRIORITY_ALL );
+
     int endIndex = endsWithNewline ? buildLength - 1 : buildLength;
     if (captionMode != CC_MODE_ROLL_UP) {
-      return captionStringBuilder.substring(0, endIndex);
+      return new SpannableStringBuilder(captionStringBuilder.subSequence(0, endIndex));
```
ojw28: It's unclear why you're not just returning captionStringBuilder.subSequence(0, endIndex) directly, and having the return type be CharSequence. It looks like it should work fine.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -158,14 +191,15 @@ protected boolean isReady() {
...
   }
 
-  private void invokeRenderer(String text) {
+  private void invokeRenderer(SpannableStringBuilder text) {
     if (Util.areEqual(lastRenderedCaption, text)) {
       // No change.
       return;
     }
-    this.lastRenderedCaption = text;
+    final SpannableStringBuilder textToRender = text == null ? null : new SpannableStringBuilder(text);
```
ojw28: This line looks unnecessary. Have invokeRenderer accept a CharSequence and just pass it directly to the target?

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -58,16 +67,40 @@
...
   private int captionRowCount;
-  private String caption;
-  private String lastRenderedCaption;
+  private SpannableStringBuilder caption;
+  private SpannableStringBuilder lastRenderedCaption;
   private ClosedCaptionCtrl repeatableControl;
 
+  private static final String TAG = "Eia608Captions";
+
+  // Incoming styling commands are ordered, a higher priority command clears any previous lower
```
ojw28: Just for my own education: I assume this is defined in the spec, as opposed to being a decision you made :)?

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -58,16 +67,40 @@
...
+  // priority commands. Similarly, output SPANs cannot partially overlap, so we need to close all
+  // lower priority SPANs in case of opening a higher priority one. Having these preconditions we
+  // do not need stack or more complex tracking of styles
+  private enum STYLE_PRIORITY {
```
ojw28: We avoid using enum by policy. Please use private static final integers.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -334,21 +477,32 @@ private String getDisplayCaption() {
       return null;
     }
 
+    // Every styling span should be closed at end of line
+    closeOpenSpans( STYLE_PRIORITY.STYLE_PRIORITY_ALL );
```
ojw28: This doesn't look right. There's no guarantee that a line has ended when getDisplayCaption is called. In particular, this would not be true when in paint-on caption mode.

I think it you might need to do something like adding spans up-front using SPAN_EXCLUSIVE_INCLUSIVE, then replacing them SPAN_EXCLUSIVE_EXCLUSIVE spans when you actually reach an end-of-line.

ojw28:
Added a bunch of comments. Looks like the right approach in general. I think a little fine-tuning may be required though! See in particular the comment about paint-on caption mode.

zsmatyas:
Thanks! Updates in progress.

I have issues with paint-on captions in general, as it does not make sense with digital transmission, character alignment and fonts any more. As each character was exactly defined how to look like and the character width was uniform, overwriting a character in the middle of the row did not cause any necessary repositioning of any other characters. But today, this is not true. I tested two versions: 
- saving every text and modifying them before repaint: every character position to the right of the overwritten character is constantly changing: the text is not readable.
- creating and showing a new caption over the first one: as character width is not uniform, the overlap is unpredictable leading to unexpected results. So, at the moment, I just hope no one wants to use paint-on mode any more...

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: I read the spec many times, but it is quite ambiguous.
This order of priority of styling commands is more clear in the legal requirement: Closed caption decoder requirements for analog television receivers:

https://www.law.cornell.edu/cfr/text/47/79.101
Section: (h) Character Attributes— Point (1)/ii
"The color attribute has the highest priority and can only be changed by the Mid-Row Code of another color. Italics has the next highest priority. If characters with both color and italics are desired, the italics Mid-Row Code must follow the color assignment. Any color Mid-Row Code will turn off italics. If the least significant bit of a Preamble Address Code or of a color or italics Mid-Row Code is a 1 (high), underlining is turned on."
"Thus, for example, if a red, italicized, underlined, flashing character is desired, the attributes must be received in the following order: a red Mid-Row or Preamble Address Code, an italics Mid-Row Code with underline bit, and the Flash On command. The character will then be preceded by three spaces (two if red was assigned via a Preamble Address Code)."

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: I need the subclass of CharSequence (SpannableStringBuilder) to carry the formatting information as well, and I create a new instance before sending it to the renderer to avoid concurrent modifications from this thread as now we operate on references. Am I missing something?

library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
zsmatyas: This is the sign that I already removed most of the debugging before submitting the code ;)

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: Do you mean that the variable can be defined as the more generic supertype and I can still initialize it to SpannableStringBuilder? I guess it should work...

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: Done.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: The function signature should be OK with the supertype, but SpannableStringBuilder seems to be necessary to carry the formatting information.

I just found a copy constructor of SpannableStringBuilder for this case:
http://developer.android.com/reference/android/text/SpannableStringBuilder.html#SpannableStringBuilder%28java.lang.CharSequence,%20int,%20int%29

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
zsmatyas: There are issues present, among others:
- Paint On and Roll Up captions does not make too much sense without correct positioning as continuous movement renders text hardly readable. (Positioning is an upcoming change-set after these).
- Erase commands are not implemented correctly, so captions might stay on the screen longer than desired.
- Carriage returns are handled with adding line breaks to the text, leading to incorrect positioning. Line breaks do not match the behavior of the standard, as it defines rows of (somewhat independent) characters, line breaks are artificial handling of rows.
All these will be handled in an upcoming change.

I can update these lines to handle Paint On and Roll Up captions more correctly without positioning, unfortunately, I have never tested these as Paint On samples are unreasonable, while Roll Up samples tend to not have formatting.

library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
zsmatyas: I added getIndentation() while I was trying to figure out the meaning of the bits of the stream. Although it is not useful without full positioning, it might help anyone how is attempting to update the code. 
Now, when you stop at a break point, Android Studio is able to write out every piece of information about the incoming Commands instead of a single HEX value. Shall I remove it? (Either ways I will add most of it back with the next change set implementing correct positioning).

zsmatyas:
Any comments on the second version? :)

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608TrackRenderer.java
```diff
@@ -58,16 +67,40 @@
...
+  // defined only to give a name to the highest priority (means "every spans").
+  private static final int STYLE_PRIORITY_ALL = 4;
+
+  // Characters (text to display) and control codes (formatting instructions) are arriving
+  // one-by-one, and we want to translate this incoming byte stream into a character sequence with
+  // spans. We need the start and the end position to correctly set up a span, so we will use the
+  // following variables to store information about previously received formatting instructions.
+  // When we get to a point where we should end or change formatting, we can set up the spans
+  // between the stored start position and the current position (new control code, end of line, etc)
+  private int styleSwitchUnderlineStartPos;
```
zsmatyas: These need to be inited to -1, default 0 means that start this Tag from pos 0 incorrectly.

library/src/main/java/com/google/android/exoplayer/text/eia608/ClosedCaptionCtrl.java
ojw28: It would be preferable to remove it until it's needed, yes. It makes it easier to find "the change that added positioning" if its contained within a single change, as opposed to some of it arriving first :).

ojw28:
Hm, something weird happened to this pull request that's resulted in the version I previously commented on being no longer available. Did you do a force push or similar? For future reference, it's preferable not to do this. I can't find any way of looking at the delta between that version and the latest one as a result :).

That aside, apologies for the delay. If you could remove all the debug code + code that's unnecessary for now, then I'll patch it in and give it another detailed look. Thanks!

zsmatyas:
Yes, I force pushed, as the head of the Dev branch moved, and my fork did not seem to have those changes. How else should I have done the fast forward on the fork?

I have updated the second patch to correct the possible bug when closing tags, but some of those changes got removed by the 3rd commit, as the render() is called at different points.  

Now, with the 3rd commit, every debug function is used by toString(), so the debugger can show the Closed Caption commands instead of the Byte values. LogCat will not contain these lines until you turn on 
boolean DEBUG = false;
in Eia608TrackRenderer.java

Unused functions are:
- getPreambleAddressCodeChannel(), as I have not met any samples with multiple text channels in a single text stream.
- isModePopOn() - only added to be symmetric with isModeRollUp() and isModePaintOn().

ojw28:
Not sure how else you should have done it; I don't work too much with Git/GitHub directly :)! The simplest solution is presumably just to not fast forward your fork (since it's a feature branch, there's no particular reason why you need to continually merge the latest unrelated changes into it).

I'll take a look tomorrow.

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608CueBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+  private float position;
+  private float line;
+  private int rowIndex;
+
+  public Eia608CueBuilder() {
+    reset();
+  }
+
+  public Eia608CueBuilder(Eia608CueBuilder other) {
+    text = other.text;
```
zsmatyas: Here we need a new object, not just a new reference :(

zsmatyas:
There is one more issue present, not directly in the code changes. The specification defines 15 independent rows for captions and lets you display them in the middle 80% of the screen only (to avoid overscan problems of old analog televisions). But the default font size and background settings of the captions in ExoPlayer are slightly bigger than the one defined in the standard, so you can only fit 14 lines into the middle 80% without overlap.

As dropping a line is not an option, and overscan should be not an issue any more, I currently scale the 15 lines to the middle 90% of the screen, so there is no overlap, but the vertical positions are not 100% accurate. This is marked in the code with comments.

(Similarly, as the characters are not uniform width like they were in the 80s when the specification was defined, only the first characters of every line will be correctly positioned horizontally, the rest of the line will only depend on the font. This seemed to be enough for all the test samples, modes and use cases I have met till now)

We can update the default fonts, background around the captions, but I did not want to modify things like that in this change set.

zsmatyas:
Any news? :)

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608CueBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+  private float position;
+  private float line;
+  private int rowIndex;
+
+  public Eia608CueBuilder() {
+    reset();
+  }
+
+  public Eia608CueBuilder(Eia608CueBuilder other) {
+    text = new SpannableStringBuilder(other.text);
```
zsmatyas: I have no idea what changed, but this line started throwing exception with the same content if other.text is null. So an extra check is probably necessary here :)

library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608CueBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    rowIndex = rowIdx; // saved for roll-up feature
+
+    // 10% of screen is left of for safety reasons (analog overscan)
+    // the leftover 80% is divided into 15 equal rows. The problem is that the font size
+    // must match the row line height for this, so at the moment, I scale to 90% instead, to avoid
+    // overlap of the borders around the rows.
+    // -1 as the row and column indices are 1 based in the spec
+    this.line = (rowIdx - 1) / 15f * 0.9f + 0.05f;
```
zsmatyas: This needs minor updates to get closer to the original standard. To fit 15 lines into the safe area, previously I set 15 points (position is the top left corner of a line) evenly in the middle 80% of the screen. But positioning the bottom line exactly to 80% would cause the text of that last line go under its top left corner - effectively outside the intended safe area. 

So let's distribute "80% - font size" evenly: 
```
final float FONT_SIZE_PERCENTAGE = 0.8f / 15f;
this.line = (rowIdx - 1) / 15f * (0.8f - FONT_SIZE_PERCENTAGE) + 0.1f;
```
But this pulls all the lines closer to each other, and the default subtitle rendering might cause overlapping edges with the default font and subtitle background setting. I solved this by reordering the calls in cuePainter:
- Draw all cue backgrounds first
- then Draw all texts
This way there are no issues of any overlaps with the default fonts and settings.

PR #1324, opened 2016-02-29
UdpDataSource: use Uri's methods to access host, port parts

gpinigin:
Current implementation doesn't work at least for udp://@239.0.1.2:1234, moreover parsing is already done by Uri


ojw28:
Thanks!

PR #1342, opened 2016-03-06
showing srt subtitle in backward and forward seek

bluemix:
if there is an srt subtitle, and the video is seek backward, the subtitle stops from updating; the edited code is a simple fix to this problem.


ojw28:
I see what you're doing now, but I don't think it's necessary. There was an issue like the one you describe in 1.5.4, but it was fixed in 1.5.5 in https://github.com/google/ExoPlayer/commit/813031b8452b611c1edb93571f8a1daaaebe49ad. Were you using 1.5.4 when you were seeing this issue? If so, just updating to 1.5.5 should resolve the issue.

bluemix:
yes, I see it now. I was using 1.5.4, and I saw 1.5.5 and the problem is fixed :)
thanks anyway :)

PR #1345, opened 2016-03-07
OkHttp: missing 'Content-Type' header causes NullPointerException even if content type predicate isn't set

gpinigin:


gpinigin:
Closed: Incorrect target branch

PR #1366, opened 2016-03-13
Rename demo-misc-vp9-opus-sw to demo-ext

erdemguven:
With the inclusion of flac extension to demo-misc-vp9-opus-sw, its name needs to be more generic. Also updated the README.md.


erdemguven:
I signed it!

PR #1367, opened 2016-03-13
Fix flac readme.

erdemguven:
Added missing link and moved flac download instructions to a separate block.

### DIFF
--- a/extensions/flac/README.md
+++ b/extensions/flac/README.md
@@ -31,13 +31,17 @@ FLAC_EXT_PATH="${EXOPLAYER_ROOT}/extensions/flac/src/main"
 *   Download the [Android NDK][] and set its location in an environment
     variable:
 
+[Android NDK]: https://developer.android.com/tools/sdk/ndk/index.html
+
 ```
 NDK_PATH="<path to Android NDK>"
 ```
 
-*   Download and extract flac-1.3.1 as "${FLAC_EXT_PATH}/jni/flac" folder `curl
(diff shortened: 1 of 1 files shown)


extensions/flac/README.md
```diff
@@ -31,13 +31,17 @@ FLAC_EXT_PATH="${EXOPLAYER_ROOT}/extensions/flac/src/main"
...
 ```
 NDK_PATH="<path to Android NDK>"
 ```
 
-*   Download and extract flac-1.3.1 as "${FLAC_EXT_PATH}/jni/flac" folder `curl
-    http://downloads.xiph.org/releases/flac/flac-1.3.1.tar.xz | tar xJ && \ mv
-    flac-1.3.1 flac`
+*   Download and extract flac-1.3.1 as "${FLAC_EXT_PATH}/jni/flac" folder
+
+<pre><code>
```
ojw28: Is there a reason why this isn't

```
curl http://downloads.xiph.org/releases/flac/flac-1.3.1.tar.xz | tar xJ && mv flac-1.3.1 flac
```

I don't think we use <pre><code> anywhere else (and it seems to put an unusual amount of vertical space into the box).

ojw28: NB - The comment above is kind of confusing because GitHub renders it. Look at how "NDK_PATH="<path to Android NDK>" is put in a box above to see what I mean!

erdemguven: I used pre and code blocks as Android Studio markdown viewer somehow doesn't escape &s in ''' block. Noticing it works fine in Github I changed it to '''.

PR #1372, opened 2016-03-14
Add zlib compression support to WebmExtractor

IanDBird:
This change adds support for zlib compression to the `WebmExtractor` for basic samples. Zlib compression can be used for tracks which represents different subtitles (in my actual case, PGS). These are non video/audio tracks since those are already compressed and would not benefit from additional compression.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
@@ -44,6 +44,8 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.zip.DataFormatException;
+import java.util.zip.Inflater;
 
 /**
  * An extractor to facilitate data retrieval from the WebM container format.
@@ -152,6 +154,9 @@ public final class WebmExtractor implements Extractor {
   private static final int LACING_FIXED_SIZE = 2;
   private static final int LACING_EBML = 3;
 
+  private static final int CONTENT_COMPRESSION_ZLIB = 0;
+  private static final int CONTENT_COMPRESSION_HEADER_STRIPPING = 3;
+
   /**
    * A template for the prefix that must be added to each subrip sample. The 12 byte end timecode
    * starting at {@link #SUBRIP_PREFIX_END_TIMECODE_OFFSET} is set to a dummy value, and must be
@@ -190,6 +195,7 @@ public final class WebmExtractor implements Extractor {
   private final ParsableByteArray seekEntryIdBytes;
   private final ParsableByteArray sampleStrippedBytes;
   private final ParsableByteArray subripSample;
+  private final ParsableByteArray sampleDecompressedBytes;
 
   private long segmentContentPosition = UNKNOWN;
   private long segmentContentSize = UNKNOWN;
@@ -235,6 +241,7 @@ public final class WebmExtractor implements Extractor {
   private int sampleBytesWritten;
   private boolean sampleRead;
   private boolean sampleSeenReferenceBlock;
+  private Inflater sampleInflator;
 
   // Extractor outputs.
   private ExtractorOutput extractorOutput;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -402,6 +410,12 @@ public int read(ExtractorInput input, PositionHolder seekPosition) throws IOExce
       case ID_CONTENT_ENCODING:
         // TODO: check and fail if more than one content encoding is present.
         return;
+      case ID_CONTENT_COMPRESSION:
+        // A number of muxers (e.g. mkvmerge v4.0.0) do not write EBML elements which correspond to the default values
+        // as defined in the spec. They assume that if they provide a ID_CONTENT_COMPRESSION with no specific
+        // ID_CONTENT_COMPRESSION_ALGORITHM, zlib will be used.
+        currentTrack.contentCompression = CONTENT_COMPRESSION_ZLIB;
```
vigneshvg: instead of doing this, you can add a hasContentCompression boolean and set it to true here. contentCompression can then default to 0 rather than being explicitly set here (since that's the default according to the spec anyway [1]).

[1] https://matroska.org/technical/specs/index.html#ContentCompAlgo

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -235,6 +241,7 @@
   private int sampleBytesWritten;
   private boolean sampleRead;
   private boolean sampleSeenReferenceBlock;
+  private Inflater sampleInflator;
```
vigneshvg: isn't this a per-track setting? i.e. a file can have one compressed track and one non-compressed track right?

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
IanDBird: Sure, will change.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
IanDBird: Content compression is a per-track setting. The test file i'm using myself contains a H264 and DTS track (uncompressed) and two separate zlib compressed PGS tracks. I generated my own test file using mkvtoolnix (https://mkvtoolnix.download/) which allows the compression to be specified.

I assume you're asking this question to query why I made the `sampleInflator` scoped for the extractor, and not one per track (inside the `Track` instance)? I guess my assumption was that each individual sample is read independently, and we don't need to worry about asynchronously reading two samples at the same time (is this valid?). Since `resetSample` is called between sample reads from different tracks, the inflator instance is reset.

The alternative would be to create an Inflator instance per-compressed-track and keep it around. We would then re-use it for all sample reads rather than using the one inside the Extractor which keeps on being reset/recreated between samples of different tracks.

Is that preferable? I'm happy to change this as you see best, just want to make sure I understand before making (and testing) the changes.

ojw28:
It's a little unclear whether this is the best approach v.s. just shoving something into the format to indicate that the samples are compressed, and leaving it to the renderers to do the decompression.

One argument for this is that it keeps the data compressed in the buffer that sits in between, decompressing each sample only when it's needed (similarly to how we don't decode video frames until they're needed). An argument against is that doing it in the extractor solves the problem without the downstream renderers having to do anything. I'll have a think.

IanDBird:
Yeah, in my own (image subtitle) track renderer, I actually had it decompressing the sample before attempting to parse it. The reason I thought about moving it into the extractor was:
 - There was no existing mechanism for the extractor to flag the sample as compressed, so the renderer just tried and if failed, considered it already inflated (not ideal)
 - I assume zlib compression can be applied to other track samples, for example embedded subrip.

IanDBird:
@ojw28 - Any more thoughts on whether you want these changes, or a different approach?

PR #1374, opened 2016-03-15
Add support to identify VobSub/PGS tracks from MKV

IanDBird:
This change adds identification of PGS/VobSub tracks to MKV. With this inclusion, it's possible for applications to provide their own `TrackRenderer`s in order to handle these MimeTypes.

You'll notice that VobSub requires us to later consume the `codecPrivate` data. As mentioned by the spec (https://www.matroska.org/technical/specs/subtitles/images.html), in the case for VobSub, this contains the IDX manifest.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
@@ -82,6 +82,8 @@ public final class WebmExtractor implements Extractor {
   private static final String CODEC_ID_DTS_LOSSLESS = "A_DTS/LOSSLESS";
   private static final String CODEC_ID_FLAC = "A_FLAC";
   private static final String CODEC_ID_SUBRIP = "S_TEXT/UTF8";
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1254,6 +1258,14 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
         case CODEC_ID_SUBRIP:
           mimeType = MimeTypes.APPLICATION_SUBRIP;
           break;
+        case CODEC_ID_VOBSUB:
+          mimeType = MimeTypes.APPLICATION_VOBSUB;
+          initializationData = new ArrayList<>(1);
```
ojw28: Please use Collections.singletonList(codecPrivate);

ojw28:
Otherwise looks fine; thanks!

IanDBird:
@ojw28 - Thanks for the quick review, i've pushed the change you suggested.

Issue #178, opened 2014-11-27
Exoplayer unable to stream certain AAC encoded audio with .m3u8


spark85:
I'm porting an audio player project from Vitamio to ExoPlayer.  I've been able to get it working, except I have two audio sources that will not play.  There is no noticable difference between them and other samples that playback perfectly fine.  The issue is not sample length since very long running samples play as well.   Here are links to the sample file: 

http://s3.amazonaws.com/instarad_audio/20/208888/index.m3u8
http://s3.amazonaws.com/instarad_audio/20/208237/index.m3u8

Here is the error --> 

```
java.lang.IllegalStateException
        at android.media.MediaCodec.dequeueOutputBuffer(Native Method)
        at com.google.android.exoplayer.MediaCodecTrackRenderer.drainOutputBuffer(MediaCodecTrackRenderer.java:720)
        at com.google.android.exoplayer.MediaCodecTrackRenderer.doSomeWork(MediaCodecTrackRenderer.java:420)
        at com.google.android.exoplayer.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:398)
        at com.google.android.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:203)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:157)
        at android.os.HandlerThread.run(HandlerThread.java:61)
        at com.google.android.exoplayer.util.PriorityHandlerThread.run(PriorityHandlerThread.java:40)
```

11-27 17:50:05.423  28529-28571/com.google.android.exoplayer.demo E/OMXNodeInstance﹕ OMX_FreeBuffer for buffer h

I'm creating these samples as TYPE_HLS_MEDIA, either with AUDIO_AAC encoding or  VIDEO_H264. 

My best guess is that these samples were recorded with an unusuall sample rate, or they are somehow corrupt. 

All samples will be encoded as AAC streaming through .m3u8 format.  Are there other Renders/Encoders I should be looking at? 

Thank you, 
David

martinbonnin:
Just for reference, I just came across the same issue with AAC main profile (object type 1). Apparently, it is not supported either: See https://github.com/mstorsjo/fdk-aac/blob/047376af8c47d45c78204112a7d1b873c93822bc/libMpegTPDec/src/tpdec_asc.cpp#L1318

PR #1385, opened 2016-03-21
added a workaround for some streams that do not (cannot?) signal the correct AAC object type in their ADTS headers

martinbonnin:
The AAC ADTS header codes objectType on 2 bits which is not enough for object types above 4 like HE-AAC (AAC_SBR = 5). This patch is certainly not a perfect solution but at least it makes some streams playable.  Some other projects have similar workarounds. For an example: https://github.com/dailymotion/hls.js/blob/ce96495c701482308d2cb57d50259bcd8e2bc506/src/demux/adts.js#L51.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/ts/AdtsReader.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/ts/AdtsReader.java
@@ -55,6 +55,7 @@ import java.util.Collections;
   private final ParsableBitArray adtsScratch;
   private final ParsableByteArray id3HeaderBuffer;
   private final TrackOutput id3Output;
+  private final int overrideAudioObjectType;
 
   private int state;
   private int bytesRead;
@@ -79,10 +80,15 @@ import java.util.Collections;
    * @param id3Output A {@link TrackOutput} to which ID3 samples should be written.
    */
   public AdtsReader(TrackOutput output, TrackOutput id3Output) {
+    this(output, id3Output, -1);
+  }
+
+  public AdtsReader(TrackOutput output, TrackOutput id3Output, int overrideAudioObjectType) {
     super(output);
     this.id3Output = id3Output;
     id3Output.format(MediaFormat.createId3Format());
     adtsScratch = new ParsableBitArray(new byte[HEADER_SIZE + CRC_SIZE]);
+    this.overrideAudioObjectType = overrideAudioObjectType;
     id3HeaderBuffer = new ParsableByteArray(Arrays.copyOf(ID3_IDENTIFIER, ID3_HEADER_SIZE));
     setFindingSampleState();
   }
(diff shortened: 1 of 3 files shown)


martinbonnin:
This pull requests solves https://github.com/google/ExoPlayer/issues/178 in my case.

ojw28:
*Or*, people could stop putting media into containers that can't actually house it properly?

martinbonnin:
Yeaaa, I know :-/... 
Still, is there a solution to transport HE-AAC correctly inside HLS ?  From what I read, AAC inside TS files uses ADTS by construction. And ADTS does not work for HE-AAC... 

If HLS does not support HE-AAC, that's a bummer.

ojw28:
To "correctly" transport HE-AAC in ADTS you'd need to put a value of 5 into a field that can only represent the range [1,4]. Hence I disagree with the statement that there must be a way to transport HE-AAC correctly. There's a discussion [here](https://github.com/dailymotion/hls.js/issues/224). I would concur with the statement there, that:

> it is impossible to correctly signal HE-AAC using ADTS header

I mean, you can continue to use outdated, unsuitable, inefficient and computationally expensive container formats on Android because other industry players mandate their use, or you can deliver a high quality experience on Android devices and save the use of such container formats only for devices where they're required (which might actually help advance the industry ;)).

martinbonnin:
Well, the thing is that having mp4 for Android instead of TS might actually result in a lower experience for end users because of the CDN and network caches being less efficient, even if the container itself is more efficient. This is just a guess, this is typically the kind of thing that is hard to measure but everything is not black and white. Plus the fact that some content exists like this and it would be nice to have a way to be compatible with it.

Ffmpeg has implicit AAC decoding (beware the archeological thread !! ;-)) : http://ffmpeg.org/pipermail/ffmpeg-devel/2009-December/062737.html

HLS passes the object type in the CODECS attribute of the manifest.

I get your point but there might be solutions to this very specific problem.


On Tue, Mar 22, 2016 at 7:59 AM, ojw28 < notifications@github.com [notifications@github.com] > wrote:
To "correctly" transport HE-AAC in ADTS you'd need to put a value of 5 into a field that can only represent the range [1,4]. Hence I disagree with the statement that there must be a way to transport HE-AAC correctly. There's a discussion here [https://github.com/dailymotion/hls.js/issues/224] . I would concur with the statement there, that:

it is impossible to correctly signal HE-AAC using ADTS header

I mean, you can continue to use outdated, unsuitable, inefficient and computationally expensive container formats on Android because other industry players mandate their use, or you can deliver a high quality experience on Android devices and save the use of such container formats only for devices where they're required (which might actually help advance the industry ;)).

kientux:
@martinbonnin Can you point me what's the value to put to `overideAacObjectType`? Where can I find that number? Do I need to extract it from hls source?

ojw28:
- For what it's worth it is possible to transmux MP4 to TS at the CDN level, and I think some CDN providers support this. Quite a few major streaming providers have moved to DASH now, and I've yet to hear any negatives (quite the opposite in fact).
- The CODECS attribute is optional, and even with it, I don't think there's enough information to properly reconstruct the csd buffer (could be wrong).
- From an ideological point of view I think I'd argue that you should penalize legacy formats (and devices that require them) by preferring to evict them from the CDN, but I acknowledge that this isn't a particularly practical suggestion ;).
- Since Android doesn't support AAC Main we'll likely "fix" this issue by blindly mapping audioObjectType == 1 to AAC LC and hoping for the best. I'll push a change to do this soon.

martinbonnin:
@kientux you can use '2', it's AAC_LC and should be somewhat compatible with AAC_SBR.

PR #1390, opened 2016-03-23
Feature/positional subtitles

beazlr02:
This is a patch to add region support to the TTML subtitles parser (specifically, it adds support for percentages in the offset attribute of the region element). See https://www.w3.org/TR/ttaf1-dfxp/#layout-vocabulary-region.

The TtmlParser class has been updated to support the region element, and the TtmlNode classes has been updated to support the offset attribute of this element.

Whereas previously there was a single SpannableStringBuilder which built up the active subtitle for the current time splice, there is now a RegionTrackingFormattedTextManager which tracks the current region as the TtmlParser executes and creates builders as new regions are encountered. These builders are then returned as individual Cues.


ojw28:
Hey. We've pushed TTML positioning support to the dev branch now. Apologies for the overlap in work. Please take a look and let us know of any issues/shortcomings. If you'd like to contribute tests covering specifics of the support you're going to be relying on, those would be appreciated also!

PR #1397, opened 2016-03-25
UdpDataSource: allow to setup socket timeout, change default timeout from infinite to 8000 milliseconds

gpinigin:


ojw28:
Thanks!

PR #1409, opened 2016-04-01
Fourcc vc1 metadata parse support

drhill:
Shield TV supports VC1 hardware decoding.  This allows Exoplayer to use the decoder for MKV files with VC1.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
@@ -1216,13 +1216,9 @@ public final class WebmExtractor implements Extractor {
           nalUnitLengthFieldLength = hevcData.second;
           break;
         case CODEC_ID_FOURCC:
-            Pair<List<byte[]>, Long> fourccData = parseFourCCPrivate(
-                new ParsableByteArray(codecPrivate));
-            if (fourccData.second != FOURCC_COMPRESSION_VC1)
-                throw new ParserException("Unrecognized codec identifier for FourCC.");
-            mimeType = MimeTypes.VIDEO_VC1;
-            initializationData = fourccData.first;
-            break;
+          mimeType = MimeTypes.VIDEO_VC1;
+          initializationData = parseFourCcVc1Private(new ParsableByteArray(codecPrivate));
+          break;
         case CODEC_ID_VORBIS:
           mimeType = MimeTypes.AUDIO_VORBIS;
           maxInputSize = VORBIS_MAX_INPUT_SIZE;
@@ -1314,33 +1310,30 @@ public final class WebmExtractor implements Extractor {
       this.output.format(format);
     }
 
-    private static Pair<List<byte[]>, Long> parseFourCCPrivate(ParsableByteArray buffer)
+    private static List<byte[]> parseFourCcVc1Private(ParsableByteArray buffer)
             throws ParserException {
-        try {
-            buffer.setPosition(buffer.getPosition()+16); // size, skip width(4), height(4),
-                                                         // planes(2), and bitcount(2)
-            long biCompression = buffer.readLittleEndianUnsignedInt();
-            buffer.setPosition(buffer.getPosition()+20); // skip sizeimage(4), xpel/meter(4),
-                                                       // ypel/meter(4), clrUsed(4), clrImportant(4)
-            List<byte[]> privateData = new ArrayList<>();
-            final int baseBytesLeft = buffer.bytesLeft();
-            byte[] scratchBytes = new byte[baseBytesLeft];
-            buffer.readBytes(scratchBytes, 0, baseBytesLeft);
+      try {
+        buffer.skipBytes(16); // size(4), skip width(4), height(4), planes(2), bitcount(2)
+        long biCompression = buffer.readLittleEndianUnsignedInt();
+        if (biCompression != FOURCC_COMPRESSION_VC1)
+          throw new ParserException("Unrecognized compression for FourCC.");
 
-            int realStartPos;
-            for(realStartPos = 0; realStartPos < (baseBytesLeft - 4); ++realStartPos) {
+        buffer.skipBytes(20); // skip sizeimage(4), xpel/meter(4) ypel/meter(4), clrUsed(4), clrImportant(4)
+        final int baseBytesLeft = buffer.bytesLeft();
+        byte[] scratchBytes = new byte[baseBytesLeft];
+        buffer.readBytes(scratchBytes, 0, baseBytesLeft);
 
-                if (scratchBytes[realStartPos] == 0x00 && scratchBytes[1 + realStartPos] == 0x00 &&
-                        scratchBytes[2 + realStartPos] == 0x01 && scratchBytes[3 + realStartPos] == 0x0f)
-                    break;
-            }
-            byte[] bytes = new byte[baseBytesLeft - realStartPos];
-            System.arraycopy(scratchBytes, realStartPos, bytes, 0, baseBytesLeft - realStartPos);
-            privateData.add(bytes);
-            return Pair.create(privateData, biCompression);
-        } catch (ArrayIndexOutOfBoundsException e) {
-            throw new ParserException("Error parsing FourCC codec private");
+        int realStartPos;
+        for (realStartPos = 0; realStartPos < (baseBytesLeft - 4); ++realStartPos) {
+
+          if (scratchBytes[realStartPos] == 0x00 && scratchBytes[1 + realStartPos] == 0x00 &&
+                  scratchBytes[2 + realStartPos] == 0x01 && scratchBytes[3 + realStartPos] == 0x0f)
+            break;
         }
+        return Collections.singletonList(Arrays.copyOfRange(scratchBytes, realStartPos, baseBytesLeft));
+      } catch (ArrayIndexOutOfBoundsException e) {
+        throw new ParserException("Error parsing FourCC VC1 codec private");
+      }
     }
 
     /**
--- a/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
@@ -1216,9 +1216,9 @@ public final class WebmExtractor implements Extractor {
           nalUnitLengthFieldLength = hevcData.second;
           break;
         case CODEC_ID_FOURCC:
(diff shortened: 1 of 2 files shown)


drhill:
I signed it!

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,61 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
       this.output.format(format);
     }
 
+    private static class BITMAPINFOHEADER {
```
ojw28: See comment on other PR about getting rid of unnecessary code. This whole thing is basically the same as skipping some bytes, checking a constant (biCompression) and skipping some more bytes. It's not necessary to define a class containing a whole bunch of stuff that's unused to do this.

drhill:
Updated.  Sent a clip via link in email (VC1 && 16bit pcm).

Anything else you want me to clean up?

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,35 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+            long biCompression = buffer.readLittleEndianUnsignedInt();
+            buffer.setPosition(buffer.getPosition()+20); // skip sizeimage(4), xpel/meter(4),
+                                                       // ypel/meter(4), clrUsed(4), clrImportant(4)
+            List<byte[]> privateData = new ArrayList<>();
+            final int baseBytesLeft = buffer.bytesLeft();
+            byte[] scratchBytes = new byte[baseBytesLeft];
+            buffer.readBytes(scratchBytes, 0, baseBytesLeft);
+
+            int realStartPos;
+            for(realStartPos = 0; realStartPos < (baseBytesLeft - 4); ++realStartPos) {
```
ojw28: Where are the bytes beyond biClrImportant defined (do you have a link to some documentation you can provide, since the documentation I found seems to suggest that biClrImportant is the last thing that should be present).

Also, why is it necessary to search for 0x00,0x00,0x01,0x0f? It appears that on the piece of test media this ends up skipping one byte, which seems a little odd.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,35 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
       this.output.format(format);
     }
 
+    private static Pair<List<byte[]>, Long> parseFourCCPrivate(ParsableByteArray buffer)
+            throws ParserException {
+        try {
+            buffer.setPosition(buffer.getPosition()+16); // size, skip width(4), height(4),
```
ojw28: buffer.skipBytes(16) is neater ;) (ditto below where you're skipping).

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,35 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+
+            int realStartPos;
+            for(realStartPos = 0; realStartPos < (baseBytesLeft - 4); ++realStartPos) {
+
+                if (scratchBytes[realStartPos] == 0x00 && scratchBytes[1 + realStartPos] == 0x00 &&
+                        scratchBytes[2 + realStartPos] == 0x01 && scratchBytes[3 + realStartPos] == 0x0f)
+                    break;
+            }
+            byte[] bytes = new byte[baseBytesLeft - realStartPos];
+            System.arraycopy(scratchBytes, realStartPos, bytes, 0, baseBytesLeft - realStartPos);
```
ojw28: Use Arrays.copyOfRange.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,35 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+            privateData.add(bytes);
```
ojw28: You can also just use Collections.singletonList here, instead of making an ArrayList and adding a single item to it.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1303,6 +1314,35 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+            return Pair.create(privateData, biCompression);
+        } catch (ArrayIndexOutOfBoundsException e) {
```
ojw28: How would index out of bounds happen? Seems like it shouldn't (in which case don't try/catch it)?

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
drhill: https://matroska.org/technical/specs/codecid/index.html
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&ved=0ahUKEwi6ko7ikfjLAhXEjZQKHQQdAVwQFgghMAE&url=https%3A%2F%2Fhaali.su%2Fmkv%2Fcodecs.pdf&usg=AFQjCNHwS8o7g09E0qzFNfwPxkY9a_wlmQ&sig2=gbb7_rCon1r_oJWrUN48VA&bvm=bv.118443451,d.dGo&cad=rja

Search for V_MS/VFW/FOURCC (and vc1).  The remaining bytes are private data.  I believe the first byte is a reserve before Profile/PPS/SPS and if you pass the full private data in you get a pink screen which is not necessarily enjoyable viewing experience. ;)

This was on a Shield TV device.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
ojw28: Ignore this; looks like you're just copying what's done elsewhere.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1212,6 +1215,14 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
           initializationData = hevcData.first;
           nalUnitLengthFieldLength = hevcData.second;
           break;
+        case CODEC_ID_FOURCC:
+            Pair<List<byte[]>, Long> fourccData = parseFourCCPrivate(
+                new ParsableByteArray(codecPrivate));
+            if (fourccData.second != FOURCC_COMPRESSION_VC1)
```
ojw28: You could move this check directly into parseFourCCPrivate and then just have the method return the List<byte[]>.

ojw28:
I added some comments to address and am unsure about the for loop. Otherwise looks pretty good to merge. Thanks!

General point - Please try and make your code match the existing style. Specifically indentation (2 char indent rather than 4, unless it's actually a broken line in which case we use 4 chars). We also always put a space either side of an operator, so always `1 + 2` rather than `1+2`.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
drhill: FourCC isn't always going to be VC1 format though.  Would you want me to return the list only on valid compression (vc1)?

drhill:
Thanks.  I'll fix them in a couple of hours.  Android Studio defaults to 4 char indent.  Easy enough to fixes.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
ojw28: Is there any documentation about the existence of a reserved byte anywhere (or even some other project skipping it)? If that's what it actually is, we should just skip a byte rather than having a for loop, probably? It would be good to understand exactly what's happening here! Thanks.

NB - Even if the for loop lives, it might be best just to look for 0x00 0x00 0x01? I don't think there's value in checking the 0x0F?

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
ojw28: There's no value in returning the list if it's not VC1, since you end up immediately throwing it away anyway, I think?

drhill: Sure.  But I set the mimetype to VC1 at that point.  If there is no list returned do you want me to throw a parserexception?

ojw28: To clarify, I was suggesting you could just throw ParserException in parseFourCCPrivate in the case that biCompression isn't the value you're looking for. Then this block just becomes:

```
mimeType = MimeTypes.VIDEO_VC1;
initializationData = parseFourCCPrivate(new ParsableByteArray(codecPrivate));
```

You could optionally rename the method to something like `parseFourCcVc1Private` if that makes it clearer under what conditions it's happy.

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
drhill: I'll be honest.  After I tried just passing the whole thing (and it went pink) I just went and double checked how Kodi is handling it because they had the same pink issue when they hooked up the hardware decoder when Nvidia added it.  My thought is when in doubt check Kodi.

https://github.com/xbmc/xbmc/blob/7f13ce1ff612283bef6dfc090c017a9e6416ac3a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp

Koying handles the VC1 extra data at line 453 (AV_CODEC_ID_VC1).  If I had to guess the 0x00,0x00,0x01,0x0f is a profile flag value.  I'm assuming he looked for that and discarded anything in front as the vc1 coded for Nvidia must be looking for the profile value first (another assumption by me).

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
drhill: Yeah, I dig good function names. :)

Will do.

drhill:
Changes up! :)

PR #1410, opened 2016-04-01
Add 16 bit PCM audio track detected including MS ACM for >2 channels 

drhill:
Seems that 1 or 2 channel tracks come in directly as PCM_INT_LIT while >2 channel tracks are wrapped in MS ACM to preserve channel mapping.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
@@ -157,12 +157,6 @@ public final class WebmExtractor implements Extractor {
   private static final int LACING_FIXED_SIZE = 2;
   private static final int LACING_EBML = 3;
 
-  private static final byte[] A_MS_ACM_GUID_PCM = {   (byte) 0x01, (byte) 0x00, (byte) 0x00, (byte) 0x00,
-                                                      (byte) 0x00, (byte) 0x00, (byte) 0x10, (byte) 0x00,
-                                                      (byte) 0x80, (byte) 0x00, (byte) 0x00, (byte) 0xAA,
-                                                      (byte) 0x00, (byte) 0x38, (byte) 0x9B, (byte) 0x71 };
-  private static final byte ACM_FORMATTAG_A_PCM_INT_LIT = 0x0001;
-
   /**
    * A template for the prefix that must be added to each subrip sample. The 12 byte end timecode
    * starting at {@link #SUBRIP_PREFIX_END_TIMECODE_OFFSET} is set to a dummy value, and must be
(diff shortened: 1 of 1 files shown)


ojw28:
Please can you provide sample media for both codec types? Ideally also for both 16-bit and non-16-bit. You can email some samples to dev.exoplayer@gmail.com if you're unable to provide them publicly. Thanks!

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1255,7 +1270,25 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+        case CODEC_ID_ACM:
+
+          WAVEFORMATEX struct = parseMsAcmCodecPrivate(new ParsableByteArray(codecPrivate));
+          if (struct.wFormatTag == ACM_FORMATTAG_A_PCM_INT_LIT ||
+                  (struct.wFormatTag == WAVEFORMATEXTENSIBLE.WAVE_FORMAT_EXTENSIBLE
+                  && (struct instanceof WAVEFORMATEXTENSIBLE) && Arrays.equals(((WAVEFORMATEXTENSIBLE)struct).SubFormat, A_MS_ACM_GUID_PCM))) {
+
+              mimeType = MimeTypes.AUDIO_RAW;
+              if (audioBitDepth != 16)
+                  throw new ParserException("ExoPlayer is only capable of handling 16-bit audio.");
```
ojw28: As a side note, we're working on adding support for re-sampling of 8/24-bit audio to. This will likely happen all the way down in AudioTrack. Or in other words, don't bother working on this, since it's near the top of our list already ;).

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1437,6 +1470,96 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
       }
     }
 
+    private static class WAVEFORMATEX {
```
ojw28: This is a huge amount of code for something that basically checks a few values. Can we just do the few-lines version?

ojw28: And delete all the constants and variables that aren't used for anything.

drhill:
Cleaned up.  I proved samples via email link for 16bit PCM and 24 bit PCM (mono and multichannel).  A_MS_ACM is tricky because the clips I sent were originally that (except the mono 24 bit), but ffmpeg changed the MKV header info when I cut the clip down.  I'll look for another way to get a clean clip that has A_MS_ACM that isn't full length as the shortest I have would be 40+ minutes of HD video & lossless audio.

Is there any other changes you would like me to make?

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1437,6 +1461,33 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
       }
     }
 
+    public static final int WAVEFORMAT_SIZE = 18;
```
ojw28: All this stuff should be private rather than public (including the method). The constants should be defined at the top of the file (method at the bottom as it is now).

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1437,6 +1461,33 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+    public static final int WAVE_FORMAT_EXTENSIBLE = 0xFFFE;
+    public static final int WAVEFORMATEX_SIZE = 22 + WAVEFORMAT_SIZE;
+    public static final int GUID_SIZE = 16;
+
+    private static final byte ACM_FORMATTAG_A_PCM_INT_LIT = 0x0001;
+    private static final byte[] A_MS_ACM_GUID_PCM = {   (byte) 0x01, (byte) 0x00, (byte) 0x00, (byte) 0x00,
```
ojw28: It's might be cleaner to represent this as:

```
private static final UUID A_MS_ACM_GUID_PCM = new UUID(0xXXXXXXXXXXXXXXXXL, 0xXXXXXXXXXXXXXXXXL);
```

library/src/main/java/com/google/android/exoplayer/extractor/webm/WebmExtractor.java
```diff
@@ -1437,6 +1461,33 @@ public void initializeOutput(ExtractorOutput output, int trackId, long durationU
...
+
+    public static int parseMsAcmCodecPrivate(ParsableByteArray buffer)
+            throws ParserException {
+
+        int bytesLeft = buffer.bytesLeft();
+        int wFormatTag = buffer.readLittleEndianUnsignedShort();
+        if (wFormatTag == WAVE_FORMAT_EXTENSIBLE && bytesLeft == WAVEFORMATEX_SIZE) {
+
+          buffer.setPosition(WAVEFORMAT_SIZE + 6); // skip waveformat(18), unionsamples(2), channelmask(4)
+          byte[] SubFormat = new byte[GUID_SIZE];
```
ojw28: This would then be:

```
UUID subFormatUuid = new UUID(buffer.readLong(), buffer.readLong());
if (subFormatUuid.equals(A_MS_ACM_GUID_PCM)) {
  ....
}
```

ojw28:
Also looks good otherwise, thanks!

ojw28:
[NB - Matching style of code comment applies to this one too]

drhill:
Changes up!

ojw28:
Thanks for the changes! FYI I made some tweaks in https://github.com/google/ExoPlayer/commit/841c10f12163dad34066b81b9a8058e2c77fd1ea. Let me know if I broke anything; I confess to not having tested the result...!

drhill:
Np.  I'll give it a test when I get home this afternoon.

drhill:
Works fine, thanks!

PR #1441, opened 2016-04-14
Expose NAME of transmuxed audio tracks as trackId

RikHeijdens:
This change exposes the NAME of the audio rendition as trackId for transmuxed audio tracks, similar to external audio renditions.

Example:
Given the following playlist segment from "Apple master playlist advanced" in the Samples:

```
#EXTM3U
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="bipbop_audio",LANGUAGE="eng",NAME="BipBop Audio 1",AUTOSELECT=YES,DEFAULT=YES
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="bipbop_audio",LANGUAGE="eng",NAME="BipBop Audio 2",AUTOSELECT=NO,DEFAULT=NO,URI="alternate_audio_aac_sinewave/prog_index.m3u8"
```

The first audio rendition's MediaFormat.trackId would be null, and the second audio rendition's trackId would be "BipBop Audio 2". As of this change the first audio rendition's trackId will be set to "BipBop Audio 1", similar to the external audio rendition.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
@@ -15,12 +15,12 @@
  */
 package com.google.android.exoplayer;
 
+import android.annotation.SuppressLint;
+import android.annotation.TargetApi;
+
 import com.google.android.exoplayer.util.Assertions;
 import com.google.android.exoplayer.util.MimeTypes;
 import com.google.android.exoplayer.util.Util;
-
-import android.annotation.SuppressLint;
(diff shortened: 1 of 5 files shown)


ojw28:
Thanks for this! This is something we've fixed already in our (currently still internal) 2.x branch. We're trying to minimize changes into 1.x at this stage, so I'm not sure whether we'll merge this. Will leave it open for now though.

ojw28:
See comment above.

PR #1485, opened 2016-04-29
Pcm float encoding

drhill:
call to allow the default encoding to be whatever is set by user.  Default is still 16bit.  In my case PCM_FLOAT_ENCODING.  It worked fine with a decoded TrueHD track, using ffmpeg, except the 5.1 track was passed as 2.0.  Is this a limitation of Android?


ojw28:
I don't really understand this change. If a track as a bit depth other than 16, then we should be parsing it from the container and propagating it properly through the format object. The AudioTrack may also need to do re-sampling in the case that the bit depth is not supported. @dcower was planning to look at this at some point.

drhill:
I use this with an extension audio renderer I wrote that decodes using ffmpeg.  From my observation ffmpeg decodes everything in a float planar though once I observed signed 32bit planar.   In the decoder I do my conversion from planar to packed where I could also do resampling/channel reconfiguration, but I keep those the same.  If I were to feed AudioTrack any non 16bit audio (in this case float) I only get white noise.  Sadly, as I mentioned, I only get 2 channels no matter what the track is originally after feeding the data to AudioTrack.

ojw28:
As per my response above, the solution to this needs to be more like the one you've used in #1490.

PR #1490, opened 2016-05-01
Pcm24bit

drhill:
enabled downres of pcm signed LE 24bit audio.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaCodecTrackRenderer.java
@@ -935,6 +935,9 @@ public abstract class MediaCodecTrackRenderer extends SampleSourceTrackRenderer 
     if (codecNeedsMonoChannelCountWorkaround) {
       format.setInteger(android.media.MediaFormat.KEY_CHANNEL_COUNT, 1);
     }
+
+    if (this.format.bitdepth > 0)
+      format.setInteger(MediaFormat.BitDepthKey, this.format.bitdepth);
     onOutputFormatChanged(codec, format);
     codecCounters.outputFormatChangedCount++;
   }
--- a/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
@@ -147,6 +147,13 @@ public final class MediaFormat {
   private int hashCode;
   private android.media.MediaFormat frameworkMediaFormat;
 
+  /**
+   * The bit depth of the track, or {@link #NO_VALUE} if unknown or not applicable.
+   */
+  public final int bitdepth;
+
+  public final static String BitDepthKey = "bitdepth";
+
   public static MediaFormat createVideoFormat(String trackId, String mimeType, int bitrate,
       int maxInputSize, long durationUs, int width, int height, List<byte[]> initializationData) {
     return createVideoFormat(trackId, mimeType, bitrate, maxInputSize, durationUs, width, height,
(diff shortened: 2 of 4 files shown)


ojw28:
Hey! Thanks for this. Looks pretty good. Could you please:

- Link to some sample content that's 24 bit PCM.
- Make 8 bit PCM work too ;)? It should be very similar. Note that ENCODING_PCM_8BIT is not guaranteed to be supported by all devices. We probably want to attempt to use it, but if constructing the AudioTrack results in IllegalArgumentException then we should re-sample to 16 bit instead.

drhill:
I previously sent you a Google Drive link with a few different samples.  Two of them were 24bit pcm audio.  They were:
H264_PCM_24bit_KB2.mkv
H264_PCM_24bit_Mono_G1954.mkv

I can resend the link again if you need me to.

As I have to rebase this branch for any merge anyway I can certainly take a look at 8bit audio, but I actually don't have any samples for that as I don't believe any blu rays would have that.  Have a link to a sample?  Maybe I can look to integrate 32 bit float for >16bit tracks that are 2 channel only in the same vain and drop the other PR.  Android only supports two channels for 32bit float, correct?

Off topic but not unrelated.  Is there any chance you will be adding TrueHD passthrough anytime soon?  My initial straight forward attempt failed with massive underruns and thus choppy, at best, audio.  That and VC1 support in the Nexus Player are my only missing features... and I doubt we are going to get VC1 in hardware for the NP and the cpu can't handle a software decode (1080p).

drhill:
I pushed some changes for a refactor.  PCM 8 bit is untested as I don't have a sample.  I put hooks for when you pass 32bit float into audiotrack.  There is a commented out flag to try to set float encoding <=2 channel 24bit pcm and the conversion function has a commented out stub for the conversion... I tried.  I really spent a lot of time tonight on it and I'm sure it is something trivial I overlooked, but I couldn't get 24bit int -> 32bit float.  At one point I was close with very audible static but from then on I only managed to make things worse.  :shrug:  I'm sure you'll know what I'm missing.

I'm going to close my PR and any further 32bit development (and cleanup) should come from here.

cdenney9100:
On May 4, 2016 7:07 PM, "drhill" <notifications@github.com> wrote:
>
> I pushed some changes for a refactor. PCM 8 bit is untested as I don't
have a sample. I put hooks for when you pass 32bit float into audiotrack.
There is a commented out flag to try to set float encoding <=2 channel
24bit pcm and the conversion function has a commented out stub for the
conversion... I tried. I really spent a lot of time tonight on it and I'm
sure it is something trivial I overlooked, but I couldn't get 24bit int ->
32bit float. At one point I was close with very audible static but from
then on I only managed to make things worse. :shrug: I'm sure you'll know
what I'm missing.
>
> I'm going to close my PR and any further 32bit development (and cleanup)
should come from here.

ojw28:
Cool; thanks! I've been able to clean this up a little and also enable 8/24-bit PCM coming through WavExtractor. Both seem to work just fine. It should be fairly easy to support different PCM bit depths contained in MP4 as well (just how easy depends on where the bit depth is located in the container).

My cleanup will temporarily strip the 32-bit stuff out again, but will give a good platform to work on top of. No need to adjust this PR. I'll merge it, then merge my cleanup on top, and we can go from there!

drhill:
Cool.  Thanks!

ojw28:
Built on top of in https://github.com/google/ExoPlayer/commit/de1ccea3756f28a74922db3a15c92bd68cde3829.

drhill:
Oh no!!!  You were dropping the High Byte and not the low byte.  Should look like this if you don't want the speakers to yell angrily at you.

case C.ENCODING_PCM_24BIT:
        // 24->16 bit resampling. Drop the least significant byte.
        for (int i = offset; i < limit; i += 3) {
          resampledBuffer.put(buffer.get(i + 1));
          resampledBuffer.put(buffer.get(i + 2));
        }
        break;

drhill:
Also, I'm a little confused why you set buffer = resampleBuffer, and then allow 
      buffer.position(offset); which could be different than 0 (unless I'm missing something here) and then later on decide whether to use buffer or resampleBuffer.  It seems buffer = resampleBuffer isn't needed.  Right?

ojw28:
- Please send a pull request to fix the byte being dropped. Thanks!
- buffer is used (e.g. through the Util.SDK_INT < 21 code path), so I think it does need to be set to resampledBuffer. buffer.position(offset) is a no-op when the buffer is the resampled one, since through that code path the offset is set to resampledBuffer.position(). The handling of position/limit is pretty confusing in AudioTrack right now. We've cleaned it up quite a bit for 2.x. We hope we'll be able to push an initial version toward the end of next week (or possibly the start of the week after).

drhill:
* fix is PR 1508.
* I looked at it again after I wrote that and saw it wasn't a problem then started reading more and got a headache. Then I decided I didn't hate myself enough so I looked into TrueHD passthrough again and got annoyed again. ;)

Issue #1414, opened 2016-04-04
Add rawResponse field to BaseChunkSampleSourceEventListener.onLoadCompleted


Skrewtape:
Our backend can insert several custom fields into the manifest of a video stream, which we need to interpret at playback time.  It doesn't make sense to have ExoPlayer read, parse and understand these fields, but it would be nice if we could read them as they're downloaded by the player.

In order to solve this problem, we'd like to add a field called "rawResponse" to the method signature for the onLoadCompleted event callback, which would contains the raw bytes downloaded for that request, and would be populated for any files that have the type set to Chunk.TYPE_MANIFEST; i.e. DASH or HLS manifest files.

Alternately, if you can think of an alternate way to accomplish this same goal, we'd be more than willing to explore that path.

ojw28:
Please provide more detail. Are you using DASH, SmoothStreaming or HLS? Is this for live streaming or for on-demand? I don't think the callback to which you refer is invoked for manifest load or refresh for DASH or SmoothStreaming.

Skrewtape:
We need this functionality for live HLS playback, but I imagine it'd also make sense for DASH or SS.  Is there a different callback that could be used to handle all the cases?

ojw28:
For DASH and SmoothStreaming custom data can be embedded directly into the media in EMSG boxes, which we'll be adding support for in 2.x.

I'm not really sure what the best approach would be. Can you provide an example of what these fields look like and do, how other players provide you with access, and when you expect to receive them?

Skrewtape:
The fields are custom tags in the HLS source -- they're proprietary, and give us playback-time metadata about the video as it's playing.

For example, on iOS, we use the Resource Loader Delegate to get access to the raw HLS variant data after it's downloaded.  We see them as each HLS variant is downloaded and can update our internal state as playback proceeds.

ojw28:
Question - Is there a reason why you don't burn your metadata into the stream as ID3, which I think would be the normal way to do something like this in the context of HLS. See [this](https://developer.apple.com/library/ios/documentation/AudioVideo/Conceptual/HTTP_Live_Streaming_Metadata_Spec/Introduction/Introduction.html#//apple_ref/doc/uid/TP40010435-CH1-DontLinkElementID_11), for example. ExoPlayer already supports metadata delivery via ID3.

Skrewtape:
[...]

PR #1500, opened 2016-05-05
Adding rawRequest parameter to onLoadCompleted callback method

Skrewtape:
This change implements the feature requested in #1414


Skrewtape:
We signed a CLA as a corporation: thePlatform, Inc.

ojw28:
We really don't want HLS specific stuff propagating through classes like `BaseChunkSampleSourceEventListener`, documented with "if this is HLS". This approach doesn't scale and is generally quite confusing.

It feels like it would be better if `HlsChunkSource` were to have a listener interface for this. You could modify `MediaPlaylistChunk` to stash the raw data, and invoke the listener to pass the data from `HlsChunkSource.onChunkLoadCompleted`.

Skrewtape:
Thanks, that's a good suggestion.  Our dev who works on this stuff is on vacation this week, but we'll have a new PR for you next week with this implementation.

PR #1508, opened 2016-05-07
pcm24LE->16 needs to skip the first byte (lower) for a channel, not the last (higher)

drhill:


ojw28:
Hmm. I'm not sure about this. I have a 24-bit WAV sample that played fine with the code how it was, and is basically white noise after this change. Are you certain this change is correct, and have you tested it on some sample of your own?

ojw28:
Sample: http://download.naimlabel.com/test/30803470-8b73-43fa-a29b-570964d7eef7.aspx

drhill:
Yep, I only questioned it after testing and having loud white noise on my end which is now fixed.  You can test with the samples I gave you before.  I know mediainfo says it, but are your sample is actually Little Endian?  I know mine were Little Endian.  I've also read that pcm24 can be packed (http://wiki.multimedia.cx/?title=PCM)  but haven't seen that with my files.

ojw28:
Yeah, I gave yours a try. I'm pretty sure the sample is little endian. I don't think it would have played correctly before this change if this were not the case. It's possible there's an off-by-one error in the Wav extractor for the 24-bit case, or something like that.

ojw28:
Right, there's a "pad byte" at the start of the data chunk in WAV, which we're probably not skipping properly.

ojw28:
Urgh, ignore me, there isn't. It's probably something like that though. I'll investigate!

ojw28:
NB - There was a bug in WavExtractor that was causing data to be read from the wrong place. Will push a fix shortly. So all is good!

drhill:
Ok.  Good to know.  Thanks!

PR #1510, opened 2016-05-08
MediaFormat implements Parcelable

b95505017:
So that we could export ExoPlayer track format to another process

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
+++ b/library/src/main/java/com/google/android/exoplayer/MediaFormat.java
@@ -15,14 +15,14 @@
  */
 package com.google.android.exoplayer;
 
+import com.google.android.exoplayer.util.Assertions;
+import com.google.android.exoplayer.util.MimeTypes;
+import com.google.android.exoplayer.util.Util;
+
 import android.annotation.SuppressLint;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/android/exoplayer/MediaFormat.java
```diff
@@ -15,22 +15,25 @@
  */
 package com.google.android.exoplayer;
 
+import android.annotation.SuppressLint;
```
ojw28: Please revert the change of import order (so android ones are still below the com.google ones).

library/src/main/java/com/google/android/exoplayer/MediaFormat.java
```diff
@@ -214,6 +217,29 @@ public static MediaFormat createId3Format() {
         C.UNKNOWN_TIME_US);
   }
 
+  /* package */ MediaFormat(Parcel in) {
```
ojw28: Would be nice if the ordering here is the same as in the constructor below, rather than being slightly different ;).

b95505017: All done! :)

PR #1521, opened 2016-05-11
Discontinuity freeze bug

algobunny:
What was happening is that the provider stores Timestamp adjusters by the discontinuity index.
These indexes tend of be of low values (0, 1, 3, etc.) and different media within the same live stream will have the same indexes. But ff course they will have different timestamp adjusters.

However new timestamp adjusters are never created because the provider looks up the adjuster for the the old media, and attempts to use that. Thus resulting in the Discontinuity Detected expects x got y message. 

We don't want to use the same PtsTimestampAdjuster adjuster for different media content.


ojw28:
This doesn't look right to me. It will cause a new adjusted to be created for every chunk (since ever chunk has a different start time). That's not correct behavior.

It's invalid for an HLS live stream to be re-using the same discontinuity sequence numbers over time. It sounds like your streams are omitting EXT-X-DISCONTINUITY-SEQUENCE from the media playlists, which must be included as per the HLS spec if an EXT-X-DISCONTINUITY is ever removed from the media playlist.

algobunny:
I signed it!

ojw28:
See comment above.

algobunny:
@ojw28 The playlists do have #EXT-X-DISCONTINUITY-SEQUENCE:<d>
However I do agree it the bandage in this PR might not be the best solution. As it is undoing the optimizations provided by PtsTimestampAdjusterProvider.

I do think our team is hitting the edge case described in HlsChunkSource:

```
      if (timestampAdjuster == null) {
        // The master source has yet to instantiate an adjuster for the discontinuity sequence.
        // TODO: There's probably an edge case if the master starts playback at a chunk belonging to
        // a discontinuity sequence greater than the one that this source is trying to start at.
        return;
      }

```

Any suggestions on how this can be implemented? 

On our use case we have a live stream, at about every ~15 minute interval it stops, plays a few ads then returns to stream. We have discovered our streams no longer work after commit: https://github.com/google/ExoPlayer/commit/00f8deda3d7b558794986fcc6be47d49ea3a5b2b

ojw28:
That edge case is pretty unlikely to happen, and I don't think it would cause the behaviour you're seeing. I still suspect the source media is broken, since we haven't had other reports of this issue. You'll have to file an issue and provide links to sample content if you want us to do some debugging. It's not viable to discuss how to solve something in a pull request when we don't understand what the problem is ;)...

PR #1525, opened 2016-05-12
Added TIT2 and APIC ID3 frames

pittenga:


pittenga:
I signed it! Company: MichiganLabs

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
```diff
@@ -0,0 +1,30 @@
+package com.google.android.exoplayer.metadata.id3;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+
+/**
+ * Created by apittenger on 5/11/16.
+ */
+
+public class APICFrame extends Id3Frame{
```
ojw28: Please use 2 character indentation to be consistent with the rest of ExoPlayer's codebase.

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
```diff
@@ -0,0 +1,30 @@
+package com.google.android.exoplayer.metadata.id3;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+
+/**
+ * Created by apittenger on 5/11/16.
```
ojw28: Please add proper Javadoc (ditto for other new file).

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
```diff
@@ -0,0 +1,30 @@
+package com.google.android.exoplayer.metadata.id3;
```
ojw28: These files need to carry the same copyright header comment as the rest of the code base.

ojw28:
Please see inline comments. You also need to ensure your GitHub account is properly associated with your corp CLA (the googlebot will pick this up if it's the case).

pittenga:
Thanks for the notes. Cleaned everything up and googelbot looks like it picked up the CLA (I had to change my primary email to be my company email instead of my personal...maybe a googlebot enhancement to also look at secondary email(s)).

library/src/main/java/com/google/android/exoplayer/metadata/id3/Id3Parser.java
```diff
@@ -107,7 +107,40 @@ public boolean canParse(String mimeType) {
...
+
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        int firstZeroIndex = indexOf(frame, 0, (byte) 0);
+        String mimeType = new String(frame, 0, firstZeroIndex, "ISO-8859-1");
+        byte pictureType = frame[firstZeroIndex+1];
+
+        int descriptionStartIndex = firstZeroIndex + 2;
+        int descriptionEndIndex = indexOf(frame, descriptionStartIndex, (byte) 0);//indexOfEOS(frame, descriptionStartIndex, encoding);
```
ojw28: Isn't it correct to be using encoding here, according to the spec, as per the piece of code that's commented out?

ojw28:
Can you provide some sample content that contains these ID3 frames?

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.google.android.exoplayer.metadata.id3;
+
+/**
+ * APIC (Attached Picture) ID3 frame.
+ */
+public class APICFrame extends Id3Frame {
+  public static final String ID = "APIC";
+
+  public final byte encoding;
```
ojw28: Why is encoding needed here? I would have thought it's only relevant when parsing (we don't propagate encoding in any of the other frame classes)

library/src/main/java/com/google/android/exoplayer/metadata/id3/Id3Parser.java
```diff
@@ -107,7 +107,40 @@ public boolean canParse(String mimeType) {
...
+                objectDataSize);
+        id3Frames.add(new APICFrame((byte)encoding, mimeType, pictureType, description, objectData));
+      }else if (frameId0 == 'T' && frameId1 == 'I' && frameId2 == 'T' && frameId3 == '2') {
+        // Check frame ID == TIT2
+        int encoding = id3Data.readUnsignedByte();
+
+        byte[] frame = new byte[frameSize - 1];
+        id3Data.readBytes(frame, 0, frameSize - 1);
+
+        String description = new String(frame, 0, indexOfEOS(frame, 0, encoding), "ISO-8859-1");
```
ojw28: It looks wrong to be hardcoding "ISO-8859-1" here. Shouldn't you be using:
String charset = getCharsetName(encoding);

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.exoplayer.metadata.id3;
+
+/**
+ * APIC (Attached Picture) ID3 frame.
+ */
+public class APICFrame extends Id3Frame {
```
ojw28: Should be final (ditto for TIT2Frame). Also, by convention, we only capitalize the first letter even if the whole thing is normally in caps. So should be ApicFrame and Tit2Frame.

library/src/main/java/com/google/android/exoplayer/metadata/id3/Id3Parser.java
pittenga: No. The last parameter of indexOf takes the byte you are looking for (in our case, the "0" byte that separates the description from the binary data after it). indexOfEOS takes the encoding.

library/src/main/java/com/google/android/exoplayer/metadata/id3/APICFrame.java
pittenga: It's not. Will remove

library/src/main/java/com/google/android/exoplayer/metadata/id3/Id3Parser.java
pittenga: Just copying what was done in the GEOB and PRIV frames. Good point though. Will update.

library/src/main/java/com/google/android/exoplayer/metadata/id3/Id3Parser.java
pittenga: Compared this to the spec though and we actually should be using EOS here. Will update.

pittenga:
Sample content provided in Id3ParserTest.java

ojw28:
Oops, I think I might have merged this before you'd finished! In any case, I've done some cleanup in https://github.com/google/ExoPlayer/commit/388acb39d2145257ef52ebc76bf89d8e49309f5b, including generalizing the TIT2 frame parsing to work for any T000-TZZZ excl TXXX frame.

pittenga:
That works. Thanks!

PR #1540, opened 2016-05-18
Adding onManifestLoadCompleted method to HlsChunkSource.EventListener

Skrewtape:
This change implements the feature requested in #1414, with the modifications requested


ojw28:
We'll clean this up a little bit for consistency, but otherwise looks good. Will merge as is; the listener method name might change shortly, so be on the lookout for that. Thanks!

PR #1548, opened 2016-05-23
Remove break statement when end list tag found by hls playlist parser

dansoonie:
The http live streaming draft(https://www.ietf.org/id/draft-pantos-http-live-streaming-19.txt) specifies that the EXT-X-ENDLIST may appear anywhere in the media playlist file. In case of ExoPlayer's HlsPlaylistParser's parseMediaPlaylist() would stop parsing the file when the end list tag is found due to the break statement. The end list tag may appear anywhere if the break statement is removed.


ojw28:
Thanks!

Issue #368, opened 2015-03-25
Audio modification - speed, volume boost


vandzi:
Is it possible with ExoPlayer to change playback speed,  boost volume or other audio modifications? Thanks!

ojw28:
- There's a discussion in #26 about changing playback speed. It's not something we're planning to support directly any time soon, although as discussed there, it's possible in theory to extend ExoPlayer to do this.
- You can adjust the volume using (this)[https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/MediaCodecAudioTrackRenderer.java#L63].
- You can attach effects by doing something similar to the 4th post in #252. Note that you may need to enable the workaround as described at the end of that thread if you want to do this.

PR #1551, opened 2016-05-24
Add interface and implementation for volume control

dansoonie:
Someone issued google/ExoPlayer#368 regarding audio modification and a response was posted that MediaCodecAudioTrackRenderer can be used to carry out the task. But the ability to set volume for a player was thought to be a common task. The MediaPlayer provided in Android SDK also has an interface to set its volume, so an attempt was made to add interface and implementation for volume control. Only difference with MediaPlayer with volume controlling interface is that it also has a getVolume method along with setVolume.

Volume controls are added to PlayerActivity in order to demonstrate use of the interface and correct implementation.

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer/demo/PlayerActivity.java
+++ b/demo/src/main/java/com/google/android/exoplayer/demo/PlayerActivity.java
@@ -69,6 +69,7 @@ import android.widget.Button;
 import android.widget.MediaController;
 import android.widget.PopupMenu;
 import android.widget.PopupMenu.OnMenuItemClickListener;
+import android.widget.SeekBar;
 import android.widget.TextView;
 import android.widget.Toast;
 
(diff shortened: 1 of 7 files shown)


ojw28:
Making this type of change to the core `ExoPlayer` API is problematic due to the way that renderers are injected. The `setVolume` method simply wont work when using any audio renderers other than `MediaCodecAudioTrackRenderer`, for example. The dependency (as a result of the cast) from `ExoPlayer` to `MediaCodecAudioTrackRenderer` is also problematic from a structural point of view' `ExoPlayer` isn't supposed to care what type of renderers have been injected into it.

In 2.x we'll be adding a `SimpleExoPlayer` that extends `ExoPlayer` and injects a set of renderers suitable for most use cases. We'll be adding a `setVolume` method to that class.

ojw28:
Or to paraphrase: It's by design that `ExoPlayer` doesn't look like `MediaPlayer`. It's deliberately more bare-bones and agnostic as to what components are injected into it. In 1.x `DemoPlayer` in the demo app is the API that looks like `MediaPlayer`. In 2.x it will be `SimpleExoPlayer` that looks like `MediaPlayer`.

dansoonie:
@ojw28, I feared that was the case, but I wanted a simple interface for controlling volume and it seemed like some other people also wanted it too. I'll just have to use it for my own good for now. Thank you for pointing out  the philosophy of the design. Will be looking forward to seeing and using SimpleExoPlayer.

PR #1561, opened 2016-05-30
prevent null exception in case previousTsLoadable == null

giladna:
In order to avoid the following NPE:

Happens from time to time when pressing back button during playback.

 E/AndroidRuntime: FATAL EXCEPTION: ExoPlayerImplInternal:Handler
                                                                             Process: com.kaltura.basicplayerdemo, PID: 15803
                                                                             java.lang.NullPointerException: Attempt to read from field 'long com.google.android.exoplayer.chunk.MediaChunk.endTimeUs' on a null object reference
                                                                                 at com.google.android.exoplayer.hls.HlsSampleSource.getNextLoadPositionUs(HlsSampleSource.java:754)
                                                                                 at com.google.android.exoplayer.hls.HlsSampleSource.maybeStartLoading(HlsSampleSource.java:687)
                                                                                 at com.google.android.exoplayer.hls.HlsSampleSource.restartFrom(HlsSampleSource.java:665)
                                                                                 at com.google.android.exoplayer.hls.HlsSampleSource.onLoadCanceled(HlsSampleSource.java:441)
                                                                                 at com.google.android.exoplayer.upstream.Loader$LoadTask.handleMessage(Loader.java:240)
                                                                                 at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                                 at android.os.Looper.loop(Looper.java:148)
                                                                                 at android.os.HandlerThread.run(HandlerThread.java:61)
                                                                                 at com.google.android.exoplayer.util.PriorityHandlerThread.run(PriorityHandlerThread.java:40)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/hls/HlsSampleSource.java
+++ b/library/src/main/java/com/google/android/exoplayer/hls/HlsSampleSource.java
@@ -756,7 +756,9 @@ public final class HlsSampleSource implements SampleSource, SampleSourceReader, 
       return pendingResetPositionUs;
     } else {
       return loadingFinished || (prepared && enabledTrackCount == 0) ? -1
-          : currentTsLoadable != null ? currentTsLoadable.endTimeUs : previousTsLoadable.endTimeUs;
+         : (currentTsLoadable  != null) ? currentTsLoadable.endTimeUs
+         : (previousTsLoadable != null) ? previousTsLoadable.endTimeUs : -1;
+
     }
   }
 


ojw28:
@itaykinnrot - Are you able to reproduce this in the ExoPlayer demo app, without modification? If so, please file a new issue along with a full bug report (captured just after reproducing the issue) and the other information requested in the issue template. Also, what version of ExoPlayer are you using?

We've never managed to reproduce this ourselves. Whilst it's obviously easy to fix the NPE by adding a null check, we worry that this is simply masking some other underlying issue. Specifically, how is `enabledTrackCount` greater than zero if the playback is exiting, which is a condition for the stack trace posted in the description to occur.

itaykinnrot:
The issue occur when the app is going to background and then return - the stream gets reloaded and sometime the issue occur. hard to reproduce but clients report it -Thanks,

ojw28:
We need to understand how it actually occurs, not just when. For example, what's the sequence of execution that causes the player to get into this state? We have no way of determining whether this is the correct patch to fix the root cause if we don't fully understand how the issue arises.

PR #1562, opened 2016-05-31
Support for MPEG Audio Layer II

tresvecesseis:
Currently is difficult to determine if a Android box support MPEG Audio other than Layer III so in order to support all kind of android devices (our service, and the majority of the DVB TV services, use MP2 for the audio tracks) we had to implement a software decoder inside exoplayer, as of last summer all the relevant patents on MPEG Audio have expired so it should not be a problem to include this support in the library but I'm not sure about the libmpg123 library distribution or if you will accept an extension without the needed native libraries?


tresvecesseis:
I signed it!

PR #1581, opened 2016-06-08
Support seeking on CBR files when a partial Xing header is supplied but no Table of Contents

Plasma:
Hi,

This PR makes a change in XingHeader.create() to force a fallback to ConstantBitrateSeeker (by returning NULL in XingSeeker) when the Xing/Info header field is present, but does not contain the Table of Contents (TOC) bytes.

This allows CBR files to be seekable even if the Xing/Info header is absent or does not contain the optional TOC bytes.

By returning NULL, we end up using the ConsistentBitrateSeeker via https://github.com/google/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/extractor/mp3/Mp3Extractor.java#L313

In my experience, the implementation details of which players (across Chrome browser, Firefox, IE, iOS, etc) support seeking in CBR files varies, and various platforms and software packages can output files with or without these headers filled in completely.

By supporting this edge case, more audio files will become seekable when supported.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/extractor/mp3/XingSeeker.java
+++ b/library/src/main/java/com/google/android/exoplayer/extractor/mp3/XingSeeker.java
@@ -54,7 +54,7 @@ import com.google.android.exoplayer.util.Util;
         sampleRate);
     if ((flags & 0x06) != 0x06) {
       // If the size in bytes or table of contents is missing, the stream is not seekable.
-      return new XingSeeker(firstFramePosition, durationUs, inputLength);
+      return null;
     }
 
     long sizeBytes = frame.readUnsignedIntToInt();


andrewlewis:
Hi. Thanks for the PR. Some questions:

- Is a stream that has a Xing header without a table of contents definitely always CBR?
- Is this written anywhere authoritative, or do you know of any other players that make this assumption?
- Do you have any information on how common streams with a Xing header and not TOC are?

The current logic in ExoPlayer is not completely consistent because we treat streams with Xing headers as seekable but CBR if they are missing a frame count, and unseekable if the the frame count is present but the TOC is missing.

The Android platform in contrast treats files with Xing headers missing either the frame count or the TOC as unseekable ([see XINGSeeker.cpp](https://android.googlesource.com/platform/frameworks/av/+/master/media/libstagefright/XINGSeeker.cpp#47)).

Plasma:
Hi Andrew,

> Is a stream that has a Xing header without a table of contents definitely always CBR?

I don't believe so. A XING header can exist on a VBR file and omit the TOC flag from its supported flags, as per the specification. I am using http://www.codeproject.com/Articles/8295/MPEG-Audio-Frame-Header#XINGHeader as a good source of reference.

I think this is where heuristics come in to play. In the example I have for my file, I assume CBR based upon having access to frame count and file length. In my case, we've forked the code and its working for us.

> Is this written anywhere authoritative, or do you know of any other players that make this assumption?

Not off the top of my head, no, but I do know that ExoPlayer fails to seek the file I tested while the standard MediaPlayer does seek without issue. This is where players seem to differ slightly from what edge cases are supported.

As mentioned earlier, Chrome browser last year decided to change how it handled seeking (requiring a XING header even one that had no fields) due to a code change, causing SoundCloud and other services to no longer seek -- audio needed to be re-encoded.

> Do you have any information on how common streams with a Xing header and not TOC are?

Not really, no. I'd probably just sample a few audio files from popular online services, but in our experience players differ in how they support seeking and its an implementation detail that's a bit unfortunate, probably due to the spec being a bit grey in this area.

ojw28:
Please re-send this pull request into the `dev-v2` branch if it's still applicable. I'm not sure this is a good idea though; if media is supposed to be seekable, then it should contain sufficient information to enable seeking without the player having to resort to heuristics. @andrewlewis - any comments on the above?

andrewlewis:
Given the lack of authoritative information on how to interpret this case, I think we should continue to be conservative and treat it as a signal that the file is VBR so can't be seeked without a TOC.

If we see that the vast majority of files with a Xing header lacking a TOC are actually CBR, then we should make this change, but I don't think we have enough data to justify it yet.

Applications that can control the input media format and care about seeking should use another container, as VBR MP3 seeking is approximate anyway.

PR #1629, opened 2016-06-23
Feature/clean build dependencies

danielgomezrico:
Make easy to update the compile sdk number on each module or build tools version or any other dependency.


ojw28:
We don't take changes directly into the master branch. We currently take bug fixes only in `dev` and other changes into `dev-2.X.X-experimental`

PR #1637, opened 2016-06-28
Delete redundant final modifiers.

hotchemi:
- Delete redundant final modifiers.


hotchemi:
I signed it!

ojw28:
Thanks, however we're ramping down development of v1 and so only intend on accepting bug fixes at this point. If there's an equivalent change in the 2.x.x experimental branch, feel free to propose a change there.

hotchemi:
Got it.

PR #1668, opened 2016-07-11
Fallback to DTS passthrough when HD isn't supported

IanDBird:
The following PR handles the passthrough scenario when DTS-HD content is being played, but the attached passthrough device only supports DTS. Since the HD part of the audio codec is implemented as an extension, it's possible to silently drop it by simply configuring the `AudioTrack` to plain DTS.


andrewlewis:
As discussed, it's probably fine to merge a workaround for this but I will follow up to check if it's safe to treat DTS-HD as DTS first.

ojw28:
This pull request was closed automatically as a result of branch restructuring. Apologies for the inconvenience. Please send a new pull request to `dev-v1` if appropriate, I'm not entirely sure where the discussion above ended up. Thanks!

PR #1682, opened 2016-07-19
Fix H.264 intra refresh live stream playback problem (issue #1654)

### LINKED ISSUE #1654: Unable to play h.264 intra-refresh video from the middle of a live stream
## Steps to reproduce
1. Generate a live stream with h.264 intra-refresh in format MPEG-TS over HTTP. E.g. on Windows:
   
   ```
   ffmpeg -f gdigrab -framerate 10 -i desktop -filter:v fps=10 -filter:v scale=640:-1 -f mpegts -muxdelay 0.1 -c:v libx264 -pix_fmt yuv420p -b:v 400K -g 40 -tune zerolatency -x264opts intra-refresh=1:slice-max-size=1300 udp://236.0.0.5:7020?pkt_size=188
   
   "C:\Program Files (x86)\VideoLAN\VLC\vlc.exe" -vvv -I dummy "udp://@236.0.0.5:7020" ":sout=#duplicate{dst=http{mux=ffmpeg{mux=mpegts},dst=:8080/}}"
   ```
2. Play the stream `http://<IP>:8080` with ExoPlayer demo (You need to add an entry); or record part of the stream by `wget` and play the TS file with ExoPlayer demo. I attached a TS file (h264-intra-refresh-middle.ts) and you can play it directly.
## Actual result

The screen of ExoPlayer demo is black, and the playBackState changes from "buffering" to "ended"; the dimension of the video is correct. Aways reproducible.
## Expected result

ExoPlayer demo play the video normally (there may be black screen for several seconds at the beginning, like VLC).
## Other info
- If I disable intra-refresh (remove `-x264opts intra-refresh=1:slice-max-size=1300` from the ffmpeg command line), or record the stream to a file (`a.ts`) from the very beginning, ExoPlayer  demo can play it without problem:

```
ffmpeg -f gdigrab -framerate 10 -i desktop -filter:v fps=10 -filter:v scale=640:-1 -f mpegts -muxdelay 0.1 -c:v libx264 -pix_fmt yuv420p -b:v 400K -g 40 -tune zerolatency -x264opts intra-refresh=1:slice-max-size=1300 pipe: > a.ts
```
## Device

Lenovo 2 A8-50F, Android 5.1.
## ExoPlayer library version

 r1.5.9, from the git master. 
## attachements
- test case: https://github.com/duanyao/ExoPlayer/blob/test/test/h.264-intra-refresh-middle-stream/h264-intra-refresh-middle.ts
- bugreport: https://github.com/duanyao/ExoPlayer/blob/test/test/h.264-intra-refresh-middle-stream/bugreport.txt

duanyao:
Fix #1654 .

Follow @andrewlewis's suggestion to synthesize a key frame for the stream. The main limitaion is that not all resolutions that MediaCodec decoders support are also supported by MediaCodec encoders, so high resolution video stream may still not playable (the limit is ~1280x720 on my device). 

Videos with color formats other than YUV420 may also not work. I have no idea how to figure out the actually color format of an input buffer and always assume YUV420.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/C.java
+++ b/library/src/main/java/com/google/android/exoplayer2/C.java
@@ -143,6 +143,11 @@ public interface C {
   int BUFFER_FLAG_DECODE_ONLY = 0x80000000;
 
   /**
+   * H.264 may be with intra-refresh enabled.
+   */
+  int BUFFER_FLAG_INTRA_REFRESH = 0x02000000;
+
+  /**
    * A return value for methods where the end of an input was encountered.
    */
   int RESULT_END_OF_INPUT = -1;
--- a/library/src/main/java/com/google/android/exoplayer2/decoder/Buffer.java
+++ b/library/src/main/java/com/google/android/exoplayer2/decoder/Buffer.java
@@ -87,7 +87,7 @@ public abstract class Buffer {
    * @param flag The flag to check.
    * @return Whether the flag is set.
    */
-  protected final boolean getFlag(int flag) {
+  public final boolean getFlag(int flag) {
     return (flags & flag) == flag;
   }
 
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/DefaultTrackOutput.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/DefaultTrackOutput.java
@@ -521,7 +521,7 @@ public final class DefaultTrackOutput implements TrackOutput {
         pendingSplice = false;
       }
       if (needKeyframe) {
-        if ((flags & C.BUFFER_FLAG_KEY_FRAME) == 0) {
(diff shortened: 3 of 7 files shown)


duanyao:
I signed CLA.

ojw28:
This pull request was closed automatically as a result of branch restructuring. Apologies for the inconvenience.

I think @andrewlewis had some concerns about side effects of this change, which perhaps he can comment on.

andrewlewis:
Thanks. This is a very cool workaround! However, I don't think we should integrate it into the core library:
- As you point out, it is not safe to rely on devices having an encoder for the format they need to decode.
- As written, there are no guarantees that the encoded frame will be compatible with the stream to be decoded. For example, the encoded frame might use a different profile than the stream to play.
- Applications using this will need to signal out of band how many frames to drop before the frame has been completely refreshed. To avoid this limitation in general, it would be necessary to parse the incoming stream and figure out what macroblocks had been refreshed, which would involve integrating an H.264/AVC bitstream parser into the library.

Issue #1216, opened 2016-02-01
Change playback rate of video track


jmgirven:
I would like to reduce the playback rate of a video to 50%, or increase it to 200%. #26 is a similar thread, but contrary to what they are trying to do, I don't really care about the audio. One of the first comments is:

> To clarify, do you mean with audio (sped up but corrected to still be at the correct pitch) or without?

My answer is without audio. Does this make a significant different to the complexity of the answer?

I am primarily interested in locally stored mp4 files. I am looking to support Android 16 and above.

It sounds like the playback speed of the video is tied to the speed of the audio. Is that always true? I have tried using this [gist](https://gist.github.com/jedhoffmann/b99b384be76e353dc08f) to slow the audio, but as @jedhoffmann says in his comment on #26, he hasn't factored in video. I find that with video and slow motion, the playback becomes very jerky (far more than should be expected from having fewer frames rendered to the screen per second).

This [sample](https://github.com/skonb/MediaCodecSpeedControllableTrackRenderer) does suggest it accounts for video (albeit based on an older version of exoplayer), but the main contribution of `mCurrentSpeed` to the code is to modify the `joiningDeadlineUs`. Is this what I am missing to get smooth playback in slow motion? Naively, it sounded like an optimisation rather that being key.

AquilesCanta:
So, for the sake of simplicity let us say you have only a Video track. In that case, of course, the playback speed of video is linked to real time (since there is no audio). I am tempted to think that the provided gist is a bit of an overkill. 

I would try to use an implementation of MediaClock that suits your needs. Something like the  StandaloneMediaClock but with your adaptations to support the desired playback speed. You shouldn't find yourself in need of playback implementation modifications. Keep me updated if you feel you have met an Exoplayer limitation.

jmgirven:
OK, thanks. It seems #571 is related too then.

jmgirven:
[First effort](https://gist.github.com/jmgirven/4e0ba2c7d584c8695549)

AquilesCanta:
Does it work well? Have you tried just using your own MediaClock implementation instead of modifying the renderer?

jmgirven:
[...]

Issue #571, opened 2015-06-22
Frame rendering timing with MediaClock


Arqu:
Considering the new implementation of MediaClock, the timing of frames rendered seems no longer to be handled by MediaCodecVideoTrackRendered but more by the the MediaCodecAudioTrackRenderer with the MediaClock implementation. Since I need to "adjust" the frame rate on my own in some of my cases (I know this is a bit unrelated to the general implementation), I wonder how can we manage playback speed/rendering timing now (currently it is ok to disable or distort the audio for me).

ojw28:
There hasn't been any functional change in terms of how timing is implemented. MediaCodecAudioTrackRenderer has always driven the timing. Previously it indicated this by returning true from isTimeSource. Now it does so by returning a clock instance from getMediaClock. This is solely a restructuring of the code, not a change in how it works.

Given your use cases are custom, you'll have to figure them out on your own!

Arqu:
Thanks managed to figure it out with getMediaClock and getPositionUs().

PR #1691, opened 2016-07-22
Implement variable playback speed.

### LINKED ISSUE #26: Change playback rate (video + audio only use cases)
Hello guys. Is it possible to change playback rate of video track e.g play faster or play slow ?

duanyao:
Add a new API to `ExoPlayer` - `setPlaybackSpeed(fload)/getPlaybackSpeed()`, and also add `getPlaybackSpeed()` to `MediaClock` . If the media has audio and API level >= 23, `android.media.PlaybackParams` is used to change the playback speed; otherwise `StandaloneMediaClock.setPlaybackSpeed(fload)`  is used instead. If API level < 23 and speed is not 1, the audio is jumpy, though the video plays well.

Might fix #26, #1216, #571 .

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/ExoPlayer.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ExoPlayer.java
@@ -366,4 +366,13 @@ public interface ExoPlayer {
    */
   int getBufferedPercentage();
 
+  /**
+   * @return the speed factor: speed_of_playback / speed_of_real_clock
+   */
+  float getPlaybackSpeed();
+
+  /**
+   * @param speed the speed factor: speed_of_playback / speed_of_real_clock
+   */
+  void setPlaybackSpeed(float speed);
 }
--- a/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java
@@ -206,6 +206,12 @@ import java.util.concurrent.CopyOnWriteArraySet;
         : (int) (duration == 0 ? 100 : (bufferedPosition * 100) / duration);
   }
 
+  @Override
+  public float getPlaybackSpeed() { return internalPlayer.getPlaybackSpeed(); }
+
+  @Override
+  public void setPlaybackSpeed(float speed) { internalPlayer.setPlaybackSpeed(speed); }
+
   // Not private so it can be called from an inner class without going through a thunk method.
   /* package */ void handleEvent(Message msg) {
     switch (msg.what) {
--- a/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -30,6 +30,8 @@ import com.google.android.exoplayer2.util.StandaloneMediaClock;
 import com.google.android.exoplayer2.util.TraceUtil;
 import com.google.android.exoplayer2.util.Util;
 
+import android.media.PlaybackParams;
+import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Message;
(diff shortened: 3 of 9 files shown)


ojw28:
This pull request was closed automatically as a result of branch restructuring. Apologies for the inconvenience.

I don't think we really want to take pull requests where the result is known to be broken though (i.e. "audio is jumpy" doesn't sound like it would be of sufficient quality for us to accept). Note also that in V2 there's a new `SimpleExoPlayer.setPlaybackParams` method, which should cleanly cover most use cases on API level 23 and above.

duanyao:
`SimpleExoPlayer.setPlaybackParams` does not work for video without audio, and doesn't work on API < 23 either. Addtionally `setPlaybackParams` is not quite a straightforward API for setting playback speed, and lacking a getter.

The issue of "audio is jumpy" of this PR just happens on API < 23, and I think it is better than nothing for now. Fixing this issue requires quite a lot of work and may need some third party library to resample the audio. Do you think I should do this in this PR?

ojw28:
I believe the majority of the demand for this feature is for playbacks that include audio. For example podcasts, audio books, and educational videos such as recorded lectures. I don't see the video only case as high priority at this time.

I disagree that jumpy audio is better than nothing. As an application developer I'd rather ship a feature that works properly or not ship that feature at all. As a maintainer of ExoPlayer, I don't feel comfortable adding an API that we'd know to be broken and have no plan to ever fix (see below).

You're correct to note that a full solution prior to API level 23 requires a non-trivial effort and most likely integration with a third party library. This is not something we plan to support and we would be unlikely to accept a pull request adding such functionality. A dependency on a third party library would be unacceptable for the core ExoPlayer library. If it were structured as an extension that would be more acceptable, but we'd be unlikely to accept a pull request regardless due to the ongoing maintenance cost. Note also that a dependency on a third party library may be problematic from a licensing point of view.

So, I think the current situation, where we solve the problem for playbacks with audio on API level 23 and above, is pretty much where we want to be right now. It's never been our goal to be all things to all people, and staying focused on supporting core use cases really well is our priority. There is of course nothing to prevent a sufficiently determined application developer from implementing this feature on top of ExoPlayer for their own use.

duanyao:
Yes video without audio is less common, but if variable speed is left unimplement for this case, applicaions based on ExoPlayer would behave inconsistently with/without audio, even on API >= 23, which would surprise users.

Similarly, applicaions based on ExoPlayer can change speed on API >= 23 but can't otherwise, which is also broken from users' perspective. I don't known whether this is more broken than jumpy audio, but I'd rather implement it and leave the choice to applicaion developers and users (our project definitely needs it).

Anyway, it is up to you to decide.

ojw28:
Just to respond to the points above:

> Yes video without audio is less common, but if variable speed is left unimplement for this case, applicaions based on ExoPlayer would behave inconsistently with/without audio, even on API >= 23, which would surprise users.

Most applications based on ExoPlayer aren't ever required to play video without audio. Consider YouTube, Facebook, any premium streaming service. All content in such cases is transcoded to include audio (even if the audio is silence). So video only media is a fairly niche use case, the primary example that springs to mind is security camera live streams. Applications that need to play video with audio _and_ video without audio _and_ expose variable speed playback is a very niche use case. Application developers can opt to not expose variable speed playback if they think the inconsistency is confusing, or can opt to do their own integration with a third party library to deliver the feature consistently if they think it's really important.

> Similarly, applicaions based on ExoPlayer can change speed on API >= 23 but can't otherwise, which is also broken from users' perspective. I don't known whether this is more broken than jumpy audio, but I'd rather implement it and leave the choice to applicaion developers and users (our project definitely need it).

It's quite common for applications to expose additional features on newer devices. Newer devices can support HD video whilst older ones can't. Ditto for things like 360 video, VR and 60fps content. Many games wont work at all on older devices. I think most users understand this.

Thanks

yangwuan55:
@all Please see here ,I have a implementation of variable speed for exoplayer,support the preversion of 23.

PR #1773, opened 2016-08-24
PAC and Midrow code handling for EIA-608 captions

RikHeijdens:
This is an effort to implement preamble control codes and mid row style changes for EIA-608 captions.

Currently the implementation does not completely follow the EIA-608 spec, but should provide reasonable closed caption positioning and styling for most cases.

You can test this by Selecting the `Apple master playlist` stream under HLS.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608CueBuilder.java
+++ b/library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608CueBuilder.java
@@ -1,0 +1,203 @@
+package com.google.android.exoplayer2.text.eia608;
+
+import com.google.android.exoplayer2.text.Cue;
+import com.google.android.exoplayer2.util.Assertions;
+
+import android.text.Layout;
+import android.text.SpannableStringBuilder;
+import android.text.Spanned;
+import android.text.style.CharacterStyle;
+import android.text.style.ForegroundColorSpan;
+import android.text.style.StyleSpan;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * A Builder for EIA-608 cues.
+ */
+/* package */ final class Eia608CueBuilder {
+
+  private static final int BASE_ROW = 15;
+
+  /**
+   * The caption string.
+   */
+  private SpannableStringBuilder captionStringBuilder;
+
+  /**
+   * The caption styles to apply to the caption string.
+   */
+  private HashMap<Integer, CharacterStyle> captionStyles;
+
+  /**
+   * The row on which the Cue should be displayed.
+   */
+  private int row;
+
+  /**
+   * The indent of the cue - horizontal positioning.
+   */
+  private int indent;
+
+  /**
+   * The setTabOffset offset for the cue.
+   */
+  private int tabOffset;
+
+  public Eia608CueBuilder() {
+    row = BASE_ROW;
+    indent = 0;
+    tabOffset = 0;
+    captionStringBuilder = new SpannableStringBuilder();
+    captionStyles = new HashMap<>();
+  }
+
+  /**
+   * Sets the row for this cue.
+   * @param row the row to set.
+   */
+  public void setRow(int row) {
+    Assertions.checkArgument(row >= 1 && row <= 15);
+    this.row = row;
+  }
+
+  public int getRow() {
+    return row;
+  }
+
+  /**
+   * Rolls up the Cue one row.
+   * @return true if rolling was possible.
+   */
+  public boolean rollUp() {
+    if (row < 1) {
+      return false;
+    }
+    setRow(row - 1);
+    return true;
+  }
+
+  /**
+   * Sets the indent for this cue.
+   * @param indent an indent value, must be a multiple of 4 within the range [0,28]
+   */
+  public void setIndent(int indent) {
+    Assertions.checkArgument(indent % 4 == 0 && indent <= 28);
+    this.indent = indent;
+  }
+
+  public void tab(int tabs) {
+    tabOffset += tabs;
+  }
+
+  /**
+   * Indents the cue position with amountOfTabs.
+   * @param tabOffset the amount of tabs the cue position should be indented.
+   */
+  public void setTabOffset(int tabOffset) {
+    this.tabOffset = tabOffset;
+  }
+
+  /**
+   * Appends a character to the current Cue.
+   * @param character the character to append.
+   */
+  public void append(char character) {
+    captionStringBuilder.append(character);
+  }
+
+  /**
+   * Removes the last character of the caption string.
+   */
+  public void backspace() {
+    if (captionStringBuilder.length() > 0) {
+      captionStringBuilder.delete(captionStringBuilder.length() - 1, captionStringBuilder.length());
+    }
+  }
+
+  /**
+   * Opens a character style at the current cueIndex.
+   * Takes care of style priorities.
+   *
+   * @param style the style to set.
+   */
+  public void setCharacterStyle(CharacterStyle style) {
+    int startIndex = getSpanStartIndex();
+    // Close all open spans of the same type, and add a new one.
+    if (style instanceof ForegroundColorSpan) {
+      // Setting a foreground color clears the italics style.
+      closeSpan(StyleSpan.class); // Italics is a style span.
+    }
+    closeSpan(style.getClass());
+    captionStyles.put(startIndex, style);
+  }
+
+  /**
+   * Closes all open spans of the spansToApply class.
+   * @param spansToClose the class of which the spans should be closed.
+   */
+  private void closeSpan(Class<? extends CharacterStyle> spansToClose) {
+    for (Integer index : captionStyles.keySet()) {
+      CharacterStyle style = captionStyles.get(index);
+      if (spansToClose.isInstance(style)) {
+        if (index < captionStringBuilder.length()) {
+          captionStringBuilder.setSpan(style, index,
+            captionStringBuilder.length(), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+        }
+        captionStyles.remove(index);
+      }
+    }
+  }
+
+  /**
+   * Applies all currently opened spans to the SpannableStringBuilder.
+   */
+  public void closeSpans() {
+    // Check if we have to do anything.
+    if (captionStyles.size() == 0) {
+      return;
+    }
+
+    for (Integer startIndex : captionStyles.keySet()) {
+      // There may be cases, e.g. when seeking where the startIndex becomes greater
+      // than what is actually in the string builder, in that case, just discard the span.
+      if (startIndex < captionStringBuilder.length()) {
+        CharacterStyle captionStyle = captionStyles.get(startIndex);
+        captionStringBuilder.setSpan(captionStyle, startIndex,
+          captionStringBuilder.length(), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+      }
+      captionStyles.remove(startIndex);
+    }
+  }
+
+  public Cue build() {
+    closeSpans();
+    float cueLine = 10 + (5.33f * row);
+    float cuePosition = 10 + (2.5f * indent);
+    cuePosition = (tabOffset * 2.5f) + cuePosition;
+    return new Cue(captionStringBuilder, Layout.Alignment.ALIGN_NORMAL, cueLine / 100, Cue.LINE_TYPE_FRACTION,
+      Cue.ANCHOR_TYPE_START, cuePosition / 100, Cue.TYPE_UNSET, Cue.DIMEN_UNSET);
+  }
+
+  private int getSpanStartIndex() {
+    return captionStringBuilder.length() > 0 ? captionStringBuilder.length() - 1 : 0;
+  }
+
+  public static List<Cue> buildCues(List<Eia608CueBuilder> builders) {
+    if (builders.isEmpty()) {
+      return Collections.emptyList();
+    }
+    LinkedList<Cue> cues = new LinkedList<>();
+    for (Eia608CueBuilder builder : builders) {
+      if (builder.captionStringBuilder.length() == 0) {
+        continue;
+      }
+      cues.add(builder.build());
+    }
+    return cues;
+  }
+
+}
--- a/library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608CueBuilder.java
+++ b/library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608CueBuilder.java
@@ -1,14 +1,14 @@
 package com.google.android.exoplayer2.text.eia608;
 
-import com.google.android.exoplayer2.text.Cue;
-import com.google.android.exoplayer2.util.Assertions;
-
 import android.text.Layout;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
 import android.text.style.CharacterStyle;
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,32 +501,197 @@ private void handleMiscCode(byte cc2) {
...
+    applySpan(); // Apply any spans.
+
+    // Parse the "next row down" flag.
+    boolean nextRowDown = (cc2 & 0x20) != 0;
+    if (nextRowDown) {
+      // TODO: We should create a new cue instead, this may cause issues when
+      // the new line receives it's own PAC which we ignore currently.
+      // As a result of that the new line will be positioned directly below the
+      // previous line.
+      maybeAppendNewline();
```
RikHeijdens: Ideally, we should create a new Cue here, but I'm not sure how we can display Cues within different Eia608Subtitle objects on the screen at the same time. Maybe we shouldn't try to instantiate new Eia608Subtitle objects in `dequeueOutputBuffer()` all the time, and instead stream the cues through a single instance of Eia608Subtitle?

ojw28:
@cdrolle FYI

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,32 +501,197 @@ private void handleMiscCode(byte cc2) {
...
+          break;
+        case 0x3:
+          // Row 12 or 13 (Bottom)
+          cueLine = CUE_LINE_MAP[11];
+          break;
+      }
+    }
+  }
+
+  private void handleMidrowCode(byte cc1, byte cc2) {
```
RikHeijdens, line 531: I don't have a stream that uses mid row codes to change the foreground/background color to anything but white/black, so I'm not sure if this actually works.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,32 +501,201 @@ private void handleMiscCode(byte cc2) {
...
+    }
+  }
+
+  private void handleTabOffset(byte cc2) {
+    // Formula for tab offset handling adapted from:
+    // https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#x1-preamble-address-code-pac
+    // We're ignoring any tab offsets that do not occur at the beginning of a new cue.
+    // This is not conform the spec, but works in most cases.
+    if (captionStringBuilder.length() == 0) {
+      switch (cc2) {
```
zsmatyas: This entire block can be simplified to a single line: 
tabOffset = cc2 - 0x20;

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -86,6 +99,12 @@
   private static final byte CTRL_CARRIAGE_RETURN = 0x2D;
   private static final byte CTRL_ERASE_NON_DISPLAYED_MEMORY = 0x2E;
 
+  private static final byte CTRL_TAB_OFFSET_CHAN_1 = 0x17;
+  private static final byte CTRL_TAB_OFFSET_CHAN_2 = 0x1F;
+  private static final byte CTRL_TAB_OFFSET_1 = 0x21;
```
zsmatyas: These can be removed.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
zsmatyas: Check the implementation here:
https://github.com/google/ExoPlayer/pull/1312/files#diff-e9fd670327a94df1ab96714b095f4985R397

Beware the order of the priorities.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -159,23 +178,79 @@
     0xC5, 0xE5, 0xD8, 0xF8, 0x250C, 0x2510, 0x2514, 0x2518
   };
 
+  // Maps EIA-608 PAC row numbers to WebVTT cue line settings.
+  // Adapted from: https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#x1-preamble-address-code-pac
+  private static final float[] CUE_LINE_MAP = new float[] {
```
zsmatyas: If you want to use a map and the standard has the row values in specific, but not linear order like: (1, 3, 2).

And than you reorder everything with a switch case, like you map the value 0x1 to the index 4:

case 0x1:  // Row 5 or 6
   cueLine = CUE_LINE_MAP[4];

Then you could put those values into the map in the same order as the standard specifies, and than you can remove the switch and directly index the map:
cueLine[ 0x1 ];

See previous implementation:
https://github.com/google/ExoPlayer/pull/1312/files#diff-babdf8243982d21611d6cb14f7df1bf9R104

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,32 +501,201 @@ private void handleMiscCode(byte cc2) {
...
+          break;
+        case CTRL_TAB_OFFSET_3:
+          tabOffset += 3;
+          break;
+      }
+    }
+  }
+
+  private int getSpanStartIndex() {
+    return captionStringBuilder.length() > 0 ? captionStringBuilder.length() - 1 : 0;
```
zsmatyas: Will this call length() twice or is the compiler clever enough to call it only once?

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,32 +501,201 @@ private void handleMiscCode(byte cc2) {
...
+
+  private void flushCaptionBuilder() {
+    captionStringBuilder.clear();
+    captionStringBuilder.clearSpans();
+  }
+
   private void backspace() {
     if (captionStringBuilder.length() > 0) {
-      captionStringBuilder.setLength(captionStringBuilder.length() - 1);
+      captionStringBuilder.replace(captionStringBuilder.length() - 1, captionStringBuilder.length(), "");
```
zsmatyas: Probably more efficient:
https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#delete(int,%20int)

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -463,19 +719,19 @@ private String getDisplayCaption() {
...
     if (captionMode != CC_MODE_ROLL_UP) {
-      return captionStringBuilder.substring(0, endIndex);
+      return captionStringBuilder.subSequence(0, endIndex);
     }
 
     int startIndex = 0;
     int searchBackwardFromIndex = endIndex;
     for (int i = 0; i < captionRowCount && searchBackwardFromIndex != -1; i++) {
-      searchBackwardFromIndex = captionStringBuilder.lastIndexOf("\n", searchBackwardFromIndex - 1);
+      searchBackwardFromIndex = captionStringBuilder.toString().lastIndexOf("\n", searchBackwardFromIndex - 1);
```
zsmatyas: This entire logic can be removed as soon as every line is implemented with a separate Cue object, but until then:
what is the problem with the lastIndexOf() of StringBuilder? Why do you use the String one?
https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#lastIndexOf(java.lang.String,%20int)

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: I think it should be clever enough, but I didn't check the byte code. Might be worth a little investigation.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: According to the [documentation](https://developer.android.com/reference/android/text/SpannableStringBuilder.html#delete%28int, int%29) delete is just a convenience method, but I agree, it would probably fit better here.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: [SpannableStringBuilder](https://developer.android.com/reference/android/text/SpannableStringBuilder.html#delete%28int, int%29) does not offer the lastIndexOf method.

And yes, most likely it can be removed after we figured out: https://github.com/google/ExoPlayer/pull/1773#discussion_r76045863

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: Thanks - I didn't know about your implementation.
I'll address all the comments tomorrow.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: This should take care of the order of priorities: https://github.com/google/ExoPlayer/pull/1773/commits/7cd94819c7a4a0222022cbfd49a1d77f30aa5812#diff-7fa75103163e987aeb1657e153a4604fR596

Do you by any chance have a stream that uses midrow codes to change styles? Apple's HLS sample doesn't seem to do it.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -159,23 +175,72 @@
...
+    Color.YELLOW,
+    Color.MAGENTA,
+    Color.BLACK // Only used by Mid Row style changes, for PAC an value of 0x7 means italics.
+  };
+
+  // Transparency is defined in the first byte of an integer.
+  private static final int TRANSPARENCY_MASK = 0x80FFFFFF;
+
+  private static final int STYLE_ITALIC = Typeface.ITALIC;
+  private static final float DEFAULT_CUE_LINE = CUE_LINE_MAP[4]; // Row 11
```
zsmatyas: the 4th element does not seem to be line 11. Is the comment matching the code?

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
zsmatyas: You are right, sorry. I incorrectly assumed SpannableStringBuilder being a subclass of StringBuilder

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
RikHeijdens: Thanks, good spot!

It will be fixed in the next commit, I'm currently working on displaying every line as a separate cue.

RikHeijdens:
Since https://github.com/google/ExoPlayer/pull/1773/commits/71d83f3e8478f05271845eb36a0a3f1bfec98808 every caption line is now a separate cue and being streamed through a single instance of `Eia608Subtitle`, at this point of time that causes a couple of issues:
- Cue background boxes overlap
- Paint-on/Roll-up captions are not being drawn correctly

Furthermore, I'm not happy about horizontal positioning. The EIA-608 standard assumes a maximum of 32 characters per line, because of that horizontal positioning is off in general. I'm not sure what would be the best approach to handle it. I checked AV Player's behavior and they seem to align cues with less than 1 empty row between them. Should we mimic that behavior?

zsmatyas:
Good job, you are almost there. :) If you can wait a few days, I might be able to push my V1 changes ported to V2, I just lowered priority working on these as I did not see too much chance that my pull request gets accepted.

To address the issues:
- Overlap is caused because with the correct positioning and default sizes there must be an overlap. (You cannot fit 15 thicker lines into the same area that was using narrower lines). I solved this by slicing the rendering into 2 phases: backgrounds first then the texts. It is just a few lines of code.
-  Paint on cannot work nicely with fonts that are not unified width. Replacing any character with one of different with will re-align the entire line, the text continuously moves, you cannot read it. 
- For Roll up cations the line count, new line and the "show the text" commands must be implemented according to the spec. Remove hacks like maybeAppendNewLine(). These were probably included in my commits for the V1 implementation. The output of my version (colors/positioning/timing) matched the other players for android that handles 608.

Bad news: 708 will complicate things much more :)

RikHeijdens:
@zsmatyas 

Roll-up/Paint-on captions should follow the spec pretty close now, but for some reason the SubtitlePainter does not paint the entire CueText. For example with the following [stream](http://playertest.longtailvideo.com/adaptive/cpcweb/cpcdemo.m3u8) it only paints the first two characters of the paint-on cue, when the SubtitlePainters receive the entire Cue. I'm not sure why that is happening, as it should recalculate the layout.

RikHeijdens:
OK, disregard my previous comment. That should be fixed now :)

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608CueBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+  public int getRow() {
+    return row;
+  }
+
+  /**
+   * Rolls up the Cue one row.
+   * @return true if rolling was possible.
+   */
+  public boolean rollUp() {
+    if (row < 1) {
```
yichengliu: It seems the condition here should be `if (row <= 1)`. Otherwise when `row` equals `1`, `setRow(0)` will be called and there will be an exception.

library/src/main/java/com/google/android/exoplayer2/text/eia608/Eia608Decoder.java
```diff
@@ -414,68 +453,102 @@ private void handleMiscCode(byte cc2) {
...
       case CTRL_CARRIAGE_RETURN:
-        maybeAppendNewline();
+        // Each time a Carriage Return is received, the text in the top row of the window is erased
+        // from memory and from the display. The remaining rows of text are each rolled up into the
+        // next highest row in the window, leaving the base row blank and ready to accept new text.
+        if (captionMode == CC_MODE_ROLL_UP) {
+          for (Eia608CueBuilder cue : cues) {
+            // Roll up all the other rows.
+            if (!cue.rollUp()) {
+              cues.remove(cue);
```
yichengliu: This code can cause `ConcurrentModificationException` when `rollUp()` returns false.

RikHeijdens: Thanks @YichengLiu I've addressed both this issue and your other comment.

RikHeijdens:
@ojw28 @cdrolle I noticed that recently some changes where made to the text.eia608/cea packages - is this something that eventually will be merged or are you working on your own implementation?

ojw28:
I'm aware there are some issues with the merged result of this. We'll get those fixed up shortly!

ojw28:
Thanks for the contribution btw!

PR #1776, opened 2016-08-25
[FIXED] Switch width & height of vertical video when checking the capability of video

uni2nd:
When playing a vertical video that has a RESOLUTION as 720x1280,
width & height of that video need to be reversed when calling MediaCodecInfo.VideoCapabilities.areSizeAndRateSupported() & MediaCodecInfo.VideoCapabilities.isSizeSupported().

You can see the problem at the below example.

The playlist of input video. (HLS)
# EXTM3U
# EXT-X-VERSION:3
# EXT-X-STREAM-INF:BANDWIDTH=1813682,CODECS="avc1.66.31,mp4a.40.2",RESOLUTION=720x1280

blahblahblah.m3u8

Especially, a **Sony Experia M4** can decode up to 1920 x 1088 according to VideoCapability's info,
but supportFormat() method returns false if the video size is 720 x 1280.

Thanks.

ps. Actually ExoPlayer v1.x also has this problem. So if you think that this is acceptable, then you need to look at VideoFormatSelectorUtil.isFormatPlayable() in  ExoPlayer v1.x.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
@@ -185,18 +185,28 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
       return FORMAT_UNSUPPORTED_SUBTYPE;
     }
 
+    // If vertical video, width & height is reversed.
+    // So we need to switch them when checking capability to decode.
+    // Please see MediaCodecInfo.VideoCapabilities.isSizeSupported()
+    int width = format.width, height = format.height;
+    if (width < height) {
+      int tempWidth = width;
+      width = height;
+      height = tempWidth;
+    }
+
     boolean decoderCapable;
-    if (format.width > 0 && format.height > 0) {
+    if (width > 0 && height > 0) {
       if (Util.SDK_INT >= 21) {
         if (format.frameRate > 0) {
-          decoderCapable = decoderInfo.isVideoSizeAndRateSupportedV21(format.width, format.height,
-              format.frameRate);
+          decoderCapable = decoderInfo.isVideoSizeAndRateSupportedV21(width, height,
+                  format.frameRate);
         } else {
-          decoderCapable = decoderInfo.isVideoSizeSupportedV21(format.width, format.height);
+          decoderCapable = decoderInfo.isVideoSizeSupportedV21(width, height);
         }
         decoderCapable &= decoderInfo.isCodecSupported(format.codecs);
       } else {
-        decoderCapable = format.width * format.height <= MediaCodecUtil.maxH264DecodableFrameSize();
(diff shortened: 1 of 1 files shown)


uni2nd:
I signed it!

uni2nd:
I signed it!

ojw28:
Interesting. To check we're thinking along the same lines, this is effectively a platform bug, isn't it? VideoCapabilities shouldn't claim that the device doesn't support decoding 720x1280 video if it can successfully decode this resolution. It sounds like the device is under-reporting its capabilities for the vertical video case?

I'm not saying we shouldn't work around the issue; just trying to check we have consensus that the platform is doing the wrong thing? I'll follow up internally if so and make sure this doesn't happen going forward.

ojw28:
By the way, do you see this issue on any devices other than the Sony M4?

uni2nd:
@ojw28 OK, let me check more devices and let you know.

BTW about the consensus,
I think I can say that this might be a platform bug as you said,
but it might also be an ExoPlayer's bug.
It's the problem that there is no standard for vertical video.
As you know, we say 720P as when width is 720 px, not height commonly.
But it's not always correct when we say vertical video.
I mean that we need more robust rules can be applied to vertical video if we say whether this is a platform's or exo's issue.
Converting width & height is also a kind of this problem.
I'm thinking It's... it's a little complicated... :(

Please let me know if you all arrive at a conclusion about this issue.
Thanks.

uni2nd:
I signed it!

uni2nd:
I signed it!

uni2nd:
I signed it!

ojw28:
Did you have a chance to test on any more devices? I'm pretty sure it is a platform issue; the fact that video is normally horizontal doesn't mean the API can ignore input for which this isn't the case.

There's a danger that flipping the dimensions around might cause other issues (e.g. if there exists a device can legitimately play horizontal video but not vertical video with the dimensions flipped). So I think we do need results from other devices before deciding what to do here. If the issue is isolated to the M4 then we should add a workaround just for that device.

uni2nd:
Sorry for late answer.

I'm sorry but I couldn't have a time to find out more details,
but I got a lot of bug report that some of our users couldn't see our vertical video (720x1280).
And they all used the previous version that this fix is not applied.

The bug report says that the models are
SHV-E330S, MyPhone_MY28S, MyPhoneMY27, GT-I9500, D450, Flare_J1, A74A, AS8, and so on...
In principle, I totally agree with you with platform bug, but in practice, partially no.

Anyway, if you decide to do like that, please let Android team know about this issue.
Thanks for your help!! :)

ojw28:
Hi. We decided to merge an equivalent change, having found some more occurrences of this issue. The change is https://github.com/google/ExoPlayer/commit/2e3ffe1e94a16d6ca1e5c09f818969cbcd531abe and has been merged into `dev-v2`. Closing this request. Thanks for your input!

uni2nd:
I'm really good to hear that.
Thanks for your comment!!

PR #1778, opened 2016-08-25
Add method to ExoPlayer.Listener for notifying when seek completes

Lien:
Track when seek is pending. When work is done and renderers are ready, send
MSG_SEEK_COMPLETE if seek is pending to notify listeners when seek completes.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer/ExoPlayerImplInternal.java
+++ b/library/src/main/java/com/google/android/exoplayer/ExoPlayerImplInternal.java
@@ -45,6 +45,7 @@ import java.util.concurrent.atomic.AtomicInteger;
   public static final int MSG_STATE_CHANGED = 2;
   public static final int MSG_SET_PLAY_WHEN_READY_ACK = 3;
   public static final int MSG_ERROR = 4;
+  public static final int MSG_SEEK_COMPLETE = 1000;
 
   // Internal messages
   private static final int MSG_PREPARE = 1;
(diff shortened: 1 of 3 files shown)


Lien:
This change is used within the Periscope for Android app so that we are notified when seek completes. This allows us to properly sync Periscope chats and hearts with the new video position.

library/src/main/java/com/google/android/exoplayer/ExoPlayerImplInternal.java
```diff
@@ -500,6 +507,8 @@ private void seekToInternal(long positionMs) throws ExoPlaybackException {
     try {
       if (positionMs == (positionUs / 1000)) {
         // Seek is to the current position. Do nothing.
+        pendingSeek = false;
+        eventHandler.sendEmptyMessage(MSG_SEEK_COMPLETE);
```
ojw28, line 511: Note - The seek to the "current position" might not have completed yet, in which case I think this ends up firing the event too soon. For example if you do a seek to t=5000 twice in quick succession:
- The first call will update positionUs at L516 (just below) and pendingSeek will be set to true.
- The second call will arrive almost immediately whilst the player is still buffering, set pendingSeek back to false again and fire the event.

It feels like it would be more correct to remove the two lines being added here, or turn pendingSeek into an integer counter so that you can fire the correct number of MSG_SEEK_COMPLETE events once the seek really has completed.

ojw28:
Thanks for the pull request. We're no longer accepting feature contributions into V1 because nearly all of our work has shifted to V2. Please feel free to propose a PR to the `dev-v2` branch if and when you need to do so.

PR #1797, opened 2016-09-03
Support ID3/Apple metadata parsing in MP3 and MP4 files

cbfiddle:
Supports parsing of ID3 in MP3 files and Apple metadata in MP4 files.
Gapless info is extracted from the parsed metadata.
A new class Metadata is added that includes GaplessInfo.
GaplessInfoHolder used only in the one place where it is needed.
Metadata is added to Format.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/Format.java
+++ b/library/src/main/java/com/google/android/exoplayer2/Format.java
@@ -21,6 +21,8 @@ import android.media.MediaFormat;
 import android.os.Parcel;
 import android.os.Parcelable;
 import com.google.android.exoplayer2.drm.DrmInitData;
+import com.google.android.exoplayer2.extractor.GaplessInfo;
+import com.google.android.exoplayer2.metadata.Metadata;
 import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.Util;
 import java.nio.ByteBuffer;
@@ -100,6 +102,11 @@ public final class Format implements Parcelable {
    * DRM initialization data if the stream is protected, or null otherwise.
    */
   public final DrmInitData drmInitData;
+  /**
+   * Static metadata
+   */
+  public final Metadata metadata;
+
 
   // Video specific.
 
@@ -196,7 +203,7 @@ public final class Format implements Parcelable {
       float frameRate, List<byte[]> initializationData) {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, width,
         height, frameRate, NO_VALUE, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, NO_VALUE, 0, null, OFFSET_SAMPLE_RELATIVE, initializationData, null);
+        NO_VALUE, NO_VALUE, 0, null, OFFSET_SAMPLE_RELATIVE, initializationData, null, null);
   }
 
   public static Format createVideoSampleFormat(String id, String sampleMimeType, String codecs,
@@ -222,7 +229,7 @@ public final class Format implements Parcelable {
     return new Format(id, null, sampleMimeType, codecs, bitrate, maxInputSize, width, height,
         frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, 0, null, OFFSET_SAMPLE_RELATIVE, initializationData,
-        drmInitData);
+        drmInitData, null);
   }
 
   // Audio.
@@ -233,7 +240,7 @@ public final class Format implements Parcelable {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, channelCount, sampleRate, NO_VALUE,
         NO_VALUE, NO_VALUE, selectionFlags, language, OFFSET_SAMPLE_RELATIVE, initializationData,
-        null);
+        null, null);
   }
 
   public static Format createAudioSampleFormat(String id, String sampleMimeType, String codecs,
@@ -260,7 +267,7 @@ public final class Format implements Parcelable {
     return new Format(id, null, sampleMimeType, codecs, bitrate, maxInputSize, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, channelCount, sampleRate, pcmEncoding,
         encoderDelay, encoderPadding, selectionFlags, language, OFFSET_SAMPLE_RELATIVE,
-        initializationData, drmInitData);
+        initializationData, drmInitData, null);
   }
 
   // Text.
@@ -269,7 +276,7 @@ public final class Format implements Parcelable {
       String sampleMimeType, String codecs, int bitrate, int selectionFlags, String language) {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, NO_VALUE, selectionFlags, language, OFFSET_SAMPLE_RELATIVE, null, null);
+        NO_VALUE, NO_VALUE, selectionFlags, language, OFFSET_SAMPLE_RELATIVE, null, null, null);
   }
 
   public static Format createTextSampleFormat(String id, String sampleMimeType, String codecs,
@@ -283,7 +290,7 @@ public final class Format implements Parcelable {
       long subsampleOffsetUs) {
     return new Format(id, null, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, selectionFlags, language, subsampleOffsetUs, null, drmInitData);
+        NO_VALUE, selectionFlags, language, subsampleOffsetUs, null, drmInitData, null);
   }
 
   // Image.
@@ -292,7 +299,7 @@ public final class Format implements Parcelable {
       int bitrate, List<byte[]> initializationData, String language, DrmInitData drmInitData) {
     return new Format(id, null, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, 0, language, OFFSET_SAMPLE_RELATIVE, initializationData, drmInitData);
+        NO_VALUE, 0, language, OFFSET_SAMPLE_RELATIVE, initializationData, drmInitData, null);
   }
 
   // Generic.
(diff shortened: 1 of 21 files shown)


ojw28:
Sorry for the delay in looking at this; it looks useful :)! Will take a look shortly.

library/src/main/java/com/google/android/exoplayer2/metadata/id3/ApicFrame.java
```diff
@@ -35,4 +39,62 @@ public ApicFrame(String mimeType, String description, int pictureType, byte[] pi
...
+    result = 31 * result + (mimeType != null ? mimeType.hashCode() : 0);
+    result = 31 * result + (description != null ? description.hashCode() : 0);
+    result = 31 * result + pictureType;
+    result = 31 * result + Arrays.hashCode(pictureData);
+    return result;
+  }
+
+  @Override
+  public void writeToParcel(Parcel dest, int flags) {
+    dest.writeString(id);
```
ojw28, line 78: It's quite confusing to work out what's going on here, given the Id3Frame constructor is reading the string whilst the ApicClass writeToParcel method is writing it. The id is also constant (here and in most, but not all, of the other concrete Id3Frame implementations), meaning it needn't be written to the parcel or read from it.

You should probably remove the `Id3Frame(Parcel in)` constructor. Where id isn't constant, have the implementation class read the id rather than Id3Frame.

Ditto for all other classes that extend Id3Frame.

library/src/main/java/com/google/android/exoplayer2/metadata/id3/ApicFrame.java
```diff
@@ -35,4 +39,62 @@ public ApicFrame(String mimeType, String description, int pictureType, byte[] pi
...
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
+    ApicFrame that = (ApicFrame) o;
+
+    if (id != null ? !id.equals(that.id) : that.id != null) return false;
```
ojw28, line 57: id is constant, so there's no value in including it in the equals implementation. Ditto for hashCode. Same comment applies to all other Id3Frame subclasses except those without constant id (e.g. TextInformationFrame).

library/src/main/java/com/google/android/exoplayer2/metadata/id3/CommentFrame.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Comment ID3 frame.
+ */
+public final class CommentFrame extends Id3Frame {
+
+  public final String language;
+  public final String text;
+
+  public CommentFrame(String language, String description, String text) {
+    super(description);
```
ojw28, line 30: This doesn't look right. Shouldn't the id COMM (or COM0 depending on another comment)?

library/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -16,76 +16,150 @@
...
+            frame = decodeGeobFrame(frameData, frameSize);
+          } else if (frameId0 == 'A' && frameId1 == 'P' && frameId2 == 'I' && frameId3 == 'C') {
+            frame = decodeApicFrame(frameData, frameSize);
+          } else if (frameId0 == 'T') {
+            String id = frameId3 != 0 ?
+                String.format(Locale.US, "%c%c%c%c", frameId0, frameId1, frameId2, frameId3) :
+                String.format(Locale.US, "%c%c%c", frameId0, frameId1, frameId2);
+            frame = decodeTextInformationFrame(frameData, frameSize, id);
+          } else if (frameId0 == 'C' && frameId1 == 'O' && frameId2 == 'M' &&
+              (frameId3 == 'M' || frameId3 == 0)) {
```
ojw28, line 143: Where is COM0 defined? I can't find it in the id3 spec.

ojw28 (commented):
Some preliminary comments. It would probably be useful to split this into smaller changes. The first one being one that makes all the Id3Frame implementations Parcelable, adds equals/hashCode, adds CommentFrame and parses them in Id3Decoder only?

library/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
```diff
@@ -136,22 +136,13 @@ public Uri getUri() {
   @Override
   public void close() throws ContentDataSourceException {
```
ojw28, line 137: Changes in this file appear to be irrelevant to this CL? Probably a merge error? Please revert if so?

library/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -409,21 +418,32 @@ public Format copyWithManifestFormatInfo(Format manifestFormat,
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, maxInputSize, width,
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding, selectionFlags,
-        language, subsampleOffsetUs, initializationData, drmInitData);
+        language, subsampleOffsetUs, initializationData, drmInitData, null);
```
ojw28, line 421: Why is this null? Should be metadata presumably?

cbfiddle:
It is not COM0, it is COM with a null byte. All version 2 tags http://id3.org/id3v2-00 have three characters.

I will look at the other issues later.

ojw28:
Ah. There are other version 2 tags that we'd need to handle as well (TXX, COM, GEO, PIC). Can we keep handling of version 2 tags for a separate change; this one has enough stuff in it already (arguably it should already be split into smaller changes).

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Id3Util.java
```diff
@@ -83,211 +90,6 @@ private static boolean canParseMetadata(int majorVersion, int minorVersion, int
         && !(majorVersion == 4 && (flags & 0x0F) != 0);
```
ojw28, line 90: If we're supporting other versions, presumably the check for major version here needs removing. Although as per comment elsewhere, we should probably do that separately.

ojw28: Ignore this. Comment doesn't make sense.

ojw28:
I've created a new branch for this work here: https://github.com/google/ExoPlayer/tree/dev-v2-id3

If you could send the next pull request to that branch, that would be great! If you don't have time to address the comments here any time soon, please just send the original change as a pull request to that branch as it was, and we can help with the comments as part of a more iterative process.

cbfiddle:
> On Oct 6, 2016, at 11:20 AM, ojw28 notifications@github.com wrote:
> 
> I've created a new branch for this work here: https://github.com/google/ExoPlayer/tree/dev-v2-id3 https://github.com/google/ExoPlayer/tree/dev-v2-id3
> If you could send the next pull request to that branch, that would be great! If you don't have time to address the comments here any time soon, please just send the original change as a pull request to that branch as it was, and we can help with the comments as part of a more iterative process.

I think I have just done that (sending the original change).

Regarding the changes to ContentDataSource, that indeed is a separate issue. I filed the original bug report, and I included a small improvement to the fix.

Regarding your comments about the id instance variable, I think your suggestions can be taken one step further. As the CommentFrame has a fixed id, it does not need an id instance variable at all. The Id3Frame class could define an abstract getter method instead of the instance variable. The id instance variable would be defined in the classes that need it, along with the Parcelable related code.

I agree with your comment about line 409 in Format.java.

  Alan

ojw28:
Merged into `dev-v2-id3`. Work should continue there, and will be merged to `dev-v2` when ready. Thanks!

library/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -262,6 +376,37 @@ private static BinaryFrame decodeBinaryFrame(ParsableByteArray id3Data, int fram
...
+    byte[] source = dataSource.data;
+    int sourceIndex = dataSource.getPosition();
+    int limit = sourceIndex + count;
+    byte[] dest = new byte[count];
+    int destIndex = 0;
+
+    while (sourceIndex < limit) {
+      byte b = source[sourceIndex++];
+      if ((b & 0xFF) == 0xFF) {
+        int nextIndex = sourceIndex+1;
```
ojw28, line 394: This looks wrong to me. sourceIndex is incremented on L392 already, so nextIndex ends up being two greater than the index at which 0xFF is found. It should be one greater, I think.

cbfiddle:
I think you are correct.

ojw28:
Thanks for the confirmation. I've pushed some improvements in https://github.com/google/ExoPlayer/commit/110c8f6f1f4459b05e9943fd6f5c8521735bbfbe.

ojw28:
@cbfiddle - Hi. We're approaching being able to merge this into `dev-v2` (we're also making `SimpleExoPlayerView` automatically show the album art, if present :)).

One question: For AtomParsers, where did you find the list of possible types? I'm really struggling to find anything authoritative for some of them. For example I seemingly can't find anything about "�trk". Are you sure it exists? Ditto for quite a few of the others.

ojw28:
http://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/QuickTime.html seems relatively authoritative. No "�trk" though. Only the variant starting with ©.

ojw28:
Ah, I think you're just handling the possibility of the © symbol getting mangled. Is that likely in practice?

cbfiddle:
I’m not finding @trk. Where is it?

ojw28:
I edited my posts (because they were wrong - heh). Sorry for the confusion. Read the web version here: https://github.com/google/ExoPlayer/pull/1797

I think my only remaining question is how likely the copyright symbol is to be mangled. Hopefully we can get this merged back into `dev-v2` in the next few days :).

cbfiddle:
Here’s the story on the non-standard tag variants. I based the submitted code on code I wrote many years ago, and I no longer remember what my sources were for the tags. My guess is that back in the day, some popular but now obsolete program wrote files with these alternative tags. I figured there might be files out there that use those tags, and no harm is done if we recognize them. It is an example of the Internet robustness principle https://en.wikipedia.org/wiki/Robustness_principle. FWIW, in my testing I found files in my library that were encoded by iTunes version 1.

ojw28:
@cbfiddle - Thanks for all the work you did on this! We've just merged the changes back into `dev-v2`. You should find album art is automatically displayed in the demo app as well, including for MP4. I'm going to post a couple of comments on this closed request, just in case you're keeping your own fork and/or are interested in a few things I discovered :).

library/src/main/java/com/google/android/exoplayer2/extractor/mp4/AtomParsers.java
```diff
@@ -423,47 +430,333 @@ private static void parseMetaAtom(ParsableByteArray data, GaplessInfoHolder out)
...
+          byte[] bb = (byte[]) value;
+          Id3Frame frame = new BinaryFrame(extendedName, bb);
+          builder.add(frame);
         }
+      }
+    }
+  }
+
+  private static Object parseDataBox(ParsableByteArray ilst, int length) {
+    int versionAndFlags = ilst.readInt();
```
ojw28, line 735: I think this ends up reading the wrong 4 bytes. At the calling sites it appears you skip over the versionAndFlags int. This the reads the following 4 bytes, which I think are defined to be empty bytes in a data box.

library/src/main/java/com/google/android/exoplayer2/extractor/mp4/AtomParsers.java
```diff
@@ -423,47 +430,333 @@ private static void parseMetaAtom(ParsableByteArray data, GaplessInfoHolder out)
...
+                                                  ParsableByteArray ilst,
+                                                  int endPosition) {
+    int length = ilst.readInt() - Atom.FULL_HEADER_SIZE;
+    int key = ilst.readInt();
+    ilst.skipBytes(4);
+    if (key == Atom.TYPE_data) {
+      Object value = parseDataBox(ilst, length);
+      if (value instanceof byte[]) {
+        byte[] bytes = (byte[]) value;
+        if (bytes.length == 8) {
```
ojw28, line 642: I have a piece of sample media where the length is only 6. I couldn't find an authoritative source to say whether this is valid, but since the final two bytes are ignored regardless, it seemed fine to handle this case.

PR #1803, opened 2016-09-07
TsExtractor support for language code in the audio tracks.

tresvecesseis:
We add support to display the language code of the audio tracks in MP2T streams allowing the player applications to display meaningful track names in the audio track selectors.


ojw28:
@AquilesCanta PTAL, thanks.

PR #1829, opened 2016-09-16
Handle 0 width and height in AspectRatioFrameLayout

amosyuen:
When in vertical or horizontal list views with wrap_content as a parameter, the height or width will be measured as 0. In this case the dimension should be calculated off of the non zero width or height.


amosyuen:
I signed it!

PR #1847, opened 2016-09-22
extract NAME in HLS STREAM-INF

b95505017:
In v1:
https://github.com/google/ExoPlayer/commit/10f3be6d509428c4bc9f4a1610dda0271dbdcdd8

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
+++ b/library/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
@@ -188,8 +188,11 @@ public final class HlsPlaylistParser implements ParsingLoadable.Parser<HlsPlayli
           width = Format.NO_VALUE;
           height = Format.NO_VALUE;
         }
+        String name = parseOptionalStringAttr(line, REGEX_NAME);
+        if (name == null) {
+          name = Integer.toString(variants.size());
+        }
         line = iterator.next();
-        String name = Integer.toString(variants.size());
         Format format = Format.createVideoContainerFormat(name, MimeTypes.APPLICATION_M3U8, null,
             codecs, bitrate, width, height, Format.NO_VALUE, null);
         variants.add(new HlsMasterPlaylist.HlsUrl(name, line, format, null, null, null));


b95505017:
@ojw28 Any idea to merge this?

AquilesCanta:
Hello @b95505017, thanks for your contribution. 

The spec does not define the name attribute for the EXT-X-STREAM-INF. Why do you want it in? It is in v1, but I removed it from v2 as part of a refactoring process for the mentioned reason.

b95505017:
There are some providers use these attributes as track info for UI stuff.
For example:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1500000,RESOLUTION=1280x720,NAME="720p"
720p/720p.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=512000,RESOLUTION=640x360,NAME="360p"
360p/360p.m3u8
```

ojw28:
Are they important providers or just providers who are doing it wrong (or both)? Can you provide examples so we can gauge importance. We'd rather not support something that's not in the spec without good reason. Also, do Apple support this, given it's not part of the spec?

b95505017:
Or is there any way to extend the playlist parser functionality to get these custom attributes without rewriting a new one? It is useful for client's quality selector.

ojw28:
It's not useful if it's only one provider who's doing it wrong, as per my response above. Please provide concrete examples of providers who are violating the HLS spec in this way, and we'll assess whether it's something we want to support or not. Else we wont be supporting this. Thanks!

b95505017:
For example this PR https://github.com/grafov/m3u8/pull/30
Wowza will provide this attribute let JWPlayer showing viewer readable text on the quality selector, e.g. 360p, 720p.
Just like this http://www.gsview.tv/channel/moea/Live/1835396.

RikHeijdens:
@ojw28 Could you clarify why exposing the String value for `NAME` would be wrong? I agree, it's debatable whether the `id` field in [Format](https://github.com/google/ExoPlayer/blob/release-v2/library/src/main/java/com/google/android/exoplayer2/Format.java#L51) should be used but the [HLS spec](https://tools.ietf.org/html/draft-pantos-http-live-streaming-20#page-24) is pretty clear to me:

> ```
>  NAME 
>  The value is a quoted-string containing a human-readable
>  description of the Rendition.  If the LANGUAGE attribute is
>  present then this description SHOULD be in that language.  This
>  attribute is REQUIRED.
> ```

ojw28:
We already parse NAME from EXT-X-MEDIA tags, as per the part of the spec to which you refer. This pull request parses NAME from EXT-X-STREAM-INF tags. The HLS spec does not define the NAME attribute on EXT-X-STREAM-INF tags.

RikHeijdens:
Ah yes, you're correct - I mixed up EXT-X-MEDIA and EXT-X-STREAM-INF.

PR #1902, opened 2016-10-06
Using okhttp specific version

b95505017:
To avoid unpredictable behavior in the future.

### DIFF
--- a/extensions/okhttp/build.gradle
+++ b/extensions/okhttp/build.gradle
@@ -37,7 +37,7 @@ android {
 
 dependencies {
     compile project(':library')
-    compile('com.squareup.okhttp3:okhttp:+') {
+    compile('com.squareup.okhttp3:okhttp:3.4.1') {
         exclude group: 'org.json'
     }
 }


Ood-Tsen:
Hi @b95505017, 
Why should we choose okhttp3 version 3.4.1 ?

b95505017:
I just point to the latest version. Using `+` may have unpredictable behavior. e.g. OkHttp API change.

ojw28:
Yeah. Note this is only the demo app. It doesn't stop other applications from using some other version of okhttp.

ojw28:
Oh wait, it's not the demo app. I don't think it's correct to make this restriction. If we did so and if some other library required some other version of okhttp, wouldn't it become impossible to resolve which version to use in any case where a developer wanted to use both?

I think the specific version of a dependency should be specified in the gradle file corresponding to whatever the end product is (typically an app). Each library should define its dependency versions as loosely as possible, to give the developer of the end product flexibility in version choice and to minimize the chances that it'll be impossible to resolve a suitable version.

b95505017:
For example, if okhttp 3.5 change some API we used in this extension, we can't control it and build will fail.
And also I think that it will cause other library which use okhttp in the same app compiled with latest version of okhttp. If there are some incompatible behavior in the future version,  it will cause other library build fail too.

ojw28:
As per above, I'm think the version should be specified in the gradle file corresponding to whatever the end product is. Gradle should be clever enough to pick the version specified there when resolving dependencies.

If libraries are overly specific in which version they require then dependency resolution will fail instead, which isn't what you'd want to happen. For example if one library specifies a dependency on `3.4.1` and some other library specifies a dependency on `3.4.0`, there is no way that Gradle can resolve a suitable version. Conversely, if each library specifies `+` Gradle can resolve a suitable version. Furthermore, if each library specifies `+` and the application specifies `3.4.1`, Gradle can still resolve a suitable version and in addition the version is suitably constrained so as not to cause unpredictable behavior.

ojw28:
Hmm. May well have been wrong about this. Discussing / figuring out exactly what best practice is :)...

PR #1944, opened 2016-10-13
bump gradle and support lib

ralphpina:


ojw28:
We are unable to bump the gradle version at this time due to a bug in the recent version that prevents our tests from running successfully. See in the [file history](https://github.com/everalbum/ExoPlayer/commits/ce29611eee453223d23c33e88a721c46a4d2ecc5/gradle/wrapper/gradle-wrapper.properties) that we've already tried. We're trying to follow up with the Android plugin for gradle people to find out if and when they're going to fix this.

ralphpina:
:+1: cool. I didn't think to look at the history. Will close.

PR #1968, opened 2016-10-20
Rename RELEASENOTES.md to CHANGELOG.md

danielgomezrico:
Just to make it more easy to find


ojw28:
This will break any links to the existing location. I think there needs to be a better reason if it's to be moved.

danielgomezrico:
@ojw28 yeah, you are right but there's just one link to that file and it's already fixed in `d9e30ec`

ojw28:
The file might be linked from anywhere. The most obvious other example I'm aware of (which granted, we could fix) is: https://bintray.com/google/exoplayer/exoplayer#release

In general, it's not possible for us to find and fix such links. Does naming the file CHANGELOG.md follow a standard naming convention that's widely accepted and documented somewhere? If so, can you point us to it. We might consider making the change in that case. We'd likely want to leave something in RELEASENOTES.md (e.g. a link to the new location, so it's not just a 404). If not, this seems little more than personal preference to me, in which case we're best off leaving it where it is.

danielgomezrico:
@ojw28 from http://keepachangelog.com/en/0.3.0/ they mention:

```
What should the change log file be named?
Well, if you can’t tell from the example above, CHANGELOG.md is the best convention so far.

Some projects also use HISTORY.txt, HISTORY.md, History.md, NEWS.txt, NEWS.md, News.txt, RELEASES.txt, RELEASE.md, releases.md, etc.

It’s a mess. All these names only makes it harder for people to find it.
```

If will break things its ok for me to ignore this PR.

PR #1979, opened 2016-10-21
Null pointer exception fixed

vitekn:
pesPayloadReader can be null here because DefaultStreamReaderFactory can return null on unknown streamId. If we have a junk transport stream in our content an exception will be thrown.


vitekn:
I've signed cla, but the label is still 'no'.

PR #1982, opened 2016-10-21
Added demo for use of ExoPlayer in RecyclerView

ralphpina:
I recently implemented ExoPlayer inside a RecyclerView and I could not find any examples, and very little (none?) documentation on how to do this. I have taken my implementation and added it to the Demo application.


eneim:
@ralphpina Interesting, I'm working on [a library](https://github.com/eneim/Toro) which has a pretty close concept with you. Can you give it a comment? Your way of implementing the VideoPlayer has both pros and cons (I would not say mine is better/worse :D, it is actually not a good way and I will change it in the future), I would like to learn from it a bit.

ralphpina:
@eneim I took a look at the library. It clearly provides a much higher level of abstraction. But I think the main implementation details are the same. It seems ExoPlayer is part of the upcoming 2.0 release?

eneim:
@ralphpina I agree about the abstraction, that was my intent. I have just finish almost 90% of my 2.1 specification (not 2.0 >.<). Yeah I will focus on ExoPlayer 2, and base on it to form my library direction :D.

ojw28:
`dev-v2-id3` was a feature branch for ID3 development, which has now been merged and is being deleted. This PR needs to be made to `dev-v2`. Closing the PR because this is required for the feature branch to be deleted.

PR #2046, opened 2016-11-07
Enable the data binding library

HSY229:
Enable the data binding library for demo module.

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -37,6 +37,10 @@ android {
         abortOnError false
     }
 
+    dataBinding {
+        enabled = true
+    }
+
     productFlavors {
         noExtensions
         withExtensions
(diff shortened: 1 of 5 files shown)


ojw28:
....Why?

HSY229:
@ojw28 It helps to noticeably minimize the code for app logic binding it with its view.

ojw28:
The overall line delta of this change appears to be positive (+432, −409) and I also think it's missing a whole file (PlayerActivityBinding), which will make it more positive when added. So it's unclear how this is minimizing anything. Have I missed something? I think I'd also argue that it makes the demo app harder to understand, since someone trying to do this will now also have to look up how an (optional) data binding library works...

HSY229:
We can write the condition to layout to handle the UI state. 
before:

``` java
private void showControls() {
    debugRootView.setVisibility(View.VISIBLE);
}
```

after:

``` java
<LinearLayout android:id="@+id/controls_root"
        ......
        android:visibility="@{viewModel.playbackState == ExoPlayer.STATE_ENDED || viewModel.playbackState == ExoPlayer.STATE_ERROR ? View.VISIBLE : View.GONE}">
```

Data binding lead us to use MVVM. I have to modify my word. It minimize the _glue_ code. 
But I think you are right, it's extra work with learn the data binding library.

ojw28:
At best this is personal preference. At worst it's more verbose than it was previously (the overall line delta is already positive, and will be significantly positive once the missing class is added), is less performant, and is (presumably?) subject to fewer compile-time checks. Closing.

PR #2068, opened 2016-11-15
Set the aspect ratio on SimpleExoPlayerView when an active player is attached

remcomokveld:
When a SimpleExoPlayerView is attached to an already playing SimpleExoPlayer the onVideoSizeChanged method would not be called. This caused a stretch when for example the surface view was recreated due to a configuration change.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/ui/SimpleExoPlayerView.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ui/SimpleExoPlayerView.java
@@ -33,6 +33,7 @@ import android.widget.ImageView;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.ExoPlaybackException;
 import com.google.android.exoplayer2.ExoPlayer;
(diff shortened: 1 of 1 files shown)


remcomokveld:
I signed it!

ojw28:
Thanks for reporting this issue. I'm not sure this is the right solution though. The use of `videoFormat.rotationDegrees` in particular may not be correct, since the decoder may or may not have already applied the rotation. We may need to make it so that `onVideoSizeChanged` is always invoked after the `Surface` has been replaced (which is what we do for `onRenderedFirstFrame`).

As a related point: `SimpleExoPlayerView` should probably apply any unapplied rotation in the case that `TextureView` is being used, which it doesn't currently do.

I've filed https://github.com/google/ExoPlayer/issues/2077 to track fixing these issues. I'll close this pull request, since the fix will look different to the proposal here.

ojw28:
I think this should be fixed in `dev-v2` by the change ref'd in #2077, although I didn't test this specific case. Please give it a try and let us know if there's still an issue!

PR #2072, opened 2016-11-16
Make AspectRatioFrameLayout open

danielgomezrico:
If I want to embend a video inside a custom view you can have a custom AspectRatioFrameLayout and use <merge> in the view xml.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
@@ -28,7 +28,7 @@ import java.lang.annotation.RetentionPolicy;
 /**
  * A {@link FrameLayout} that resizes itself to match a specified aspect ratio.
  */
-public final class AspectRatioFrameLayout extends FrameLayout {
+public class AspectRatioFrameLayout extends FrameLayout {
 
   /**
    * Resize modes for {@link AspectRatioFrameLayout}.


ojw28:
You can already use `AspectRatioFrameLayout` in a custom layout xml. Removing the `final` modifier is only needed if you actually need to extend the class to add extra functionality. Could you please describe what additional functionality you need; it may well be that extending the class isn't the best solution to whatever you're trying to solve. Thanks!

danielgomezrico:
@ojw28 I have a custom view (FrameLayout) where I inflate the view with the `AspectRatioFrameLayout`, so my Idea was to use `<merge>` inside the layout and make my custom view `AspectRatioFrameLayout` so it have less nested views.

ojw28:
I'm still not entirely sure I understand what you mean, sorry! Please can you provide an example (e.g. the actual layout file that you're using, and the simpler one that you could use if `AspectRatioFrameLayout` were non-final?

Also, I doubt removing a single view from the hierarchy will provide any measurable performance gain here, unless you're repeating this layout many times (e.g. as the layout for a grid or list entry where many entries are visible simultaneously).

danielgomezrico:
@ojw28 yeah, its a list of videos.

I have an item:

``` xml
<merge xmlns:android="http://schemas.android.com/apk/res/android"
  xmlns:app="http://schemas.android.com/apk/res-auto"
  xmlns:tools="http://schemas.android.com/tools"
  android:id="@+id/aspectRationFrameLayout"
  android:layout_width="match_parent"
  android:layout_height="match_parent"
  android:animateLayoutChanges="true"
  android:background="@color/winwin_gray_light">

  <TextureView
    android:id="@+id/videoTextureView"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    tools:visibility="gone" />

</merge>
```

And I want my view class to be AspectRatioFrameLayout

``` java
class VideoView extends AspectRatioFrameLayout {}
```

I mean, this way it ends like:

``` markdown
-> AspectRatioFrameLayout
     -> TextureView
```

Instead of 

``` markdown
-> FrameLayout
    -> AspectRatioFrameLayout
        -> TextureView
```

For every item!

Its a little ui optimization and its better because videos are not lightweight, and in this way we avoid a FrameLayout at the top of every item.

ojw28:
- Hmm. That seems like a strange way to be structuring things. Why does your `VideoView` class need to extend anything? Why not just inflate a layout xml directly, and pass the inflated root to a standalone helper class?
- I suspect that `TextureView` is so much more expensive than the extra view you're saving, that you'll see no measurable performance gain whatsoever. Have you actually seen a measurable performance difference, and if so what data are you looking at?

danielgomezrico:
@ojw28 do you have an example with a standalone helper class? 
I do it because I need to use it inside other layout files.

Its expensive but I need to have a list and scroll smoothly, Im thinking about using SurfaceView but just for sdk 23 <.

ojw28:
`SurfaceView` is expensive too, relative to the cost of the extra view. My point is that there's little point in optimising something that's a very low percentage of the total cost. Unless you can actually measure the difference, you should in my opinion just leave the additional view in your hierarchy and be done with it.

PR #2127, opened 2016-11-29
Added support for PCM u/a-law audio in FLV containers

EE-GSlomin:
Adds basic support for PCM a-law and u-law audio for FLV containers.

Some things to note:

As far as I can tell, G.711 is standardized on an 8khz sample rate even though the FLV specification states that 8k is unsupported for anything other than nellymouser. 

The G711 codec does not appear to support stereo, and should most likely ignore the header bits for FLV.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
@@ -15,6 +15,8 @@
  */
 package com.google.android.exoplayer2.extractor.flv;
 
+import android.media.AudioFormat;
+import android.media.AudioTrack;
 import android.util.Pair;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
@@ -30,6 +32,8 @@ import java.util.Collections;
 /* package */ final class AudioTagPayloadReader extends TagPayloadReader {
 
   // Audio format
+  private static final int AUDIO_FORMAT_ALAW = 7;
+  private static final int AUDIO_FORMAT_ULAW = 8;
   private static final int AUDIO_FORMAT_AAC = 10;
 
   // AAC PACKET TYPE
@@ -44,6 +48,7 @@ import java.util.Collections;
   // State variables
   private boolean hasParsedAudioDataHeader;
   private boolean hasOutputFormat;
+  private int audioFormat;
 
   public AudioTagPayloadReader(TrackOutput output) {
     super(output);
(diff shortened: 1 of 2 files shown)


EE-GSlomin:
I signed it!

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -78,8 +97,13 @@ protected boolean parseHeader(ParsableByteArray data) throws UnsupportedFormatEx
   @Override
   protected void parsePayload(ParsableByteArray data, long timeUs) {
     int packetType = data.readUnsignedByte();
-    // Parse sequence header just in case it was not done before.
-    if (packetType == AAC_PACKET_TYPE_SEQUENCE_HEADER && !hasOutputFormat) {
+
+    if (audioFormat != AUDIO_FORMAT_AAC || packetType == AAC_PACKET_TYPE_AAC_RAW) {
```
ojw28: Why have you swapped the if/else order around? It's more natural to have the blocks outputting formats before the ones outputting samples. Can we put them back into that order again?

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -58,15 +63,29 @@ public void seek() {
...
       }
-      // TODO: Add support for MP3 and PCM.
-      if (audioFormat != AUDIO_FORMAT_AAC) {
-        throw new UnsupportedFormatException("Audio format not supported: " + audioFormat);
+      // TODO: Add support for MP3.
+      if (audioFormat == AUDIO_FORMAT_ALAW || audioFormat == AUDIO_FORMAT_ULAW) {
+
+          String type = (audioFormat == AUDIO_FORMAT_ALAW) ? MimeTypes.AUDIO_ALAW : MimeTypes.AUDIO_ULAW;
+          int encoding = (encodingSize == 1) ? C.ENCODING_PCM_16BIT : C.ENCODING_PCM_8BIT;
+          int minSize = AudioTrack.getMinBufferSize(8000, AudioFormat.CHANNEL_OUT_MONO, encoding);
```
ojw28: I don't think this is the right thing to do here. This is supposed to be the maximum size of a sample, which is unrelated to how large the AudioTrack buffer needs to be. Based on L103/L104, it looks like the maximum sample size would be the maximum size of a tag payload. Is that bounded in FLV? It looks a bit like it could be up to 2^24, which is really too big to be setting here. If so, it's probably fine to leave it set to Format.NO_VALUE as is done for the AAC case.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -58,15 +63,29 @@ public void seek() {
...
+
+          Format format = Format.createAudioSampleFormat(null, type, null, minSize * 4, minSize,
+                  1, 8000, encoding, null, null, 0, null);
+          output.format(format);
+
+          hasOutputFormat = true;
+          hasParsedAudioDataHeader = true;
```
ojw28: This line looks unnecessary because it's set on L89 anyway. Delete?

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
EE-GSlomin: It's generally faster to place the most executed cases first, though I'm not sure if this actually matters anymore with modern VMs and compilers. I just assumed it was an oversight when I changed it. If it's there to maintain a style, then that's perfectly fine and can change it back.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
EE-GSlomin: As far as I can tell, the only upper limit is the 24 bit size as you pointed out. Will change to NO_VALUE.

ojw28: Sounds good; thanks.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
ojw28: I don't think the difference is going to be measurable, so I think we'd prefer the order that's best for readability in this case (i.e. how it was before). Thanks!

EE-GSlomin:
No problem, that should take care of it.

ojw28:
Urgh, sorry. This needs to go into `dev-v2`, not `release-v2` branch. Please open a pull request to that branch instead, and we'll get it merged. Thanks!

PR #2132, opened 2016-12-01
allow access to child classes

drhill:
Allow classes extending SimpleExoPlayer to use the existing ComponentListener


ojw28:
I don't understand why this is needed. The `ComponentListener` is already passed to the `buildXRenderers` methods where needed. See [here](https://github.com/google/ExoPlayer/blob/894ae1a310cc2e1fe5ba59eb56c7fbeb751036d5/library/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java#L625). What's the use case where you need some other way to access it?

drhill:
Lol, sorry.  Copy and paste did me in when I grabbed the existing buildAudioRenderers and made a minor change in my new class.  That function uses componentListener and not the passed in eventListener.

PR #2133, opened 2016-12-01
add function to change subtitles style

drhill:


ojw28:
There are a whole bunch of useful setX methods on SubtitleView (looks like there are 9 such methods). It seems inconsistent to expose only one of them, but exposing all of them is going to bloat SimpleExoPlayerView quite a bit. How about a `getSubtitleView` method that just returns the view? That doesn't feel ideal either, but is perhaps the "less bad" option?

drhill:
That seems fine.  Ultimately I used this just to add outlining to the text so you can read it no matter what picture is displayed behind.  I'll make the change later.  Thanks!

ojw28:
We'll try and merge some of these at roughly the same time (it makes handling them a little easier for us), but this looks ready to go when that happens. Thanks!

drhill:
Great thanks.  Let me know if I need to do anything else.

PR #2135, opened 2016-12-01
add mp2 audio

drhill:
add mp2 audio hook for ffmpeg

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
@@ -1406,6 +1406,9 @@ public final class MatroskaExtractor implements Extractor {
           initializationData = Collections.singletonList(codecPrivate);
           break;
         case CODEC_ID_MP2:
+          mimeType = MimeTypes.AUDIO_MPEG_L2;
+          maxInputSize = MP3_MAX_INPUT_SIZE;
+          break;
         case CODEC_ID_MP3:
           mimeType = MimeTypes.AUDIO_MPEG;
           maxInputSize = MP3_MAX_INPUT_SIZE;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1403,6 +1405,7 @@ public void initializeOutput(ExtractorOutput output, int trackId) throws ParserE
           mimeType = MimeTypes.AUDIO_AAC;
           initializationData = Collections.singletonList(codecPrivate);
           break;
+        case CODEC_ID_MP2:
```
ojw28, line 1408: I think it's fine/good to add MP2 to the list, but if you use `AUDIO_MPEG` then we'll end up with the mp3 decoder on regular Android devices being used to try and play mp2 streams, which will fail. I'm pretty sure you need to be using `AUDIO_MPEG_L2` instead for this. `FfmpegLibrary.getCodecName` appears to treat the two in the same way anyway, so changing it should "just work".

It's unclear to me whether `maxInputSize` should be set to anything or not. I'm not sure where `MP3_MAX_INPUT_SIZE` came from. @andrewlewis will know :).

andrewlewis: The constant came from the platform [MP3Extractor.cpp](https://android.googlesource.com/platform/frameworks/av/+/master/media/libstagefright/MP3Extractor.cpp#431). Should be safe to use it here too as per the comment (and we might even be able to use a smaller value, without rounding to the next power of two, but I'm not sure this is a good idea).

Unrelated to this change, we could set `MP3_MAX_INPUT_SIZE = MpegAudioHeader.MAX_FRAME_SIZE_BYTES` to avoid having the magic number.

PR #2136, opened 2016-12-01
truehd support

drhill:
pass through support for Dolby TrueHD

### DIFF
--- a/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
+++ b/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
@@ -96,7 +96,8 @@ import java.util.List;
         parsableExtraData.setPosition(extraData.length - 4);
         sampleRate = parsableExtraData.readUnsignedIntToInt();
       }
-      hasOutputFormat = true;
+      if (!"truehd".equals(codecName) || sampleRate > 0)
+        hasOutputFormat = true;
     }
     outputBuffer.data.position(0);
     outputBuffer.data.limit(result);
--- a/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
+++ b/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
@@ -96,8 +96,7 @@ import java.util.List;
         parsableExtraData.setPosition(extraData.length - 4);
         sampleRate = parsableExtraData.readUnsignedIntToInt();
       }
-      if (!"truehd".equals(codecName) || sampleRate > 0)
-        hasOutputFormat = true;
+      hasOutputFormat = true;
     }
     outputBuffer.data.position(0);
     outputBuffer.data.limit(result);
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
@@ -125,9 +125,14 @@ public final class Ac3Util {
       121, 139, 174, 208, 243, 278, 348, 417, 487, 557, 696, 835, 975, 1114, 1253, 1393};
 
   /**
+   * Value to identify TrueHD packet
+   */
+  private static final int TRUEHD_SYNCWORD = 0xba6f72f8;
+
+  /**
    * Sample hold count for TRUE HD frames
    */
-  public static final int TRUEHD_SAMPLE_COMMIT_COUNT = 16;
+  public static final int TRUEHD_SAMPLE_COMMIT_COUNT = 8;
 
 
   /**
@@ -310,19 +315,15 @@ public final class Ac3Util {
 
   public static int parseTrueHDSyncframeAudioSampleCount(ByteBuffer buffer) {
 
-    if (buffer.limit() < 13)
+    if (buffer.limit() < 12)
       return 0;
     int ratebits;
     int streamType = buffer.getInt(4);
-    if (streamType == 0xbb6f72f8) { // mlp
-      ratebits = buffer.get(10);
-    } else if (streamType == 0xba6f72f8) { // truehd
+    if (streamType == TRUEHD_SYNCWORD)
       ratebits = buffer.get(8);
-    } else
+    else
       return 0;
-
-    int frameSize = 40 << (ratebits & 7);
-    return frameSize * TRUEHD_SAMPLE_COMMIT_COUNT;
+    return (40 << (ratebits & 7)) * TRUEHD_SAMPLE_COMMIT_COUNT;
   }
 
   private Ac3Util() {}
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
@@ -125,6 +125,12 @@ public final class Ac3Util {
       121, 139, 174, 208, 243, 278, 348, 417, 487, 557, 696, 835, 975, 1114, 1253, 1393};
 
   /**
+   * Sample hold count for TRUE HD frames
+   */
+  public static final int TRUEHD_SAMPLE_COMMIT_COUNT = 16;
+
+
+  /**
    * Returns the AC-3 format given {@code data} containing the AC3SpecificBox according to
    * ETSI TS 102 366 Annex F. The reading position of {@code data} will be modified.
    *
@@ -300,6 +306,23 @@ public final class Ac3Util {
     } else { // sampleRate == 48000
       return 4 * bitrate;
     }
+  }
+
+  public static int parseTrueHDSyncframeAudioSampleCount(ByteBuffer buffer) {
+
+    if (buffer.limit() < 13)
+      return 0;
+    int ratebits;
+    int streamType = buffer.getInt(4);
+    if (streamType == 0xbb6f72f8) { // mlp
+      ratebits = buffer.get(10);
+    } else if (streamType == 0xba6f72f8) { // truehd
+      ratebits = buffer.get(8);
+    } else
+      return 0;
+
+    int frameSize = 40 << (ratebits & 7);
+    return frameSize * TRUEHD_SAMPLE_COMMIT_COUNT;
   }
 
   private Ac3Util() {}
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
@@ -125,9 +125,9 @@ public final class Ac3Util {
       121, 139, 174, 208, 243, 278, 348, 417, 487, 557, 696, 835, 975, 1114, 1253, 1393};
 
   /**
-   * Value to identify TrueHD packet
+   * Min size to identify TrueHD packet
    */
-  private static final int TRUEHD_SYNCWORD = 0xba6f72f8;
+  private static final int TRUEHD_SYNCWORD_MIN_LEN = 12;
 
   /**
    * Sample hold count for TRUE HD frames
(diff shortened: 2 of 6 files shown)


drhill:
Are there any changes that need to be done for this to be merged?

library/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -152,7 +152,11 @@ private C() {}
    */
   @SuppressWarnings("InlinedApi")
   public static final int ENCODING_DTS_HD = AudioFormat.ENCODING_DTS_HD;
-
+  /**
+   *
+   */
+  @SuppressWarnings("InlinedApi")
+  public static final int ENCODING_TRUE_HD = 14;
```
andrewlewis: dev-v2 now targets API 25 so please use [AudioFormat.ENCODING_DOLBY_TRUEHD](https://developer.android.com/reference/android/media/AudioFormat.html#ENCODING_DOLBY_TRUEHD) and update javadoc to match the others.

library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1608,4 +1628,118 @@ private static boolean parseMsAcmCodecPrivate(ParsableByteArray buffer) throws P
 
   }
 
+  private class TrueHDSampleHolder {
```
andrewlewis: Is there any other signaling in Matroska that indicates that a sample contains a TrueHD syncframe? Or is there a guarantee that 16 samples batched together will contain a syncframe at the start of one of the samples?

If so I think it might be possible to simplify/remove this. Calls to TrackOutput.sampleData and TrackOutput.sampleMetadata don't have to be matching in the sense that you can call sampleData a few times then call sampleMetadata once for the batch. So assuming you don't need to synchronize to a syncframe in the extractor you can just write data using the 'normal' readToOutput path then batch the samples using one call to sampleMetadata in commitSampleToOutput.

The platform AudioTrack should correctly synchronize to a syncframe internally (please let me know if this doesn't work for TrueHD) so the only remaining issue is that our AudioTrack wrapper needs to be able to read the syncframe header to get the size of the encoded sample in frames. Assuming a batch of 16 samples read in this extractor contains a syncframe, can we just move the code to find the syncframe header to the AudioTrack or Ac3Util so it only gets executed once after the flush? That seems preferable to introducing an extra copy here. Thanks!

library/src/main/java/com/google/android/exoplayer2/C.java
drhill: done

library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
drhill: I'll look into seeing if this can be simplified more.  Previous attempts at doing batching in AudioTrack didn't work.

andrewlewis: I think it's good to do the batching in MatroskaExtractor by calling sampleMetadata once every 16 samples in commitSampleToOutput. But it would be preferable to locate the start of the syncframe and get the encoded frame size in AudioTrack/Ac3Util. This all assumes that you are guaranteed to fit a syncframe in 16 samples, which I'm not sure about.

drhill: latest changes get rid of the extra copying.  I trimmed the every 16 samples down to 8 to make it more responsive, but when you get lower the shield tv keeps getting buffer underruns.  The shield tv also has problems if the first data sent to its truehd passthrough decoder isn't a sync frame.  I'll look into it further if I can do a better job and get the encoded frame size.

drhill:
rebased to latest dev-v2.  There is still some funkyness if you jump around in a true hd clip with the latest code.  It didn't happen before when I was buffering, but without the buffering it gets weird.  Going to look into this again soon (hopefully).

drhill:
fixed my funkyness.  I wasn't passing all samples before and that caused issues.  Now I pass all samples, before I find a sync frame I pass them one at a time like normal.  Once a sync frame is found I call the sample metadata once every 8 samples.  On a seek it gets reset and looks for a sync frame again.  Eight samples was chosen because I haven't noticed any drop outs or underruns and when I had it at 16 seeking was less responsive.  I've got nothing else but trial and error with an nvidia shield on that one.

drhill:
I was not aware as I'm on a slightly older version.  It will have to wait as last Friday my now former day development job decided they needed to do massive layoffs to juke the stats again.  So it may be a few weeks until I can address this.

drhill:
I rebased with the latest dev-v2 and it works for me fine.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
```diff
@@ -96,7 +96,8 @@ public FfmpegDecoderException decode(DecoderInputBuffer inputBuffer,
         parsableExtraData.setPosition(extraData.length - 4);
         sampleRate = parsableExtraData.readUnsignedIntToInt();
       }
-      hasOutputFormat = true;
+      if (!"truehd".equals(codecName) || sampleRate > 0)
```
andrewlewis: If the sample rate can be zero here, it looks like the FfmpegAudioRenderer's output format can have a zero sample rate, so the AudioTrack can be configured with a zero sample rate. Is there a way to guarantee that we have a non-zero sample rate at this point instead?

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioTrack.java
```diff
@@ -1439,6 +1439,8 @@ private static int getEncodingForMimeType(String mimeType) {
         return C.ENCODING_DTS;
       case MimeTypes.AUDIO_DTS_HD:
         return C.ENCODING_DTS_HD;
+      case MimeTypes.AUDIO_TRUEHD:
```
andrewlewis: Around L588 do we need to set a buffer size for TrueHD output or is the buffer size for DTS fine to use? (In the latter case, please update the comment to refer to TrueHD in addition to DTS and DTS HD.)

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1931,4 +1947,68 @@ private static boolean parseMsAcmCodecPrivate(ParsableByteArray buffer) throws P
...
+            throws IOException, InterruptedException {
+      samplesSize += size;
+      ++counter;
+      if (isFirstSampleCheck())
+        this.blockFlags = blockFlags;
+      if (!gotSyncFrame) {
+        byte[] bytes = new byte[12];
+        input.peekFully(bytes, 0, 12);
+        input.resetPeekPosition();
+        if (0 < Ac3Util.parseTrueHDSyncframeAudioSampleCount(ByteBuffer.wrap(bytes).order(ByteOrder.LITTLE_ENDIAN))) {
```
andrewlewis: Could parseTrueHDSyncframeAudioSampleCount take a byte[] to avoid wrapping every time?

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1931,4 +1947,68 @@ private static boolean parseMsAcmCodecPrivate(ParsableByteArray buffer) throws P
...
+
+    void testSample(ExtractorInput input, int size, @C.BufferFlags int blockFlags, Track track)
+            throws IOException, InterruptedException {
+      samplesSize += size;
+      ++counter;
+      if (isFirstSampleCheck())
+        this.blockFlags = blockFlags;
+      if (!gotSyncFrame) {
+        byte[] bytes = new byte[12];
+        input.peekFully(bytes, 0, 12);
```
andrewlewis: This extractor supports header stripping compression where some specified bytes are prepended to every output sample. I'm not sure it makes sense to use that with TrueHD audio tracks, given how the samples are fragmented. If it were possible, we'd need to read from sampleStrippedBytes first (see readToTarget). If it's definitely not possible, maybe worth asserting that track.sampleStrippedBytes is null.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -355,6 +357,8 @@
   @C.BufferFlags
   private int blockFlags;
 
+  private int lastReadResult = Extractor.RESULT_CONTINUE;
```
andrewlewis: Maybe just store private boolean readEndOfInput, set to true if continueReading is false in read()?

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
```diff
@@ -302,6 +313,19 @@ private static int getAc3SyncframeSize(int fscod, int frmsizecod) {
     }
   }
 
+  public static int parseTrueHDSyncframeAudioSampleCount(ByteBuffer buffer) {
+
+    if (buffer.limit() < 12)
+      return 0;
```
andrewlewis: Use C.INDEX_UNSET instead of 0?

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1931,4 +1947,68 @@ private static boolean parseMsAcmCodecPrivate(ParsableByteArray buffer) throws P
...
+        }
+      }
+      this.track = track;
+    }
+
+    boolean commit(long timeUs, boolean forceCommit) {
+      if (isFirstSampleCheck())
+        this.timeUs = timeUs;
+
+      boolean commit = counter == Ac3Util.TRUEHD_SAMPLE_COMMIT_COUNT || letFirstIn || forceCommit;
```
andrewlewis: I'm confused about letFirstIn -- why does the first sample we read not need to get batched into a group of TRUEHD_SAMPLE_COMMIT_COUNT samples?

I thought the way this should work is that each sample that is eventually output by this extractor starts with a syncframe and has TRUEHD_SAMPLE_COMMIT_COUNT samples from the container concatenated together (except for the last sample, which may be truncated). This logic doesn't appear to wait for a syncframe before starting output.

andrewlewis (commented):
Thanks!

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioTrack.java
drhill: Adding to the comment as any truehd tracks I've played seem fine with the DTS buffer size.  On a side note I have another pull request for a buffer size change because in 7.1 tracks I get audio dropouts on DTS-HD MA tracks.  5.1 tracks seem fine.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac3Util.java
drhill: Sure, but I think I have to add a check for framesPerEncodedSample == 0 || framesPerEncodedSample == C.INDEX_UNSET where we call getFramesPerEncodedSample around Line 747.  Or we can do a framesPerEncodedSample <= 0 which seems bad.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
drhill: Yep

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
drhill: Yeah... that is much easier to deal with that way.  I think part of that is a holder from a more needlessly complex implementation.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
drhill: Edit:  Yeah this needs to be in here because of certain media (older VC1 with TrueHD) requires it for proper playback (and seeking).  I suppose the version of software that created the MKV file might be to blame as it seems to be awhile between where we seek to and when we find our first sync frame, but other software plays it fine.   I'll rename the variable skipBatchingUntilSyncFrameFound.  It's not ideal, but throwing away the data before finding a sync flag causes issues too and the only way for clean playback was to just commit data (as any other track type/codec does) until we find a sync frame.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
drhill: I'll put in the code to check sampleStrippedBytes if it has data, but I don't think I have anything to actually test if that case were to ever happen.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
drhill: I'm taking that change out.  I think it was caused by some previous bad logic in the batching.  I can't reproduce the case where I ended up with a zero sample rate anymore.... which is good I think.

drhill:
Closing since it was already merged in.  Thanks!

PR #2140, opened 2016-12-01
Added FLV PCM law support

EE-GSlomin:
Re-opening against the correct target.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
@@ -15,6 +15,8 @@
  */
 package com.google.android.exoplayer2.extractor.flv;
 
+import android.media.AudioFormat;
+import android.media.AudioTrack;
 import android.util.Pair;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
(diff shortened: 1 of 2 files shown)


EE-GSlomin:
My personal project depends on the ExoMedia framework which depends on ExoPlayer1.5 instead of 2.0. I've already merged these changes into my 1.5 branch. I could open another pull for dev-v1. Is anything being backported into the 1.x branches still?

ojw28:
We're only back-porting bug fixes, not features. And only the subset of bug fixes that can be easily back-ported (i.e. we're not back-porting bug fixes that would require also back-porting other structural changes).

You might want to look at migrating to ExoPlayer V2. You may find you no longer need the ExoMedia layer, since ExoPlayer V2 provides higher level APIs of its own (e.g. SimpleExoPlayer, SimpleExoPlayerView).

EE-GSlomin:
Fair enough. I came to similar conclusions about ExoMedia, we'll probably drop it in the near future.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -58,15 +63,26 @@ public void seek() {
   protected boolean parseHeader(ParsableByteArray data) throws UnsupportedFormatException {
     if (!hasParsedAudioDataHeader) {
       int header = data.readUnsignedByte();
-      int audioFormat = (header >> 4) & 0x0F;
+      audioFormat = (header >> 4) & 0x0F;
       int sampleRateIndex = (header >> 2) & 0x03;
+      int encodingSize = header & 0x01;
       if (sampleRateIndex < 0 || sampleRateIndex >= AUDIO_SAMPLING_RATE_TABLE.length) {
```
ojw28, line 69: Hah at this check by the way. sampleRateIndex is guaranteed to be in the range [0-3] because of the `& 0x03` where it's calculated. The values in `AUDIO_SAMPLING_RATE_TABLE` aren't ever used :). I'm going to push a change that cleans some of this up a little!

PR #2150, opened 2016-12-04
Script (and changes) to help build the ffmpeg extension

stuckless:
- Changes enable that you can publish the ffmpeg extension to the maven local repo
- Changes enable that you specify a build version on the commandline
- Script auto-configures NDK and allows for extra ffmpeg configure args to be passed.


ojw28:
I think we'd rather just stick with the instructions in the README, and require developers to explicitly follow the steps (creating their own script in the process, if they want to streamline repeating the process). In particular I don't think it's a good idea to bury details of which decoders are being enabled in the script. The developer really should be making an explicit choice.

PR #2207, opened 2016-12-16
Fix a small mistake in opus_multistream_decoder

supercairos:
As per documentation, frame_size should be :

> The number of samples per channel of available space in pcm. If this is less than the maximum packet duration (120 ms; 5760 for 48kHz), this function will not be capable of decoding some packets. In the case of PLC (data==NULL) or FEC (decode_fec=1), then frame_size needs to be exactly the duration of audio that is missing, otherwise the decoder will not be in the optimal state to decode the next incoming packet. For the PLC and FEC cases, frame_size mustbe a multiple of 2.5 m

### DIFF
--- a/extensions/opus/src/main/jni/opus_jni.cc
+++ b/extensions/opus/src/main/jni/opus_jni.cc
@@ -110,7 +110,7 @@ DECODER_FUNC(jint, opusDecode, jlong jDecoder, jlong jTimeUs,
   int16_t* outputBufferData = reinterpret_cast<int16_t*>(
       env->GetDirectBufferAddress(jOutputBufferData));
   int sampleCount = opus_multistream_decode(decoder, inputBuffer, inputSize,
-      outputBufferData, outputSize, 0);
+      outputBufferData, inputSampleCount, 0);
   // record error code
   errorCode = (sampleCount < 0) ? sampleCount : 0;
   return (sampleCount < 0) ? sampleCount


supercairos:
I signed it!

ojw28:
@vigneshvg - Could you take a look at this? Thanks! Note: The fix will have to be submitted into dev-v2, not release-v2.

vigneshvg:
Is there a specific case that this patch fixes?

From what i see in other uses of the decoder (ffmpeg [1], android platform [2]) they are all hardcoded to 960*6 = 5760. Considering that the existing calculation seems to make sense.

[1] https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libopusdec.c#L142
[2] https://android.googlesource.com/platform/frameworks/av/+/master/media/libstagefright/codecs/opus/dec/SoftOpus.cpp#461

ojw28:
@supercairos - Could you comment on the above? Thanks.

supercairos:
Your (old) code was working. I'm just quoting the [Opus documentation](http://opus-codec.org/docs/opus_api-1.1.3/group__opus__multistream.html#gaa4b89541efe01970cf52e4a336db3ad0) that state:
> 	frame_size	int: The number of samples per channel of available space in pcm. If this is less than the maximum packet duration (120 ms; 5760 for 48kHz), this function will not be capable of decoding some packets. In the case of PLC (data==NULL) or FEC (decode_fec=1), then frame_size needs to be exactly the duration of audio that is missing, otherwise the decoder will not be in the optimal state to decode the next incoming packet. For the PLC and FEC cases, frame_size must be a multiple of 2.5 ms. 

In your case you're giving him the number of total bytes for each sample instead of the number of samples:
` const jint outputSize = inputSampleCount * kBytesPerSample * channelCount;`
Witch, I don't know if it could cause any errors in the opus decoding process.
That's why I proposed that patch.

Now looking at your sample source code, I guess my assumption were also wrong, and you should always use the maximum allowed frame size. That is as per documentation:
> If this is less than the maximum packet duration (120 ms; 5760 for 48kHz), this function will not be capable of decoding some packets.

120ms@48khz = (20 * (48 000 / 1000) * 6 = 960 * 6 = 5760 frames witch is what Android and FFMPEG are using.

I guess we can close this pull request and create a new one that use 
`#define MAX_FRAME_SIZE (960 * 6)` as input in the `opus_multistream_decode` function.

PR #2219, opened 2016-12-17
add imageview into simpleexoplayerview to display image subtitles 

drhill:


library/src/main/java/com/google/android/exoplayer2/text/ImageCue.java
```diff
@@ -0,0 +1,41 @@
+package com.google.android.exoplayer2.text;
+
+import android.graphics.Bitmap;
+
+public class ImageCue extends Cue {
+
+  final private long start_display_time;
+  final private int x;
```
ojw28: A lot of these parameters appear to be duplicating what's already in the Cue base class. Why can't the cue be sized using the existing fields (line, position, size) with (lineType=LINE_TYPE_FRACTION). It feels a lot like this should be supported just by adding a Bitmap field directly into Cue, that can be set instead of (text, textAlignment).

A few asides:
- It's unclear what units these variables are in.
- We tend to just make final immutable fields public, rather than having getter methods.
- "final private" -> "private final" (or "public final" as per comment above) as per the [Google style guide](https://google.github.io/styleguide/javaguide.html#s4.8.7-modifiers).

library/src/main/java/com/google/android/exoplayer2/text/ImageCue.java
```diff
@@ -0,0 +1,41 @@
...
+  final private int y;
+  final private int bitmap_height;
+  final private int bitmap_width;
+  final private int plane_height;
+  final private int plane_width;
+  final private Bitmap bitmap;
+  final private boolean isForced;
```
ojw28: We already support forcing at the Format level (Format.selectionFlags can include C.SELECTION_FLAG_FORCED).

library/src/main/res/layout/exo_simple_player_view.xml
```diff
@@ -36,6 +36,12 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent"/>
 
+    <ImageView
```
ojw28: This doesn't generalize properly to support multiple image cues at once. Any solution will likely need SubtitleView/SubtitlePainter to handle drawing directly with proper support for this, similarly to what happens in the text case.

library/src/main/java/com/google/android/exoplayer2/text/ImageCue.java
drhill: Ok, would you like it to map as:

line = y
position = x
size = width
size_height (new) = height
bitmap (new)

drop start_display_time since it isn't currently used, isForced see below

library/src/main/java/com/google/android/exoplayer2/text/ImageCue.java
drhill: Trying to wrap my head around this without actively debugging.  There can be subs that are forced in a non forced stream... but if we don't have the stream enabled I guess that shouldn't matter.  So we don't need it.

library/src/main/res/layout/exo_simple_player_view.xml
drhill: Ok, will look into the changing to painters next.

drhill:
Other than now have some conflicts, is that good?

Looking at it now it was certainly much smaller and simpler than expected just adding a bitmap to the Cue and painter.

library/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -79,12 +80,18 @@
...
+  public final Bitmap bitmap;
+  /**
    * The position of the {@link #lineAnchor} of the cue box within the viewport in the direction
    * orthogonal to the writing direction, or {@link #DIMEN_UNSET}. When set, the interpretation of
    * the value depends on the value of {@link #lineType}.
    * <p>
    * For horizontal text and {@link #lineType} equal to {@link #LINE_TYPE_FRACTION}, this is the
    * fractional vertical position relative to the top of the viewport.
+   * <p>
+   * If {@link #bitmap} is not null then this value is used to indicate the top position
```
ojw28: There's no reason to special case this. Why not define it to be the position of the lineAnchor (i.e. top, middle or bottom of the bitmap) as in the text case?

The only thing that's special about a bitmap is that lineType must be LINE_TYPE_FRACTION (LINE_TYPE_NUMBER doesn't make sense, since we don't know how tall a line is if the text is in bitmap form).

library/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -119,6 +126,8 @@
    * For horizontal text, this is the horizontal position relative to the left of the viewport. Note
    * that positioning is relative to the left of the viewport even in the case of right-to-left
    * text.
+   * <p>
+   * If {@link #bitmap} is not null then this value is used to indicate the left position
```
ojw28: Ditto re using the anchor.

library/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -134,10 +143,26 @@
   /**
    * The size of the cue box in the writing direction specified as a fraction of the viewport size
    * in that direction, or {@link #DIMEN_UNSET}.
+   * <p>
+   * If {@link #bitmap} is not null then this value is used to indicate the width
```
ojw28: I don't think this comment is needed. Size is the same for text and for the bitmap.

library/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -134,10 +143,26 @@
...
    */
   public final float size;
 
   /**
+   * The height size of the cue box when a {@link #bitmap} is set specified as a fraction of the
+   * viewport size in that direction, or {@link #DIMEN_UNSET}.
+   */
+  public final float size_height;
```
ojw28: What's this used for? We already have the width. Can't the height be inferred? I doubt it would make sense to allow the bitmaps to be deformed into another aspect ratio, which is the only case I can think of where this isn't redundant.

drhill: Trying to get this straight in my head again.

We can grab width and height from the bitmap, but the passed in width (size) and height (size_height) should be the fractional value of the width and height based on the original plane width and height.(correct?).  So a image sub from a 1080p plane would pass in the percentage of the screen the bitmap takes up so it can be resized on a 720p painter view...  Hard to test because everything I have is 1080p.

Would it be better to change the parameter to plane_height (and add one for plane_width and not use size) and do all the calculations in subtitlepainter?

ojw28: `size` should just be the fraction of the video's width occupied by the bitmap. `position` and `line` should be the fractional positions of the respective anchors. I don't think you need `size_height` at all.

drhill:
Made requested changes to reduce comments and use only size.  Calculated from bitmap width and the original plane width.  Aspect ratio is kept so only width is needed.

SteinerOk:
Any plans to merge it?

IanDBird:
@ojw28 - Are there plans to merge this? We actually have a similar modification in our own fork when we added PGS / VobSub support. We could implement the majority of the functionality outside of ExoPlayer, but this one requirement required us to modify ExoPlayer directly. Would be nice to have framework support for image subtitles, even if there isn't actually shipped parsers.

PR #2244, opened 2016-12-22
compare strings should use equals insteadof ==

michalliu:
This bug only exists on v1, v2 fixed already.


michalliu:
I signed it!

ojw28:
This is not a bug; it's fine to use referential equality with strings if you know that it's sufficient, which is the case here. It's true that it's less confusing just to use `.equals`, but we're accepting bug fixes only into V1 at this point, which this is not.

PR #2247, opened 2016-12-22
improve the performance of far seek

WeiChungChang:
http://programmingmemojohnchang.blogspot.tw/2016/12/improve-performance-of-short-seek-of_22.html

#2253 

**Please read in detail by the link above which includes the result of experimental comparison**.

Here we do optimize to the upstream side.

Originally if the seek target is behind sampleQueue at the upstream side, we will clean up the whole sampleQueue. Then do refetch from the media chunk which contains the search target.

Obviously, there is a waste in downloading redundant data. As the result, if we keep the the data of the last partially downloaded media chunk, it could save some. 

If the conditions below are satisfied, we could try to locate the last key frame within sampleQueue and drop the out-of-range (decoded only) samples at renderer.
1. The search target locates within the last media chunk.
2. The nearest key frame precedes the search target has been either within sampleQueue or sent to decoded (or rendered).

To take use of this fact, we create a new function named skipToLastKeyframe(). It tries to find the last key sample within the sampleQueue. Also, we create a function named isWithinLastChunk() to check whether the seek target locates within the last media chunk.
Finally, we drop the out-of-range (decode only) samples at renderer.

The experimental result shows a speed up ~252.5%. for 4K MPEG DASH streaming source.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
@@ -202,6 +202,8 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
   private boolean outputStreamEnded;
   private boolean waitingForKeys;
 
+  private long latestResetPosition;
+
   protected DecoderCounters decoderCounters;
 
   /**
@@ -230,6 +232,7 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
     outputBufferInfo = new MediaCodec.BufferInfo();
     codecReconfigurationState = RECONFIGURATION_STATE_NONE;
     codecReinitializationState = REINITIALIZATION_STATE_NONE;
+    latestResetPosition = C.TIME_UNSET;
   }
 
   @Override
@@ -388,6 +391,7 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
   protected void onPositionReset(long positionUs, boolean joining) throws ExoPlaybackException {
     inputStreamEnded = false;
     outputStreamEnded = false;
+    latestResetPosition = positionUs;
     if (codec != null) {
       flushCodec();
     }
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -202,6 +202,8 @@ private static String buildCustomDiagnosticInfo(int errorCode) {
   private boolean outputStreamEnded;
   private boolean waitingForKeys;
 
+  private long latestResetPosition;
```
ojw28: I don't think you need to make any changes to this class. You should find that buffer.isDecodeOnly() is already true for the buffers you're skipping, which means they'll be skipped anyway. This flag should be being set in DefaultTrackOutput.readData because decodeOnlyUntilUs should be set to the last seek position.

Try reverting this class and seeing if everything still works. If so, great. If not, is something wrong with my reasoning above, or is something broken?

library/src/main/java/com/google/android/exoplayer2/source/chunk/ChunkSampleStream.java
```diff
@@ -114,15 +114,30 @@ public long getBufferedPositionUs() {
...
+  
     lastSeekPositionUs = positionUs;
+    if (isWithinLastChunk(positionUs)) {
+      result = sampleQueue.skipToLastKeyframe(positionUs);
+    } else {
+      result = sampleQueue.skipToKeyframeBefore(positionUs);
+    }
     // If we're not pending a reset, see if we can seek within the sample queue.
-    boolean seekInsideBuffer = !isPendingReset() && sampleQueue.skipToKeyframeBefore(positionUs);
+    boolean seekInsideBuffer = (!isPendingReset()) && (result == true);
```
ojw28: I think the changes here are more complicated than they need to be. Why not just add a boolean argument to skipToKeyframeBefore that determines whether it's allowed to skip to the last keyframe if the position argument exceeds the last timestamp in the queue. The only change you'd need to make in skipToKeyframeBefore would be to also check this boolean argument as part of the check:

```
      if (timeUs > largestQueuedTimestampUs) {
        return C.POSITION_UNSET;
      }
```

Then here, you can simply do:

```
boolean seekInsideBuffer = !isPendingReset()
    && sampleQueue.skipToKeyframeBefore(positionUs, isWithinLastChunk(positionUs));
```

You could then revert most of this change, whilst still achieving the optimization. It's probably preferable to keep a single argument version of skipToKeyframeBefore that calls skipToKeyframeBefore(position, false), to avoid having to change any other calling sites.

ojw28 (commented):
Hi. Thanks for the pull request and detailed description! This looks like a good idea, although I also think it can be achieved in a much simpler way. Please take a look at the comments attached. Thanks!

WeiChungChang:
Please ignore the last wrong commit;

I have made a new patch as #2318 

Thanks~

PR #2258, opened 2016-12-25
Add default artwork support to SimpleExoPlayerView

svenkonings:
Add support for a default artwork image that is displayed if no artwork can be found in the metadata.


svenkonings:
I signed it!

PR #2273, opened 2016-12-31
Add default artwork support to SimpleExoPlayerView

svenkonings:
Add support for a default artwork image that is displayed if no artwork can be found in the metadata.


ojw28:
Please update the class level javadoc of SimpleExoPlayerView to include information about the new attribute (around line 67/68). Otherwise this looks great, so we'll get this merged once that's done. Thanks!

svenkonings:
@ojw28 I added a description for the new attribute. Should I also update the description of use_artwork to mention that an artwork is also displayed if the default_artwork is set, or is this clear enough?

ojw28:
Looks good, thanks!

PR #2281, opened 2017-01-02
Minimal SSA/ASS Subtitle Support

jcable:
Hi, please find for consideration this pull request which adds SSA/AAS subtitle decoding.

The Script Info section is not decoded. The Styles and Events sections are fully decoded.

\N tags in the text are converted to newlines and all other text tags are removed. The full rich text and the full style information is kept in the Cue object and is therefore available for enhanced rendering but this is not implemented.

Unit tests are included.

Thanks to ojw28 for essential advice in getting timestamp handling correct.


library/src/main/java/com/google/android/exoplayer2/text/ssa/Style.java
```diff
@@ -0,0 +1,283 @@
...
+
+import java.util.Map;
+
+import static android.R.attr.angle;
+
+/**
+ * Created by cablej01 on 27/12/2016.
+ */
+
+public class Style {
```
ojw28, line 11: The way we typically support styling of subtitles is by attaching Span objects directly onto the Cue text, which needs to be a CharSequence (you can use SpannableStringBuilder directly) rather than just a String. Take a look at what we do for webvtt to get the idea (see in particular WebvttCueParser).

I think you should use the same approach for SSA. In particular, there are built in Span objects for font, background color, foreground color, size adjustment, bold, italic, underline, strike-out etc, which the existing subtitle rendering components will render automatically without you having to do anything.

For anything that's not supported by existing Span objects I guess it is sensible to extend Cue to add them (assuming they apply to the whole Cue). If they might apply to only part of the Cue, a custom Span would be appropriate. This stuff is only useful if you think someone's going to build a UI component that actually uses this information.

ojw28:
This will be much easier to review and merge as a sequence of smaller pull requests. Could we start with just the extractor change (and corresponding tests / test assets)? That looks pretty easy to get merged as a first step.

I posted a comment about the decoder part that probably requires some work. It's fine to send the decoder pieces a single pull request, although it might be simpler if you send two; one adding basic support for plain text SSA, and the second adding in all the styling stuff. Up to you! Thanks.

library/src/main/java/com/google/android/exoplayer2/text/ssa/SSADecoder.java
```diff
@@ -0,0 +1,235 @@
...
+ */
+
+
+public class SSADecoder extends SimpleSubtitleDecoder {
+    private static final String TAG = "SSADecoder";
+    private static String defaultDialogueFormat = "Start, End, Layer, Style, Name, MarginL, MarginR, MarginV, Effect, Text";
+    private static String defaultStyleFormat = "Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding";
+    private String[] dialogueFormat;
+    private String[] styleFormat;
+    private Map<String,Style> styles = new HashMap<>();
```
ojw28: I'd expect the decoder to be completely stateless (i.e. styles would be a local variable in decode). This approach will definitely be fine for a side-loaded caption file. I'm not sure what happens in the MKV case. Does the header from which this data is parsed get prefixed onto every sample as it's output, and if not already is that possible to do? If so, this approach will be fine for that case too.

jcable:
Hi Olly - I'll make the changes you suggest to allow the first pull request and work on from there.

Julian

library/src/main/java/com/google/android/exoplayer2/text/ssa/Style.java
jcable: Sounds good. I just wanted a branch point where everything was captured and titles would render without style. Happy to change the approach to fit the architecture.

library/src/main/java/com/google/android/exoplayer2/text/ssa/SSADecoder.java
jcable: I thought it was better to move the header from the mkv decoder to the SSADecoder just once

 If you prefer the header state stored in the mkv decoder and passed every time that is possible but much less efficient as there are typically many styles defined and re-parsing them for every cue seems very wasteful. 

If it were possible to pass something other than a byte[] to the decoder it would be fine, but the subtitle decoder isn't called directly by the container decoder so I have to recreate an octet stream to pass through the demuxer. Passing the styles every time probably increases the effective bit rate between the container decoder and the subtitle decoder by a factor of 10 or 20.

library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1453,6 +1502,10 @@ public void initializeOutput(ExtractorOutput output, int trackId) throws ParserE
         case CODEC_ID_SUBRIP:
           mimeType = MimeTypes.APPLICATION_SUBRIP;
           break;
+        case CODEC_ID_ASS:
+          mimeType = MimeTypes.TEXT_SSA;
+          ssaHeader = codecPrivate;
```
ojw28: Is this the header that you don't want to prefix onto each sample? If so, put it into the format object (by doing: `initializationData = codecPrivate`) and don't prefix it onto any sample at all.

library/src/main/java/com/google/android/exoplayer2/text/ssa/SSADecoder.java
ojw28: See other comment; you should probably put the header into the Format init data instead, if that works. This will require some plumbing of the format data from TextRenderer to the decoder.

library/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
jcable: OK, I can get the header into the format object and I can see it is there in the decoder factory createDecoder call in onStreamChanged in the TextRenderer.

But the decode method of the decoder is called from the decoder thread, not from the text. 

If I plumb the format data into the decoder from the TextRenderer doesn't that make the decoder stateful again? The subtitle objects are created in the decoder thread and i can't see any way of passing Format data into each call to the decoder that would make it stateless.

The styles need to be accessible in the decoder at decode time if I'm going to create the spans at the time the cues are created.

jcable:
Had a go. I think this is heading in the direction you suggest. I need a bit more guidance.

jcable:
Having slept on it I realise that if I put a `<span class="style name">` into the text the decoder doesn't need to keep the styles. Is that the sort of thing you have in mind?

Should separate lines of text in a subtitle go into separate cues?

jcable:
I've  noticed a difference in the timestamps passed to Subtitle.getNextEventTimeIndex between release-v2 and dev-v2. In release-v2 it works the way you explained and it fine when Cues are timed from 0 within a Subtitle (within a block?). In dev-v2 the timestamps passed look more absolute from the beginning of the stream. Here are some log outputs from getNextEventTimeIndex:

release-v2:

 I/SSASubtitle: time 27254
 I/SSASubtitle: time 14284
 I/SSASubtitle: time 393533
 I/SSASubtitle: time 285
 I/SSASubtitle: time 2029
 I/SSASubtitle: time 35277
 I/SSASubtitle: time 9651
 I/SSASubtitle: time 22227
 I/SSASubtitle: time 2122
 I/SSASubtitle: time 3891
 I/SSASubtitle: time 9881
 I/SSASubtitle: time 9933

dev-v2;

I/SSASubtitle: time 4236599
 I/SSASubtitle: time 9610250
 I/SSASubtitle: time 33021021
 I/SSASubtitle: time 108873334
 I/SSASubtitle: time 119647937
 I/SSASubtitle: time 121220273
 I/SSASubtitle: time 126186667
 I/SSASubtitle: time 127568776
 I/SSASubtitle: time 129281769
 I/SSASubtitle: time 130566168
 I/SSASubtitle: time 133366078
 I/SSASubtitle: time 135076803

ojw28:
> Having slept on it I realise that if I put a into the text the decoder doesn't need to keep the styles. Is that the sort of thing you have in mind?

I'm not sure what this means; I think there might be some words missing from your sentence :)?

> Should separate lines of text in a subtitle go into separate cues?

Not unless they're positioned differently. If it's just a line break then it should be a single cue that contains a line break.

jcable:
Finally worked out what I had done to make my comment meaningless - unfamiliarity with markdown. It should now make sense.

jcable:
I had a look at how webvtt does things and as far as I can tell the style definitions are embedded in the text before the first cue. This seems very similar to the SSA/ASS approach so I don't understand what should be changed next.

library/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java
```diff
@@ -18,13 +18,20 @@
...
 import com.google.android.exoplayer2.text.ttml.TtmlDecoder;
 import com.google.android.exoplayer2.text.tx3g.Tx3gDecoder;
 import com.google.android.exoplayer2.text.webvtt.Mp4WebvttDecoder;
 import com.google.android.exoplayer2.text.webvtt.WebvttDecoder;
 import com.google.android.exoplayer2.util.MimeTypes;
 
+import java.io.UnsupportedEncodingException;
+import java.util.List;
+
+import static android.R.attr.initialKeyguardLayout;
```
ojw28: What are these imports for?

library/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java
```diff
@@ -76,7 +84,13 @@ public SubtitleDecoder createDecoder(Format format) {
         if (clazz == null) {
           throw new IllegalArgumentException("Attempted to create decoder for unsupported format");
         }
-        if (format.sampleMimeType.equals(MimeTypes.APPLICATION_CEA608)
+        if (format.sampleMimeType.equals(MimeTypes.TEXT_SSA)) {
+          byte[] header = format.initializationData.get(1);
+          String dlgfmt = new String(format.initializationData.get(0), "UTF-8");
```
ojw28: Please avoid abbreviations. I don't understand what "dlgfmt" means, and I doubt most other people will either ;).

ojw28:
This change is more likely to be merged if it's split into two pieces. One that adds basic support without styling in an obviously correct way, and a second that adds styling. I still don't think the styling stuff in this change is in line with what we do elsewhere, and I think it'll make things much easier if we can merge the non-controversial bit and discuss that separately.

library/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java
jcable: Something Android Studio did when i wasn't looking ...

changes removed

library/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java
jcable: That name wasn't adding any value. I've removed it.

jcable:
That's fine. Can you suggest the rough shape of the changes you would like to get to the style-less version and I'll make those and then we can take stock.

ojw28:
Sorry for the long silence on this. I sync'd this on top of latest dev-v2, cleaned up some pieces, removed the styling and merged. So what's in `dev-v2` now can serve as the "taking stock" point! Thanks.

PR #2285, opened 2017-01-03
Use vector drawables for the SimpleExoPlayer player control buttons.

colinrtwhite:
This is a simple pull request to update the SimpleExoPlayer player controls to use the default Material Design vector drawable icons provided by Google.

Here's what the UI looks like with the changes:
screenshot_20170102-231434


colinrtwhite:
I signed it!

ojw28:
Wont this crash applications running below API level 21 unless (a) they're using the support library and have configured their gradle file appropriately, or (b) they're using Android Studio? I don't think we want to place these kind of limitations on how the library can be used.

Can we add the vector drawables under `drawable-v21`, but also retain the existing png drawables for backward compatibility?

colinrtwhite:
The Gradle build process should automatically generate PNG drawables if the user's minSDK is below API 21 using the [Vector Asset Studio](https://developer.android.com/studio/write/vector-asset-studio.html) (no configuration required). However, I agree it makes sense to include PNGs of these vectors in case they aren't using Gradle.

I've updated my PR to include PNGs for devices < API 21.

ojw28:
Thanks!

PR #2314, opened 2017-01-11
Fix vector drawable usage on API level 21 or above

svenkonings:
According to the [API guide on providing resources](https://developer.android.com/guide/topics/resources/providing-resources.html) _Screen pixel density (dpi)_ has a higher precedence than _Platform Version (API level)_. This means that currently the png drawables are preferred over the vector drawables, even on API level 21 or above.
In API level 21 a new dpi qualifier called `anydpi` was added. This value takes precedence over the other dpi qualifiers, but will only be used on API level 21 or above. The guide recommends this qualifier for vector drawables.


ojw28:
Is it equivalent to do drawable-anydpi-v21? It might be clearer just because it still keeps the API level in the directory name. Thoughts?

svenkonings:
Yes, that's equivalent. I agree it's clearer to explicitly mention the API level. Currently it's somewhat confusing which one will be used if you haven't read the guide. I will change it.

ojw28:
Thanks. We'll merge this shortly.

Issue #2253, opened 2016-12-22
Enhance the far seek  response of streaming


WeiChungChang:
as #2247 

http://programmingmemojohnchang.blogspot.tw/2016/12/improve-performance-of-short-seek-of_22.html

For 4K movie, if the seek locates within the last download media chunk but has not been in sampleQueue yet. We will download redundant data currently.

It makes the response time of seek slower. Also, it leads to waste of network flow.

Here a request to review the problem is delivered.

Thanks~

ojw28:
A solution for this issue was merged in https://github.com/google/ExoPlayer/pull/2318.

PR #2318, opened 2017-01-12
Improve far seek performance of chunkSampleStream

WeiChungChang:
Rework to make the patch accordingly by @ojw28 kindly suggestions #as #2247.

Issue report: #2253

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/DefaultTrackOutput.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/DefaultTrackOutput.java
@@ -229,10 +229,26 @@ public final class DefaultTrackOutput implements TrackOutput {
    * Attempts to skip to the keyframe before the specified time, if it's present in the buffer.
    *
    * @param timeUs The seek time.
+   * @param skipToLastKey Skip to last key regardless the seek time is out of range .
+   * @return Whether the skip was successful.
+   */
+  public boolean skipToKeyframeBefore(long timeUs, boolean skipToLastKey) {
+    long nextOffset = infoQueue.skipToKeyframeBefore(timeUs, skipToLastKey);
+    if (nextOffset == C.POSITION_UNSET) {
+      return false;
+    }
+    dropDownstreamTo(nextOffset);
+    return true;
+  }
+
+  /**
+   * Attempts to skip to the keyframe before the specified time, if it's present in the buffer.
+   *
+   * @param timeUs The seek time.
    * @return Whether the skip was successful.
    */
   public boolean skipToKeyframeBefore(long timeUs) {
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/google/android/exoplayer2/source/chunk/ChunkSampleStream.java
```diff
@@ -115,14 +115,29 @@ public long getBufferedPositionUs() {
...
    * Seeks to the specified position in microseconds.
    *
    * @param positionUs The seek position in microseconds.
    */
   public void seekToUs(long positionUs) {
     lastSeekPositionUs = positionUs;
     // If we're not pending a reset, see if we can seek within the sample queue.
-    boolean seekInsideBuffer = !isPendingReset() && sampleQueue.skipToKeyframeBefore(positionUs);
+    boolean seekInsideBuffer = !isPendingReset() && 
+        sampleQueue.skipToKeyframeBefore(positionUs, isWithinLastChunk(positionUs));
```
ojw28: I'm not sure you need to be as sophisticated as checking whether the position is within the last chunk. Isn't it sufficient just to pass `positionUs < getNextLoadPositionUs()` as the second argument? I don't think passing true does any harm in cases where `positionUs` is *earlier* than the last chunk.

This allows you to remove your `isWithinLastChunk` and `getCurrentLoadPosition` methods ;).

library/src/main/java/com/google/android/exoplayer2/extractor/DefaultTrackOutput.java
```diff
@@ -229,10 +229,26 @@ public long getLargestQueuedTimestampUs() {
...
+
+  /**
+   * Attempts to skip to the keyframe before the specified time, if it's present in the buffer.
+   *
+   * @param timeUs The seek time.
    * @return Whether the skip was successful.
    */
   public boolean skipToKeyframeBefore(long timeUs) {
-    long nextOffset = infoQueue.skipToKeyframeBefore(timeUs);
+    long nextOffset = infoQueue.skipToKeyframeBefore(timeUs, false);
```
ojw28: You can replace the whole body of this method with just:
`return skipToKeyframeBefore(timeUs, false)`

ojw28 (commented):
Looking good; thanks! I think you can make a few further simplifications, but otherwise think we're good to get this merged. Please take a look.

WeiChungChang:
Dear @ojw28:
Thanks a lot for your suggestion to make it clearer and simpler.
I have made the new patch as 'refine_to_make_flow_simpler':
https://github.com/google/ExoPlayer/pull/2318/commits/2214a4c152fab862c928875f8265e77676d3652b
Thank you!

PR #2326, opened 2017-01-14
Apple Lossless (ALAC) support

devint1:
Support Apple Lossless Audio Codec via the FFmpeg extension

### DIFF
--- a/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
+++ b/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
@@ -19,6 +19,7 @@ import com.google.android.exoplayer2.decoder.DecoderInputBuffer;
 import com.google.android.exoplayer2.decoder.SimpleDecoder;
 import com.google.android.exoplayer2.decoder.SimpleOutputBuffer;
 import com.google.android.exoplayer2.util.MimeTypes;
+import com.google.android.exoplayer2.util.ParsableByteArray;
 import java.nio.ByteBuffer;
 import java.util.List;
 
@@ -88,7 +89,9 @@ import java.util.List;
     if (!hasOutputFormat) {
       channelCount = ffmpegGetChannelCount(nativeContext);
       if ("alac".equals(codecName)) {
-        sampleRate = ByteBuffer.wrap(extraData, extraData.length - 4, 4).getInt();
+        ParsableByteArray parsableExtraData = new ParsableByteArray(extraData);
+        parsableExtraData.setPosition(extraData.length - 4);
+        sampleRate = parsableExtraData.readUnsignedIntToInt();
       } else {
         sampleRate = ffmpegGetSampleRate(nativeContext);
       }
--- a/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
+++ b/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
@@ -87,7 +87,11 @@ import java.util.List;
     }
     if (!hasOutputFormat) {
       channelCount = ffmpegGetChannelCount(nativeContext);
-      sampleRate = ffmpegGetSampleRate(nativeContext);
+      if ("alac".equals(codecName)) {
+        sampleRate = ByteBuffer.wrap(extraData, extraData.length - 4, 4).getInt();
+      } else {
+        sampleRate = ffmpegGetSampleRate(nativeContext);
+      }
       hasOutputFormat = true;
     }
     outputBuffer.data.position(0);
@@ -123,6 +127,7 @@ import java.util.List;
   private static byte[] getExtraData(String mimeType, List<byte[]> initializationData) {
     switch (mimeType) {
       case MimeTypes.AUDIO_AAC:
+      case MimeTypes.AUDIO_ALAC:
       case MimeTypes.AUDIO_OPUS:
         return initializationData.get(0);
       case MimeTypes.AUDIO_VORBIS:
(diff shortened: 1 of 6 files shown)


extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
```diff
@@ -87,7 +87,11 @@ public FfmpegDecoderException decode(DecoderInputBuffer inputBuffer,
     }
     if (!hasOutputFormat) {
       channelCount = ffmpegGetChannelCount(nativeContext);
-      sampleRate = ffmpegGetSampleRate(nativeContext);
+      if ("alac".equals(codecName)) {
```
ojw28: Why does alac need a special case for this :(? Otherwise this change looks fine; thanks!

devint1: The ALAC decoder in FFmpeg does not set `sample_rate` on the native context the same way the Vorbis, Opus, and FLAC decoders do. Because of this, the sample rate needs to be obtained from the last 4 bytes in the ALAC cookie somehow. FFmpeg just [skips them](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/alac.c#L541).

ojw28: Right. Is it a bug that it's not set though, or is is actually intentional? I'd like to see FFMPEG get patched if it's a bug. It's fine to also leave this here as a workaround, but it would be good to be able to ref an FFMPEG issue in a comment to explain why this is here.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
```diff
@@ -87,7 +87,11 @@ public FfmpegDecoderException decode(DecoderInputBuffer inputBuffer,
...
+        sampleRate = ByteBuffer.wrap(extraData, extraData.length - 4, 4).getInt();
```
ojw28: Is this equivalent to:
```
ParsableByteArray parsableExtraData = new ParsableByteArray(extraData);
parsableExtraData.setPosition(extraData.length - 4);
sampleRate = parsableExtraData.readUnsignedIntToInt();
```
It's IMO preferable to use that if so, since it checks the int really is unsigned.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
devint1: I did some digging and couldn't find anything that indicated that it was intentionally not being set. It may be a bug since the behavior is inconsistent with the other decoders. I can submit a ticket to FFmpeg if needed.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
devint1: Yes, these are equivalent. I'll update this accordingly.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
ojw28: Please submit a ticket and ref it in a comment here. That way someone who encounters this code in the future can see why there's a special case, and update the comment to why if it turns out it was intentional, or remove the special case a few years from now if the ticket has been fixed :).

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
```diff
@@ -87,7 +88,17 @@ public FfmpegDecoderException decode(DecoderInputBuffer inputBuffer,
...
-      sampleRate = ffmpegGetSampleRate(nativeContext);
+
+      // Unlike the other decoders, ALAC does not set the sample rate, so we must get it ourselves.
+      // See https://trac.ffmpeg.org/ticket/6096
+      if ("alac".equals(codecName)) {
+        ParsableByteArray parsableExtraData = new ParsableByteArray(extraData);
+        parsableExtraData.setPosition(extraData.length - 4);
+        sampleRate = parsableExtraData.readUnsignedIntToInt();
+      } else {
+        sampleRate = ffmpegGetSampleRate(nativeContext);
```
ojw28: One final question. Does this return 0 currently? If so, and given it looks like the FFMPEG issue is going to be fixed, I was wondering whether we could structure the code so the workaround is only performed if using a version of FFMPEG before the fix lands, like (warning: untested, so please test this!):

```
sampleRate = ffmpegGetSampleRate(nativeContext);
if (sampleRate == 0 && "alac".equals(codecName)) {
  // ALAC decoder did not set the sample rate in earlier versions of FFMPEG.
  // See https://trac.ffmpeg.org/ticket/6096
  ParsableByteArray parsableExtraData = new ParsableByteArray(extraData);
  parsableExtraData.setPosition(extraData.length - 4);
  sampleRate = parsableExtraData.readUnsignedIntToInt();
}
```

ojw28:
All looks good. See one inline comment. Once addressed we'll get this merged.

extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegDecoder.java
devint1: Yes, it currently returns 0. Looks like FFmpeg [zeroes out the context](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/options.c#L95) so should be good. Tested with no issues.

PR #2329, opened 2017-01-14
Opus: use fixed max size in Opus decoding

supercairos:
As per [documentation](http://opus-codec.org/docs/opus_api-1.1.3/group__opus__multistream.html#gaa4b89541efe01970cf52e4a336db3ad0) of the opus_multistream_decode: 
> **_frame_size_**: int: The number of samples per channel of available space in pcm. **If this is less than the maximum packet duration (120 ms; 5760 for 48kHz), this function will not be capable of decoding some packets**. In the case of PLC (data==NULL) or FEC (decode_fec=1), then frame_size needs to be exactly the duration of audio that is missing, otherwise the decoder will not be in the optimal state to decode the next incoming packet. For the PLC and FEC cases, frame_size must be a multiple of 2.5 ms. 

Sample reference code 
[https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libopusdec.c#L142](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libopusdec.c#L142)
[https://android.googlesource.com/platform/frameworks/av/+/master/media/libstagefright/codecs/opus/dec/SoftOpus.cpp#461](https://android.googlesource.com/platform/frameworks/av/+/master/media/libstagefright/codecs/opus/dec/SoftOpus.cpp#461)

Original pull request : #2207

### DIFF
--- a/extensions/opus/src/main/jni/opus_jni.cc
+++ b/extensions/opus/src/main/jni/opus_jni.cc
@@ -59,6 +59,7 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved) {
 }
 
 static const int kBytesPerSample = 2;  // opus fixed point uses 16 bit samples.
+static const int kMaxOpusOutputPacketSizeSamples = 960 * 6;// Maximum packet size used in Xiph's opusdec.
 static int channelCount;
 static int errorCode;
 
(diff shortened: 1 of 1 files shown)


ojw28:
@vigneshvg - Please comment again. Thanks.

extensions/opus/src/main/jni/opus_jni.cc
```diff
@@ -59,6 +59,7 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved) {
 }
 
 static const int kBytesPerSample = 2;  // opus fixed point uses 16 bit samples.
+static const int kMaxOpusOutputPacketSizeSamples = 960 * 6;// Maximum packet size used in Xiph's opusdec.
```
vigneshvg, line 62: nit: two spaces after the semi-colon. actually the comment can go since there's enough justification for this line in the commit message.

vigneshvg (changes requested):

supercairos:
Hi @vigneshvg, I will fix my code tomorrow or Wednesday.

ojw28:
@supercairos - Are you planning on updating this change?

Issue #2361, opened 2017-01-23
Support CENC ClearKey


yqritc:
Hi, I'm trying to figure out how to play dash content using clearkey.
It seems mediadrm in Android Framework support clearkey.
https://android.googlesource.com/platform/frameworks/av/+/master/drm/mediadrm/plugins/clearkey

I've found widvine and plyaready samples in demo application, but I couldn't find clearykey one.
ExoPlayer does not support clearkey?
If not, any plan to support in the future?

thanks

yqritc:
The implementation of clear key seems to be merged!! thanks!!

PR #2372, opened 2017-01-25
Added support for CENC ClearKey

wasabeef:
Support CENC ClearKey
https://w3c.github.io/encrypted-media/format-registry/initdata/cenc.html

See issue https://github.com/google/ExoPlayer/issues/2361

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/drm/DefaultDrmSessionManager.java
+++ b/library/src/main/java/com/google/android/exoplayer2/drm/DefaultDrmSessionManager.java
@@ -35,6 +35,7 @@ import com.google.android.exoplayer2.drm.ExoMediaDrm.OnEventListener;
 import com.google.android.exoplayer2.drm.ExoMediaDrm.ProvisionRequest;
 import com.google.android.exoplayer2.extractor.mp4.PsshAtomUtil;
 import com.google.android.exoplayer2.util.Assertions;
+import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.Util;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
@@ -101,6 +102,11 @@ public class DefaultDrmSessionManager<T extends ExoMediaCrypto> implements DrmSe
   public static final int MODE_DOWNLOAD = 2;
   /** Releases an existing offline license. */
   public static final int MODE_RELEASE = 3;
+
+  /**
+   * The format to use when ClearKey encryption.
+   */
+  private static final String CENC_INIT_DATA_FORMAT = "cenc";
 
   private static final String TAG = "OfflineDrmSessionMngr";
 
@@ -335,6 +341,21 @@ public class DefaultDrmSessionManager<T extends ExoMediaCrypto> implements DrmSe
           // Extraction failed. schemeData isn't a Widevine PSSH atom, so leave it unchanged.
         } else {
           schemeInitData = psshData;
+        }
+      }
+      if (C.CENC_UUID.equals(uuid)) {
+        // If "video/mp4" and "audio/mp4" are not supported as CENC schema, change it to "cenc".
+        // Before 7.1.x in API 25, "video/mp4" and "audio/mp4" are not supported.
+        if (MimeTypes.VIDEO_MP4.equals(schemeMimeType) || MimeTypes.AUDIO_MP4.equals(
+            schemeMimeType)) {
+          if (Util.SDK_INT >= 26) {
+            // Nothing to do.
+          } else if (Util.SDK_INT == 25 && !MediaDrm.isCryptoSchemeSupported(uuid,
+              schemeMimeType)) {
+            schemeMimeType = CENC_INIT_DATA_FORMAT;
+          } else if (Util.SDK_INT <= 24) {
+            schemeMimeType = CENC_INIT_DATA_FORMAT;
+          }
         }
       }
     }
--- a/demo/src/main/assets/media.exolist.json
+++ b/demo/src/main/assets/media.exolist.json
@@ -278,6 +278,18 @@
     ]
   },
   {
+    "name": "ClearKey DASH",
+    "samples": [
+      {
+        "name": "Big Buck Bunny (CENC ClearKey)",
+        "uri": "http://html5.cablelabs.com:8100/cenc/ck/dash.mpd",
+        "extension": "mpd",
+        "drm_scheme": "cenc",
+        "drm_license_url": "https://wasabeef.jp/demos/cenc-ck-dash.json"
+      }
+    ]
+  },
+  {
     "name": "SmoothStreaming",
     "samples": [
       {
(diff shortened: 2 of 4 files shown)


library/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1192,7 +1192,7 @@ private static DrmInitData getDrmInitDataFromAtoms(List<Atom.LeafAtom> leafChild
         if (uuid == null) {
           Log.w(TAG, "Skipped pssh atom (failed to extract uuid)");
         } else {
-          schemeDatas.add(new SchemeData(uuid, MimeTypes.VIDEO_MP4, psshData));
+          schemeDatas.add(new SchemeData(uuid, MimeTypes.CENC, psshData));
```
ojw28: This doesn't appear to be necessary? I have a hunch it might break Widevine/PlayReady on some earlier API levels also, although I didn't confirm this. If it works without, can we just leave it how it was and also remove the CENC constant from MimeTypes.java?

ojw28 (commented):
Looks good aside from one comment (see inline). Thanks!

library/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
wasabeef: Thank for your replay.

It accept "video/mp4" only on Android 7.1.1  or higher.
https://android.googlesource.com/platform/frameworks/av/+/01a977aa75c296c5467024a7beac50cbd4ed0335%5E%21/#F4
https://android.googlesource.com/platform/frameworks/av/+/android-5.0.0_r1/drm/mediadrm/plugins/clearkey/InitDataParser.cpp#48

Should I change the definition class?

library/src/main/java/com/google/android/exoplayer2/util/MimeTypes.java
```diff
@@ -81,6 +81,8 @@
   public static final String APPLICATION_CAMERA_MOTION = BASE_TYPE_APPLICATION + "/x-camera-motion";
   public static final String APPLICATION_EMSG = BASE_TYPE_APPLICATION + "/x-emsg";
 
+  public static final String CENC = "cenc";
+
```
wasabeef: Should I create a new class?

ojw28:
Thanks for clarification in your previous response (although it seems to have disappeared somehow). I don't think you should modify `MimeTypes` as part of this change. I'm not convinced it's a good idea to change `FragmentedMp4Extractor` either.

It feels to me that you're trying to work around a bug in the underlying CDM implementation on older versions of Android. The correct place to do that is probably `DefaultDrmSessionManager.acquireSession`. There's already a workaround in that method for the Widevine CDM, so you can add a second workaround there that changes `schemeMimeType` if all of the following are true:

1. `Util.SDK_INT < 23`
1. `C.CENC_UUID.equals(uuid)`
1. `MimeTypes.VIDEO_MP4.equals(schemeMimeType)`

That way the workaround will be applied correctly regardless of where the media is coming from (e.g. if another extractor implementation is used).

wasabeef:
Sorry for the late reply. 

Oh, I see. Why is SDK_INT 23?
It seems the `video/mp4` has been added since android-7.1.1_r1.

**7.1.0_r7**
https://android.googlesource.com/platform/frameworks/av/+/android-7.1.0_r7/drm/mediadrm/plugins/clearkey/InitDataParser.cpp#48

**android-7.1.1_r1**
https://android.googlesource.com/platform/frameworks/av/+/android-7.1.1_r1/drm/mediadrm/plugins/clearkey/InitDataParser.cpp#49

library/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
abouda: You should also add this for the case where the pssh is parsed from the manifest. That is in `DashManifestParser#parseContentProtection`.
Otherwise if you follow @ojw28 ´s tip about doing this in the DrmSessionManager, that would cover both cases.

ojw28:
`Util.SDK_INT < 26` then, since that would suggest not all devices with `SDK_INT == 25` will include the change, but that all devices with `SDK_INT == 26` or greater will?

wasabeef:
It's difficult to judge with `SDK_INT == 25`, In this case, I'd like to use `MediaDrm#isCryptoSchemeSupported`.

What do you think?
```
if (C.CENC_UUID.equals(uuid) && MimeTypes.VIDEO_MP4.equals(schemeMimeType)) {
  // If "video/mp4" is not supported as CENC schema, change it to "cenc".
  if (Util.SDK_INT < 19 || !MediaDrm.isCryptoSchemeSupported(uuid, MimeTypes.VIDEO_MP4)) {
    schemeMimeType = CENC_INIT_DATA_FORMAT;
  }
}
```

library/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
wasabeef: That's right.

abouda:
@wasabeef Why checking `Util.SDK_INT < 19`? Afaik CENC scheme is not supported on android prior to API 21.
Anyway you will get a `UnsupportedDrmException` as soon as you try to instantiate the `DrmSessionManager`. So I guess you can remove the version check here.

library/src/main/java/com/google/android/exoplayer2/drm/DefaultDrmSessionManager.java
```diff
@@ -337,6 +343,12 @@ public void setMode(@Mode int mode, byte[] offlineLicenseKeySetId) {
           schemeInitData = psshData;
         }
       }
+      if (C.CENC_UUID.equals(uuid) && MimeTypes.VIDEO_MP4.equals(schemeMimeType)) {
```
ojw28: You should handle audio/mp4 as well for completeness, I think. I'd also rather there were a capping SDK level after which the workaround is not applied. We shouldn't be applying workarounds for platform issues to future platform releases; we should be making sure the problems are fixed in the platform (which it appears has been done).

wasabeef: @ojw28 

I'm sorry for the late reply. I updated again.

ojw28:
Thanks! I simplified this somewhat in https://github.com/google/ExoPlayer/commit/3bb08e58f68ee1cc5a288b601c5fec3068cd83da. I think it's equivalent, except that it always replaces schemeMimeType to "cenc" on API level 25. I think this is fine - my main concern was just not applying the workaround indefinitely, which is achieved by capping to not apply on API levels above 25. Let me know if I accidentally broke something!

ojw28:
Also corrected some naming in https://github.com/google/ExoPlayer/commit/5fe5076c8686a09608bbef8797e1c16fa8d80fd8.

wasabeef:
No problem. 
I'm sorry that I didn't understand your intentions correctly. 🙇

wasabeef:
Don't you like [MediaDrm.isCryptoSchemeSupported](https://goo.gl/bMglsU) ?

ojw28:
I don't dislike it. It just seems unnecessarily complicated to use it just on one API level.

wasabeef:
Right, got it. Thanks!

ojw28:
@wasabeef - The sample asset you added in this request appears to no longer work. Could you update or remove it? Thanks.

PR #2379, opened 2017-01-27
Added standard maven deploy plugin to allow deploy artifact to local repository

tprochazka:
It is not fully default maven deploy plugin, but its [modification](https://github.com/dcendents/android-maven-gradle-plugin) to better compatibility with Android AAR.

By "gradlew :library:install" it is possible to publish any version of ExoPlayer library to local repo and then use it project just by adding mavenLocal() to the repositories section.

Maybe it is not ideal way how to implement it, because you are creating custom variables like releaseVersion in the project.ext space, instead of using directly existing project variable "version", but I don't wanted to do bigger changes. This works good.

Just normally in the dev branch should be -SNAPSHOT in the version to distinguish dev build from the released version in the bintray. -SNAPSHOT should be removed just for the tagged commit which will be deployed to  bintray.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:2.2.1'
         classpath 'com.novoda:bintray-release:0.3.4'
+        classpath 'com.github.dcendents:android-maven-gradle-plugin:1.4.1'
     }
 }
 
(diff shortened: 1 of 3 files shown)


tprochazka:
I signed it!

ojw28:
I'm somewhat unclear what the aim is here, or exactly what the desired workflow is. If you want to depend on a local snapshot, why not just build the .aar and then do "File -> New -> New Module -> Import .JAR/.AAR Package" in Android Studio? Please clarify. Thanks!

ojw28:
We've made it easy to depend directly on a local checkout of ExoPlayer from another Android Studio project. See the "Locally" section of our [README](https://github.com/google/ExoPlayer/blob/dev-v2/README.md)) for details. So I don't think is needed any more.

PR #2399, opened 2017-01-31
specifying default extractors explicitly

emakar:
so that proguard is possible to shrink unused Extractors

similar to v2 change: https://github.com/google/ExoPlayer/commit/f9f95d638bbe23f7c9f584e7500a5d1f8c7c0feb


ojw28:
We're only accepting bug fixes into V1 at this point. Move to V2 for shiny new things ;).

emakar:
I'd be glad to move to V2, however we are stuck on dex limit (and cannot afford multidex for various reasons). That said, I've checked V2 methods output and even with unused extractors being shrinked it's still far above V1's. With mp3, mp4 extractors V1 (with this change) gives 628 methods, while V2 - about 1K.
_________
I understand this could hardly be considered as a bug fix, however the change fixes 'outside single dex issue' for me and makes library a little bit more comfy, leaving space for 'soft' migration to V2 in the future

ojw28:
Not being able to use multidex for various (unspecified) reasons doesn't sound like an ExoPlayer problem. This is something you'll likely need to resolve locally. Sorry.

emakar:
That's sad, but sounds reasonable :)
I'll try to look at V2 closer, there should possibly be more space for similar changes. Thanks

PR #2444, opened 2017-02-12
Updated Gradle, Compile and Build Tools Version

twisstosin:
Updated Gradle, Compile and Build Tools Version.

Max Compatibility With Android API 25.


library/src/main/res/layout/exo_playback_control_view.xml
```diff
@@ -58,7 +58,7 @@
     <TextView android:id="@id/exo_position"
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
-      android:textSize="14sp"
+      android:textSize="13sp"
```
ojw28: Why?

twisstosin: Appears too big on device with overridden fonts.

ojw28: What are overridden fonts? How does one override fonts on a device?

twisstosin: Non-Stock android ROMS,
and apps with overridden fonts

For instance
```
FontsOverride.setDefaultFont(this, "DEFAULT", "fonts/proximanova_regular.ttf");
```
overrides the default font with custom font in fonts asset folder.

Where FontsOverride is
```
public final class FontsOverride {

    public static void setDefaultFont(Context context,
                                      String staticTypefaceFieldName, String fontAssetName) {
        final Typeface regular = Typeface.createFromAsset(context.getAssets(),
                fontAssetName);
        replaceFont(staticTypefaceFieldName, regular);
    }

    protected static void replaceFont(String staticTypefaceFieldName,
                                      final Typeface newTypeface) {
        if (Build.VERSION.SDK_INT >= 21) {
            Map<String, Typeface> newMap = new HashMap<String, Typeface>();
            newMap.put("sans-serif", newTypeface);
            try {
                final Field staticField = Typeface.class
                        .getDeclaredField("sSystemFontMap");
                staticField.setAccessible(true);
                staticField.set(null, newMap);
            } catch (NoSuchFieldException e) {
                e.printStackTrace();
            } catch (IllegalAccessException e) {
                e.printStackTrace();
            }
        } else {
            try {
                final Field staticField = Typeface.class
                        .getDeclaredField(staticTypefaceFieldName);
                staticField.setAccessible(true);
                staticField.set(null, newTypeface);
            } catch (NoSuchFieldException e) {
                e.printStackTrace();
            } catch (IllegalAccessException e) {
                e.printStackTrace();
            }
        }
    }
}
```

ojw28: I don't think we should be changing our layout for this case. Making the font the right size for custom ROMs at the cost of making it too small for the 99.9% case is not acceptable. I'd hypothesize that fonts that cause this issue are not sized correctly, although this is just a guess.

As an aside, the sizes used here are just taken from the Android platform playback controls view, so any app using the platform provided controls presumably encounters the same issue when using fonts in this way.

Happy to merge the other parts of this pull request, but please revert the modified font size part.

ojw28: Note also that you can override the layout in your own application code if you're programmatically replacing the font wish to do so. See [here](https://medium.com/google-exoplayer/customizing-exoplayers-ui-components-728cf55ee07a).

twisstosin: Alrighty cool, will revert the modified font size.

PR #2464, opened 2017-02-16
Make DrmSessionException constructor public to enable creating custom DrmSessionManager implementations.

jschamburger:
I noticed that in the current dev-v2 branch it is not possible to implement a custom `DrmSessionManager` which implements the `DrmSession` interface in a reasonable way.
The problem is that DrmSessionException has a package-local constructor, which means that the implementation of `DrmSession.onError` can't return anything other than `null` (unless, of course, the implementation would be in the same package, like `DefaultDrmSessionManager`).


ojw28:
Heh, this is an oversight. Thanks! We'll merge shortly.

jschamburger:
Anytime. :-) 
I stumbled upon this when I was trying your fix for https://github.com/google/ExoPlayer/issues/1865 which caused trouble with my DrmSessionManager based on the v2.1.1 StreamingDrmSessionManager. Btw, the fix is working great.

Issue #2445, opened 2017-02-13
Rethink handling of "unseekable" mp3 files


rustyshelf:
Using ExoPlayer v2.2.0 and trying to play this file: [sample.mp3](http://omnystudio.com/d/clips/3ea926be-9be7-482a-8ac3-a44f016e43d1/fb1f46f6-b3d6-4068-8ab5-a44f016e5066/f58994ce-4213-41a4-af2e-a714006ce1b2/audio.mp3?utm_source=Podcast&in_playlist=7f09e143-5892-47cc-a7e2-a4e4015c8eee) it reports as un-seekable. However this same file seeks fine in the Android MediaPlayer and various other players as well.

Doing an afinfo on the macOS command line doesn't seem to turn up anything interesting about the file:
```
File:           Feminist Fight Club.mp3
File type ID:   MPG3
Num Tracks:     1
----
Data format:     2 ch,  44100 Hz, '.mp3' (0x00000000) 0 bits/channel, 0 bytes/packet, 1152 frames/packet, 0 bytes/frame
                no channel layout.
estimated duration: 3887.560687 sec
audio bytes: 62201388
audio packets: 148821
bit rate: 128000 bits per second
packet size upper bound: 1052
maximum packet size: 417
audio data file offset: 417
optimized
----
```

This seems like a bug, but I'm unsure about where to hunt for whats going wrong?

Edit: Just to rule out our implementation I tried it in the sample project as well, same result, it disables the seek bar and buttons.

AquilesCanta:
Have you seen #1299?

rustyshelf:
@AquilesCanta I have now (I did try and search for similar issues but didn't find any), my vote @ojw28 would be to assume constant bit rate and allow seeking, or at the very least have this as a configuration parameter. We're working on a new version of our app at the moment and not being able to seek (or even resume from the middle of) files like this would be a deal breaker :(

ojw28:
Are you in control of the media being played? If not then where is it typically coming from?

We're unlikely to add support for this case. As per #1299 it's unlikely MediaPlayer is doing a particularly good job of allowing seeking (there's likely a significant trade-off of one kind or another involved).

rustyshelf:
[...]

Issue #1299, opened 2016-02-23
seekTo restarts audio stream


victordigio:
In some audios, when I call to seekTo, playback is restarted, and the audio start from the beginning. I think there is something wrong with the audio, but with MediaPlayer works fine.

Audio link: http://www.ivoox.com/primer-palo-19-02-16-al-infierno-los_mn_10502707_api_1.mp3
ExoPlayer version: 1.5.0
Device and android version: any

ojw28:
The file contains a Xing header that's missing a table of content. If there were a table of content, the file would be seekable.

I doubt MediaPlayer is doing a particularly good job of allowing seeking. It's most likely indexing the entire file, which will significantly increase startup latency and will not scale properly to large files. In ExoPlayer we've opted to only support seeking in files that actually contain the information necessary to seek in an efficient and scalable way.

@andrewlewis - As an aside, I briefly wondered whether we should support seeking in this kind of file using a constant bitrate assumption. I gave it a try but seeking was inaccurate by over 40s in some cases for the provided sample, so I'm pretty sure this isn't a good idea. Let me know if you disagree though :).

victordigio:
Well, I guess if audio is not seekable, at least it should keep current position instead of set to 0. It would be great if could work like MediaPlayer on this cases.

ojw28:
We won't be making it work like MediaPlayer, for the reasons specified above.

We should allow an application to query whether the media is seekable, so that it can avoid calling seekTo if the media is not seekable except in the case where it wishes to restart the playback. I've filed #1300 to track this.

zgxzgxzg:
How embodied in the code inside

PaulWoitaschek:
I have a similar case but I'm not sure if this is not acutally a bug @ojw28 
I'll send the file by mail.

ExoPlayer 2.1.1

 > ffprobe -v quiet -print_format json -show_format -show_streams Chapter\ 03.mp3 

```json
{
    "streams": [
        {
            "index": 0,
            "codec_name": "mp3",
            "codec_long_name": "MP3 (MPEG audio layer 3)",
            "codec_type": "audio",
            "codec_time_base": "1/24000",
            "codec_tag_string": "[0][0][0][0]",
            "codec_tag": "0x0000",
            "sample_fmt": "s16p",
            "sample_rate": "24000",
            "channels": 2,
            "channel_layout": "stereo",
            "bits_per_sample": 0,
            "r_frame_rate": "0/0",
[...]

PR #2468, opened 2017-02-17
Add Flags to MP3 Extractor to force CBR Seeker

danysantiago:
Added flags to the mp3 extractor to control behavior of the extractor.
Added FLAG_ENABLE_CONSTANT_BITRATE_SEEKING to let the extractor know that the CBR seeker is desired regardless of the mp3 being VBR, with or without a Xing header.

Relates to:
https://github.com/google/ExoPlayer/issues/1299
https://github.com/google/ExoPlayer/issues/2445

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
+++ b/library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
@@ -31,8 +31,13 @@ import com.google.android.exoplayer2.metadata.Metadata;
 import com.google.android.exoplayer2.metadata.id3.Id3Decoder;
 import com.google.android.exoplayer2.util.ParsableByteArray;
 import com.google.android.exoplayer2.util.Util;
+
+import android.support.annotation.IntDef;
+
 import java.io.EOFException;
 import java.io.IOException;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
 /**
  * Extracts data from an MP3 file.
@@ -50,6 +55,19 @@ public final class Mp3Extractor implements Extractor {
     }
 
   };
+
+  /**
+   * Flags controlling the behavior of the extractor.
+   */
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef(flag = true, value = {FLAG_ENABLE_CONSTANT_BITRATE_SEEKING})
+  public @interface Flags {}
+
+  /**
+   * Flag to force enable seeking using a constant bitrate assumption in cases where seeking would
+   * otherwise not be possible.
+   */
+  public static final int FLAG_ENABLE_CONSTANT_BITRATE_SEEKING = 1;
 
   /**
    * The maximum number of bytes to search when synchronizing, before giving up.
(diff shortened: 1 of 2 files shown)


danysantiago:
I signed it!

danysantiago:
Please let me know what would be the appropriate way to write tests for these changes.
Not sure how `Mp3ExtractorTest` works and there didn't seem to be simple hooks to test that part of the seeker creation.

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
```diff
@@ -93,16 +114,27 @@
    * Constructs a new {@link Mp3Extractor}.
    */
   public Mp3Extractor() {
-    this(C.TIME_UNSET);
+    this(0, C.TIME_UNSET);
```
ojw28: Just do "this(0);"?

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
```diff
@@ -329,7 +361,8 @@ private Seeker setupSeeker(ExtractorInput input) throws IOException, Interrupted
       frame.setPosition(xingBase);
       headerData = frame.readInt();
     }
-    if (headerData == XING_HEADER || headerData == INFO_HEADER) {
+    if ((flags & FLAG_ENABLE_CONSTANT_BITRATE_SEEKING) == 0
```
ojw28: I don't think it's right that you're preventing the creation of the XingSeeker in all cases when the flag is set. If there's a Xing header that *does* have a table of contents, it's going to be better to use the XingSeeker than to force constant bitrate seeking...

It seems more appropriate to leave this code unchanged and instead modify the
```
if(seeker == null) {
  ...
}
```
block at the bottom of this method to be more like:
```
if (seeker == null || (!seeker.isSeekable()
    && (flags & FLAG_ENABLE_CONSTANT_BITRATE_SEEKING) != 0))) {
  ...
}
```

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
```diff
@@ -52,6 +57,19 @@
...
 
   /**
+   * Flags controlling the behavior of the extractor.
+   */
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef(flag = true, value = {FLAG_ENABLE_CONSTANT_BITRATE_SEEKING})
+  public @interface Flags {}
+
+  /**
+   * Flag to enable using the {@link ConstantBitrateSeeker} for cases such as Xing Header present
```
ojw28: Flag to force enable seeking using a constant bitrate assumption in cases where seeking would otherwise not be possible.

Note: You shouldn't @link to ConstantBitrateSeeker because it's package private, so there's no page generated to link to when the Javadoc is actually generated (we only generate Javadoc for public and protected visibilities).

danysantiago: Done, thanks for the explanation about which jdoc is generated.

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
danysantiago: Yup, updated.

library/src/main/java/com/google/android/exoplayer2/extractor/mp3/Mp3Extractor.java
danysantiago: That makes sense, thanks, updated.

danysantiago:
Thanks!

PR #2475, opened 2017-02-19
Added proper dpad seekbar positioning

bigexxx:


ojw28:
Could you clarify what is not "proper" about dpad seeking if you just make the SeekBar focusable?

bigexxx:
yep, just make seekbar focusable and workable

ojw28:
Are you saying just making it focusable is sufficient, or are you saying that this pull request is required? If so, what does this pull request achieve that just making the SeekBar focusable (and nothing else) does not?

bigexxx:
no, i additionally added seekEndedAction Runnable and setOnKeyListener on progressBar for handling dpad keycodes, after clicking right or left, dpad is seeked and posted delayed action to commit changes to player

ojw28:
@andrewlewis Is working on making some other changes to the seek bar, which likely mean this change wont be necessary.

ojw28:
The dev branch has a new `TimeBar` class now. Is this change still needed, or does that behave in the desired way? If it is still needed, the implementation should probably be moved from `PlaybackControlView` into `DefaultTimeBar`?

andrewlewis:
The `DefaultTimeBar` we'll push shortly (probably today) implements similar timeout behavior to this PR: after pressing the left/right directional buttons the seek will occur after a timeout, unless the user presses the center/enter button which causes the seek to happen immediately.

PR #2503, opened 2017-02-26
Rtmp client

b95505017:
RTMP DataSource extension


demo/build.gradle
```diff
@@ -45,4 +45,5 @@ dependencies {
     withExtensionsCompile project(path: ':extension-flac')
     withExtensionsCompile project(path: ':extension-opus')
     withExtensionsCompile project(path: ':extension-vp9')
+    compile 'net.butterflytv.utils:rtmp-client:0.2.6.1'
```
ojw28: We don't want this dependency from the demo app. For all other cases where we've taken an external dependency, we've done so by creating an extension (see the six existing extensions [here](https://github.com/google/ExoPlayer/tree/release-v2/extensions)).

If you restructure this change so that `RtmpDataSource` is in a proper extension then we'll consider accepting a pull request. We probably don't want any samples in the demo app by default.

demo/src/main/java/com/google/android/exoplayer2/demo/player/RtmpDataSource.java
```diff
@@ -0,0 +1,55 @@
+package com.google.android.exoplayer2.demo.player;
```
ojw28: You need to put the standard header present in other ExoPlayer source files at the top here

demo/src/main/java/com/google/android/exoplayer2/demo/player/RtmpDataSource.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.android.exoplayer2.C;
+import com.google.android.exoplayer2.upstream.DataSource;
+import com.google.android.exoplayer2.upstream.DataSpec;
+
+import net.butterflytv.rtmp_client.RtmpClient;
+
+import java.io.IOException;
+
+/**
+ * Created by faraklit on 08.01.2016.
```
ojw28: Needs proper Javadoc

demo/src/main/java/com/google/android/exoplayer2/demo/player/RtmpDataSource.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class RtmpDataSource implements DataSource {
+
+
+    private final RtmpClient rtmpClient;
```
ojw28: Please use 2 character indentation as in other ExoPlayer source files

demo/src/main/java/com/google/android/exoplayer2/demo/player/RtmpDataSource.java
```diff
@@ -0,0 +1,55 @@
...
+    private Uri uri;
+
+    private RtmpDataSource() {
+        rtmpClient = new RtmpClient();
+    }
+
+    @Override
+    public Uri getUri() {
+        return uri;
```
ojw28: Hm. This is always null? I think it needs setting in `open` and clearing in `close`?

demo/src/main/java/com/google/android/exoplayer2/demo/player/RtmpDataSource.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.android.exoplayer2.upstream.DataSpec;
+
+import net.butterflytv.rtmp_client.RtmpClient;
+
+import java.io.IOException;
+
+/**
+ * Created by faraklit on 08.01.2016.
+ */
+public class RtmpDataSource implements DataSource {
```
ojw28: Please make final (`public final class`)

library/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
```diff
@@ -51,6 +52,7 @@
   private static final int STATE_SKIPPING_TO_TAG_HEADER = 2;
   private static final int STATE_READING_TAG_HEADER = 3;
   private static final int STATE_READING_TAG_DATA = 4;
+  private static final int STATE_SEEK_FLV = 5;
```
ojw28: Are these FLV changes (the ones to this file and to ScriptTagPayloadReader) actually related to adding RTMP support? If not they should be in a separate pull request.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
```diff
@@ -288,7 +303,24 @@ public long getDurationUs() {
 
   @Override
   public long getPosition(long timeUs) {
+
+
+    double timeMs = timeUs / 1000000;
```
ojw28: There are only 1000 microseconds in a millisecond, so either the conversion or the variable naming is wrong here? I suspect the naming (i.e. timeMs should be timeSecs)?

library/src/main/java/com/google/android/exoplayer2/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -79,18 +86,33 @@ protected void parsePayload(ParsableByteArray data, long timeUs) throws ParserEx
...
-    }
-    // Set the duration to the value contained in the metadata, if present.
-    Map<String, Object> metadata = readAmfEcmaArray(data);
-    if (metadata.containsKey(KEY_DURATION)) {
-      double durationSeconds = (double) metadata.get(KEY_DURATION);
-      if (durationSeconds > 0.0) {
-        durationUs = (long) (durationSeconds * C.MICROS_PER_SECOND);
+//    if (type != AMF_TYPE_ECMA_ARRAY) {
+//      //   Should never happen.
+//      //aom: actually it can happen for live streams
```
ojw28: Please delete the code block if that's the case

library/src/main/java/com/google/android/exoplayer2/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -215,4 +244,19 @@ private static Object readAmfData(ParsableByteArray data, int type) {
     }
   }
 
+  public void setKeyFrameTimes(List<Double> keyFrameTimes) {
```
ojw28: I'm unclear why you need these setter methods. Why can't you just assign the variables directly?

library/src/main/java/com/google/android/exoplayer2/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -172,11 +194,18 @@ private static String readAmfString(ParsableByteArray data) {
    * @return The value read from the buffer.
    */
   private static HashMap<String, Object> readAmfEcmaArray(ParsableByteArray data) {
-    int count = data.readUnsignedIntToInt();
-    HashMap<String, Object> array = new HashMap<>(count);
-    for (int i = 0; i < count; i++) {
+    //ecma array is not the strict array, so it shouldn't be trusted number of item count
```
ojw28: Rather than this comment, it would probably be clearer to do:
int approximateCount = data.readUnsignedIntToInt();

It's probably still worth passing the value to the HashMap constructor as an initial size as well, as was previously the case.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -215,4 +244,19 @@ private static Object readAmfData(ParsableByteArray data, int type) {
...
 
+  public void setKeyFrameTimes(List<Double> keyFrameTimes) {
+    this.keyFrameTimes = keyFrameTimes;
+  }
+
+  public void setKeyFrameFilePositions(List<Double> keyFrameFilePositions) {
+    this.keyFrameFilePositions = keyFrameFilePositions;
+  }
+
+  public List<Double> getKeyFrameFilePositions() {
```
ojw28: It would be nice if these getters were moved to be next to the getDurationUs.

library/src/main/java/com/google/android/exoplayer2/extractor/flv/ScriptTagPayloadReader.java
```diff
@@ -79,18 +86,33 @@ protected void parsePayload(ParsableByteArray data, long timeUs) throws ParserEx
...
+      Map<String, Object> metadata = readAmfEcmaArray(data);
+      if (metadata.containsKey(KEY_DURATION)) {
+        double durationSeconds = (double) metadata.get(KEY_DURATION);
+        if (durationSeconds > 0.0) {
+          durationUs = (long) (durationSeconds * C.MICROS_PER_SECOND);
+        }
+      }
+      if (metadata.containsKey(KEY_FRAMES)) {
+        HashMap<String, List<Double>> keyFrames = (HashMap<String, List<Double>>) metadata.get(KEY_FRAMES);
+        setKeyFrameTimes(keyFrames.get(KEY_FRAME_TIMES));
```
ojw28: It might be nice to convert the values into long microsecond and byte offset arrays here, rather than keeping them as doubles. That way the public getter methods, for time at least, are all in the same unit (i.e. microseconds).

ojw28 (commented):
Thanks for the pull request. I think it needs splitting in two (one request for the FLV changes, and another for the RTMP feature). Please also see inline comments. Thanks!

ojw28:
@b95505017 - Do you have any plans to split this request, address the comments and sign the CLA? Else I'll go ahead and close this. There's definitely nothing we can do with it without the CLA being signed, and I'd rather not leave it open indefinitely.

b95505017:
I've signed the CLA already, but @mekya. He is the main author about those commits, I just made them rebase v2.

mekya:
I have signed CLA and I am going to try to fix the issues in your( @ojw28 ) comments in a week. I am sorry I am busy in these days. Btw, Thank you @b95505017

b95505017:
@ojw28 I refactor everything, please take a look again thanks!

ojw28:
This is looking much better now; thanks! I have some concerns about the underlying library being used. I've filed a couple of issues on that library directly. If you're able to help with, or comment on, those issues directly, that might help speed things up! The issues are:

https://github.com/ant-media/LibRtmp-Client-for-Android/issues/32
https://github.com/ant-media/LibRtmp-Client-for-Android/issues/31
https://github.com/ant-media/LibRtmp-Client-for-Android/issues/30

b95505017:
Actually the best choice of rtmp client is using the pure java Socket version instead of librtmp I think, but the concept will be based on [SimpleRtmp](https://github.com/faucamp/SimpleRtmp) which may hit the copyright. The pure java Socket version will be more extendable and have more control about the rtmp behavior including [redirect behavior](https://github.com/ossrs/srs/issues/369). 

Not sure what is the best choice for ExoPlayer project.

ojw28:
If there's a pure Java RTMP library available somewhere under a sensible license, that's distributed via jcenter, then it's fine to depend on that instead. I'd rather not implement the whole thing inside the extension if possible; having a dependency is preferable. If the whole thing is implemented inside the extension then it needs to be original work, not copy/paste or derivative of some other project, as per the CLA.

ojw28:
I got this working reasonably well, but I found another issue (https://github.com/ant-media/LibRtmp-Client-for-Android/issues/33). I think we'll probably need the issues we've raised to be fixed in LibRtmp before we're able to merge this. Alternatively, we could depend on a different project if there's a better alternative.

b95505017:
That's weird, they claim have all ABIs.
>It is probably the smallest(~60KB) rtmp client for android. It calls librtmp functions over JNI interface. With all cpu architectures(arm, arm7, arm8, x86, x86-64, mips) its size is getting about 300KB

mekya:
@b95505017 @ojw28 
This librtmp java wrapper library is developed with a specific project that does not use any 64 bit library so it works on both 32-bit and 64-bit devices. In order to support all cpu arch., we just need to add only a configuration parameter when building. 

In addition, before I develop this library, I googled a RTMP Java library. I found some libraries however they were complex and their sizes are a few MBs.

ojw28:
@mekya - Thanks for the improvements so far! There are a couple of remaining issues (mostly minor things at this point, but now might be a good time to make those tweaks if you agree they're good ideas). Once we have those + a release with the 64bit variants, I think we should be good to go ahead and merge an updated version of this change.

mekya:
@ojw28 yeah I agree with you, I think it is the right time to make these minor changes as well. 
I am gonna try to fix https://github.com/ant-media/LibRtmp-Client-for-Android/issues/34 and https://github.com/ant-media/LibRtmp-Client-for-Android/issues/31

is it ok? Any other issue?

Bests,
A. Oguz

ojw28:
I don't think I have any other suggestions, no. Thanks!

ojw28:
@mekya - Do you have any kind of time estimate for the remaining changes + an updated release? No worries if not; just curious :).

ojw28:
Ping @mekya :).

mekya:
64 bytes from @mekya icmp_seq=0 ttl=30 days time=20 days :P

I am sorry for late reply @ojw28 , please check the latest commits in the repo

ojw28:
We still have one open issue on LibRtmp here, which is affecting the minimum sdk version we can set for the extension to 21 (not particularly useful). Tracked by https://github.com/ant-media/LibRtmp-Client-for-Android/issues/39. That said, I think we can merge + cleanup at this point.

PR #2517, opened 2017-02-28
Dev v2 pcm float

drhill:
As of the latest release of the Shield TV now supports outputting PCM 32 bit float for multi channel audio (prior versions only supported 2 channels no matter how many channels were passed).  The Resampling buffer was updated to convert 24bit audio to 32bit float.  Also, the ffmpeg extension was updated to output 32bit float when supported as well.  Might as well convert any >16bit audio streams to 32 bit instead of dithering.

### DIFF
--- a/extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
+++ b/extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
@@ -79,8 +79,7 @@ AVCodecContext *createContext(JNIEnv *env, AVCodec *codec,
  * Decodes the packet into the output buffer, returning the number of bytes
  * written, or a negative value in the case of an error.
  */
-int decodePacket(AVCodecContext *context, AVPacket *packet,
-                 uint8_t *outputBuffer, int outputSize, jboolean use32BitFloatOutput);
+int decodePacket(AVCodecContext *context, AVPacket *packet, uint8_t *outputBuffer, int outputSize);
 
 /**
  * Outputs a log message describing the avcodec error number.
@@ -119,7 +118,7 @@ DECODER_FUNC(jlong, ffmpegInitialize, jstring codecName, jbyteArray extraData, j
 }
 
 DECODER_FUNC(jint, ffmpegDecode, jlong context, jobject inputData,
-    jint inputSize, jobject outputData, jint outputSize, jboolean use32BitFloatOutput) {
+    jint inputSize, jobject outputData, jint outputSize) {
   if (!context) {
     LOGE("Context must be non-NULL.");
     return -1;
@@ -142,8 +141,7 @@ DECODER_FUNC(jint, ffmpegDecode, jlong context, jobject inputData,
   av_init_packet(&packet);
   packet.data = inputBuffer;
   packet.size = inputSize;
-  return decodePacket((AVCodecContext *) context, &packet, outputBuffer,
-                      outputSize, use32BitFloatOutput);
+  return decodePacket((AVCodecContext *) context, &packet, outputBuffer, outputSize);
 }
 
 DECODER_FUNC(jint, ffmpegGetChannelCount, jlong context) {
@@ -162,7 +160,7 @@ DECODER_FUNC(jint, ffmpegGetSampleRate, jlong context) {
   return ((AVCodecContext *) context)->sample_rate;
 }
 
-DECODER_FUNC(jlong, ffmpegReset, jlong jContext, jbyteArray extraData, jboolean use32BitFloatOutput) {
+DECODER_FUNC(jlong, ffmpegReset, jlong jContext, jbyteArray extraData) {
   AVCodecContext *context = (AVCodecContext *) jContext;
   if (!context) {
     LOGE("Tried to reset without a context.");
@@ -171,6 +169,8 @@ DECODER_FUNC(jlong, ffmpegReset, jlong jContext, jbyteArray extraData, jboolean 
 
   AVCodecID codecId = context->codec_id;
   if (codecId == AV_CODEC_ID_TRUEHD) {
+
+    const bool use32BitFloatOutput = context->request_sample_fmt == OUTPUT_FORMAT_32FLOAT;
     // Release and recreate the context if the codec is TrueHD.
     // TODO: Figure out why flushing doesn't work for this codec.
     releaseContext(context);
@@ -232,7 +232,7 @@ AVCodecContext *createContext(JNIEnv *env, AVCodec *codec,
 }
 
 int decodePacket(AVCodecContext *context, AVPacket *packet,
-                 uint8_t *outputBuffer, int outputSize, jboolean use32BitFloatOutput) {
+                 uint8_t *outputBuffer, int outputSize) {
   int result = 0;
   // Queue input data.
   result = avcodec_send_packet(context, packet);
@@ -277,7 +277,7 @@ int decodePacket(AVCodecContext *context, AVPacket *packet,
       av_opt_set_int(resampleContext, "in_sample_rate", sampleRate, 0);
       av_opt_set_int(resampleContext, "out_sample_rate", sampleRate, 0);
       av_opt_set_int(resampleContext, "in_sample_fmt", sampleFormat, 0);
-      av_opt_set_int(resampleContext, "out_sample_fmt", use32BitFloatOutput ? OUTPUT_FORMAT_32FLOAT : OUTPUT_FORMAT, 0);
+      av_opt_set_int(resampleContext, "out_sample_fmt", context->request_sample_fmt, 0);
       result = avresample_open(resampleContext);
       if (result < 0) {
         logError("avresample_open", result);
@@ -287,7 +287,7 @@ int decodePacket(AVCodecContext *context, AVPacket *packet,
       context->opaque = resampleContext;
     }
     int inSampleSize = av_get_bytes_per_sample(sampleFormat);
-    int outSampleSize = av_get_bytes_per_sample(use32BitFloatOutput ? OUTPUT_FORMAT_32FLOAT : OUTPUT_FORMAT);
+    int outSampleSize = av_get_bytes_per_sample(context->request_sample_fmt);
     int outSamples = avresample_get_out_samples(resampleContext, sampleCount);
     int bufferOutSize = outSampleSize * channelCount * outSamples;
     if (outSize + bufferOutSize > outputSize) {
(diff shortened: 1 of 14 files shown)


drhill:
rebased with latest dev-v2

drhill:
Anything I can do to get this pulled?  Audio quality improves with this for 24 bit PCM streams and when using ffmpeg to convert audio that is > 16 bit.

library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
```diff
@@ -129,7 +138,7 @@ public boolean configure(int sampleRateHz, int channelCount, @Encoding int encod
 
   @Override
   public boolean isActive() {
-    return Math.abs(speed - 1f) >= CLOSE_THRESHOLD || Math.abs(pitch - 1f) >= CLOSE_THRESHOLD;
+    return isActive && (Math.abs(speed - 1f) >= CLOSE_THRESHOLD || Math.abs(pitch - 1f) >= CLOSE_THRESHOLD);
```
andrewlewis: `AudioTrack.setPlaybackParameters` currently assumes that it's possible to change the playback parameters unless it's using passthrough output. When using `ENCODING_PCM_FLOAT` it will behave as if the parameters were applied, but this processor will actually be inactive in that case so the parameters won't be applied.

Thinking about `AudioProcessor`s more generally in the context of this change: if not in passthrough mode, at the moment `AudioTrack` always applies `ResamplingAudioProcessor` first to get audio into 16-bit PCM format then applies the other processors in turn, all of which expect 16-bit PCM input. In the current design, it's expected that audio processors other than the resampler should input and output PCM 16-bit encoding, and they will throw if they get something else.

If the main use case for this change is to decode to 32-bit float directly from `FfmpegAudioRenderer`, I wonder if we can simplify things by making `ENCODING_PCM_FLOAT` work like passthrough, at least initially. (Perhaps with a new boolean `encodedOutput` in `AudioTrack`, which is always true if `passthrough` is true but also true if writing float data.) Then the audio processors don't need to be modified. Then in later changes we could add support for resampling 24/32-bit PCM to float as a special case, then potentially handle float input in the other audio processors if there is enough demand to justify the added complexity. What do you think?

library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
```diff
@@ -850,6 +855,17 @@ public static int inferContentType(String fileName) {
...
 
   /**
+   * Returns the whether the device can output 32bit float audio
+   *
+   * @param capabilities The audio capabilities of the device.
+   * @param channelCount The channel count.
+   */
+  public static boolean canHandle32BitFloatAudio(AudioCapabilities capabilities, int channelCount) {
+    return (capabilities != null && capabilities.supportsEncoding(C.ENCODING_PCM_FLOAT)
+     && ((Util.SDK_INT > 20 && channelCount <= 2) || (Util.SDK_INT >= 24)));
```
andrewlewis: Any idea whether the change in support at API 24 was specific to NVIDIA Shield devices, or an Android platform limitation? I couldn't find any information about this internally. Have you found any documentation or changelogs that mention this, or is it based on observing the behavior on different API versions? Thanks.

extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
```diff
@@ -201,13 +203,13 @@ AVCodec *getCodecByName(JNIEnv* env, jstring codecName) {
...
 AVCodecContext *createContext(JNIEnv *env, AVCodec *codec,
-                              jbyteArray extraData) {
+                              jbyteArray extraData, jboolean use32BitFloatOutput) {
   AVCodecContext *context = avcodec_alloc_context3(codec);
   if (!context) {
     LOGE("Failed to allocate context.");
     return NULL;
   }
-  context->request_sample_fmt = OUTPUT_FORMAT;
+  context->request_sample_fmt = use32BitFloatOutput ? OUTPUT_FORMAT_32FLOAT : OUTPUT_FORMAT;
```
andrewlewis, line 212: Can this be read out of the `AVCodecContext` passed into `ffmpegDecode`, `decodePacket` and `ffmpegReset`, to avoid adding the extra parameter to those methods?

extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
```diff
@@ -275,7 +277,7 @@ int decodePacket(AVCodecContext *context, AVPacket *packet,
       av_opt_set_int(resampleContext, "in_sample_rate", sampleRate, 0);
       av_opt_set_int(resampleContext, "out_sample_rate", sampleRate, 0);
       av_opt_set_int(resampleContext, "in_sample_fmt", sampleFormat, 0);
-      av_opt_set_int(resampleContext, "out_sample_fmt", OUTPUT_FORMAT, 0);
+      av_opt_set_int(resampleContext, "out_sample_fmt", use32BitFloatOutput ? OUTPUT_FORMAT_32FLOAT : OUTPUT_FORMAT, 0);
```
andrewlewis: If we requested 32-bit float but the decoder output 16-bit PCM, will this upsample it? It seems preferable to just fall back to 16-bit PCM output in that case.

andrewlewis (commented):
Thanks!

library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
drhill: You are right.  This is "passthrough" really.  But there are two use cases for this change.  First is to playback 24bit pcm audio without any dithering so converting to 32bit float is necessary.  Second is to allow FFmpegAudioRenderer to decode any HD audio streams to >16bit audio.

I'll look to make the recommended changes (and below one) next week.  Thanks.

library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
drhill: I have no idea.  I only have a Shield hooked up to a receiver and it's up to date.  I would have never thought to put my nexus players (running 7.1 now) in my theater.  I'll take a gander through change logs on this later.  Wouldn't surprise me if it was a shield only limitation at the earlier apis.

extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
drhill: Will look into it.

extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
drhill: I believe the answer is yes.  That is something that is probably an easy test I can do later.

drhill:
I removed the unneeded use32BitFloatOutput parameters in ffmpeg_jni.cc.  Tested the ffmpeg decoding of 16 bit sources, they are properly resampled to 32bit which then plays fine with this.  

I was thinking seeing how I would adapt PCM Float output as "passthrough" in the code.  It felt a little sloppy.  Long term wouldn't you want to use the sonic processor on all PCM output?  I split my changes in resample processor to a new float specific resampler  and made audiotrack check output encoding before configuring/adding the processors into the used list.  When setplaybackparameters is called I had it check sonicprocessor's output capabilities to what we are using.

I don't particularly know where to look for old code changes (which may be Nvidia shield only) which fixed mutlichannel output with PCM float in API 24+.  We could drop the second check... if it's Shield only and they haven't updated to the latest firmware then who's fault is it if they are only getting 2 channel output.

andrewlewis:
Thanks for making those changes.

The part of this change that gets 32-bit float output from FFmpeg and queues it to the AudioTrack just like passthrough mode looks pretty good and is fairly minimal. For that case I think we should also ideally only get FFmpeg to provide 32-bit float output if the resolution of the input is > 16-bit (there's no point upsampling lower resolution input).

For handling 24-bit and 32-bit integer input, I'm not yet convinced it's worth adding the extra complexity to the non-passthrough audio processing path. I think it's unlikely that any of our audio processors apart from the resampler will handle non-16-bit integer input directly for a while, as it's quite uncommon and is likely to be a lot of work. Do you have a use case where you get 24-bit or 32-bit integer input that doesn't go via the FfmpegAudioRenderer (where it can be output as either 32-bit float or 16-bit integer as required)? If we do handle this, I wonder if we should just add a separate path from the existing audio processors.

By the way, if you'd prefer not to iterate further on this PR, I'm happy to do the remaining fixes for code style etc. and submit it. If that applies I'd be grateful if you could share some sample media (via email perhaps). Thanks!

drhill:
I can look at only enabled 32 bit ffmpeg when >16 bit input.  Edit:  Should add add the audiobitdepth flag from matroskaextractor into format?

I do have 24 bit lpcm tracks that I can share short samples via google drive

andrewlewis:
Sorry it's taken so long to get round to this.

Are you happy for me to make the changes separately to support PCM float output from FfmpegAudioRenderer, handled just like passthrough mode in AudioTrack (now renamed DefaultAudioSink)? If FfmpegAudioRenderer is set to use float output, there would be no support for changing playback speed/pitch or applying any other audio processors, and no resampling to/from float in audio processors. We can address those use cases later if it's worth it, but until then the app could just choose non-float output mode when creating FfmpegAudioRenderer.

Are the MatroskaExtractor changes important for this feature? I'm unclear why we need a bit depth for E-AC3, TrueHD etc..

drhill:
Sure.  I can test the changes you make for the two use cases (>16bit DTS/DD decoded tracks and >16 PCM tracks) I have and go from there.

The MatroskaExtractor changes are so we know if the original tracks were >16 bit tracks.  If they are 16bit tracks no use doing float output.

drhill:
I resync'd to the latest (as of last week) exoplayer dev-v2.  None of the changes fixed these two use cases.  I added a new pull request to get float output in ffmpeg.  That would get one use case covered.  

Just need to look at how to handle 24bit audio in and float out (with no data loss) cleanly.   I can close this and look to handle that differently.  Ok?

andrewlewis:
`FfmpegAudioRenderer` now has a constructor with an `enableFloatOutput` parameter, and passing `true` should cause 32-bit float output to be used (see [`shouldUseFloatOutput`](https://github.com/google/ExoPlayer/blob/b8aedfb/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegAudioRenderer.java#L122) for the conditions for enabling this).

I'm not sure whether that functionality is included in the version you tested. Please could you clarify what else would need to be done to support your use cases? Thanks, and apologies again for the delay!

drhill:
FfmpegAudioRenderer now outputs 32bit audio for non PCM audio.  I'll close this pull request and we can handle the input PCM audio side in the other PR

PR #2547, opened 2017-03-10
Support comma separators in ISO-8601 date/time format for milliseconds

benwilber:
Presently the regex for parsing ISO-8601 formatted datetimes for supporting `EXT-X-PROGRAM-DATE-TIME` in HLS only supports period (`.`) separators.  Some HLS stream packagers use commas (`,`) to separate milliseconds.

Examples:

`2014-08-06T11:00:00.000Z`

`2014-08-06T11:00:00,000Z`

[ISO 8601:2004](https://www.iso.org/standard/40874.html) states that parsers should support both delineators, with a preference for commas.

[Additional information](https://en.wikipedia.org/wiki/ISO_8601)

> A decimal mark, either a comma or a dot (without any preference as stated in resolution 10 of the 22nd General Conference CGPM in 2003,[16] but with a preference for a comma according to ISO 8601:2004)

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/util/Util.java
+++ b/library/src/main/java/com/google/android/exoplayer2/util/Util.java
@@ -94,7 +94,7 @@ public final class Util {
   private static final String TAG = "Util";
   private static final Pattern XS_DATE_TIME_PATTERN = Pattern.compile(
       "(\\d\\d\\d\\d)\\-(\\d\\d)\\-(\\d\\d)[Tt]"
-      + "(\\d\\d):(\\d\\d):(\\d\\d)(\\.(\\d+))?"
+      + "(\\d\\d):(\\d\\d):(\\d\\d)([\\.,](\\d+))?"
       + "([Zz]|((\\+|\\-)(\\d\\d):?(\\d\\d)))?");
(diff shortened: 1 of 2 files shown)


benwilber:
I signed it!

library/src/main/java/com/google/android/exoplayer2/util/Util.java
```diff
@@ -94,7 +94,7 @@
   private static final String TAG = "Util";
   private static final Pattern XS_DATE_TIME_PATTERN = Pattern.compile(
       "(\\d\\d\\d\\d)\\-(\\d\\d)\\-(\\d\\d)[Tt]"
-      + "(\\d\\d):(\\d\\d):(\\d\\d)(\\.(\\d+))?"
+      + "(\\d\\d):(\\d\\d):(\\d\\d)([\\.,]?(\\d+))?"
```
ojw28: I don't understand what the additional "?" is for. I'm not sure it's correct and the updated test passes without it. Please clarify or remove.

Otherwise this change looks good. Thanks!

benwilber: Yep you're correct, it was redundant.  I removed it.

PR #2606, opened 2017-03-26
Fix OMX crash for Odroid Android builds . Ignore buffer release error

dessatel:
Exynos OMX decoder for ODROID XU4 returns errors on buffer release when HLS bitrate switch happens.
This will release codec instance and continue without error

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
+++ b/library/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
@@ -879,8 +879,15 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
           return false;
         }
       } else {
-        outputIndex = codec.dequeueOutputBuffer(outputBufferInfo,
-            getDequeueOutputBufferTimeoutUs());
+          try{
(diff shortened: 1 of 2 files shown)


ojw28:
Unconditionally swallowing exceptions isn't really an acceptable fix for this issue. At the very least the workaround needs to be targeted to the specific devices, codecs and api levels that the issue is known to affect (as appropriate). See for example how `codecNeedsFlushWorkaround` is targeted to SDK levels, specific decoder names and devices.

As a second point, this is a fairly extreme workaround. It's likely there exists a less aggressive way of achieving the same thing. For example, is forcing `codecIsAdaptive` to `false` sufficient? If so, I'd suggest creating a targeted workaround in `MediaCodecRenderer` that does something more like that.

As a final point, pull requests should be made to `dev-v2`, not `release-v2`. Thanks!

ojw28:
One more thing: If you're going to provide a fix for something, please describe the fix in sufficient detail for us to understand it (e.g. at the very least a bug report that includes a full stack trace demonstrating what goes wrong). Thanks!

dessatel:
Thanks, good points. Will check on codecNeedsFlushWorkaround

PR #2612, opened 2017-03-27
Disables codecIsAdaptive for Odroid-XU4

dessatel:
Disables codecIsAdaptive for Odroid-XU3/4.

Crash log on bitrate change in HLS is attached.
[ordoid-xu4-crash.zip](https://github.com/google/ExoPlayer/files/872251/ordoid-xu4-crash.zip)

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
@@ -1171,14 +1171,14 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
         && "OMX.MTK.AUDIO.DECODER.MP3".equals(name);
   }
   /**
-   * Returns whether the decoder is known to be non adaptive.
+   * Returns whether the decoder is known to be non Adaptive.
    * <p>
    * If false is returned then we explicitly override codecIsAdaptive,
    * setting it to false.
    *
    * @param name The decoder name.
    * @param format The input format.
-   * @return True if the device is known  to be non adaptiv .
+   * @return False if the device is known  to be non adaptive .
    */
   private static boolean codecSupportsAdaptive(String name, Format format) {
     return !(
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
@@ -337,7 +337,7 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
     }
 
     String codecName = decoderInfo.name;
-    codecIsAdaptive = decoderInfo.adaptive && codecSupportsAdaptive(codecName, format);
+    codecIsAdaptive = decoderInfo.adaptive && (codecNeedsDisableAdaptationWorkaround(codecName)==false);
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -1170,5 +1170,19 @@ private static boolean codecNeedsMonoChannelCountWorkaround(String name, Format
...
+   * If false is returned then we explicitly override codecIsAdaptive,
+   * setting it to false.
+   *
+   * @param name The decoder name.
+   * @param format The input format.
+   * @return False if the device is known  to be non adaptive .
+   */
+  private static boolean codecSupportsAdaptive(String name, Format format) {
+    return !(
+        (Util.SDK_INT == 19 && Util.MODEL.equals("ODROID-XU3")
```
ojw28: If this device ever had anything prior to API level 19, consider <= instead of ==. It was probably broken on earlier API levels too (unless you know otherwise).

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -1170,5 +1170,19 @@ private static boolean codecNeedsMonoChannelCountWorkaround(String name, Format
...
+   * Returns whether the decoder is known to be non Adaptive.
+   * <p>
+   * If false is returned then we explicitly override codecIsAdaptive,
+   * setting it to false.
+   *
+   * @param name The decoder name.
+   * @param format The input format.
+   * @return False if the device is known  to be non adaptive .
+   */
+  private static boolean codecSupportsAdaptive(String name, Format format) {
```
ojw28: - Can we call this codecNeedsDisableAdaptationWorkaround for consistency with the methods above this one, and to make it clear this is a workaround and what it does.
- The Format arg is unused. Can we remove it (and the corresponding @param in the javadoc)?
- You'll need to get rid of the negating `!` below so that the new method name makes sense. The javadoc will need updating also.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -337,7 +337,7 @@ protected final void maybeInitCodec() throws ExoPlaybackException {
     }
 
     String codecName = decoderInfo.name;
-    codecIsAdaptive = decoderInfo.adaptive;
+    codecIsAdaptive = decoderInfo.adaptive && codecSupportsAdaptive(codecName, format);
```
ojw28: && !codecNeedsDisableAdaptationWorkaround(codecName)

codecSupportsAdaptive is misleading as a method name because it returns true for pretty much everything, including all non-adaptive decoders :).

ojw28 (commented):
Thanks. This is looking much better. Some small suggestions attached. Once they're done we'll get this merged :).

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
dessatel: Will do! Though sounds to me that we are just shuffling double negative from function to caller . But that's your call.

dessatel:
All suggestions are in yesterday submits

PR #2656, opened 2017-04-07
Add support for multiple subtitle tracks per PID and subtitle type id…

tresvecesseis:
…entification

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
@@ -20,9 +20,12 @@ import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.extractor.ExtractorOutput;
 import com.google.android.exoplayer2.extractor.TrackOutput;
 import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.TrackIdGenerator;
+import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.EsInfo;
+import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.LanguageInfo;
 import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.ParsableByteArray;
-import java.util.Collections;
+
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -30,22 +33,19 @@ import java.util.List;
  */
 public final class DvbSubtitleReader implements ElementaryStreamReader {
 
-  private final String language;
-  private final List<byte[]> initializationData;
+  private final List<LanguageInfo> languages;
+  private List<TrackOutput> outputTracks = new ArrayList<>();
 
-  private TrackOutput output;
   private boolean writingSample;
   private int bytesToCheck;
   private int sampleBytesWritten;
   private long sampleTimeUs;
 
   /**
-   * @param language The subtitle language code.
-   * @param initializationData Initialization data to be included in the track {@link Format}.
+   * @param esInfo Information associated to the elementary stream.
    */
-  public DvbSubtitleReader(String language, byte[] initializationData) {
-    this.language = language;
-    this.initializationData = Collections.singletonList(initializationData);
+  public DvbSubtitleReader(EsInfo esInfo) {
+    this.languages = esInfo.languagesInfo;
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


tresvecesseis:
@ojw28 , this two features could be tested with the sample "frenchTV_SD_2_x_subs_SD_in_one_PID.ts"

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
```diff
@@ -30,22 +32,49 @@
...
-  public DvbSubtitleReader(String language, byte[] initializationData) {
-    this.language = language;
-    this.initializationData = Collections.singletonList(initializationData);
+  public DvbSubtitleReader(TsPayloadReader.EsInfo esInfo) {
+    String language;
+    List<byte[]> initializationData;
+
+    int pos = 2;
+
+    while (pos + 8 <= esInfo.descriptorBytes.length) {
```
ojw28: This blindly assumes `EsInfo.descriptorBytes` only contains DVB descriptor tags. I'm not sure that's valid.

In any case, I think `TsExtractor readEsInfo` is probably the right place to be doing this parsing, given it's checking the tag type and doing some parsing already. Can we just change that method to build the list of DvbSubtitleTrack objects, each containing language + init-data, rather than just init-data that it sets currently. Then pass that list through EsInfo and I think everything is done?

tresvecesseis: @ojw28 , OK, I will implement in  `TsExtractor readEsInfo` the parsing of the subtitling descriptor (section 6.2.41 of ETSI 300 468):

```
subtitling_descriptor(){
  descriptor_tag              8 uimsbf
  descriptor_length           8 uimsbf
  for (i= 0;i<N;i++){
    ISO_639_language_code    24 bslbf
    subtitling_type           8 bslbf
    composition_page_id      16 bslbf
    ancillary_page_id        16 bslbf
  }
}
```

what do you think about the language code and language features, we should put whatever each  standard dictates or we should have a common list of language and feature codes and translate to them and even localize them?

ojw28: Sounds good to me. You'll probably find that `Util.normalizeLanguageCode(languageCode)` will do the trick of converting the language code into a standard format?

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
```diff
@@ -56,9 +56,24 @@ public void seek() {
   @Override
   public void createTracks(ExtractorOutput extractorOutput, TrackIdGenerator idGenerator) {
     idGenerator.generateNewId();
-    this.output = extractorOutput.track(idGenerator.getTrackId(), C.TRACK_TYPE_TEXT);
-    output.format(Format.createImageSampleFormat(idGenerator.getFormatId(),
-        MimeTypes.APPLICATION_DVBSUBS, null, Format.NO_VALUE, initializationData, language, null));
+
+    TrackOutput output;
```
ojw28, line 60: [General comment for future changes] It's much clearer if you declare variables in the scopes they're actually used. For example by doing
```
LanguageInfo language = languages.get(i);
```
on L64. Ditto for output on L71.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
```diff
@@ -56,9 +56,24 @@ public void seek() {
...
+    LanguageInfo language;
+
+    for (int i = 0; i < languages.size(); i++) {
+      language = languages.get(i);
+      idGenerator.generateNewId();
+
+      if (((language.programElementType & 0xF0 ) >> 4 ) == 2 ) {
+        language.languageCode += " for hard of hearing";
```
ojw28, line 68: [General comment for future changes] Similarly, this kind of mutable state is quite confusing and error prone. This mutates state passed through the constructor of this method at some later stage. If state is immutable it becomes way easier to work out what's going on.

ojw28: Note: The "hard of hearing" thing really needs passing as a proper flag in Format. The text form only makes sense in English. I'll revert this piece for now; and follow up with a proper flag in the format object later.

ojw28:
Thanks for the patch!

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbSubtitleReader.java
tresvecesseis: @ojw28 , it was my first though but I did a quick hack until I had time to implement the needed flags for the audio and subtitle tracks

PR #2590, opened 2017-03-22
Add RTMP support

begeekmyfriend:
I have implemented RTMP in pure Java for ExoPlayer with the HKS-TV RTMP live stream link added into `media.exolist.json` as a demo for watching.

The commits include:

- RTMP implementation modified from [SimpleRTMP](https://github.com/faucamp/SimpleRtmp).
  - Amf0 data format;
  - RTMP packet header format;
  - RTMP chunk stream format;
  - RTMP packet decoder;
  - RTMP connection and session.
- RTMP data source porting which is consistent with other upstream protocol such as HTTP.
- FLV stream extractor adaptation for RTMP to parse FLV stream within RTMP packets.
- RTMP uri parser.

My PR works well under HKS-TV RTMP live stream link.

PR #2683, opened 2017-04-13
Added mp3 support to FLV extractor

tylerjroach:
MP3 support has been added to the flv extractor. This will be more important if #2590 or #2503 is accepted for rtmp streams with mp3 audio. This commit will add mp3 audio support to both of those pull requests as well as standard flv playback.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
@@ -15,6 +15,7 @@
  */
 package com.google.android.exoplayer2.extractor.flv;
 
+import android.util.Log;
 import android.util.Pair;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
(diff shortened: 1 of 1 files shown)


tylerjroach:
I signed it!

library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/AudioTagPayloadReader.java
```diff
@@ -55,8 +64,19 @@ protected boolean parseHeader(ParsableByteArray data) throws UnsupportedFormatEx
     if (!hasParsedAudioDataHeader) {
       int header = data.readUnsignedByte();
       audioFormat = (header >> 4) & 0x0F;
-      // TODO: Add support for MP3.
-      if (audioFormat == AUDIO_FORMAT_ALAW || audioFormat == AUDIO_FORMAT_ULAW) {
+      if (audioFormat == AUDIO_FORMAT_MP3) {
+        int sampleRateIndex = (header >> 2) & 0x03;
+        if (sampleRateIndex < 0 || sampleRateIndex >= AUDIO_SAMPLING_RATE_TABLE.length) {
```
ojw28: This looks like it's impossible. Given L68, sampleRateIndex can only be 0,1,2 or 3 here, I think? If you agree, please delete this block (and the TAG constant near the top).

ojw28 (commented):
This change looks good! Please see the one comment inline and remove the dead code if you agree. Thanks!

rkrishnan2012:
I consent to my commits being used.

PR #2686, opened 2017-04-14
Fixes OOM that can occur from improperly formatted data

tylerjroach:
This prevents an OOM error that can occur. We have seen situations where dataOffset isn't stopped at 9. The offset should be 9 according to flv specifications. This can cause the offset to grow too large and cause OOM errors. My co-worker @rkrishnan2012 can clarify the issue further.


rkrishnan2012:
I signed it!

rkrishnan2012:
I signed it!

rkrishnan2012:
I signed it!

rkrishnan2012:
I signed it!

rkrishnan2012:

PR #2687, opened 2017-04-14
Fixes OOM that can occur from improperly formatted flv data

tylerjroach:
This prevents an OOM error that can occur. We have seen situations where dataOffset isn't stopped at 9. The offset should be 9 according to flv specifications. This can cause the offset to grow too large and cause OOM errors. My co-worker @rkrishnan2012 can clarify the issue further.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
@@ -110,6 +110,7 @@ public final class FlvExtractor implements Extractor, SeekMap {
     input.peekFully(scratch.data, 0, 4);
     scratch.setPosition(0);
     int dataOffset = scratch.readInt();
+    if (dataOffset > 9) dataOffset = 9;
 
     input.resetPeekPosition();
     input.advancePeekPosition(dataOffset);


rkrishnan2012:

library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
```diff
@@ -110,6 +110,7 @@ public boolean sniff(ExtractorInput input) throws IOException, InterruptedExcept
     input.peekFully(scratch.data, 0, 4);
     scratch.setPosition(0);
     int dataOffset = scratch.readInt();
+    if (dataOffset > 9) dataOffset = 9;
```
ojw28, line 113: Data offset is explicitly allowed to be more than 9 to accommodate arbitrary data between the defined header and body. It can never be less than 9. If it were always exactly 9 then there wouldn't be a need for this field at all; it's *explicitly* there to allow for larger values. So I'm not convinced this is a correct fix...

ojw28:
Confirmed @rkrishnan2012 has signed the CLA.

library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/FlvExtractor.java
tylerjroach: @ojw28 Do you have any suggestions for best fixing this. According to flv specifications, 9 should be the value in both flv0 and flv1. There were never any newer flv versions released. 

https://www.adobe.com/content/dam/Adobe/en/devnet/flv/pdfs/video_file_format_spec_v10.pdf

In our code, we had previously set an arbitrary number of 200 as a limit ```(dataOffset > 200)```. It was a quick win when we were plagued by OOM exceptions that can be caused from the data offset being huge. I hate to throw a random number in there, but there should be some sort of max number that is agreed upon.

ojw28: What kind of values are you actually seeing (e.g. is it always the same huge number)? Where are these dodgy media files actually coming from?

ojw28: Also, do the media files in question play correctly with the workaround in place, or do they fail to play in some other less bad way (e.g. because sniff ends up returning false)?

tylerjroach: We allow streams from any method, which would commonly include android, iphone, obs, xsplit, tricaster, vidiu, etc. We've had a commit like this in place for a long time now so I'll see if I have any previous records. @rkrishnan2012 may be able to provide more context as he is the one that worked on the fix originally. I'm not sure if these videos would end up playing or not. I do know #2688 was the more prevalent OOM cause.

ojw28: Ok, finding out more sounds good. I'm not sure we have enough information to know what the best fix (or otherwise) is at this point. For example, if the files always have some specific value and fail to play anyway, then it would be much better to check for that value specifically and just return false.

tylerjroach:
Our crash logs don't go back that far. I'll close this for now. Our crash could have been a side effect from another issue. If I revert this commit and see any issues crop up, I'll reopen with more details.

PR #2688, opened 2017-04-14
Fix potential OOM that can occur reading first NAL packet before seq header

tylerjroach:
If a NALU packet is read before a header has been seen, nalUnitLengthFieldLength will be 0 and get stuck in the ```while (data.bytesLeft() > 0)``` where an OOM exception can occur.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
@@ -93,7 +93,7 @@ import com.google.android.exoplayer2.video.AvcConfig;
           avcConfig.initializationData, Format.NO_VALUE, avcConfig.pixelWidthAspectRatio, null);
       output.format(format);
       hasOutputFormat = true;
-    } else if (packetType == AVC_PACKET_TYPE_AVC_NALU) {
+    } else if (packetType == AVC_PACKET_TYPE_AVC_NALU && hasOutputFormat) {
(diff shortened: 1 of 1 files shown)


rkrishnan2012:

library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
```diff
@@ -93,14 +93,15 @@ protected void parsePayload(ParsableByteArray data, long timeUs) throws ParserEx
...
-    } else if (packetType == AVC_PACKET_TYPE_AVC_NALU) {
+    } else if (packetType == AVC_PACKET_TYPE_AVC_NALU && hasOutputFormat) {
       // TODO: Deduplicate with Mp4Extractor.
       // Zero the top three bytes of the array that we'll use to decode nal unit lengths, in case
       // they're only 1 or 2 bytes long.
       byte[] nalLengthData = nalLength.data;
       nalLengthData[0] = 0;
       nalLengthData[1] = 0;
       nalLengthData[2] = 0;
+      if(nalUnitLengthFieldLength == 0) nalUnitLengthFieldLength = 4;
```
ojw28: I don't understand what this is for. Given L96, why would nalUnitLengthFieldLength ever be 0 here?

ojw28 (commented):
Looks like L96 is the only one that actually needs changing here. Please take a look at the comment inline and revert L104 if you agree. Thanks!

ojw28:
Confirmed @rkrishnan2012 has signed the CLA.

PR #2693, opened 2017-04-17
Fix Sub-tile Overlap issue

WeiChungChang:
Fix VTT sub-title overlap issue.

The **_detailed issue description & report_** could be provided later **_if it is necessary_**.


ojw28:
I don't understand this change. I'm dubious it's correct given the method description includes "Returns whether or not this cue should be placed in the default position" and you're changing it to return true for something that specifically requests a non-default position... Please provide a proper description.

PR #2741, opened 2017-04-27
It’s possible that url has query parameters after .m3u8, for example …

frontyard:
…Wowza HLS DVR urls end with .m3u8?DVR.

This fix will strip any query parameters before trying to guess the content type.


ojw28:
If you're testing a URL/URI then you should be using inferContentType(Uri uri) instead of inferContentType(String fileName). The method that takes the Uri already does this for you.

frontyard:
I signed it!

PR #2756, opened 2017-04-30
Include Pixel Aspect Ratio in DebugTextViewHelper

kiall:
Add the video Pixel Aspect Ratio to the DebugTextViewHelper in order to help debug issues related to PAR changes mid-stream.

This info is useful to know and have available, so that my apps users (when they have debugging enabled) can see both the resolution and stretch applied.


ojw28:
This isn't useful information for the many apps that will only ever be playing content with par=1. Can we show it only if it's not equal to 1 or Format.NO_VALUE (both of which are the boring case)?

kiall:
Yea, that makes total sense. I'll update the PR to conditionally include this only when it's relevant.

ojw28:
Are you planning to update this? Else we will close it. Thanks.

kiall:
Apologies, I am, I'm travelling for work the last few weeks, so haven't had a chance just yet. I'll be back to the office Monday!

PR #2768, opened 2017-05-04
DefaultHttpDataSource: Fix can not play through a local proxy

michalliu:
When playing the video which is a local address eg: http://127.0.0.1 or htttp://10.66.66.66. Such address can be served by NanoHTTPD or something similar.

It’s commonly in a enterprise environment that everything is behind a proxy. when user have a proxy setting on the phone. The playback will fail because the proxy server can not connect to a local adress such as 127.0.0.1

So, we should bypass the local proxy setting in such scenario.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/DefaultHttpDataSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/DefaultHttpDataSource.java
@@ -29,9 +29,14 @@ import java.io.InterruptedIOException;
 import java.io.OutputStream;
 import java.lang.reflect.Method;
 import java.net.HttpURLConnection;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
 import java.net.NoRouteToHostException;
 import java.net.ProtocolException;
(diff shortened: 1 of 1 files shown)


ojw28:
I'm not an expert in this area, but I'm not convinced this is a good idea. Some specific concerns that spring to mind:

- Aren't there cases where it *would* be preferable for local requests to still go through a proxy? For example if the proxy was itself local (I'm unsure if what I've written makes sense, just a hunch!).
- It's unclear whether isLocalAddress may involve a non-trivial overhead. I suspect it may do, which would adversely affect the common case.
- In the enterprise environment you speak of, I think you can probably just avoid the problem in the first place by configuring the proxy setting to exclude local addresses. Proxy settings on a phone I'm using include a "Bypass proxy for" setting. Isn't that the correct solution for this problem? Else you're relying on *every* individual application that might make local requests to explicitly ignore the proxy setting, which is non-standard behavior.

michalliu:
I agree with you isLocalAddress is a overhead. The user should add a ignore policy on phone settings if needed.

PR #2814, opened 2017-05-11
Support AVC Level 5.2

sillywab8:
Samsung S7 and Sony XZs with 7.1.1 updates are advertising the capability to support AVC Level 5.2 frame sizes. Without this proposed update these devices' videos will play at only 480p.

Level 5.1 and 5.2 support same frame size according to https://en.wikipedia.org/wiki/H.264/MPEG-4_AVC.


sillywab8:
I signed it!

sillywab8:
I Confirm.

ojw28:
This is modifying 784 files in a dead release candidate branch. I'm pretty sure that's not what you intended to do? Pull requests should go to the `dev-v2` branch.

PR #2820, opened 2017-05-12
Add support for AVC Level 5.2

sillywab8:
Samsung S7 and Sony XZs with 7.1.1 updates are advertising the capability to support AVC Level 5.2 frame sizes. Without this proposed update these devices' videos will play at only 480p. 
See https://github.com/google/ExoPlayer/commit/83b43a6fe620f9ce74066f6b06c35e134b07d094

Level 5.1 and 5.2 support same frame size according to https://en.wikipedia.org/wiki/H.264/MPEG-4_AVC.

The constant was set in _MediaCodecInfo.java_
```        public static final int AVCLevel52      = 0x10000;```

but never checks against:
```     
...
      case CodecProfileLevel.AVCLevel5: return 22080 * 16 * 16;
      case CodecProfileLevel.AVCLevel51: return 36864 * 16 * 16;  <-- only 5.1
      default: return -1;
    }
  }
```

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
@@ -429,6 +429,7 @@ public final class MediaCodecUtil {
       case CodecProfileLevel.AVCLevel42: return 8704 * 16 * 16;
       case CodecProfileLevel.AVCLevel5: return 22080 * 16 * 16;
       case CodecProfileLevel.AVCLevel51: return 36864 * 16 * 16;
+      case CodecProfileLevel.AVCLevel52: return 36864 * 16 * 16;
       default: return -1;
     }
   }

### CODE REVIEW

**Comment by ojw28** [comment]
On a Galaxy S7 at least, I don't think this method is ever called. The same is probably true for the Sony XZ as well. Do you actually see the behavior you describe on real Galaxy S7 and Sony XZ devices? If so, what does the call stack look like and what version of Android are they using?

These questions aside, yes, we should merge this. I doubt it'll make a difference for the devices mentioned though...?

**Comment by sillywab8** [comment]
The call stack starts from our HLSRendererBuilder when we are determining variant indices.

I collected this log from _maxH264DecodableFrameSize()_ method in _MediaCodecUtil.java_ so the device is hitting the method:
```
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities number of profile levels: 5
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities profile level: 65536
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities profile level: 65536
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities profile level: 65536
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities profile level: 65536
05-06 18:00:57.507 I/com.myapp.android(24373): Codec capabilities profile level: 65536
```

I was only able to get a hold of the Samsung S7 to recreate but it was reported on an XZs as well.  This particular device is just on API 24 and its showing that it can handle AVC Level 5.2 (0x1000 ... 65536).

```
~~~ Device ~~~
  Android: 24, kernel: 3.18.31-11028965
  Manufacturer: samsung
  Model: SM-G930V
  Available Processors: 4
```

ojw28:
Are you using a really old version of ExoPlayer or something? You'd be better off using MediaCodecUtil.isSizeSupportedV21 (or the one that also takes rate) on API level 21 and above, which is what logic internal to ExoPlayer does.

sillywab8:
Yes, r1.4.2. 
v2 has significant structure changes which will require us to rewrite our video component so for right now that's on a todo list.  I think a lot of folks out there are still on v1 for that reason.  But this fix got us going for now and bought us some time.

Issue #2841, opened 2017-05-18
Cannot playback TS files by fail to identify key frame within it


WeiChungChang:
### Issue description
Cannot Playback TS files.
There is only audio & **_all video frames are dropped._**
**_It is because the frames are all deemed as "non-key"._**
However, if played on, ex: VLC player, the movie is fine.

The file's key frames could be identified by **_recovery point SEI message_**
 
_**"...a special header, called the SEI Recovery Point, tells the decoder to "start here, decode X frames, and then start displaying the video."**_ 

https://software.intel.com/en-us/forums/intel-media-sdk/topic/311715 
http://www.videoredo.net/msgBoard/showthread.php?34794-How-to-cut-H264-without-IDR-frame

**_It may need the flow to create the flow of parsing SEI message, check recovery point & identify the key frame by it.  _**   A proposal of solution is as #2842

### Reproduction steps
Put the issued files of Link to test content & playback it.

### Link to test content
https://drive.google.com/drive/folders/0B7H5vJR3qWj0djkzS2ZHeWR6ZTA

### Version of ExoPlayer being used
2.3.1

### Device(s) and version(s) of Android being used

100%, HTC x9 smartphone with Android 6.0

### A full bug report captured from the device
Playback issue & root cause is described as indicated;
Not a system issue so skip here.

AquilesCanta:
Hi, please try the latest release. We've implemented key frame recognition using the random access indicator. Hopefully this works for you. Please post below (including a non-working stream) if you still run into issues.

PR #2842, opened 2017-05-18
H264 sei msg recovery point flow

WeiChungChang:
It proposes a possible solution to #2841;

By identifying key frame according to H264 SEI message recovery point, it succeeds finding key frame & could playback the issued files.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H264Reader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H264Reader.java
@@ -39,6 +39,8 @@ public final class H264Reader implements ElementaryStreamReader {
   private static final int NAL_UNIT_TYPE_SPS = 7; // Sequence parameter set
   private static final int NAL_UNIT_TYPE_PPS = 8; // Picture parameter set
 
+  private static final int SEI_TYPE_RECOVERY_POINT = 6;
+
   private final SeiReader seiReader;
   private final boolean allowNonIdrKeyframes;
   private final boolean detectAccessUnits;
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H264Reader.java
```diff
@@ -201,6 +203,7 @@ private void endNalUnit(long position, int offset, int discardPadding, long pesT
     }
     if (sei.endNalUnit(discardPadding)) {
       int unescapedLength = NalUnitUtil.unescapeStream(sei.nalData, sei.nalLength);
+      checkRecoveryPoint(sei.nalData, sei.nalLength);
```
ojw28, line 206: sei.nalLength should probably be unescapedLength?

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H264Reader.java
```diff
@@ -209,6 +212,43 @@ private void endNalUnit(long position, int offset, int discardPadding, long pesT
...
+      ffByte = bitArray.readBits(8);
+      size += ffByte;
+    } while (ffByte == 255 /* equal to 0xFF */);
+
+    // Read the slice header using the syntax defined in ITU-T Recommendation H.264 (2013)
+    // subsection 7.3.3.
+    if (!bitArray.canReadExpGolombCodedNum()) {
+      return;
+    }
+    int recoveryFrameCount = bitArray.readUnsignedExpGolombCodedInt();
```
ojw28, line 246: Are you sure the implementation here is correct? I'm seeing large values being read here (e.g. 223). It doesn't feel right that the number would be so large?

ojw28:
Closing this because Android decoders are not required to support starting from SEI recovery points. As per https://developer.android.com/reference/android/media/MediaCodec#KeyFrames, only IDR frames are suitable starting points. Some hardware decoders do appear to support being fed from SEI recovery points, but others do not. Google's software decoder, which is the only decoder on some lower end devices, does not provide support.

PR #2904, opened 2017-06-04
Add label attribute support for DASH audio/text

itaykinnrot:
When playing a stream with multiple audio tracks, currently the only way to identify the different tracks is using the lang attribute of AdaptationSet. We have customers who would like to customize the label that is displayed to the end user for selecting the audio track. Also, multiple audio tracks are sometimes used for uses cases which are not multiple languages, for example - one audio track with background music and one without it.
While it is possible to pass this info out-of-band, we feel it makes most sense to pass it in-band in the MPD. Other protocols have support for this - in HLS you can set the NAME attribute of EXT-X-MEDIA, and in MSS there's a Name attribute on StreamIndex

### DIFF
--- a/demo/src/main/java/com/google/android/exoplayer2/demo/DemoUtil.java
+++ b/demo/src/main/java/com/google/android/exoplayer2/demo/DemoUtil.java
@@ -38,14 +38,14 @@ import java.util.Locale;
           buildResolutionString(format), buildBitrateString(format)), buildTrackIdString(format)),
           buildSampleMimeTypeString(format));
     } else if (MimeTypes.isAudio(format.sampleMimeType)) {
-      trackName = joinWithSeparator(joinWithSeparator(joinWithSeparator(joinWithSeparator(
+      trackName = joinWithSeparator(joinWithSeparator(joinWithSeparator(joinWithSeparator(joinWithSeparator(
           buildLanguageString(format), buildAudioPropertyString(format)),
           buildBitrateString(format)), buildTrackIdString(format)),
-          buildSampleMimeTypeString(format));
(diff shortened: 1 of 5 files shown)


ojw28:
Is there a way to do this within the DASH spec (or DASH IF recommendations)? It doesn't look like `@label` is actually defined anywhere in either.

itaykinnrot:
I'll try...
But i still think this is useful and some packagers already support it.
Shaka project also accept it.
Thanks,

OrenMe:
Hi, I think this is a valid use case without changing the spec.
MPD is just an XML scheme so adding an attribute or even a node wouldn't break any parser AFAIK.
And I think this is quite useful and is supported on HLS and as stated, some packagers do support this feature.

ojw28:
My main concern is that it will be standardized using some other attribute name or mechanism, that we'll have to change it, and that we'll end up annoying a bunch of people who have come to rely on it. So I'd feel much happier if the approach were documented somewhere. You could file a request asking guidance to be added to the DASH-IF guidelines [here](https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues) or [here](https://gitreports.com/issue/Dash-Industry-Forum/DASH-IF-IOP), for example.

ojw28:
Sorry for ignoring this for so long. Looking at this again, I'd be happy to merge it. However the pull request needs to go to `dev-v2` rather than `release-v2`. If you update the change to build against latest `dev-v2` and open a new pull request into that branch, then we'll get it merged in. Thanks!

Closing this one, since a new one will need opening into the correct branch.

itaykinnrot:
https://github.com/google/ExoPlayer/pull/4391

PR #2918, opened 2017-06-08
Add onPreAacquireSession event.

mofneko:
We created an event to handle the timing just before the DRM is switched.
I need to play videos that always switch to widevine and other DRM.
Widevine needs to update SurfaceView every time it plays. Therefore, we decided to detect changes in DRM in advance at this event.


mofneko:
ping @ojw28 Do you think this is a useful change?
If so, I will resolve conflicts.

ojw28:
I don't really understand what this is for. That said, I don't think it's the right way of doing whatever it is that you're trying to do. Since the event is being posted from playback thread to main thread, you don't really have any strong guarantees about when that event will end up being handled on the main thread. So if you're relying on the event to do something that's needed when switching DRM scheme, I don't think there's any strong guarantee that you'd do it in time.

Please clarify.

ojw28:
Closing due to lack of clarification.

talhatariq:
send me download links

mofneko:
ping @talhatariq
The comment seems to have been deleted, but do you still need a download link?
If so, I will prepare the data again.

PR #2921, opened 2017-06-08
add  SampleAES AVC/H264 support in ExoPlayer

zhouweiguo2017:
refined code with latest ExoPlayer code( so it might be useful for code reviewer to compare and merge);

validate again: 
      1) all source code could be built sucessfully with AndroidStudio;
      2) the generated APK could playback SampleAES H264 content well as expected;
      3) this new feature bring nothing side-effects with existing code.

developer could add code to support SampleAES H265 video, SampleAES AAC, SampleAES AC3 contents base on the implementation


ojw28:
Please can you revert the indentation changes before we look at this? Specifically (but not limited to) C.java and TsExtractor.java. If you look at the diff (by clicking "Files changed"), it appears as though you've deleted and re-added huge parts of these files because you've changed them to have 4 character indentation rather than 2 character indentation. This makes it very hard to review what you've actually changed. ExoPlayer source code uses 2 character indentation (and 4 character indentation for line continuations only).

Thanks!

zhouweiguo2017:
Hi ojw28,
Thanks for your comments. I had been refined the original commitment to convenient for code review.

you can find the modified files as following:

[https://github.com/google/ExoPlayer/pull/2921/files]

zhouweiguo2017:
Hi ojw28,

Thanks your comments.
I had been refined the code according to your comments.  and did a simple sanity check with the generated apk, it seems that the SampleAES test content could playbacked well as expected and nothing side effects with the existing test contents.

zhouweiguo2017:
7 times commit on 13-06-2017 according to ojw28's review comment.  These commits are really manual work but might important for code review.^:^

zhouweiguo2017:
7 times commit on 13-06-2017 according to ojw28's review comment. These commits are really manual work but might useful for code review.^:^

zhouweiguo2017:
7 times commit on 13-06-2017 according to ojw28's review comment. These commits are really manual work but might useful for code review.^:^

AquilesCanta:
I think we can come up with a solution that does not require modifying existing readers.

This applies to transport streams only, but we could have an HlsTsPayloadReaderFactory that returns a wrapper for the specific readers. For example, an "H264SampleEncryptedReader" would first decrypt [the encrypted parts](https://developer.apple.com/library/content/documentation/AudioVideo/Conceptual/HLS_Sample_Encryption/Encryption/Encryption.html) using an AES/CBC cipher, and would then pass that to the actual reader. Same for other readers and, eventually, extractors. What do you think?

zhouweiguo2017:
Hi AquilesCanta,

Thanks your comments.
I agree with your idea.
Indeed we should create a specific reader for SampleAES content and make the new reader to be source node of the original H264Reader, it will more OOD here.

Due to the major modification in original H264Reader is trying assemble a completed NALU from the coming ES data,  pass the HLSEncryptInfo to the original H264Reader and then make some minor modifications in original H264Reader, the benefit here is that we can reuse  the most part of the original code and following the original mechnism/framework of ExoPlayer strictly.

zhouweiguo2017:
I can't understand why the conflicts still exist after I had been fixed the conflicts and validationed in local repo(the code had been submitted after validation and sanity check). appreciate too much for explanation.

AquilesCanta:
I'll close this PR as obsolete. I think it would be easier to implement sample AES encryption using a TrackOutput implementation, with minimal to no reader modifications.

PR #2936, opened 2017-06-12
exoplayer2 - mediaplayer extension

michalliu:
This extension provides the same API,callbacks, lifecycle with android system [mediaplayer](https://developer.android.com/reference/android/media/MediaPlayer.html). It makes developers using exoplayer in their project much easier since most of apps are using android mediaplayer currently.

We use this extension in our production environment without a problem.

### DIFF
--- a/extensions/mediaplayer/build.gradle
+++ b/extensions/mediaplayer/build.gradle
@@ -1,0 +1,42 @@
+// Copyright (C) 2016 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


ojw28:
I don't think this is a good idea. ExoPlayer's API is better thought out and can be improved over time. It's preferable for developers to incur a one time cost switching over to it. For developers who need to be able to switch between ExoPlayer and MediaPlayer, it's preferable to write a wrapper around MediaPlayer that implements ExoPlayer's API, not the other way around.

michalliu:
Ok, just keep a reference here in case of someone needs it

michalliu:
I created a repository for this extension (with demo). I believe this extension is useful. https://github.com/michalliu/exoplayer2-mediaplayer-extension

PR #2960, opened 2017-06-16
make hd audio buffer larger to prevent audio dropouts

drhill:
This isn't something that is always reproducible though it happens often enough, but eventually when watching a movie with HD audio passthrough (DTS-HD MA) audio will drop out for a second.


andrewlewis:
Looks good, thanks. Is there any particular reasoning behind the factor of six (giving a buffer size of ~ 300 KB)?

drhill:
Trial and error.  Multiplying by 4 still had drop outs.  5 too I believe.  I can't even say for a fact that 6 is the cure for all tracks, because even when using my receiver with another device (blu ray player) you get infrequent drop outs but when I started using 6 no longer happened to the point you noticed it.  Prior it would happen on every 7.1 DTS HD MA track at least once it not more.

drhill:
Closed since already merged in.  Thanks!

PR #2970, opened 2017-06-19
make defaultloadcontrol extensible

michalliu:
`DefaultLoadControl` is nice for most usages. It’s better to make it extensible so users can add more features to it. eg. load dynamic configs over network. It can be done by using a wrapper over it but it’s easier by inheriting


ojw28:
Why don't you just fork in these cases? It's less than 100 lines of code (excluding comments). If you want custom behaviors, it's kinda unlikely in my opinion that you want to inherit some of the base behaviors (and be at the mercy of those changing from release to release).

michalliu:
Okay, It's just a suggestion, make a non-final class to final will cause compiler error if someone already extends it, vice-versa is not. feel free to close this pull request, I have no problem with it.

ojw28:
Doesn't seem like a big deal to fork for the use case you describe. We like to keep things final where possible, since it helps to limit the (already very large) library API surface. Thanks.

PR #2991, opened 2017-06-23
Remove user agent override from OkHttpDataSourceFactory

alexbirkett:
Like any other header, the `User-Agent` can be configured on an `OkHttpClient`
instance e.g.:

```kotlin
val okHttpClient = OkHttpClient.Builder()
        .addInterceptor { chain ->
            val request = chain.request().newBuilder().addHeader(USER_AGENT, userAgent).build()
            chain.proceed(request)
        }
        .addInterceptor(httpLoggingInterceptor)
        .build()
```

The `OkHttpDataSource` required that a `User-Agent` was passed in as a parameter
overriding any `User-Agent` already configured on the `OkHttpClient`. This made
it impossible to configure the `User-Agent` in once place if the `OkHttpClient`
is also used elsewhere in the app.

The OkHttpClient is the appropriate place to configure headers that are not
specific to a particular request.

### DIFF
--- a/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
+++ b/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
@@ -46,7 +46,6 @@ public class OkHttpDataSource implements HttpDataSource {
   private static final AtomicReference<byte[]> skipBufferReference = new AtomicReference<>();
 
   private final Call.Factory callFactory;
-  private final String userAgent;
   private final Predicate<String> contentTypePredicate;
   private final TransferListener<? super OkHttpDataSource> listener;
   private final CacheControl cacheControl;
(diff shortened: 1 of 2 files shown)


extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
```diff
@@ -280,7 +276,6 @@ private Request makeRequest(DataSpec dataSpec) {
       }
       builder.addHeader("Range", rangeRequest);
     }
-    builder.addHeader("User-Agent", userAgent);
```
ojw28, line 283: Why not just add a null check here, remove the assertion from the constructor, and update the Javadoc to indicate that null is allowed? It seems unnecessary to remove the functionality, which is useful if you want to use one userAgent for requests made by the player, and another for requests made elsewhere by your app.

ojw28 (commented):
See inline comment. Fixes need to go into the `dev-v2` also, thanks.

extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
alexbirkett: I'm not sure why you'd want to share an `OkHttpClient` instance and mutate what is essentially its config. This seems like the wrong place to set the `User-Agent` to me. I get that a breaking change could be annoying. I'd rather make it optional than not have any change.

ojw28: What about for the use case I described (it seems pretty reasonable to me that an app might want to include info about the ExoPlayer version in the user agent being sent in requests made by the player)? You could probably argue that you should just use a separate `OkHttpClient` instance. I'm not sure what the implications of that are, but it feels at least possible that there might be some efficiency loss there (e.g. if the okhttp implementation uses a separate connection pool for each client).

In any case, let's go for the less invasive change where we just make the user agent optional, in the `dev-v2` branch. Let me know if you plan to send such a change, or alternatively whether you'd like us to push one directly.

Thanks for bringing the issue to our attention :)!

alexbirkett: I can send a new pr

PR #3015, opened 2017-07-03
Frugal in terms of bandwidth

get2abhi:
A data saver tweak in the existing exoplayer. Most of the logic is implemented in _**determineIdealSelectedIndex**_

FrugalAdaptiveTrackSelection.java is a clone of AdaptiveTrackSelection.java. The only difference is in the _**determineIdealSelectedIndex**_ method

### DIFF
--- a/demo/src/main/assets/media.exolist.json
+++ b/demo/src/main/assets/media.exolist.json
@@ -1,5 +1,14 @@
 [
   {
+    "name" : "RnD",
+    "samples": [
+      {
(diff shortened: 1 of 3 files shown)


ojw28:
This code looks like it would be a valid thing to create and inject from an individual application. It doesn't look particularly re-usable though (e.g. it seems to cap bit-rate to an arbitrary fixed limit), so I don't think it makes sense for it to be part of the core library.

PR #3041, opened 2017-07-10
Change module prefix to use RTMP extension in other projects

Schmak:


ojw28:
We just simultaneously fixed this, heh. Thanks though :).

ojw28:
Actually, something else has gone wrong with our merging script that has caused the extension to largely disappear. Investigating.

ojw28:
Fixed.

PR #3084, opened 2017-07-20
Use modulePrefix for the library-core dependency

brol1dev:
When working with the local ExoPlayer repository and using extensions, I had to add the `modulePrefix` to the `library-core` dependency so my app project could recognize it.

I saw that all other modules use `modulePrefix` and mediasession is the one missing it.

### DIFF
--- a/extensions/mediasession/build.gradle
+++ b/extensions/mediasession/build.gradle
@@ -25,7 +25,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


brol1dev:
I signed it! (CLA)

ojw28:
Thanks for spotting this! We've merged an equivalent fix.

Issue #2552, opened 2017-03-12
FfmpegAudioRenderer fails to correctly handle channel layout changes in MPEG-TS streams


kiall:
Possibly related to / a regression of https://github.com/google/ExoPlayer/issues/2111

### Issue description

When FfmpegAudioRenderer is used for decoding an AC3 audio track, and that track's channel layout changes abruptly (as is common on MPEG-TS streams during commercial breaks etc), audio and video begins to stutter.

It's possible this isn't AC3 specific, but I don't have any other samples available to verify.

### Reproduction steps

* Add provided sample into the demo player media.exolist.json, example:

        {
          "name": "AC3 Audio Layout Change (TS)",
          "uri": "http://10.5.1.21/ac3-channel-layout-change.ts"
        },

* Build the demo app, with the ffmpeg extension enabled.
* Play the sample, wait for approx 1:01 and observe stuttering.

### Link to test content

https://www.dropbox.com/s/0uvzypw9ckusnqr/ac3-channel-layout-change.ts?dl=0

### Version of ExoPlayer being used

* ExoPlayer r2.2.0
* FFmpeg git sha 6c7a0876fefdd3c919afcabf493492cb4946ca6a

### Device(s) and version(s) of Android being used

Reproduces on:

* Android TV Emulator - v7.1.1
* nVidia Shield - Android 7.0, Shield SW 5.1 (24.28.403.124)

### A full bug report captured from the device

https://www.dropbox.com/s/py4bpiqds4pbb3g/bugreport-NYC-2017-03-12-22-55-21.zip?dl=0

kiall:
It's also worth noting that if the audio is allowed to passthrough to the TV attached to the shield for decoding (and downmixing to 2 channel), the sample plays back fine.

kiall:
FYI - I'll be travelling, and away from any Android TV dev stuff starting Friday for 2 weeks - any responses   needed will be slow, apologies.

Let me know if you need anything and I'll do my best!

kiall:
Hey guys, anything you need from me to confirm this issue etc? Thanks!

AquilesCanta:
Sorry for the delay on this. I think I know what the issue might be. I'll run some tests and update this issue.

kiall:
Awesome, thanks @AquilesCanta :)

AquilesCanta:
A fix will be available in the next push.

ojw28:
Fixed in `dev-v2`. Please give it a try!

[...]

PR #3085, opened 2017-07-21
Support 7.1 EAC3 in MP4, TS, DASH and HLS formats

ybai001:
This update will provide below supports to ExoPlayer
- MPEG-4 with 7.1-channel EAC3 playback
- ID3v2 with 7.1-channel EAC3 playback, which used for HLS late binding
- Parsing DASH manifest with Dolby specific element

Change-Id: If7ed0999ababccd5ecff38dcdd5bcbc089529142


ybai001:
I signed it!

ojw28:
Thanks for sending this! If this is under a new corporate CLA, there may be a bit of a delay before the bot detects that it's been signed.

ybai001:
Hi, ojw28,

I was told that Dolby has signed contributor license agreement with Google. And I don't need to sign the individual contributor license.
Dolby administrator has added me into the Dolby contributor group and he said it was OK for me to contribute code from that time.
Is there anything else I need to do? :)

Thanks.

andrewlewis:
It takes a few days for corporate CLAs to update. I'll follow up to check the status of this one. Thanks!

andrewlewis:
@ybai001 The commit author email address is not in the CLA group. Please could you amend the commit be authored by your ybai@... email address, and make sure that address is associated with your GitHub account? Or, alternatively, arrange to have the email address from the commit added to the CLA group?

ybai001:
@andrewlewis Thanks. I know the reason. I used my long email address (yanning.bai@...) to do the commit. But my admin added my short email address (ybai@...) into the CLA group. I'll ask him to add my long email address to CLA group too.

ojw28:
Looks good now. We'll take a look at reviewing this asap. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
```diff
@@ -188,18 +192,36 @@ private boolean skipToNextSync(ParsableByteArray pesBuffer) {
...
+      maybeReconfigure = false;
+    }
     sampleSize = frameInfo.frameSize;
     // In this class a sample is an access unit (syncframe in AC-3), but the MediaFormat sample rate
     // specifies the number of PCM audio samples per second.
     sampleDurationUs = C.MICROS_PER_SECOND * frameInfo.sampleCount / format.sampleRate;
+    // If the (E-)AC-3 stream is 7.1-channel, the dependent sub-stream shares the sampleDuration
+    // with independent sub-stream.
+    if (frameInfo.streamType == Ac3Util.STREAM_TYPE_TYPE1) {
+      sampleDurationUs = 0;
```
ojw28, line 223: This means timeUs is only incremented in each packetStarted call. Is that intentional? I'm not really sure I understand the comment above.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
```diff
@@ -188,18 +192,36 @@ private boolean skipToNextSync(ParsableByteArray pesBuffer) {
   private void parseHeader() {
     headerScratchBits.setPosition(0);
     Ac3Util.Ac3SyncFrameInfo frameInfo = Ac3Util.parseAc3SyncframeInfo(headerScratchBits);
-    if (format == null || frameInfo.channelCount != format.channelCount
-        || frameInfo.sampleRate != format.sampleRate
-        || frameInfo.mimeType != format.sampleMimeType) {
+    if ((format == null || frameInfo.sampleRate != format.sampleRate
+        || frameInfo.mimeType != format.sampleMimeType)
```
ojw28, line 196: Hm. The idea of this block was (I think) to output a new format whenever something needs to change in it. That doesn't look like it's the case any more. In particular, a change to only channelCount doesn't look like it will cause a new format to be output. I'm not sure if I'm just misunderstanding the goal, but it feels like this code should look more like:

```
Ac3Util.Ac3SyncFrameInfo frameInfo = Ac3Util.parseAc3SyncframeInfo(headerScratchBits);
int formatChannelCount = frameInfo.channelCount;
if (frameInfo.streamType == Ac3Util.STREAM_TYPE_TYPE1
    && format.channelCount == 6 && frameInfo.channelCount == 4) {
  formatChannelCount = 8;
}
if (format == null || formatChannelCount != format.channelCount
   || frameInfo.sampleRate != format.sampleRate
   || frameInfo.mimeType != format.sampleMimeType) {
  .... Create and output the format
}
```

Does that make any sense?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -901,6 +903,33 @@ protected static String parseString(XmlPullParser xpp, String name, String defau
     return value == null ? defaultValue : value;
   }
 
+  protected static int parseDolbyChannelConfiguration(XmlPullParser xpp, String name,
+      int defaultValue) {
+    String value = xpp.getAttributeValue(null, name).toUpperCase();
+    int channels;
```
ojw28: nit - Move this line to just above the switch statement (~L913)

ojw28 (commented):
Looks good! I'm a little unclear about the Ac3Reader change. Please clarify. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ojw28: I guess my suggestion will NullPointerException when checking format.channelCount==6. It could probably be fixed though!

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
```diff
@@ -188,18 +192,36 @@ private boolean skipToNextSync(ParsableByteArray pesBuffer) {
...
       format = Format.createAudioSampleFormat(trackFormatId, frameInfo.mimeType, null,
           Format.NO_VALUE, Format.NO_VALUE, frameInfo.channelCount, frameInfo.sampleRate, null,
           null, 0, language);
       output.format(format);
     }
+
+    // If the (E-)AC-3 stream contains I0 + D0, and I0.channels = 6, D0.channels = 4,
+    // then it's 7.1-channel (E-)AC-3 stream.
+    // TODO: handle other (E-)AC-3 streams?
+    if (maybeReconfigure == true && frameInfo.streamType == Ac3Util.STREAM_TYPE_TYPE1
```
ojw28, line 208: nit - `maybeReconfigure == true` is the same as just `maybeReconfigure` :).

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ojw28: I think it's fairly likely I just don't understand how this works! What do these headers look like? I guess there must be more than one of them?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
ybai001: OK

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ybai001: sure

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ybai001: Dolby's 7.1 EAC3 stream consists of I(Independent frame) and D(dependent frame).  These two frames cross appear in the stream, i.e. I,D,I,D,......
When first I frame occurs, format = null, streamType = TYPE0. So one format is created. At this time, channelCount = 6. 
Then first D frame occurs, streamType = TYPE1, channelCount = 4. So one new format is created to replace the old one.
When second I frame occurs, format != null. So no new format is created.
When second D frame occurs, maybeReconfigure = false. So no new format is created.
......
I remember channel count in each I frame (and D frame) can't be changed in AC3/EAC3 stream. But I need to double check with the original author of this piece of code. If this is not true, I think we need to modify this code a little.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -901,6 +903,33 @@ protected static String parseString(XmlPullParser xpp, String name, String defau
     return value == null ? defaultValue : value;
   }
 
+  protected static int parseDolbyChannelConfiguration(XmlPullParser xpp, String name,
+      int defaultValue) {
+    String value = xpp.getAttributeValue(null, name).toUpperCase();
```
ojw28: This probably gives a lint warning, because toUpperCase behaves differently depending on the phone's locale. I'd suggest using `Util.toLowerInvariant` instead, which will convert to lower case explicitly using the US English locale to make sure it behaves the same regardless of the phone locale. Obviously you'd need to make the constants below lowercase too. Alternatively, you can use toUpperCase(Locale.US) here.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ybai001: Dolby's 7.1 EAC3 stream consists of I(Independent frame) and D(dependent frame). These two frames cross appear in the stream, i.e. I,D,I,D,......
One I frame and D frame is an access unit, which is a basic unit for decoder. ( For example, for 7.1ch EAC3 stream, I frame includes L,R,C,LFE,Ls',Rs' channels. Decoder drops Ls' and Rs' channel data since they are used for 5.1 downmix output. D frame includes Ls,Rs,Lb,Rb. The final output is L,R,C,LFE,Ls,Rs,Lb,Rb (total 8 channels) ) 
When first I frame occurs, sampleDurationUs is calculated. 
Then first D frame occurs, sampleDurationUs is 0 since D frame doesn't increase more samples. It just add more channels for previous I frame. I+D is a whole access unit from decoder perspective.

I hope it explains the code above. :)

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
ybai001: OK, thank you for pointing it out.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
ybai001: I asked the original author today. He said he didn't think the channel count should be changed in EAC3 stream. (Although it can be changed in spec. But the real stream in market doesn't change it.) Shall we support this case? If yes, I have below updated code:

    Ac3Util.Ac3SyncFrameInfo frameInfo = Ac3Util.parseAc3SyncframeInfo(headerScratchBits);
    int formatChannelCount = frameInfo.channelCount;
    if (format != null && frameInfo.streamType == Ac3Util.STREAM_TYPE_TYPE1
        && lastStreamType == Ac3Util.STREAM_TYPE_TYPE0
        && lastChannelCount == 6 && formatChannelCount == 4) {
      formatChannelCount = 8;
    }    
    if (format != null && frameInfo.streamType == Ac3Util.STREAM_TYPE_TYPE0
        && lastStreamType == Ac3Util.STREAM_TYPE_TYPE1    
        && lastChannelCount == 4 && formatChannelCount == 6) {
      formatChannelCount = 8;
    }    
    if (format == null || formatChannelCount != format.channelCount
       || frameInfo.sampleRate != format.sampleRate
       || frameInfo.mimeType != format.sampleMimeType) {
      format = Format.createAudioSampleFormat(trackFormatId, frameInfo.mimeType, null,
          Format.NO_VALUE, Format.NO_VALUE, formatChannelCount, frameInfo.sampleRate, null,
          null, 0, language);
      output.format(format);
    }
    lastStreamType = frameInfo.streamType;
    lastChannelCount = frameInfo.channelCount;

ojw28: I think we should avoid the case where two formats are output, one with the incorrect channel count, then a following one with the correct channel count. We should be waiting until we know what the format should be, and just outputting that.

Also, if "I,D" is an access unit, should the reader be outputting "I,D" in a single sample, rather than two separate samples as appears to be the case currently? Note this removes the need for setting sampleDurationUs to 0 below.

By the way, if you want to get the changes in the other two files submitted ahead of this discussion being concluded, feel free to send those as a separate pull request. I think they're ready to merge.

Thanks!

ybai001: Let me think about it and discuss with my colleagues. (Seems that it is hard for TS stream.)  If there is no good solution, I'll submit a new pull request for other two files at first.

ybai001: Hi, ojw28,

After discuss with my colleagues, all of us think it is difficult to output "I,D" in a single sample based on the current ExoPlayer logic. The reasons is both I frame and D frame share the same sync word "0x0B77". There is no unique sync word to identify the beginning of one access unit. (Unless you parse the header in bitstream.) This is the limitation of EAC3 format. If we want to outputting "I,D" in a single sample, it seems that we need to re-design the state machine in "consume" method. (Maybe you have better idea? :)) This modification is big. 

BTW, one information for your reference. Based on the EAC3 spec, in theory, one access unit can include maximum 8 I frames with 64 D frames, which are I0,D0,D1,...,D7,I1,D0,D1,...,D7,I2,D0,D1,...,D7,...I7,D0,D1,...,D7  
But in reality, I didn't find this kind of stream. In most cases, I think supporting "I,D" is enough. (This is for 7.1 EAC3 stream)

Finally, it is up to you to decide whether accepting our current modification. If NOT, please feel free to let me know and I'll submit a new pull request for other two files modification.

And thanks for your carefully code review and detailed comments. It's truly helpful. 

Yanning

ojw28: Thanks for all the information! I think we'd like to accept all of the changes, but we'd also like to think about this particular change in a bit more detail before we do, and we'll need to find some time for that :).

It would be good to split the changes to the other two files into a separate pull request, just so that we can get those changes merged rather than them being held up whilst we figure out this final piece. If you could leave this pull request intact, so the discussion thread doesn't disappear anywhere, that would be great, and we'll follow up here regarding the final piece once we've mulled it over a bit more.

Thanks for your help!

ybai001: OK, I'll submit a new pull request and keep this one at the same time.

ybai001: The new pull request 3161 is created.

ojw28:
@ybai001 - Is it possible for you to resolve the conflicts on the change, and get it to the point where it consists of just what we didn't merge separately? Let's try and come to a conclusion here soon :). Thanks!

ybai001:
Sure, I'll do it.

ybai001:
@ojw28 conflicts are resolved. Let's continue discussion. :)

ojw28:
Thanks! Discussing internally; we'll post an update here soon :).

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac3Reader.java
andrewlewis: Issue #2552 asked for support for the channel configuration changing during playback so it would be good to make sure this still works.

I wonder whether grouping together "I,D" into a single output sample could actually be fairly easy to do:
- It looks like we do actually now parse the header for every syncframe, which means that we know the stream type of the syncframe being read.
- At the moment we just call output.sampleMetadata when we've read sampleSize bytes, but instead we could delay calling it until we read a syncframe with stream type == 0. It would be necessary to keep track of the number of bytes read from the current syncframe and the number of bytes read for the access unit separately.
- We'd also need a way to make sure we output the last sample in the stream (as we won't read a next syncframe for this one).
- This also means we can delay the call to output.format until we've read all constituent syncframes of the access unit, which avoids the problem of outputting two formats. (The first sample would likely just get dropped in the AudioTrack when it's reconfigured with the second format, which is bad.)

I haven't tried this, so maybe something will come up that makes it impossible, but worth attempting I think. Thanks!

ybai001: andrewlewis, thanks for this idea. If you can group together "I,D" into a single output sample, that is fine.

ojw28: Whether it's ok to group "I,D" into a single output sample I think depends on whether the Dolby decoder can accept input buffers with "I,D" grouped in a single sample. Can it?

ybai001: Yes, it can. The Dolby decoder's behavior is something as below. (pseudo code)

dolby_decoder_wrapper_process(const char* p_inBuffer, const int inBufferLength, char* p_outBuffer, int* p_bytesConsumed)
{
    ......
    int timesliceComplete;
    dolby_decoder_api_addbytes(p_inBuffer, inBufferLength, p_bytesConsumed, &timesliceComplete);
    if (timeslicecomplete) { 
        dolby_decoder_api_process_timeslice(p_outBuffer);
    }
    ......
}
The first Dolby decoder API fetches the input data. As long as it gets enough data (one full timeslice and one another full I0), it will set timesliceComplete = true.
For example, in following cases, it will set timesliceComplete = true.
(1) I0, I0  (this is general 5.1 case)
(2) I0, D0, I0 (this is general 7.1 case) 
(3) I0, D0, I1, D0, I0
Once timesliceComplete is true, it will call process() API to do decoding. (Decoder will decoder the first timeslice. The second I0 just keeps in internal buffer and will be decoded next time.)  Otherwise, it just waits for more data to be inputted.

andrewlewis: @ybai001 It sounds like grouping "I,D" together as described above will work in that case. Do you have any plans to update the pull request to implement this? Thanks!

ybai001: @andrewlewis We have no plan to do this update recently due to the internal priority. You may close this pull request at first. (This pull request includes three parts of update. The first & third parts of update have been merged by https://github.com/google/ExoPlayer/pull/3161.) After we implement this new logic, we wil submit a new pull-request. Is it OK for you?

andrewlewis: Sounds good. Thanks.

andrewlewis:
For the record, closing per discussion in comments on `Ac3Reader` (see above).

Issue #2565, opened 2017-03-16
Support CEA-608 embedded in H262 in MPEG-TS


jvdave23:
### Issue description
I have a few sample ts files that have CEA 608 closed captioning available in them. I know this because when I play it using VLC player for Android, it is able to successfully display the captions. However, VLC player has frame rate issues when rendering the captions which is why I'd like to use Exoplayer. I've hosted the files on a server, and then tried using the demo app to play the files. The demo app doesn't recognise the captions in the file.  

To confirm that it wasn't a server issue, I used one of the samples provided in the exoplayer app :  

https://tungsten.aaplimg.com/VOD/bipbop_adv_example_v2/master.m3u8

I downloaded the first 6 to 7 segments of the m3u8 file, concatenated them into a bigger ts file, uploaded it to same server as a ts file, and tried playing it using the Exoplayer demo app. The player was able to successfully render the captions from it using ExtractorMediaPeriod.

I found the point at which my sample and the concatenated sample differ in the code. At TsExtractor::run() , TsPayloadReader payloadReader is null which is why payloadReader.consume() is never called which would eventually lead to reading the captions. I don't understand what the payloadReader is so I was hoping to get some information from the dev team on that. 

### Reproduction steps
I added my sample ts files as entries in the media.exolist.json which I will send to dev.exoplayer@gmail.com. Once you have the json file in your demo app, you will see a new subcategory that has all the sample files being used in the investigation of this issue.

### Link to test content
I will send the test links to the email provided.

### Version of ExoPlayer being used
r2.1.1

### Device(s) and version(s) of Android being used
Tested on Samsung Galaxy Tab S2 - Android version 6.0.1
Tested on Nexus 5X - Android version 7.1.1
Tested on Client provided device with codecs to play video and audio - version 5.0.2

### A full bug report captured from the device
Will provide if needed but don't think it is relevant.

jvdave23:
Some more info about the sample files:

You may run into issues playing the video. It is MPEG2 video and MPEG1-L2 audio. There is one additional device we are using which is provided by our client. It has a custom Android OS on it, and some codecs that might not be available on all Android devices.

AquilesCanta:
[...]

PR #3114, opened 2017-07-27
Support CEA-608 closed-captions in H262/MPEG-TS.

goffioul:
Issue: https://github.com/google/ExoPlayer/issues/2565


AquilesCanta:
Looks nice! Thanks. Will get to this soon. Possibly next week or the one after that.

SVPA-MaxWu:
any plan to support 708 as well?

goffioul:
AFAIK 708 is supported, although I never tested it myself. However it's not used by default and/or it depends on whether your stream has the proper data in the stream descriptor (in case of MPEG/TS). To have 708 decoder enabled, you have to make sure that the SeiReader object is created with the appropriate mimetype: if your stream does not have appropriate descriptors and you don't provide the list of mimetypes explicitly when creating SeiReader, then it'll default to 608 only.

In all cases, this is independent from how the data is extracted from the stream, which was the point of this pull request.  608 and 708 embeds data similarly in MPEG2 and H264 streams. It's how the data is interpreted afterwards that makes a difference.

ojw28:
@AquilesCanta - Planning to take a look at this?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/CeaUtil.java
```diff
@@ -41,7 +65,7 @@
    * @param seiBuffer The unescaped SEI NAL unit data, excluding the NAL unit start code and type.
    * @param outputs The outputs to which any samples should be written.
    */
-  public static void consume(long presentationTimeUs, ParsableByteArray seiBuffer,
+  private static void consumeSei(long presentationTimeUs, ParsableByteArray seiBuffer,
```
AquilesCanta, line 68: Could we leave the javadocs for consume?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/CeaUtil.java
```diff
@@ -33,6 +37,26 @@
...
   private static final int USER_DATA_TYPE_CODE = 0x3;
 
+  public static void consume(long presentationTimeUs, ParsableByteArray buffer,
+      TrackOutput[] outputs, int mode) {
+    switch (mode) {
+      case MODE_H262:
+        consumeUserData(presentationTimeUs, buffer, outputs);
+        break;
+      case MODE_H264:
+        consumeSei(presentationTimeUs, buffer, outputs);
```
AquilesCanta, line 47: Considering the modes are actually only 2 and they are always known at compile time, why not just have a consumeUserData and consumeSei made public? If so, public methods should be documented.

okulagin:
@goffioul  any thought when comments left by @AquilesCanta will be addressed?

goffioul:
@okulagin I don't really have the time at the moment. It should be trivial to address @AquilesCanta 's comments, though. Feel free to do it, if you wish.

ojw28:
It looks likely support for this will be added by https://github.com/google/ExoPlayer/pull/4308.

PR #3136, opened 2017-08-04
Add possibility of forcing a specific license URL in HttpMediaDrmCallback

jeoliva:
- Renamed some license URL related variables to keep consistency across the code.
- Added a new parameter to HttpMediaDrmCallback that enables forcing defaultLicenseURL as the license acquisition URL.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
@@ -47,29 +47,29 @@ public final class HttpMediaDrmCallback implements MediaDrmCallback {
   }
 
   private final HttpDataSource.Factory dataSourceFactory;
-  private final String defaultUrl;
+  private final String defaultLicenseUrl;
+  private final boolean forceDefaultLicenseUrl;
   private final Map<String, String> keyRequestProperties;
 
   /**
-   * @param defaultUrl The default license URL.
+   * @param defaultLicenseUrl The default license URL.
    * @param dataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
    */
-  public HttpMediaDrmCallback(String defaultUrl, HttpDataSource.Factory dataSourceFactory) {
-    this(defaultUrl, dataSourceFactory, null);
+  public HttpMediaDrmCallback(String defaultLicenseUrl, HttpDataSource.Factory dataSourceFactory) {
+    this(defaultLicenseUrl, false, dataSourceFactory, null);
   }
 
   /**
-   * @deprecated Use {@link HttpMediaDrmCallback#HttpMediaDrmCallback(String, Factory)}. Request
-   *     properties can be set by calling {@link #setKeyRequestProperty(String, String)}.
-   * @param defaultUrl The default license URL.
+   * @param defaultLicenseUrl The default license URL.
+   * @param forceDefaultLicenseUrl Whether to force use of {@code defaultLicenseUrl} even for key
+   *     requests that include their own license URL.
    * @param dataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
-   * @param keyRequestProperties Request properties to set when making key requests, or null.
    */
-  @Deprecated
-  public HttpMediaDrmCallback(String defaultUrl, HttpDataSource.Factory dataSourceFactory,
-      Map<String, String> keyRequestProperties) {
+  public HttpMediaDrmCallback(String defaultLicenseUrl, boolean forceDefaultLicenseUrl,
+      HttpDataSource.Factory dataSourceFactory, Map<String, String> keyRequestProperties) {
     this.dataSourceFactory = dataSourceFactory;
-    this.defaultUrl = defaultUrl;
+    this.defaultLicenseUrl = defaultLicenseUrl;
+    this.forceDefaultLicenseUrl = forceDefaultLicenseUrl;
     this.keyRequestProperties = new HashMap<>();
     if (keyRequestProperties != null) {
       this.keyRequestProperties.putAll(keyRequestProperties);
@@ -120,8 +120,8 @@ public final class HttpMediaDrmCallback implements MediaDrmCallback {
   @Override
   public byte[] executeKeyRequest(UUID uuid, KeyRequest request) throws Exception {
     String url = request.getDefaultUrl();
-    if (TextUtils.isEmpty(url)) {
-      url = defaultUrl;
+    if (forceDefaultLicenseUrl || TextUtils.isEmpty(url)) {
+      url = defaultLicenseUrl;
     }
     Map<String, String> requestProperties = new HashMap<>();
     requestProperties.put("Content-Type", "application/octet-stream");
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/OfflineLicenseHelper.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/OfflineLicenseHelper.java
@@ -43,23 +43,44 @@ public final class OfflineLicenseHelper<T extends ExoMediaCrypto> {
    * Instantiates a new instance which uses Widevine CDM. Call {@link #release()} when the instance
    * is no longer required.
    *
-   * @param licenseUrl The default license URL.
+   * @param defaultLicenseUrl The default license URL.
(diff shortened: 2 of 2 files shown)


library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
```diff
@@ -47,29 +47,36 @@
   }
 
   private final HttpDataSource.Factory dataSourceFactory;
-  private final String defaultUrl;
+  private final String defaultLicenseUrl;
+  private final Boolean forceDefaultLicenseUrl;
```
ojw28: Why isn't this a boolean (lowercase b)? Ditto for constructor args.

library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
```diff
@@ -47,29 +47,36 @@
...
   /**
-   * @param defaultUrl The default license URL.
+   * @param defaultLicenseUrl The default license URL.
+   * @param forceDefaultLicenseUrl If true, default license URL will be used in the license
+   *    acquisition process even do there is a license URL defined in the stream manifest.
    * @param dataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
    */
-  public HttpMediaDrmCallback(String defaultUrl, HttpDataSource.Factory dataSourceFactory) {
-    this(defaultUrl, dataSourceFactory, null);
+  public HttpMediaDrmCallback(String defaultLicenseUrl, Boolean forceDefaultLicenseUrl,
```
ojw28: Please can we keep the original constructor too, so you don't have to add false to any existing call sites?

library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
```diff
@@ -47,29 +47,36 @@
...
-  private final String defaultUrl;
+  private final String defaultLicenseUrl;
+  private final Boolean forceDefaultLicenseUrl;
   private final Map<String, String> keyRequestProperties;
 
   /**
-   * @param defaultUrl The default license URL.
+   * @param defaultLicenseUrl The default license URL.
+   * @param forceDefaultLicenseUrl If true, default license URL will be used in the license
+   *    acquisition process even do there is a license URL defined in the stream manifest.
```
ojw28: "stream manifest" is too specific to one of a number of possible use cases of this class. For the purpose of this class I think it should just read something like: "Whether to force use of {@code defaultLicenseUrl} even for key requests that include their own license URL."

nit: Indent should be 4 chars for a continuation of a comment, for consistency with rest of code.

library/core/src/main/java/com/google/android/exoplayer2/drm/OfflineLicenseHelper.java
```diff
@@ -43,16 +43,20 @@
...
+   * @param forceDefaultLicenseUrl If true, default license URL will be used in the license
+   *    acquisition process even do there is a license URL defined in the stream manifest.
    * @param httpDataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
    * @return A new instance which uses Widevine CDM.
    * @throws UnsupportedDrmException If the Widevine DRM scheme is unsupported or cannot be
    *     instantiated.
    */
   public static OfflineLicenseHelper<FrameworkMediaCrypto> newWidevineInstance(
-      String licenseUrl, Factory httpDataSourceFactory) throws UnsupportedDrmException {
+      String defaultLicenseUrl, Boolean forceDefaultLicenseUrl, Factory httpDataSourceFactory)
```
ojw28: Same comments apply to this class as in HttpMediaDrmCallback.

library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
```diff
@@ -47,29 +47,36 @@
...
    * @param dataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
    */
-  public HttpMediaDrmCallback(String defaultUrl, HttpDataSource.Factory dataSourceFactory) {
-    this(defaultUrl, dataSourceFactory, null);
+  public HttpMediaDrmCallback(String defaultLicenseUrl, Boolean forceDefaultLicenseUrl,
+                              HttpDataSource.Factory dataSourceFactory) {
```
ojw28: nit - Fix indent to be consistent with code style

ojw28 (commented):
Thanks! See minor comments inline.

jeoliva:
Thanks, makes sense. Applied all the suggested changes.

library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
```diff
@@ -47,29 +47,44 @@
...
+   *     requests that include their own license URL.
+   * @param dataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
+   */
+  public HttpMediaDrmCallback(String defaultLicenseUrl, boolean forceDefaultLicenseUrl,
+      HttpDataSource.Factory dataSourceFactory) {
+    this(defaultLicenseUrl, forceDefaultLicenseUrl, dataSourceFactory, null);
+  }
+
+  /**
+   * @deprecated Use {@link HttpMediaDrmCallback#HttpMediaDrmCallback(String, boolean, Factory)}.
```
ojw28: There's not much point keeping this method, given it's deprecated and you're changing the signature (a caller will need to change their code one way or another, so they should just start using the non-deprecated ones ;)). Can we remove it and merge the body into the one above?

library/core/src/main/java/com/google/android/exoplayer2/drm/OfflineLicenseHelper.java
```diff
@@ -43,16 +43,38 @@
...
    * @param httpDataSourceFactory A factory from which to obtain {@link HttpDataSource} instances.
    * @return A new instance which uses Widevine CDM.
    * @throws UnsupportedDrmException If the Widevine DRM scheme is unsupported or cannot be
    *     instantiated.
    */
   public static OfflineLicenseHelper<FrameworkMediaCrypto> newWidevineInstance(
-      String licenseUrl, Factory httpDataSourceFactory) throws UnsupportedDrmException {
+      String defaultLicenseUrl, Factory httpDataSourceFactory)
+      throws UnsupportedDrmException {
+    return newWidevineInstance(
```
ojw28: Can this just be:

return newWidevineInstance(defaultLicenseUrl, false, httpDataSourceFactory);

ojw28 (commented):
Couple more comments, and we'll get this merged. Thanks for the quick turnaround!

library/core/src/main/java/com/google/android/exoplayer2/drm/HttpMediaDrmCallback.java
jeoliva: Makes sense!

library/core/src/main/java/com/google/android/exoplayer2/drm/OfflineLicenseHelper.java
jeoliva: I thought about that but and finally let the flag there because, although I never faced this requirement for offline content, I thought it could be useful in the same way it is for non offline content.

jeoliva: Let's remove it until there is a use case/requirement that justify this change.

jeoliva: Yes, makes sense. That method (newWidevineInstance) is not used in any other place and that change simplifies the code. Let me apply it.

jeoliva: Done! ;)

ojw28:
This is being temporarily reverted to resolve an issue (of our own making ;)) with the synchronization between our internal and external branches. Don't worry; it will be re-submitted within a day or two. Sorry for the churn!

jeoliva:
;)

jeoliva:
Hi @ojw28!

Quick question. Do you have a rough estimation regarding when this change will be merged in release branch? 

Thanks!

PR #3153, opened 2017-08-08
fix `SimpleExoPlayerView` doc error

bendecoste:
`getVideoSurfaceView` tells you that you can set `use_texture_view` to
true to use a `TextView`, but this should be `surface_type`


bendecoste:
I signed it!

ojw28:
Thanks for pointing this out. I think the correction is also inaccurate, however, since `surface_type` is not a boolean attribute. I pushed https://github.com/google/ExoPlayer/commit/ebbf9f635bdbb1b7b4ba418955bc5be207afeca6 to fix instead.

PR #3161, opened 2017-08-10
Support 7.1 EAC3 in MP4, DASH and HLS formats

ybai001:
This update will provide below supports to ExoPlayer
- MPEG-4 with 7.1-channel EAC3 playback
- ID3v2 with 7.1-channel EAC3 playback, which used for HLS late binding
- Parsing DASH manifest with Dolby specific element

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -690,7 +690,9 @@ public class DashManifestParser extends DefaultHandler
       throws XmlPullParserException, IOException {
     String schemeIdUri = parseString(xpp, "schemeIdUri", null);
     int audioChannels = "urn:mpeg:dash:23003:3:audio_channel_configuration:2011".equals(schemeIdUri)
-        ? parseInt(xpp, "value", Format.NO_VALUE) : Format.NO_VALUE;
+        ? parseInt(xpp, "value", Format.NO_VALUE) :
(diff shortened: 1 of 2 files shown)


ybai001:
This new pull request is based on the discussion in pull request 3085. I removed the 7.1 EAC3 TS handling part. 
And update code based on two ojw28's comments.
1. file DashManifestParser.java: nit - Move this line(L909) to just above the switch statement (~L913)
2. file DashManifestParser.java: line 908, using Util.toLowerInvariant instead

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -901,6 +903,33 @@ protected static String parseString(XmlPullParser xpp, String name, String defau
...
+    if (value == null) {
+      return defaultValue;
+    }
+    int channels;
+    // TODO: Parse other channel configurations
+    switch (value) {
+      case "4000":
+        channels = 1;
+        break;
+      case "A000":
```
ojw28: Now you're using toLowerInvariant on L908, these case constants also need to be lowercase :)

ybai001:
Thanks. I'm stupid. :)
Case constants are lowercase now.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
ybai001: Thanks. I'm stupid. :)
Case constants are lowercase now.

PR #3169, opened 2017-08-14
Scrubber should be antialias

LqcIce:
When I use demo in my work,I found the scrubber has some antialias.It should be antialias.Then the scrubber can be better.


LqcIce:
I signed it!

ojw28:
It's already set to anti-alias in the `dev-v2` branch (in the constructor).

LqcIce:
Thank you for telling me. I do this in release but forget to check this in dev.

PR #3187, opened 2017-08-17
Support zoom mode for AspectRatioFrameLayout

stellabei:
The AspectRatioFrameLayout does not provide resize mode that offer full screen mode respecting the aspect ratio. It is specially useful for mobile screens like Samsung S8 or LG G6 (aspect ratio 18.5:9).

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
@@ -32,7 +32,7 @@ public final class AspectRatioFrameLayout extends FrameLayout {
    * Resize modes for {@link AspectRatioFrameLayout}.
    */
   @Retention(RetentionPolicy.SOURCE)
-  @IntDef({RESIZE_MODE_FIT, RESIZE_MODE_FIXED_WIDTH, RESIZE_MODE_FIXED_HEIGHT, RESIZE_MODE_FILL})
+  @IntDef({RESIZE_MODE_FIT, RESIZE_MODE_FIXED_WIDTH, RESIZE_MODE_FIXED_HEIGHT, RESIZE_MODE_FILL, RESIZE_MODE_CROP})
   public @interface ResizeMode {}
 
   /**
@@ -51,6 +51,10 @@ public final class AspectRatioFrameLayout extends FrameLayout {
    * The specified aspect ratio is ignored.
    */
   public static final int RESIZE_MODE_FILL = 3;
+  /**
+   * The height or width is increased or decreased to crop and to obtain the desired aspect ratio.
+   */
+  public static final int RESIZE_MODE_CROP = 4;
 
   /**
    * The {@link FrameLayout} will not resize itself if the fractional difference between its natural
@@ -97,6 +101,15 @@ public final class AspectRatioFrameLayout extends FrameLayout {
   }
 
   /**
+   * Gets the resize mode.
+   *
+   * @return The resize mode.
+   */
+  public int getResizeMode() {
+    return this.resizeMode;
+  }
+
+  /**
    * Sets the resize mode.
    *
    * @param resizeMode The resize mode.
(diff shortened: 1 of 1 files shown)


ojw28:
If I remember correctly, there were a few reasons why we didn't add support for this cropping mode:

1. I seem to remember it causes larger surface allocations in memory. Which is probably fine if you're just cropping a bit of the video, but may be problematic if you're cropping a lot of it.
1. I'm unsure what happens when you allow a surface to be larger than its parent container in the view hierarchy. Is the surface correctly clipped to within the parent's bounds (fine), or does it just punch a hole through whatever it ends up overlapping with (not fine)?

How have you found using this approach in general? Does it work well across all devices you've tested on? Thanks!

stellabei:
Hi @ojw28, thanks for taking time to review my request! Here are the answers for your 2 questions:
1. Yes, it would use more memory. And it would be bad if movies can be any kind of aspect ratio, but most of movies have standard 16:9 aspect ratio. And most of mobile devices/tablet aspect ratio is similar except those S8 and LG G6 that has 18.5:9. So the cropping area is not that much. I'm considering movie aspect ratio here because most of DRM video content would be movies. Otherwise the developer can use TextureView to make this zoom-in effect instead of SurfaceView.

2. I did several experiments to answer this question. My initial experiment was on a full screen case, so I could not be sure about the effect on the parent's bounds. Then I change the parent's bounds to be smaller than the screen size to see if the overlapping are is fine. And both experiments worked fine, that is, the surface is clipped correctly to within the parent's bounds. I tested this solution on different devices, Samsung S8, Samsung S7, Motorola devices, Tablets (Nexus 9 that has a different aspect ratio) and LG devices, and all of them worked fine. 

Hope that I answered your questions! Thanks!

ojw28:
Thanks. I think we'll merge this, although with the caveats that we're not certain it will work well everywhere.

stellabei:
Thank you @ojw28

library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
```diff
@@ -51,6 +51,10 @@
    * The specified aspect ratio is ignored.
    */
   public static final int RESIZE_MODE_FILL = 3;
+  /**
+   * The height or width is increased or decreased to crop and to obtain the desired aspect ratio.
```
ojw28: Should this be "Either the width or height is increased to obtain the desired aspect ratio" :)?

library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
```diff
@@ -51,6 +51,10 @@
...
+   */
+  public static final int RESIZE_MODE_CROP = 4;
```
ojw28: I actually wonder about the variable name too. From the point of view of this class, it's not really doing any cropping. It's just making its bounds bigger. Maybe it should be called RESIZE_MODE_EXPAND or something (not that that's a particularly good name for it either...).

library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
stellabei: Yes, true. A copy paste error :-)

library/ui/src/main/java/com/google/android/exoplayer2/ui/AspectRatioFrameLayout.java
stellabei: RESIZE_MODE_ZOOM_IN or RESIZE_MODE_ASPECT_FILL? Maybe RESIZE_MODE_ASPECT_FILL is better since it is basically FILL with aspect ratio. ZOOM_IN might mislead API user to try to find a ZOOM_OUT mode... Please let me know which one is better to commit the change.

ojw28: `RESIZE_MODE_ZOOM`, without the `_IN`, would probably work pretty well :)?

stellabei: Yeap, sounds much better than ASPECT_FILL mode :)

ojw28:
Tweaked this slightly in 74b8c45e6dd9c13ed9a32a1aad1ce795535b2ac2. Don't think I broke anything, but let me know if I did!

PR #3217, opened 2017-08-29
Allow multiple PSSH boxes of different versions for same system in MP4

bangfalse:
Favors the highest version per system, or the first one of the highest version if multiple ones are found.

This allows MP4 files with Widevine PPSH boxes of both v0 and v1 to be played back successfully.


bangfalse:
I signed it!

bangfalse:
I signed it!

ojw28:
Can you help me understand why content would have both v0 and v1 pssh boxes for a single scheme? To me that kind of implies that some devices may need the v0 box and are for some reason unable to handle the v1 box. If that's true, then I wonder whether this is true for any older Android devices, in which case discarding the v0 box in the extractor wouldn't be the right thing to do. Any info to help fill in the gaps would be appreciated :). Thanks!

bangfalse:
I don't actually recall the situation which got us to having both v0 and v1 boxes in our files, but we have a large number which have these duplicate PSSH headers. I think it may have just been that we had Widevine documentation at the time which defined both versions, so we implemented both.

This code works on a Nexus 5 running 4.4.4, but I haven't done a survey of any earlier devices.

ojw28:
@wvpaf - Would you ever expect media to need to include both v0 and v1 PSSH boxes? Do both versions of the PSSH box work across all Android API levels?

bangfalse:
Can I get an update on any thinking on this? I'm happy to look for another approach if this isn't going to be taken, but we don't want to need to maintain our own fork long-term just for this one change.

wvpaf:
@ojw28, good point let me check with the team.

arashikou:
Hey there. @wvpaf asked me to take a look at this. To answer @ojw28's question above, support for PSSH box versions is down to the individual DRM Plugins and not governed by the Android API level itself, as the Android MediaDrm framework doesn't do any PSSH parsing of its own. Speaking of which, Widevine couldn't parse v1 PSSH boxes before Android M, (They hadn't been invented yet!) though you can still use the payload from a v1 PSSH box with Widevine on earlier OS releases by extracting the payload yourself rather than passing in the whole, unparsed PSSH box.

We have definitely seen content in the wild with both v0 and v1 PSSH boxes in it. (See [this Shaka Player thread](https://groups.google.com/forum/#!topic/shaka-player-users/DZuG1MzR2DA) for one example.) I can only speculate as to why people are doing this, but it's allowed by the spec and it's something we've seen people do with their content.

ojw28:
Thanks! @bangfalse - Yes, agreed you shouldn't need to maintain a fork for this. I think the right thing to do is probably to:

1. Allow `DrmInitData` to contain multiple `SchemeData` entries with the same `UUID`. If you place your version 0 PSSH before your version 1 PSSH (which I'm guessing you do, for maximum compatibility with devices that just take the first one?), then this should be all that's needed to get things working. You shouldn't need to make any changes to the FMP4 extractor.
1. Deprecate `DrmInitData.get(UUID)`, since it doesn't make much sense having completed the step above.
1. Make `DefaultDrmSessionManager` use `DrmInitData.get(index)` instead. For schemes we don't know anything special about, I guess we should just take the first `SchemeData` we find with a matching UUID. For Widevine, if there are multiple matching `SchemeData`s, we probably want to prefer a V0 PSSH box prior to M, and prefer V1 PSSH boxes post M, if we have both.

Does that sound right? If so, could you make a pull request for at least the first two steps :)? Feel free to have a stab at the third step also, although I'm happy to have a stab also if you prefer.

bangfalse:
Sorry for the delay. This does sound generally good. I will open a new pull request with an implementation that contains all 3 steps momentarily.

bangfalse:
Opened new PR as https://github.com/google/ExoPlayer/pull/3328

PR #3223, opened 2017-08-30
Add support for new codecs parameter string

Shyri:
Hi,
I was doing tests for VP9 in ISOBMFF container and I realized my tests assets were not working. I was wondering why Exoplayer test videos did work and I found the difference codecs string format. Exoplayer demos have 'codecs="vp9"' while mine has the new codecs parameter string 'codecs="vp09"' as specified here: https://www.webmproject.org/vp9/mp4/

Same think happened in shaka-player (https://github.com/google/shaka-player/issues/944)

With this change it started working again. It's simple change but it does the trick.


Shyri:
CLA Signed

Issue #3215, opened 2017-08-29
libstagefright.so SIGSEGV on various Android 7/7.1 devices after upgrading to 2.5.1


dbrain:
### Issue description
After upgrading ExoPlayer to 2.5.1 (from 2.4.4) we are seeing a lot of user reports of a `libstagefright.so` SIGSEGV. Playback starts, but a crash occurs when the player is released. This may be specific to `Widevine+DASH`, as the demo app will not crash using the `Youtube DASH` demos, while it crashes the same as our application on `Widevine+DASH` demos.

I have a feeling this may be related to the recent CENC support added:
https://github.com/google/ExoPlayer/issues/1989
But I also don't know what I'm talking about, so who knows ;)

I had a quick play at making `DefaultDrmSessionManager` return `false` for `canAquireSession` if the type was `cenc` etc. (that whole conditional), but ExoPlayer never seemed to call that method at all. Will fiddle a bit more, but thought I'd raise this ASAP.

A remote colleague has a device that can reproduce the crash (Motorola G5 Plus), but there's a bit of a lag getting debug information. Sorry if this contains limited information.

### Reproduction steps
1. Launch the demo app on an effected device
2. Start a Widevine+DASH demo (`WV: Secure SD & HD (cenc,MP4,H264)` definitely reproduces this)
3. Let playback begin
4. Press back leaving playback, releasing the player
5. Crash should occur

### Link to test content
Reproducible in the demo `WV: Secure SD & HD (cenc,MP4,H264)`, so not required.

### Version of ExoPlayer being used
2.5.1

### Device(s) and version(s) of Android being used
Seems to be 7.0/7.1 specific, but multiple devices. Easily reproducible, as it occurs from what I can tell any time when releasing the player after playback has begun.

Here are the play store device/version graphics:
![image](https://user-images.githubusercontent.com/237995/29806229-90d6f9c0-8cd1-11e7-9a9e-79a12989b6d1.png)

### A full bug report captured from the device
Here's a zipped `adb bugreport > libstagefright-widevine-dash-crash.log` from our application (`iflix.play.dbg`).
[libstagefright-widevine-dash-crash.zip](https://github.com/google/ExoPlayer/files/1259223/libstagefright-widevine-dash-crash.zip)

ojw28:
[...]

PR #3225, opened 2017-08-31
#3215 Additional secure DummySurface device exclusions

dbrain:
Proposed workaround for: https://github.com/google/ExoPlayer/issues/3215

Excludes all API 24/25 devices that do not have the `FEATURE_VR_MODE_HIGH_PERFORMANCE` system feature from having a secure `DummySurface`. This seems to be the cause of a `libstagefright.so` SIGSEGV in ExoPlayer 2.5.1 on certain 24/25 devices when releasing the player.

Verified this fixes the issue on one of the affected devices (Moto G5 Plus), but do not have access to the other 19+. My other (API: 26, 18, 19) devices lying around are also working as expected.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/DummySurface.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/DummySurface.java
@@ -43,6 +43,7 @@ import static android.opengl.GLES20.glGenTextures;
 
 import android.annotation.TargetApi;
 import android.content.Context;
+import android.content.pm.PackageManager;
 import android.graphics.SurfaceTexture;
 import android.graphics.SurfaceTexture.OnFrameAvailableListener;
 import android.opengl.EGL14;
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/video/DummySurface.java
```diff
@@ -152,9 +153,16 @@ private static void assertApiLevel17OrHigher() {
    *
    * @param context Any {@link Context}.
    */
-  @SuppressWarnings("unused") // Context may be needed in the future for better targeting.
+  @SuppressWarnings("unused")
   private static boolean deviceNeedsSecureDummySurfaceWorkaround(Context context) {
-    return Util.SDK_INT == 24 && "samsung".equals(Util.MANUFACTURER);
+    return (Util.SDK_INT == 24 && "samsung".equals(Util.MANUFACTURER))
+            || (Util.SDK_INT >= 24 && Util.SDK_INT < 26
```
ojw28, line 159: I reverted this change by accident (sorry!), although having looked at it some more I'm not sure it's right that this method returns false on API levels below 24 just because we can't check the VR flag. If secure dummy surface didn't work reliably on devices without the VR flag on API levels 24 and 25, I doubt it worked reliable prior to the flag even having existed :).

It's likely that `EGL_EXT_protected_content` wasn't advertised prior to API level 24, so perhaps this is a hypothetical case, but seems worth disabling anyway. I'll push a new CL to un-revert and make this change (and will ref it in the related issue). Thanks!

dbrain: No probs, whatever works is good by me. Previously it would only ever return `true` if the SDK was 24 and it was a Samsung device, so I figured it was safe to return false <24 until we found a <24 device that required the secure dummy surface workaround.

Issue #3224, opened 2017-08-31
AdaptiveTrackSelection failed to switch up for live hls


james12695:
I've tried the ExoPlayer demo app with my LIVE HLS with 3s segment and failed to switch up high quality track due to insufficient bufferedDurationUs in `AdaptiveTrackSelection`,  and found the similar issues #2477 and #3017 but both are marked as enhancement. So can we mark this issue as `bug` instead of `enhancement` and have a rough plan for this fix?
Thanks

ojw28:
Duplicate of #3017. Please avoid filing duplicate issues.

PR #3227, opened 2017-08-31
Fix AdaptiveTrackSelection failed to switch up

james12695:
Proposed workaround for: #3224

Update bufferedDurationUs and parameter minDurationForQualityIncreaseUs in AdaptiveTrackSelection and verified this fix for 3s, 10s and 30s HLS segment respectively.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -88,7 +88,7 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   }
 
   public static final int DEFAULT_MAX_INITIAL_BITRATE = 800000;
-  public static final int DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS = 10000;
+  public static final int DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS = 4000;
   public static final int DEFAULT_MAX_DURATION_FOR_QUALITY_DECREASE_MS = 25000;
   public static final int DEFAULT_MIN_DURATION_TO_RETAIN_AFTER_DISCARD_MS = 25000;
   public static final float DEFAULT_BANDWIDTH_FRACTION = 0.75f;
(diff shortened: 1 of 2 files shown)


james12695:
1.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
```diff
@@ -88,7 +88,7 @@ public AdaptiveTrackSelection createTrackSelection(TrackGroup group, int... trac
   }
 
   public static final int DEFAULT_MAX_INITIAL_BITRATE = 800000;
-  public static final int DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS = 10000;
+  public static final int DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS = 4000;
```
ojw28, line 91: I don't think it's a good idea just to change the default for all playbacks. This also affects on-demand playbacks, and live playbacks when the playback position is no-where near the live edge. It seems preferable to give the selector more information instead. For example if the selector were to know how close it is to the live edge, or something like that, then it could adapt its behavior accordingly.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -206,10 +206,9 @@ public void getNextChunk(HlsMediaChunk previous, long playbackPositionUs, HlsChu
     int oldVariantIndex = previous == null ? C.INDEX_UNSET
         : trackGroup.indexOf(previous.trackFormat);
     expectedPlaylistUrl = null;
-    // Use start time of the previous chunk rather than its end time because switching format will
-    // require downloading overlapping segments.
+    // Use end time instead because bufferedDurationUs will always be zero for 30s segment
```
ojw28, line 209: 30s segments are not sensible, and it does not make sense to optimize for this case.

As an example of why 30s segments aren't sensible, see the comment that was previously here: switching variant in HLS can require downloading overlapping segments. A 30s segment duration implies you'd need to wastefully download 30s of overlapping media to switch variant; which is clearly very wasteful.

Issue #3230, opened 2017-09-01
SmoothStreaming URLs that uses parenthesis "(...)" as part of the URL causes server side errors


nahojkap:
### Issue description
SmoothStreaming URLs that uses parenthesis "(...)" as part of the URL are not correctly preserved.  This kind of format has been seen when Microsoft MediaServices are used.  

For example, http://host/content.ism/manifest(filter=x) is changed by SsMediaSource.java into http://host/content.ism/manifest(filter=x)/Manifest which may produce an invalid URL towards the server side.

A simple solution would be to check for "manifest(...)" in the lastPathSegment and use the manifestUri directly if that is the case - something along the lines of

this.manifestUri = manifestUri == null ? null : lastPathSegment.contains("manifest") && lastPathSegment.contains("(") && lastPathSegment.contains(")")) ? manifestUri : lastPathSegment.equals("manifest") ? manifestUri : Uri.withAppendedPath(manifestUri, "Manifest");

### Reproduction steps
Use a manifest URL which ends with /manifest(filter=x), attempt playback, error is returned from remote server

### Link to test content
While the below link does not use the specified parameters inside the parenthesis, the URL will be rewritten by ExoPlayer and cause the error.

http://playready.directtaps.net/smoothstreaming/SSWSS720H264/SuperSpeedway_720.ism/manifest(filter=no)

### Version of ExoPlayer being used
SsMediaSource.java in current dev-v2 branch

### Device(s) and version(s) of Android being used
All devices

### A full bug report captured from the device
Not applicable

nahojkap:
Thanks @ojw28 for wrapping this up - Irma came along and messed with all our plans & I got somewhat behind.  Will let you know if we have any issues with this when we get around to integrating it.

ojw28:
Sorry to hear you were caught up in that. Thanks for your help on this, and yes, let us know if there are any remaining issues when you get back up and running.

PR #3231, opened 2017-09-01
#3230: SmoothStreaming URLs that uses parenthesis '(...)' as part of …

nahojkap:
…the URL causes server side errors

Here is the suggested fix for #3230 in a pull request.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
@@ -901,7 +901,8 @@ public final class Util {
     } else if (fileName.endsWith(".m3u8")) {
       return C.TYPE_HLS;
     } else if (fileName.endsWith(".ism") || fileName.endsWith(".isml")
-        || fileName.endsWith(".ism/manifest") || fileName.endsWith(".isml/manifest")) {
+        || fileName.endsWith(".ism/manifest") || fileName.endsWith(".isml/manifest")
+            || Uri.parse(fileName).getLastPathSegment().matches("manifest\\(.*\\)")) {
       return C.TYPE_SS;
     } else {
       return C.TYPE_OTHER;
--- a/library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
+++ b/library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
@@ -192,7 +192,9 @@ public final class SsMediaSource implements MediaSource,
       AdaptiveMediaSourceEventListener eventListener) {
     Assertions.checkState(manifest == null || !manifest.isLive);
     this.manifest = manifest;
-    this.manifestUri = manifestUri == null ? null : lastPathSegment.contains("manifest") && lastPathSegment.contains("(") && lastPathSegment.contains(")")) ? manifestUri : lastPathSegment.equals("manifest") ? manifestUri : Uri.withAppendedPath(manifestUri, "Manifest");
+    String lastPathSegment = Util.toLowerInvariant(manifestUri.getLastPathSegment());
+    // Using a regexp here to allow for URLs which contain (..) as their last part, e.g. http://host/content.ism/manifest(filter=0) - see #3230
+    this.manifestUri = manifestUri == null ? null : lastPathSegment.equals("manifest") ? manifestUri : lastPathSegment.matches("manifest\\(.*\\)") ? manifestUri : Uri.withAppendedPath(manifestUri, "Manifest");
     this.manifestDataSourceFactory = manifestDataSourceFactory;
     this.manifestParser = manifestParser;
     this.chunkSourceFactory = chunkSourceFactory;
--- a/library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
+++ b/library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
@@ -192,9 +192,7 @@ public final class SsMediaSource implements MediaSource,
       AdaptiveMediaSourceEventListener eventListener) {
     Assertions.checkState(manifest == null || !manifest.isLive);
     this.manifest = manifest;
-    this.manifestUri = manifestUri == null ? null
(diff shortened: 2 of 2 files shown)


library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
```diff
@@ -192,9 +192,7 @@ private SsMediaSource(SsManifest manifest, Uri manifestUri,
       AdaptiveMediaSourceEventListener eventListener) {
     Assertions.checkState(manifest == null || !manifest.isLive);
     this.manifest = manifest;
-    this.manifestUri = manifestUri == null ? null
-        : Util.toLowerInvariant(manifestUri.getLastPathSegment()).equals("manifest") ? manifestUri
-            : Uri.withAppendedPath(manifestUri, "Manifest");
+    this.manifestUri = manifestUri == null ? null : lastPathSegment.contains("manifest") && lastPathSegment.contains("(") && lastPathSegment.contains(")")) ? manifestUri : lastPathSegment.equals("manifest") ? manifestUri : Uri.withAppendedPath(manifestUri, "Manifest");
```
ojw28: - Does this actually compile? What is `lastPathSegment`? It doesn't seem to be defined anywhere.
- I'm not convinced about the logic. It matches all kinds of random things, like "xy)(manifest". Why not use `Util.toLowerInvariant(manifestUri.getLastPathSegment()).matches("XXX")` where "XXX" is a regular expression that more exactly describes what we consider a match?

nahojkap: You are right - apologies for that, very sloppy of me.

And a reg-exp would be a much more elegant solution - will update

nahojkap:
@ojw28 - updated the pull request - there is a duplication of the reg.exp used in Util.java and the SsMediaSource.java - not sure if you have a better, mode preferred place to put this or not.

Hope that is OK now.

library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
```diff
@@ -192,9 +192,9 @@ private SsMediaSource(SsManifest manifest, Uri manifestUri,
       AdaptiveMediaSourceEventListener eventListener) {
     Assertions.checkState(manifest == null || !manifest.isLive);
     this.manifest = manifest;
-    this.manifestUri = manifestUri == null ? null
-        : Util.toLowerInvariant(manifestUri.getLastPathSegment()).equals("manifest") ? manifestUri
-            : Uri.withAppendedPath(manifestUri, "Manifest");
+    String lastPathSegment = Util.toLowerInvariant(manifestUri.getLastPathSegment());
```
ojw28, line 195: This will NullPointerException is manifestUri is null.

library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
```diff
@@ -192,9 +192,9 @@ private SsMediaSource(SsManifest manifest, Uri manifestUri,
...
+    // Using a regexp here to allow for URLs which contain (..) as their last part, e.g. http://host/content.ism/manifest(filter=0) - see #3230
+    this.manifestUri = manifestUri == null ? null : lastPathSegment.equals("manifest") ? manifestUri : lastPathSegment.matches("manifest\\(.*\\)") ? manifestUri : Uri.withAppendedPath(manifestUri, "Manifest");
```
ojw28, line 197: It might be preferable to have a single regex that matches "manifest" and "manifest(...)". Something like `"manifest(\\(.*\\))?"` (I didn't check that's right!). Then you can get rid of the `equals("manifest")` case.

library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
```diff
@@ -901,7 +901,8 @@ public static int inferContentType(String fileName) {
     } else if (fileName.endsWith(".m3u8")) {
       return C.TYPE_HLS;
     } else if (fileName.endsWith(".ism") || fileName.endsWith(".isml")
-        || fileName.endsWith(".ism/manifest") || fileName.endsWith(".isml/manifest")) {
+        || fileName.endsWith(".ism/manifest") || fileName.endsWith(".isml/manifest")
+            || Uri.parse(fileName).getLastPathSegment().matches("manifest\\(.*\\)")) {
```
ojw28, line 905: I don't think we want to match if the .ism or .isml bit isn't there.

ojw28 (commented):
Looks nicer, although there's a NPE in there I think ;).

library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
nahojkap: You are right but by pulling together the reg-exp to check also for just 'manifest' we should be able to get rid of that local variable entirely.

library/smoothstreaming/src/main/java/com/google/android/exoplayer2/source/smoothstreaming/SsMediaSource.java
nahojkap: Will fix this

library/core/src/main/java/com/google/android/exoplayer2/util/Util.java
nahojkap: Would you prefer if this is pushed into a reg-exp as well?

ojw28: I think ideally it would be a regex check directly on fileName (i.e. avoiding parsing as a uri, particularly given the method above converts from a uri into a string to call this method :)).

ojw28:
I pushed ce7aaab3c6adc8d8b0d6723c0e59441fcca9bce2 to fix this, since this hadn't been updated in a while. Let us know on the corresponding issue if there are any problem. Thanks!

PR #3280, opened 2017-09-21
Fix primary hls playlist update

yqritc:
The current implementation does not update the `primaryHlsUrl` in `HlsPlaylistTracker` class. 
Please review the followings:

ff2ece5
- fixed `primarySnapshotAccessAgeMs` to update `primaryHlsUrl`. Due to this bug, `primaryHlsUrl` does not change and keep loading its url while loading higher variant url. 
Based on the media I used, the accesses to playlist is decreased from 40times/minute to 22times/minute.

cac16f1
- removed keepalive check for updating primary url. Because of this, `primaryHlsUrl` does not change 15 seconds and keep loading its url even though the player starts loading higher variant url within 15 second.
I couldn't think of any good reason to wait updating the primary url, so please tell me the case we need this check if there is any. 
Based on the media I used, the accesses to playlist is decreased to 15times/minute.   

thanks.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistTracker.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistTracker.java
@@ -351,7 +351,7 @@ public final class HlsPlaylistTracker implements Loader.Callback<ParsingLoadable
     }
     MediaPlaylistBundle currentPrimaryBundle = playlistBundles.get(primaryHlsUrl);
     long primarySnapshotAccessAgeMs =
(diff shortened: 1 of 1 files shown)


yqritc:
I signed it!

ojw28:
We've also submitted https://github.com/google/ExoPlayer/commit/792f046c02b45fee7a4f6f0f8ce8de41e081334d and https://github.com/google/ExoPlayer/commit/a3be9376509b9802171ae9b1b088b8f17ec2f999, that further tweak the logic. If you could verify behavior is still as expected on the latest `dev-v2`, that would be great.

yqritc:
I've confirmed the `dev-v2` works well. 
Thanks for merging pr and further improvement.

Issue #3293, opened 2017-09-25
Support initializing HLS playback without loading first chunk of default variant


natez0r:
### Issue description
When the HLSChunkSource is created, it creates a very simple track selection that it uses to determine the variant it should choose for initial loading. This defaults to the lowest quality and causes exoplayer to download unnecessary data if the actual bitrate chosen by the customizable trackselection differs from the initial selection.

### Reproduction steps
Pass a TrackSelection.Factory that chooses the highest bitrate always and observe in stetho or some other network debug tool, that the first segment from both the lowest variant and highest variant will be downloaded.

### Link to test content
Any HLS stream with multiple variants.

### Version of ExoPlayer being used
2.5.1

### Device(s) and version(s) of Android being used
Nexus 5X 7.1.2

### A full bug report captured from the device
Will follow up with bug report if necessary.

natez0r:
I plan to submit a Pull request to allow this behavior.

ojw28:
Updated the issue title to clarify this is about loading data, not about the actual initial quality that the user sees. Next steps are TBD, depending on the outcome of off-issue discussion. Thanks!

ojw28:
Issue will be resolved indirectly via https://github.com/google/ExoPlayer/pull/3318.

PR #3294, opened 2017-09-25
Allow passing of initialtrackselection for HLS.

natez0r:
To avoid situations where we download .ts files that we never intend to
play. #3293


natez0r:
WIP: thoughts?

PR #3318, opened 2017-10-02
Allow passing of HlsPlaylistParser to HlsMediaSource.

natez0r:
Also create helper method in HlsMasterPlaylist to allow the copying of
the playlist to another, but with the variants reordered based on a
passed comparator. Also added an implementation of HlsPlaylistParser
which will reorder the variants returned.

Another approach to #3293


natez0r:
@ojw28 was this more in line with what you were hoping for?

ojw28:
Yes, except with `ReorderingHlsPlaylistParser` in your codebase, rather than inside the ExoPlayer library directly. Is it flexible enough for you to do what you need to do?

natez0r:
I believe it will be. I need to test it locally now, but having some trouble building exo locally in our app (taking a bit longer than hoped)

natez0r:
was able to get this building and tested out. this is flexible enough for us. would you like me to remove the re-ordering stuff?

natez0r:
ok, updated

PR #3345, opened 2017-10-12
make videoframereleasetimehelper get refresh rate when it's enabled, …

drhill:
…so we can reuse video renderer for multiple clips with different frame rates

Frame rates need to be set before clips are prepared.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
@@ -79,6 +79,9 @@ public final class VideoFrameReleaseTimeHelper {
 
   private VideoFrameReleaseTimeHelper(double defaultDisplayRefreshRate) {
     useDefaultDisplayVsync = defaultDisplayRefreshRate != DISPLAY_REFRESH_RATE_UNKNOWN;
+    if (useDefaultDisplayVsync) {
+      vsyncSampler = VSyncSampler.getInstance();
+    }
   }
 
   /**
@@ -128,7 +131,6 @@ public final class VideoFrameReleaseTimeHelper {
   private void setSync(double defaultDisplayRefreshRate) {
 
     if (useDefaultDisplayVsync) {
-      vsyncSampler = VSyncSampler.getInstance();
       vsyncDurationNs = (long) (C.NANOS_PER_SECOND / defaultDisplayRefreshRate);
       vsyncOffsetNs = (vsyncDurationNs * VSYNC_OFFSET_PERCENTAGE) / 100;
     }
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
@@ -38,10 +38,10 @@ public final class VideoFrameReleaseTimeHelper {
   private static final long VSYNC_OFFSET_PERCENTAGE = 80;
   private static final int MIN_FRAMES_FOR_ADJUSTMENT = 6;
 
-  private final VSyncSampler vsyncSampler;
-  private final boolean useDefaultDisplayVsync;
-  private final long vsyncDurationNs;
-  private final long vsyncOffsetNs;
+  private VSyncSampler vsyncSampler;
+  private boolean useDefaultDisplayVsync;
+  private long vsyncDurationNs;
+  private long vsyncOffsetNs;
 
   private long lastFramePresentationTimeUs;
   private long adjustedLastFrameTimeNs;
@@ -71,24 +71,16 @@ public final class VideoFrameReleaseTimeHelper {
   }
 
   private VideoFrameReleaseTimeHelper(double defaultDisplayRefreshRate) {
-    useDefaultDisplayVsync = defaultDisplayRefreshRate != DISPLAY_REFRESH_RATE_UNKNOWN;
-    if (useDefaultDisplayVsync) {
-      vsyncSampler = VSyncSampler.getInstance();
-      vsyncDurationNs = (long) (C.NANOS_PER_SECOND / defaultDisplayRefreshRate);
-      vsyncOffsetNs = (vsyncDurationNs * VSYNC_OFFSET_PERCENTAGE) / 100;
-    } else {
-      vsyncSampler = null;
-      vsyncDurationNs = -1; // Value unused.
-      vsyncOffsetNs = -1; // Value unused.
-    }
+    setSync(defaultDisplayRefreshRate);
   }
 
   /**
    * Enables the helper.
    */
-  public void enable() {
+  public void enable(Context context) {
     haveSync = false;
     if (useDefaultDisplayVsync) {
+      setSync(getDefaultDisplayRefreshRate(context));
       vsyncSampler.addObserver();
     }
   }
@@ -99,6 +91,20 @@ public final class VideoFrameReleaseTimeHelper {
   public void disable() {
     if (useDefaultDisplayVsync) {
       vsyncSampler.removeObserver();
+    }
+  }
+
+  private void setSync(double defaultDisplayRefreshRate) {
+
+    useDefaultDisplayVsync = defaultDisplayRefreshRate != DISPLAY_REFRESH_RATE_UNKNOWN;
+    if (useDefaultDisplayVsync) {
+      vsyncSampler = VSyncSampler.getInstance();
+      vsyncDurationNs = (long) (C.NANOS_PER_SECOND / defaultDisplayRefreshRate);
+      vsyncOffsetNs = (vsyncDurationNs * VSYNC_OFFSET_PERCENTAGE) / 100;
+    } else {
+      vsyncSampler = null;
+      vsyncDurationNs = -1; // Value unused.
+      vsyncOffsetNs = -1; // Value unused.
     }
   }
 
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
@@ -24,6 +24,7 @@ import android.os.HandlerThread;
 import android.os.Message;
 import android.view.Choreographer;
 import android.view.Choreographer.FrameCallback;
+import android.view.Display;
 import android.view.WindowManager;
 import com.google.android.exoplayer2.C;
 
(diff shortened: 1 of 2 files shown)


ojw28:
I do not understand what this is for. The refresh rate is of the display, not of the content.

drhill:
If you create a SimpleExoPlayer instance it creates all the lower objects like MediaCodecVideoRenderer and VideoFrameReleaseTimeHelper.  VideoFrameReleaseTimeHelper only grabs the display refresh rate at construction.  This is fine if you never change the display mode (refresh rate) based on the content, but if you change the device/display's refresh rate to 23.976 and VideoFrameReleaseTimeHelper has a default refresh rate of 59.97 playback looks terrible at times.  I'm guessing it's dropped/smoothed frames but ultimately it results from VideoFrameReleaseTimeHelper.

Before I worked around this by changing the device/displays's refresh rate and when that finishes creating a new SimpleExoPlayer without leaving the activity instance before every time prepared a clip/episode/movie.  With this change enable gets called during preparation for each new clip so we grab the current refresh rate instead of the stale one.

ojw28:
Got it, although the timing of when to read the refresh rate again seems quite arbitrary. If application code were to change the refresh rate just after the read point then this wouldn't work, and an application developer shouldn't need to know about the timing of the call inside of the player.

It should be possible to register a listener to be told explicitly when the display properties change, which would allow you to update the value immediately upon a change, and not at any other time. It looks like you can register a listener on `DisplayManager` and take action in `DisplayListener.onDisplayChanged`. If you could make that approach work, I agree this is a good idea and would be happy to take such a change.

drhill:
Yes it is very arbitrary.  I wasn't aware of the listener that is a much better path to take.  Will look into it.  This should mean we need to make these few variables threadsafe as the listener could change them, correct?

ojw28:
Looks like you can easily get everything (including the callback) onto the playback thread. The listener registration method has documented guarantees about the thread on which the listener is invoked. As long as you register in the helper's `enable` method and pass null as the handler argument, everything should end up on the playback thread...

drhill:
Seems to work right now.  I put a register call in the constructor with a context because if I just put it on enable we ran into the arbitrary "well when did the refresh rate call happen" again.  And I put the unregister call in disable to free us of the listener.  Can we call enable after disable?  I assumed we do which is why the register call is there as well.  If not then that call is superfluous but harmless.

ojw28:
You still have the arbitrary "when did the refresh rate call happen" problem if the helper is disabled (causes unregister) and then enabled again (causes register, but you now have no idea if the refresh rate you had is still correct).

I think you need to query the refresh rate at the same time as adding the listener, and that both things should happen in enable. Neither should happen in the constructor. Does that work?

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -68,19 +74,12 @@ public VideoFrameReleaseTimeHelper() {
    */
   public VideoFrameReleaseTimeHelper(Context context) {
     this(getDefaultDisplayRefreshRate(context));
+    this.context = context;
```
ojw28: Use `context.getApplicationContext()` to ensure the class can't leak any shorter lived contexts that might be passed to it (it's always fine to hold onto the application context, since it's expected to be around for as long as the process remains alive).

drhill:
Oh wow, that was a pretty obvious case I missed.  Thanks.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -68,19 +74,12 @@ public VideoFrameReleaseTimeHelper() {
    */
   public VideoFrameReleaseTimeHelper(Context context) {
     this(getDefaultDisplayRefreshRate(context));
+    this.context = context.getApplicationContext();
+    registerDisplayListener();
```
ojw28: As far as I can see, there is no need to do this here, or call getDefaultDisplayRefreshRate or setSync anywhere prior to enable being called.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -100,6 +101,43 @@ public void disable() {
     if (useDefaultDisplayVsync) {
       vsyncSampler.removeObserver();
     }
+    unregisterDisplayListener();
```
ojw28: Given registerDisplayListener is in the if block in enable, this should probably be in the if block in disable also.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
drhill: If we are going to keep the default constructor (and the DISPLAY_REFRESH_RATE_UNKNOWN functionality) we need to at least call setSync in the constructor since enable only does anything if useDefaultDisplayVsync==true and after which we would reset the refresh rate and then register the listener.  Unless you want me to move useDefaultDisplayVsync=true into the constructor.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -31,17 +33,21 @@
...
   private static final double DISPLAY_REFRESH_RATE_UNKNOWN = -1;
   private static final long CHOREOGRAPHER_SAMPLE_DELAY_MILLIS = 500;
   private static final long MAX_ALLOWED_DRIFT_NS = 20000000;
 
   private static final long VSYNC_OFFSET_PERCENTAGE = 80;
   private static final int MIN_FRAMES_FOR_ADJUSTMENT = 6;
 
-  private final VSyncSampler vsyncSampler;
+  private DisplayManager.DisplayListener displayListener = null;
+  private Context context = null;
```
ojw28: It should be pretty easy to make this final

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -99,6 +102,37 @@ public void enable() {
   public void disable() {
     if (useDefaultDisplayVsync) {
       vsyncSampler.removeObserver();
+      unregisterDisplayListener();
+    }
+  }
+
+  private void registerDisplayListener() {
+    if (displayListener == null && context != null &&
+     Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
ojw28: What is this gated on M and above for? Ditto in unregister method. Looks like it needs gating to API level 17+?

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -99,6 +102,37 @@ public void enable() {
...
+        displayManager.registerDisplayListener(displayListener, null);
+      }
+    }
+  }
+
+  private void unregisterDisplayListener() {
+    if (context != null && displayListener != null &&
+     Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      DisplayManager displayManager = context.getSystemService(DisplayManager.class);
+      if (displayManager != null) {
```
ojw28: Hm - You really don't want to get here and find displayManager is null. Maybe you should defensively stash it in a variable in registerDisplayListener, so you have guaranteed access to it.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -99,6 +102,37 @@ public void enable() {
...
+        displayManager.unregisterDisplayListener(displayListener);
+        displayListener = null;
+      }
+    }
+  }
+
+  private void setSync(double defaultDisplayRefreshRate) {
+
+    if (useDefaultDisplayVsync) {
```
ojw28: Isn't this always true here?

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
ojw28: Looks like you could make it API level 17+ by using:

```
DisplayManager displayManager = (DisplayManager) context.getSystemService(Context.DISPLAY_SERVICE);
```

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
ojw28: Or just get the displayManager once in the constructor, and have it be final.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -31,17 +33,21 @@
...
   private static final long MAX_ALLOWED_DRIFT_NS = 20000000;
 
   private static final long VSYNC_OFFSET_PERCENTAGE = 80;
   private static final int MIN_FRAMES_FOR_ADJUSTMENT = 6;
 
-  private final VSyncSampler vsyncSampler;
+  private DisplayManager.DisplayListener displayListener = null;
+  private Context context = null;
+
+  private VSyncSampler vsyncSampler = null;
```
ojw28: Why is vsyncSampler no longer final? Final fields are good ;).

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
drhill: Yes it is now.

library/core/src/main/java/com/google/android/exoplayer2/video/VideoFrameReleaseTimeHelper.java
```diff
@@ -301,4 +341,30 @@ private void removeObserverInternal() {
...
+    public void onDisplayAdded(int displayId) {
+    }
+
+    @Override
+    public void onDisplayRemoved(int displayId) {
+    }
+
+    @Override
+    public void onDisplayChanged(int displayId) {
+      final int defaultDisplayId = getDefaultDisplayId(context);
```
ojw28: You can use Display.DEFAULT_DISPLAY directly (and remove getDefaultDisplayId)

drhill:
Think I addressed all those points.

PR #3371, opened 2017-10-18
Fix FLV AVCVIDEOPACKET  CTS Type Issue

andymiao858:
AVCVIDEOPACKET --> compositionTimeMs should be SI 24, not UI24.
refer to FLV 1.0 Format Document
http://wwwimages.adobe.com/content/dam/acom/en/devnet/flv/video_file_format_spec_v10.pdf
Page 10

As is
If compositionTimeMs is negative, when call data.readUnsignedInt24() will get a much larger number

Tobe
Support negative compositionTimeMs, that is to say support non sequential timestamp. 
compositionTimeMs can be both positive and negtive

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
@@ -80,6 +80,9 @@ import com.google.android.exoplayer2.video.AvcConfig;
   protected void parsePayload(ParsableByteArray data, long timeUs) throws ParserException {
     int packetType = data.readUnsignedByte();
     int compositionTimeMs = data.readUnsignedInt24();
+    // compositionTimeMs is signed int 24, change unsigned int 24 to signed int 24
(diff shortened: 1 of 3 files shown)


andymiao858:
I signed it!

library/core/src/main/java/com/google/android/exoplayer2/extractor/flv/VideoTagPayloadReader.java
```diff
@@ -80,6 +80,9 @@ protected boolean parseHeader(ParsableByteArray data) throws UnsupportedFormatEx
   protected void parsePayload(ParsableByteArray data, long timeUs) throws ParserException {
     int packetType = data.readUnsignedByte();
     int compositionTimeMs = data.readUnsignedInt24();
+    // compositionTimeMs is signed int 24, change unsigned int 24 to signed int 24
```
ojw28: Can you add a readInt24 method to ParsableByteArray instead (preferably with unit tests in ParsableByteArrayTest)? It seems like an omission; that class has methods for reading both signed and unsigned values of other lengths.

andymiao858: I have changed as you said. Please review.

andymiao858:
Please review

PR #3423, opened 2017-11-03
Dev v2 ffmpegfloat

drhill:
This should solve two use cases.

1) enabled 32 float output for ffmpeg (if supported always on, currently)
2) enable PCM stream input for u8, s16, s24, s32, f32.  I'm not sure if there is much use having u8 and s16 capabilities, but the other formats can leverage float output to avoid dithering.


andrewlewis:
32-bit float output from `FfmpegAudioRenderer` should work on `dev-v2`, though it doesn't support PCM. For PCM support, is it feasible to use `MediaCodecAudioRenderer` instead, where this is already supported? It should be straightforward to use that renderer as a fallback by putting it after `FfmpegAudioRenderer` in the list of renderers passed to the player.

drhill:
Going to close this and open a new PR with that change, which is much better than this way.  Thanks!

Issue #3426, opened 2017-11-05
ContentDataSource size determination issue


axet:
DefaultDataSource uses ContentDataSource to open urls. But ContentDataSource.open not following android documentations exactly and cause issues when opening url's in "r" mode only.

You can read ContentProvider.openFile documentation:

     * If opened with the exclusive "r" or "w" modes, the returned
     * ParcelFileDescriptor can be a pipe or socket pair to enable streaming
     * of data. Opening with the "rw" or "rwt" modes implies a file on disk that
     * supports seeking.

Since ContentDataSource uses "r" mode, ContentProvider may return Pipe instead of file and resulting bytes (total content length) will be miss detected: 

          bytesRemaining = inputStream.available();

Problem is, if ContentProvider retuns Pipe, ExoPlayer will stop playing audio after few seconds after fist seek action and close pipe before it fully readed. To fix this we may use open "rw" mode or wait until we reach EOF.

ojw28:
I put some pending comments on the proposed fix.

ojw28:
I played around with this a little by making a ContentProvider that returns a pipe. It looks like:
* If the content provider sets the length of the `AssetFileDescriptor` then things just work, since `assetFileDescriptor.getLength()` is set.
* If the content provider sets the length to `AssetFileDescriptor.UNKNOWN_LENGTH` then `inputStream.available()` doesn't give us the right thing. It doesn't look like `inputStream.getChannel().size()` gives the correct size either though. I get 0 back when I query this with my test implementation. I think we should just set `bytesRemaining` directly to `C.LENGTH_UNSET` in the case that `assetFileDescriptor.getLength()` doesn't provide a value.

axet:
1) correct
2) inputStream.getChannel().size() - only gives right size when FD is a file. So, it is safe to call for booth pipe/socket/file - it will return zero or correct size, inputStream.available() - only return correct size when FD is a file. So. channel size - is safe to use.

ojw28:
Are you certain `size()` will return 0 or the correct size? It's documented as returning a "current" size, which suggests the value may change.

axet:
I read it some where, can not find anymore, I _think_ by "current" they mean size can be changed due to write operations.

ojw28:
[...]

PR #3428, opened 2017-11-05
Update ContentDataSource.java

axet:
Checking File size makes more sense then asking for InputStream cached bytes.

#3426 explanation

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
@@ -89,7 +89,7 @@ public final class ContentDataSource implements DataSource {
         long assetFileDescriptorLength = assetFileDescriptor.getLength();
         if (assetFileDescriptorLength == AssetFileDescriptor.UNKNOWN_LENGTH) {
           // The asset must extend to the end of the file.
-          bytesRemaining = inputStream.available();
+          bytesRemaining = ((FileInputStream) inputStream).getChannel().size();
           if (bytesRemaining == 0) {
             // FileInputStream.available() returns 0 if the remaining length cannot be determined,
             // or if it's greater than Integer.MAX_VALUE. We don't know the true length in either


library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
```diff
@@ -89,7 +89,7 @@ public long open(DataSpec dataSpec) throws ContentDataSourceException {
         long assetFileDescriptorLength = assetFileDescriptor.getLength();
         if (assetFileDescriptorLength == AssetFileDescriptor.UNKNOWN_LENGTH) {
           // The asset must extend to the end of the file.
-          bytesRemaining = inputStream.available();
+          bytesRemaining = ((FileInputStream) inputStream).getChannel().size();
           if (bytesRemaining == 0) {
             // FileInputStream.available() returns 0 if the remaining length cannot be determined,
```
ojw28, line 94: This comment doesn't seem to make sense given then change being made? Does the if block itself still make sense?

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
```diff
@@ -89,7 +89,7 @@ public long open(DataSpec dataSpec) throws ContentDataSourceException {
         long assetFileDescriptorLength = assetFileDescriptor.getLength();
         if (assetFileDescriptorLength == AssetFileDescriptor.UNKNOWN_LENGTH) {
           // The asset must extend to the end of the file.
-          bytesRemaining = inputStream.available();
+          bytesRemaining = ((FileInputStream) inputStream).getChannel().size();
```
ojw28, line 92: I think it's possible for multiple assets to be in a single file. In which case is `size()` the size of just this one, or of all of them? I think you might want to subtract `position()` of the `FileChannel` to get correct behavior?

ojw28:
* We wont be able to merge this without a signed CLA.
* Pull requests need to go to the `dev-v2` branch, not `release-v2`. Please address the comments and make a new request to that branch.

axet:
1) Will wont be able to merge this without a signed CLA.

less then 5 lines of code need to sign anything? Eclipse treat small bug fixes wihtout any paper work. I do not wana bother to sing anything.

2) Ok, I close this one, but keep reffered issue open.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
axet: Not sure what do you mean, FileChannel one per file.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
axet: bytesRemaining = inputStream.available()

- for the momen when you open Pipe this call can return 0. 

bytesRemaining = ((FileInputStream) inputStream).getChannel().size();

- FileDescription for correct file in file system, this call will return correct file size. 

Again. I do not understand, why you asking this. Propbably need more testing.

ojw28:
Please do not send a second pull request if you don't intend to sign the CLA. We wont be able to accept it. Thanks.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
ojw28: I don't think the `uri` for which the source is being opened necessarily corresponds to the whole file though. Else this code wouldn't need to be using `assetFileDescriptor.getStartOffset` as above? If the `uri` only corresponds to a part of the file, that needs to be accounted for.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
ojw28: In the previous code we actually called `inputStream.available()` and then did something conditionally on the value being 0. We're not making an `inputStream.available()` call any more, so the comment doesn't make sense. Can `size()` return 0? If not why is this if block still here? If so the comment needs changing.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ContentDataSource.java
ojw28: Even if the `uri` does correspond to the whole file, you'd need to subtract `skipped` if it's non-zero.

axet: Then correct line is:

bytesRemaining = ((FileInputStream) inputStream).getChannel().size() - assetFileDescriptor.getStartOffset

axet:
@ojw28 I didnt know you trying to use "documents" to protect you from the world. I would'nt pull again.

Anyway you can't protect yourself using this technique sign - and enter. I can prove I made this code having it before the code come to the repository, and since it get there without my sign it be equals staling code. So, sign or no sign - always possible to make you/google guilty and violating copyright slavery... Just make it simple, opensource and fun. BTW i'm not going to sue anyway. Just show you how stupid "CLA" is.

Issue #3439, opened 2017-11-08
ANR when playing mp4 multiple times with Sony Bravia


yqritc:
### Issue description
After playing and releasing mp4 content multiple times, the device freeze because of ANR.
This issue can be seen by using the ExoPlayer r2.5.0 or above with demo content, "Dizzy" mp4.
Originally I've found this issue when playing mp4 in asset directory in my app.
By using my local mp4, r2.4.4 does not have this problem, but r2.5.0 or the above get ANR.
( I could not confirmed r2.4.4 with demo "Dizzy" mp4, since the player failed to play due to 302 status code)
After checking the previous issues, I've found the similar issues, #3236 #3355 (Bravia uses the same decoder with TCL TVs).
So I've tried to add the output surface workaround and then ANR seems to be fixed.
But I'm not 100% sure, this is the correct way to fix the problem.
I'm appreciate it if you can confirm this.

### Reproduction steps
Describe how the issue can be reproduced, ideally using the ExoPlayer demo app.
1. Launch demo application
2. Select "Misc" -> "Dizzy" and starts playing the content
3. After playing a few seconds, press back button and back to SampleChooserActivity
4. Repeat step 2 and step 3
5. The device freeze because of ANR

### Link to test content
Demo app "Dizzy" mp4 content

### Version of ExoPlayer being used
dev-v2
r2.5.4
r2.5.0

### Device(s) and version(s) of Android being used
Build.DEVICE: BRAVIA_ATV2, 
Build.MODEL:  BRAVIA 4K GB, 
Build.MANUFACTURER: Sony,
API level: 24

Build.DEVICE: SVP-DTV15, 
Build.MODEL: BRAVIA 4K 2015, 
Build.MANUFACTURER: Sony,
API level: 23

### A full bug report captured from the device
I will send it via email.

ojw28:
Thanks for the pull request. See one pending comment there. Happy to merge once the workaround is targeted more precisely somehow :).

ojw28:
https://github.com/google/ExoPlayer/pull/3440 was merged to fix this.

PR #3440, opened 2017-11-08
Skip using MediaCodec.setOutputSurface on Bravia

yqritc:
Issue: #3439

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
@@ -1071,8 +1071,9 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
    * If true is returned then we fall back to releasing and re-instantiating the codec instead.
    */
   private static boolean codecNeedsSetOutputSurfaceWorkaround(String name) {
-    // Work around https://github.com/google/ExoPlayer/issues/3236 and
-    // https://github.com/google/ExoPlayer/issues/3355.
+    // Work around https://github.com/google/ExoPlayer/issues/3236,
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1071,11 +1071,13 @@ private static boolean deviceNeedsAutoFrcWorkaround() {
...
   private static boolean codecNeedsSetOutputSurfaceWorkaround(String name) {
-    // Work around https://github.com/google/ExoPlayer/issues/3236 and
-    // https://github.com/google/ExoPlayer/issues/3355.
+    // Work around https://github.com/google/ExoPlayer/issues/3236,
+    // https://github.com/google/ExoPlayer/issues/3355 and
+    // https://github.com/google/ExoPlayer/issues/3439.
     return (("deb".equals(Util.DEVICE) || "flo".equals(Util.DEVICE))
         && "OMX.qcom.video.decoder.avc".equals(name))
-        || ("tcl_eu".equals(Util.DEVICE) && "OMX.MTK.VIDEO.DECODER.AVC".equals(name));
+        || (("tcl_eu".equals(Util.DEVICE) || Util.MODEL.startsWith("BRAVIA"))
```
ojw28: Util.MODEL.startsWith("BRAVIA") feels like it will blacklist all Sony TVs for as long as they stick with the BRAVIA branding. Can you come up with something a little more precise, or apply a maximum API level at which the workaround will be applied so that future BRAVIA devices don't automatically get the workaround too?

Thanks!

yqritc: @ojw28 
thanks for the feedback! I've updated BRAVIA workaround to use device name instead of models.

PR #3451, opened 2017-11-12
Add an easy way to set the shutter view background color

### LINKED ISSUE #3172: Support setting SimpleExoPlayerView background/shutter color
**- Search existing issues, including issues that are closed.**
i only found this thread: https://github.com/google/ExoPlayer/issues/2821
but the question there is regarding surface view, and not exoPlayer components.
**- Consult our FAQs, supported devices and supported formats pages.
did not find my answer there

### Issue description
i would like to change the background of SimpleExoPlayerView.
right now, while the video is loading for the first time (meaning not buffering in the middle of streaming)
the background is black. i would like to change it to white.
using `android:background` in xml or `exoPlayerView.setBackground()` in code does not work.
changing the foreground does change the color, but it also blocks (or hides) all the other UI element, including the video.

### Reproduction steps
play any video that takes time to initially load

2.5.1

LG G3 API 23.
LG K10 API 23.
Uniden prime C471 API 18.

in the layout file `exo_simple_player_view.xml`, which is found here: https://github.com/google/ExoPlayer/blob/release-v2/library/ui/src/main/res/layout/exo_simple_player_view.xml
there is a view with id "exo_shutter" which has a black background - im assuming this is the cause of my problem?
some might say to just do this (pseudo code):
```
int resId = getResources().getIdentifier("exo_shutter", "id", getPackageName());
findViewById(resId).setBackground(Color.white);
```
but it is not good practice to relay of inner id's like that... if they change in a future version i will get null pointer exception.
**unless,** can you guarantee that this will not be changed in the future and is safe to do?

nvalletta:
This pull request attempts to resolve #3172 , which requests a way to programmatically alter the SimpleExoPlayerView shutter background color.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/SimpleExoPlayerView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/SimpleExoPlayerView.java
@@ -83,6 +83,12 @@ import java.util.List;
  *         <li>Default: {@code true}</li>
  *       </ul>
  *   </li>
+ *   <li><b>{@code shutter_background_color}</b> - The background color of the {@code exo_shutter} view.
+ *       <ul>
+ *         <li>Corresponding method: {@link #setShutterBackgroundColor(int)}</li>
+ *         <li>Default: {@code 0}</li>
+ *       </ul>
+ *   </li>
  *   <li><b>{@code hide_on_touch}</b> - Whether the playback controls are hidden by touch events.
  *       <ul>
  *         <li>Corresponding method: {@link #setControllerHideOnTouch(boolean)}</li>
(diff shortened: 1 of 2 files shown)


library/ui/src/main/java/com/google/android/exoplayer2/ui/SimpleExoPlayerView.java
```diff
@@ -293,6 +301,9 @@ public SimpleExoPlayerView(Context context, AttributeSet attrs, int defStyleAttr
 
     // Shutter view.
     shutterView = findViewById(R.id.exo_shutter);
+    if (shutterView != null) {
+      shutterView.setBackgroundColor(shutterColor);
```
ojw28, line 305: I think this defaults to setting the background color to transparent. I think it'll also clear any background drawable that's been set directly on the shutter as part of a custom layout for the view. No need to update this change. I'll merge as is and add a check so that `setBackgroundColor` is only called if the attribute is actually defined.

PR #3493, opened 2017-11-23
r2.6.0

ojw28:


hare989:
Hi,

When will r2.6.0 be available on jcenter? 
https://jcenter.bintray.com/com/google/android/exoplayer/exoplayer/

ojw28:
It's already there (look right at the top). We're dropped to "r" from future version tags.

hare989:
Yup, it's there. :) 
Tnx

PR #3505, opened 2017-11-27
Removed r prefix from guide

Aqua-Ye:


ojw28:
Thanks for reporting. We're going to merge an equivalent change shortly.

Issue #3599, opened 2017-12-16
NPE in method `HlsMediaPeriod#maybeThrowPrepareError()`


Khang-NT:
### Issue description
Crash log:
```
ExoPlayerImplInternal: Internal runtime error.
            java.lang.NullPointerException: Attempt to invoke virtual method 'void com.google.android.exoplayer2.source.hls.HlsSampleStreamWrapper.maybeThrowPrepareError()' on a null object reference
            at com.google.android.exoplayer2.source.hls.HlsMediaPeriod.maybeThrowPrepareError(HlsMediaPeriod.java:98)
            at com.google.android.exoplayer2.source.DynamicConcatenatingMediaSource$DeferredMediaPeriod.maybeThrowPrepareError(DynamicConcatenatingMediaSource.java:808)
            at com.google.android.exoplayer2.ExoPlayerImplInternal.maybeThrowPeriodPrepareError(ExoPlayerImplInternal.java:970)
            at com.google.android.exoplayer2.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:544)
            at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:306)
            at android.os.Handler.dispatchMessage(Handler.java:98)
            at android.os.Looper.loop(Looper.java:154)
            at android.os.HandlerThread.run(HandlerThread.java:61)
```

I don't understand much the code base, but it shouldn't cause NPE in method `maybeThrowPrepareError` with any reason.

### Reproduction steps
It happens randomly when I play HLS media source.

### Version of ExoPlayer being used
I use version `2.6.0`.

ojw28:
Please can you provide a full bug report captured with `adb bugreport`, proper reproduction steps, and test content that can be used to reproduce this issue. All of this information is requested in the issue template.

Looking at the code for `HlsMediaPeriod`, it's hard to see how a null entry can end up in `sampleStreamWrappers`, unless there's some failure that happens before this one. If there is some previous failure, that's the root cause we need to be fixing. I'd expect such a failure to be visible in the full bug report, once provided. If there isn't, can you see how a null entry could happen?

ojw28:
@AquilesCanta - Can you see any way this can occur? If not (and unless new information is provided), then I think it's OK to go ahead and close this issue.

AquilesCanta:
> Can you see any way this can occur?

[...]

PR #3600, opened 2017-12-16
Fix NPE HlsMediaPeriod#maybeThrowPrepareError

Khang-NT:
Close https://github.com/google/ExoPlayer/issues/3599


Khang-NT:
I signed it!

ojw28:
I don't really understand why this line of code can throw a NPE in your case (let's discuss on the issue), but moving it inside a null check without understanding what's going on is not the correct way to fix it.

Issue #2777, opened 2017-05-05
Fast play will cause the HD video frozen


robin2046:
### Issue description
When testing the new feature of set playback speed, it works well with small video clips, lets say, 240p or 480p, with speed from 1/8 to 8x. However, when trying this feature with HD video, e.g. video clip captured by camera app, when speed is <1.0f it worked fine but if setting speed >= 2.0f, video will freeze after playing several frames (1-2s). And this cannot be recovered no matter how long I waited. 

### Reproduction steps
Just use the new simple Exoplayer to play a HD video clip captured by camera app, set the speed > 2.0f.

### Link to test content
Use camera app to record some HD video clip

### Version of ExoPlayer being used
r2.4.0

### Device(s) and version(s) of Android being used
Google Pixel Phone, Android Nougat 7.1

robin2046:
if possible, when the power of the device cannot support fast playing of HD video, can Exoplayer try to seek for the iFrame only?

ojw28:
Yes, this is on our list of follow-up work for variable speed playback.

robin2046:
Before the enhancement is ready, how can I detect this frozen and take action?

KirillMakarov:
Same issue. Tested on Nexus 5 with API 23, Samsung Galaxy Tab A (2016) SM-T280 API 21.
But on Samsung SM-T580 API 23 HD video with 2x is working.

ojw28:
I think this is fixed in 2.6.0. @andrewlewis - Can this be closed? Thanks!

jinminghe:
I can still reproduce this bug on latest dev-v2 branch:

In PlayerActivity::initializePlayer(), add 
`player.setPlaybackParameters(new PlaybackParameters(2.0f, 1.0f));`
Then build and install the demo app on a Google Pixel Phone with Android 7.1, run the demo app, and play a 1080p video, for example, Misc -> Screens 1080p (FMP4, H264, No Audio)

Can you see if the above steps reproduce the bug for you? Thanks.

andrewlewis:
Yes, I think this can be closed: video renderers now skip to keyframes if their decoders can't keep up.

@jinminghe That sample plays as expected at high playback speeds for me: video is not smooth but keyframes are now shown (the old behavior was that one video frame was shown then video was frozen for the duration of playback). In the demo app you should see that a frame is rendered occasionally and the "dropped to keyframe" counter labeled `dk:` gets incremented each time a frame is shown when the decoder can't keep up.

[...]

PR #3634, opened 2017-12-23
Set KEY_OPERATING_RATE for MediaCodecVideoRenderer.

jinminghe:
Related issue: #2777

Background:
[KEY_OPERATING_RATE](https://developer.android.com/reference/android/media/MediaFormat.html#KEY_OPERATING_RATE) was added in API 23 to help with high speed playback. Without setting this value, the video decoder on some devices will run slowly and cannot support 2x speed playback for HD videos. For example, it takes a Pixel ~25ms to decode a 1080p/30fps frame while we can afford at most 16ms for 2x speed.

Fix:
Allow setting KEY_OPERATING_RATE when constructing a MediaCodecVideoRenderer.


ojw28:
This is interesting. I don't think passing through the `Format` is the correct approach though. The `Format` generally describes (and is derived from) the media being played. The playback speed is more a property of the current state of the player, and is not contained within the media itself.

Is the end goal for `KEY_OPERATING_RATE` to be set correctly on the video codec given the current playback speed set through `Player.setPlaybackParameters`? If so, we should find a way to propagate the current playback speed to the video renderer, possibly as part of `RendererConfiguration` (although I haven't thought about it much). The value can then be merged into the `MediaFormat` in the `Renderer` just before the codec is configured.

jinminghe:
@ojw28 Thanks for the review! 
Yes, we should move it to the renderer, will need some more work to fix the PR.

No, the end goal is to set `KEY_OPERATING_RATE` to a default value, which is `Short.MAX_VALUE`. 
We could set `KEY_OPERATING_RATE` based on the current playback speed, for example, 30 for 1x and 60 for 2x. However, this will cause other problems when the user changes playback speed: to update `KEY_OPERATING_RATE`, we have to stop the codec and then reconfigure it, which can cause latency and make the video laggy. We also have to sync to the next key frame because we cannot restart decoding with other types of frames. This will drop a few frames. Therefore, I think we should just set `KEY_OPERATING_RATE` to `Short.MAX_VALUE` so that we don't need to change it.

ojw28:
As an aside, are you certain `KEY_OPERATING_RATE` applies to decoders? We're wondering whether it's for encoders only (we're following up internally to find out and get the documentation clarified, but just wondering whether you know in the meantime). Thanks!

jinminghe:
Yes, I built a simple app which only decoded a video (1080p/30fps) in assets and I recorded the average time to decode a frame. For Pixel on Android 7.1, the time taken to decode a frame is ~24ms without setting `KEY_OPERATING_RATE`, and after setting it to `Short.MAX_VALUE` (without changing anything else), it dropped down to ~6ms. This result is consistent.
I tried on other phones too, and observed similar behaviors though the improvement may be different.

It's easy to observe the behavior with the Exoplayer demo app too.
By following my comments in #2777

> In PlayerActivity::initializePlayer(), add
> player.setPlaybackParameters(new PlaybackParameters(2.0f, 1.0f));
> Then build and install the demo app on a Google Pixel Phone with Android 7.1, run the demo app, and play a 1080p video, for example, Misc -> Screens 1080p (FMP4, H264, No Audio)

You can see that the video is not smooth and it keeps jumping between key frames.
If you set `KEY_OPERATING_RATE` and try the above steps again, you will see the video becomes smooth.

ojw28:
Got it, thanks. I think the best solution is just to allow it to be passed as a variable through the `MediaCodecVideoRenderer` constructor?

andrewlewis:
I followed up regarding this parameter. It does indeed apply for decoders, and we should be setting it. However, it looks like the value may be used to reserve resources, and in that case we shouldn't be setting the maximum possible value, as that might result in unnecessarily increased resource usage.

As well as setting the initial effective frame rate (taking into account playback speed) when first configuring the codec, we should call [`MediaCodec.setParameters`](https://developer.android.com/reference/android/media/MediaCodec.html#setParameters(android.os.Bundle)) with a `Bundle` that sets `"operating-rate"` to the new effective frame rate when the playback speed changes.

jinminghe:
@andrewlewis 
Thanks for following up! I tried to set this parameter with `MediaCodec.setParameters` and it silently failed to apply. In my simple app to measure codec performance, it still takes a Pixel ~ 24ms to decode a frame.
To observe this in the demo app, simply add
```
Bundle bundle = new Bundle();
bundle.putInt(MediaFormat.KEY_OPERATING_RATE, 60);
codec.setParameters(bundle);
```
to `MediaCodecRender.maybeinitCodec` (after codec.start()) and follow the aforementioned steps to play a 1080p video with 2x speed, you can see that the video is still jumping between key frames. Compare this observation with the smooth playback when setting the parameter with `MediaCodec.configure`, we can confirm that `MediaCodec.setParameters` doesn't work for this parameter.

I agree that we shouldn't be using more resources than we need, but I just can't figure out how to update `MediaFormat.KEY_OPERATING_RATE` when the decoder is running, without stopping and restarting it.

Can you follow up and see if we should be able to set this parameter via `MediaCodec.setParameters`? If not, are there other ways to set this parameter when the decoder is running?

@ojw28 
Yes, we can then merge this parameter into `MediaFormat` when configuring the decoder.

andrewlewis:
@jinminghe Does it work if you first configure the codec with `KEY_OPERATING_RATE` set to e.g. 30 fps, then later on set a higher playback speed and call `MediaCodec.setParameters`? I wonder whether calling `MediaCodec.setParameters` has no effect unless `KEY_OPERATING_RATE` is set when first configuring the codec.

jinminghe:
@andrewlewis I tried that too. No, it doesn't. I also tried setting `KEY_FRAME_RATE` with `MediaCodec.setParameters` and still, no luck.
 I suspect that the resources have to be allocated when you configure the decoder and cannot be changed when it's running.

jinminghe:
@ojw28 @andrewlewis 
Updated PR to pass the value through the constructors and set the default value to be 120. Please take a look.
By the way, did you figure out how to update the value when the decoder is running? @andrewlewis

andrewlewis:
@jinminghe Unfortunately not. I'll file an internal bug for this as I see the same thing you've described: setting `KEY_OPERATING_RATE` when configuring the codec for the first time works as expected, but calling `setParameters` appears to have no effect. [Internal reference: b/71987865]

jinminghe:
@andrewlewis Is there any update for the issue? Just a note: I read through the API again, it doesn't look like we should be able to set KEY_OPERATING_RATE on the fly. See the difference between [PARAMETER_KEY_VIDEO_BITRATE](https://developer.android.com/reference/android/media/MediaCodec.html#PARAMETER_KEY_VIDEO_BITRATE) and [KEY_BIT_RATE](https://developer.android.com/reference/android/media/MediaFormat.html#KEY_BIT_RATE)

ojw28:
One update regarding `setParameters`. It appears that you need to use `putFloat` rather than `putInt` when setting the operating rate. Although you can use either during initial codec configuration ([source](https://android.googlesource.com/platform/frameworks/av/+/1b96205/media/libstagefright/ACodec.cpp#2162)), for some reason the `setParameters` code path only tries to retrieve it as a float ([source](https://android.googlesource.com/platform/frameworks/av/+/1b96205/media/libstagefright/ACodec.cpp#6341)). That looks like a bug, but one that's at least easy to work around just by using `putFloat` instead. It's worth trying that to see if it works correctly.

andrewlewis:
It looks like `setParameters` does work when specifying the operating rate via `putFloat`.

jinminghe:
Wow! Great findings! I will change the PR and fix the problem. I am so happy we figured out a proper fix!

ojw28:
How are you planning to implement this? One idea is to add a `setOperatingRate` method on the `Renderer` interface, and call it on all of the `Renderer` instances from `ExoPlayerImplInternal.setPlaybackParametersInternal` (passing `playbackParameters.speed`).

The internally, the `MediaCodec` based video renderer will need to multiply the received speed by the frame-rate defined in the format to get a rate suitable for passing to the decoder. It's a bit of an open question what rate to pass if the frame-rate of the media isn't specified.

jinminghe:
Yes, that's the plan, except that I think `setPlaybackRate` is a better name for the method. Besides updating the operating rate when the playback rate changes, we also have to keep the playback rate in the `Renderer` so that whenever we receive a new format (for cases like `ConcatenatingMediaSource `), we can update the operating rate correctly.

The unspecified frame-rate problem is interesting. Do you know why the frame-rate is not specified?

ojw28:
My only concern with `setPlaybackRate` is that it sounds like it'll actually set the rate (particularly for the audio renderer, the method name suggests passing 2 will make the renderer play twice as fast). Where-as "operating rate" sounds more related to performance configuration, which is what this method will really be doing.

Whether frame-rate is set depends on the type of media being played, the container etc. There's often no requirement that video in a container will even have a constant frame-rate, although it normally tends to be the case.

jinminghe:
Thanks. `setOperatingRate` makes sense. I have updated the PR to set the operating rate for formats with frame rates. When the format does not have a frame rate, I just leave the MediaCodec to run at default operating rate.

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
```diff
@@ -286,6 +287,19 @@ public MediaClock getMediaClock() {
     return this;
   }
 
+  @TargetApi(23)
+  @Override
+  protected void updateCodecOperatingRate(Format format) {
```
ojw28: I'm not sure there's any point doing anything for the audio case. Do you observe a difference? I strongly suspect there is none. Can we do video only for now, unless you can measure a change?

library/core/src/main/java/com/google/android/exoplayer2/BaseRenderer.java
```diff
@@ -136,6 +136,11 @@ public final void resetPosition(long positionUs) throws ExoPlaybackException {
     onPositionReset(positionUs, false);
   }
 
+  @Override
+  public void setOperatingRate(float operatingRate) {
```
ojw28: Should be final

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -464,6 +467,14 @@ protected void onPositionReset(long positionUs, boolean joining) throws ExoPlayb
     }
   }
 
+  @Override
+  protected void onOperatingRateChanged(float operatingRate) {
+    codecOperatingRate = operatingRate;
+    if (codec != null && format != null) {
```
ojw28: I think just checking the codec is sufficient. There cannot exist a codec if format is null (I think).

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -916,9 +927,22 @@ protected void onInputFormatChanged(Format newFormat) throws ExoPlaybackExceptio
...
+      }
     }
   }
 
+  /**
+   * Updates the {@link MediaCodec} operating rate.
+   * <p>
+   * The default implementation is a no-op.
+   */
+  protected void updateCodecOperatingRate(Format format) {
```
ojw28: Can we explicitly pass both the codec instance and the new operating rate to this method, so the subclass doesn't need to retrieve it with `getCodec`? It's clearer what's going on, and is more consistent with the `configureCodec` method.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -450,6 +451,8 @@ protected final MediaCodecInfo getCodecInfo() {
     return codecInfo;
   }
 
+  protected final float getCodecOperatingRate() { return codecOperatingRate; }
```
ojw28: This will not be needed if you do some of the comments below.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -969,6 +983,10 @@ protected MediaFormat getMediaFormat(
     // Set codec configuration values.
     if (Util.SDK_INT >= 23) {
       mediaFormat.setInteger(MediaFormat.KEY_PRIORITY, 0 /* realtime priority */);
+      if (format.frameRate != Format.NO_VALUE) {
+        float codecOperatingRate = getCodecOperatingRate();
```
ojw28: Can we explicitly pass operating rate down through `configureCodec` and into `getMediaFormat`, to avoid the need for this getter?

ojw28 (commented):
Posted a few comments (sorry for the long delay). Let us know if you plan to address them. If not (totally reasonable if you've moved onto other stuff) then we'll merge as-is and make the alterations our side.

Thanks again!

jinminghe:
@ojw28 Thanks for the review! I will fix them shortly!

ojw28:
@jinminghe - Thanks!

jinminghe:
@ojw28 Addressed the comments. I really like the design to pass the parameters through instead of adding extra getters!

ojw28:
Thanks for the changes! A few additional questions have come up during internal review that we might need to think about. In particular, if we end up setting the operating rate to something quite low, and then transition to playing content whose `Format` does not allow calculating an operating rate, then I think the current behavior is that we just leave the codec with a low operating rate, which is bad. It appears there's no way to reset the codec back to whatever its (unknown) default rate is, which means we may have to either invent some defaults of our own, or  release and re-instantiate the codec. Let us know if you have any thoughts on this, or happen to know of a way to reset the decoder back to its default rate!

ojw28:
@medialajos - Any thoughts? See also [Internal ref: b/111543954]. My main concern with us inventing our own defaults is that we might end up setting the operating rate to lower than the actual true default, therefore unwittingly making performance worse than had we done thing at all.

ojw28:
We came up with a kind of slightly more complicated solution to try and handle the question above in a sensible way. We'll get this merged, and then merge some changes on top that came out of that discussion.

ojw28:
Thanks for the contribution! Refinements were merged in https://github.com/google/ExoPlayer/commit/badd9356f8d617802724e58f7c7aaab40f026134. Let us know if you spot anything wrong.

medialajos:
Re: Jan 5 comment by @jinminghe :

> @andrewlewis I tried that too. No, it doesn't. I also tried setting KEY_FRAME_RATE with MediaCodec.setParameters and still, no luck.
> I suspect that the resources have to be allocated when you configure the decoder and cannot be changed when it's running.
> 

In theory, operating rate DOES NOT need to be set in the configure() to be able to be set later in setParameter(). However, support for operating rate is not required by codecs at this point, so as such we cannot enforce any behavior. I would be interested if someone could come up with a compliance test for this parameter.

PR #3635, opened 2017-12-23
add support in mediacodecaudiorenderer for 24bit pcm to float

drhill:
Kept FloatResampleAudioProcessor it's own class in case it can be leveraged/upgraded in the future.  This has to be explicitly enabled for use.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
@@ -164,10 +164,12 @@ public final class DefaultAudioSink implements AudioSink {
   public static boolean failOnSpuriousAudioTimestamp = false;
 
   @Nullable private final AudioCapabilities audioCapabilities;
+  private final boolean canConvertHiResPcmToFloat;
   private final ChannelMappingAudioProcessor channelMappingAudioProcessor;
   private final TrimmingAudioProcessor trimmingAudioProcessor;
   private final SonicAudioProcessor sonicAudioProcessor;
   private final AudioProcessor[] availableAudioProcessors;
+  private final AudioProcessor[] hiResAvailableAudioProcessors;
   private final ConditionVariable releasingConditionVariable;
   private final long[] playheadOffsets;
   private final AudioTrackUtil audioTrackUtil;
@@ -180,6 +182,7 @@ public final class DefaultAudioSink implements AudioSink {
   private AudioTrack keepSessionIdAudioTrack;
   private AudioTrack audioTrack;
   private boolean isInputPcm;
+  private boolean shouldUpResPCMAudio;
   private int inputSampleRate;
   private int sampleRate;
   private int channelConfig;
@@ -233,6 +236,8 @@ public final class DefaultAudioSink implements AudioSink {
   private boolean hasData;
   private long lastFeedElapsedRealtimeMs;
 
+
+
   /**
    * @param audioCapabilities The audio capabilities for playback on this device. May be null if the
    *     default capabilities (no encoded audio passthrough support) should be assumed.
@@ -241,7 +246,23 @@ public final class DefaultAudioSink implements AudioSink {
    */
   public DefaultAudioSink(@Nullable AudioCapabilities audioCapabilities,
       AudioProcessor[] audioProcessors) {
+    this(audioCapabilities, audioProcessors, false);
+  }
+
+  /**
+   * @param audioCapabilities The audio capabilities for playback on this device. May be null if the
+   *     default capabilities (no encoded audio passthrough support) should be assumed.
+   * @param audioProcessors An array of {@link AudioProcessor}s that will process PCM audio before
+   *     output. May be empty.
+   * @param canConvertHiResPcmToFloat Flag to convert > 16bit PCM Audio to 32bit Float PCM Audio to
+   *     avoid dithering the input audio.  If enabled other audio processors that expect 16bit PCM
+   *     are disabled
+   */
+  public DefaultAudioSink(@Nullable AudioCapabilities audioCapabilities,
+      AudioProcessor[] audioProcessors, boolean canConvertHiResPcmToFloat) {
+
     this.audioCapabilities = audioCapabilities;
+    this.canConvertHiResPcmToFloat = canConvertHiResPcmToFloat;
     releasingConditionVariable = new ConditionVariable(true);
     if (Util.SDK_INT >= 18) {
       try {
@@ -265,6 +286,8 @@ public final class DefaultAudioSink implements AudioSink {
     availableAudioProcessors[2] = trimmingAudioProcessor;
     System.arraycopy(audioProcessors, 0, availableAudioProcessors, 3, audioProcessors.length);
     availableAudioProcessors[3 + audioProcessors.length] = sonicAudioProcessor;
+    hiResAvailableAudioProcessors = new AudioProcessor[1];
+    hiResAvailableAudioProcessors[0] = new FloatResamplingAudioProcessor();
     playheadOffsets = new long[MAX_PLAYHEAD_OFFSET_COUNT];
     volume = 1.0f;
     startMediaTimeState = START_NOT_SET;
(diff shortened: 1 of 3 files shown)


andrewlewis (commented):
Thanks for the PR!

- Would it make sense for FloatResamplingAudioProcessor to support 32-bit PCM to float conversion (in addition to 24-bit)?
- I think we should try to move the resampling step into DefaultAudioSink and leave MediaCodecAudioRenderer unmodified, so that other audio renderers could also benefit from this functionality. This would also remove the need to input/output/drain the audio processor in the renderer. I suggest we add a constructor argument enableResamplingToFloat in DefaultAudioSink and apply the FloatResamplingAudioProcessor as a special case. This may be a bit tricky as we will need to disable the int PCM processors under the right circumstances. If that sounds reasonable I can take a look at doing that part of the change.

drhill:
lol... isn't that what I had a few months ago?  Both 32-bit pcm to float conversion (I removed as I have no media of that type but it is certainly a good idea) and the floatresample code in the AudioSink (much older code now).  I'll try to give it a look this weekend.  Thanks!

drhill:
Ok, got it started though tt looks a little sloppy having the hiResProcessors and the regular ones.  24bit pcm audio converts to 32 float fine.  DTS, DTS-HD, DD, TrueHD, and 16 bit PCM all seem unchanged.  Didn't test 32bit int to 32bit float as I don't have video samples of that.

andrewlewis:
Yes, this is doing float to integer conversion in the audio sink as you had a while back. I thought before that it wasn't worth the complexity to convert 24-bit and 32-bit integer input to float, but actually it is needed to make this complete. Sorry for having to split this up and iterate.

I don't think it's so bad having the int -> float processor separate from the other ones. Does it ever make sense to use the processors together? Going via 16-bit makes it not worth doing float output.

Had a quick look and this looks good. I will take a proper look next week and hopefully we can get it submitted soon. Thanks!

drhill:
No problem.  

I think the separate processors makes sense.  Maybe add a layer of extraction for each processor which encapsulates the current 16 bit processor with any float processor and have the encapsulating class direct the inputs to the proper processor or just do nothing if the input isn't supported?

`........./ 16BitResampler` 
`Resampler`
`.........\ FloatResampler`

`....................../ 16bitChannelMapperProcessor`
`ChannelMapperProcessor`
 `......................\ pass through untouched non-16bit data`

 `....................../ 16bitTrimmingAudioProcessor`
`TrimmingAudioProcessor`
 `......................\ pass through untouched non-16bit data`

etcetera
Editing formatting

library/core/src/main/java/com/google/android/exoplayer2/audio/FloatResamplingAudioProcessor.java
```diff
@@ -0,0 +1,182 @@
+package com.google.android.exoplayer2.audio;
```
andrewlewis, line 16: Please could you add an AOSP header? (Just copy from another file and replace the year with 2018.)

library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
```diff
@@ -342,15 +365,20 @@ public void configure(@C.Encoding int inputEncoding, int inputChannelCount, int
...
+     (inputEncoding == C.ENCODING_PCM_24BIT || inputEncoding == C.ENCODING_PCM_32BIT);
     if (isInputPcm) {
-      pcmFrameSize = Util.getPcmFrameSize(inputEncoding, channelCount);
+      pcmFrameSize = Util.getPcmFrameSize(shouldUpResPCMAudio
+       ? C.ENCODING_PCM_FLOAT : inputEncoding, channelCount);
     }
     @C.Encoding int encoding = inputEncoding;
     boolean processingEnabled = isInputPcm && inputEncoding != C.ENCODING_PCM_FLOAT;
     if (processingEnabled) {
+      AudioProcessor[] activeAudioProcessors = shouldUpResPCMAudio ?
```
andrewlewis: The "active" part of the name may be misleading because these audio processors won't necessarily all end up being active (in the terminology of `AudioProcessor.isActive()`).

How about renaming the final fields to something like `toIntPcmAvailableAudioProcessors` and `toFloatPcmAvailableAudioProcessors`? The names are a bit long but hopefully clearer. Or, if not doing that rename, this line could be `AudioProcessor[] availableAudioProcessors = shouldUpResPCMAudio ? hiResAvailableAudioProcessors : this.availableAudioProcessors`. Same applies for other occurrence of `activeAudioProcessors`.

library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
```diff
@@ -342,15 +365,20 @@ public void configure(@C.Encoding int inputEncoding, int inputChannelCount, int
     int channelCount = inputChannelCount;
     int sampleRate = inputSampleRate;
     isInputPcm = isEncodingPcm(inputEncoding);
+    shouldUpResPCMAudio = canConvertHiResPcmToFloat &&
+     (inputEncoding == C.ENCODING_PCM_24BIT || inputEncoding == C.ENCODING_PCM_32BIT);
     if (isInputPcm) {
-      pcmFrameSize = Util.getPcmFrameSize(inputEncoding, channelCount);
+      pcmFrameSize = Util.getPcmFrameSize(shouldUpResPCMAudio
```
andrewlewis: pcmFrameSize is used to calculate the number of submitted frames (not the number of output frames), so shouldn't this be left as it was?

library/core/src/main/java/com/google/android/exoplayer2/audio/FloatResamplingAudioProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+   */
+  private static void writePcm32bitFloat(int val, ByteBuffer buffer) {
+    float convVal = (float) (PCM_INT32_FLOAT * val);
+    int bits = Float.floatToIntBits(convVal);
+    if (bits == 0x7fc00000)
+      bits = Float.floatToIntBits((float) 0.0);
+    buffer.put((byte) (bits & 0xff));
+    buffer.put((byte) ((bits >> 8) & 0xff));
+    buffer.put((byte) ((bits >> 16) & 0xff));
+    buffer.put((byte) ((bits >> 24) & 0xff));
```
andrewlewis: Does `buffer.putLong(bits)` work here?

library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
```diff
@@ -342,15 +365,20 @@ public void configure(@C.Encoding int inputEncoding, int inputChannelCount, int
...
+       ? C.ENCODING_PCM_FLOAT : inputEncoding, channelCount);
     }
     @C.Encoding int encoding = inputEncoding;
     boolean processingEnabled = isInputPcm && inputEncoding != C.ENCODING_PCM_FLOAT;
```
andrewlewis: For the 24-bit integer PCM -> float path we can't apply playback parameters, but processingEnabled gets set to true so `setPlaybackParameters` will try to apply the playback parameters. For now shall we just add another flag `canApplyPlaybackParams` which is `processingEnabled && !shouldUpResPCMAudio` and change the check in `setPlaybackParameters` to use that?

andrewlewis (commented):
Looks good. Just some minor things to fix up then we can merge this. Thanks again!

andrewlewis:
I suggest we go with separate available audio processors references for this change as it isn't much code, and have a think about doing https://github.com/google/ExoPlayer/pull/3635#issuecomment-355745170 later on.

library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
drhill: Yes, I think that was a mixing of changes that got left in and caused discontinuity errors.  Fixed.

library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
drhill: Did the rename and changed the local activeAudioProcessors line to availableAudioProcessors

library/core/src/main/java/com/google/android/exoplayer2/audio/FloatResamplingAudioProcessor.java
drhill: Check

library/core/src/main/java/com/google/android/exoplayer2/audio/FloatResamplingAudioProcessor.java
drhill: putInt works

drhill:
Changes made.  Thanks!

PR #3673, opened 2018-01-06
PGS subtitle decoding support

drhill:
Where can I send a sample file?  83MB mkv file

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/pgs/PgsBuilder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/pgs/PgsBuilder.java
@@ -1,0 +1,232 @@
+/*
+*
+*  Sources for this implementation PGS decoding can be founder below
+*
+*    http://exar.ch/suprip/hddvd.php
+*    http://forum.doom9.org/showthread.php?t=124105
+*    http://www.equasys.de/colorconversion.html
+ */
+
+package com.google.android.exoplayer2.text.pgs;
+
+import android.graphics.Bitmap;
+
+import com.google.android.exoplayer2.text.Cue;
+import com.google.android.exoplayer2.text.Subtitle;
+import com.google.android.exoplayer2.util.ParsableByteArray;
+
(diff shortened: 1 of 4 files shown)


ojw28:
Is it possible for you to host it somewhere and send a link (to dev.exoplayer@gmail.com)? Thanks.

drhill:
Email sent

ojw28:
Got it, thanks.

library/core/src/main/java/com/google/android/exoplayer2/text/pgs/PgsBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+
+    if (list.isEmpty())
+      return new PgsSubtitle();
+
+    Cue[] cues = new Cue[list.size()];
+    long[] cueStartTimes = new long[list.size()];
+    int index = 0;
+    for (Holder curr : list) {
+      cues[index] = curr.build();
+      cueStartTimes[index++] = curr.start_time;
```
ojw28, line 71: Question: `start_time` is never assigned, and so is always 0. Does the PGS spec actually carry timestamps, or are the timestamps always provided out of band (i.e. elsewhere in the container)? 

* If it can carry timestamps then there's something missing in this implementation.
* If they're always provided out of band then it's guaranteed that there's only a single event time for each sample, and hence there's no reason to pass any timestamps to `PgsSubtitle`. This allows `PgsSubtitle` to be simplified down to something that looks identical to `DvbSubtitle`.

No need to update this pull request as I'm going through tweaking it a bit anyway, but please clarify (same will apply to any other comments). Thanks!

drhill: As far as I could tell the time stamps are only provided out of band, but I've only used this with MKV files.  It would make sense to simplify this.  It's been awhile since I wrote this.  I think I was modeling it off of SubripDecoder/SubripSubtitle class at the time.

PR #3677, opened 2018-01-08
fix code to be more understandable

TakuSemba:
I fixed logic to get bps by calculating bits and time separately instead of just multiplying 8000


TakuSemba:
I signed it!

tonihei:
Appreciate the idea of improving readability, but it's not a good idea to divide `totalElapsedTimeMs` by 1000 first because it's an integer operation and not floating point.

PR #3710, opened 2018-01-16
Improve raw resource data source

eneim:
This change provides the ability to create RawResourceDataSource to current DefaultDataSource class.

Currently there is no trivial way to combine the Power of DefaultDataSource with the use of RawResourceDataSource. Hope this change is self-explained. No unit test added.

I also update build tool to 27.0.2. The fact is from gradle build 3.x*, it implicitly define preferred version of build tool. Explicitly setting it may cause trouble. As we currently set this version explicitly, we should use latest version if possible.

> *3.x = 3.1 or maybe even 3.0, pardon me if this is only true for 3.1+.


constants.gradle
```diff
@@ -19,7 +19,7 @@ project.ext {
     minSdkVersion = 14
     compileSdkVersion = 27
     targetSdkVersion = 27
-    buildToolsVersion = '26.0.2'
+    buildToolsVersion = '27.0.2'
```
ojw28, line 22: This is unrelated to the change. Can it please be removed?

library/core/src/main/java/com/google/android/exoplayer2/upstream/RawResourceDataSource.java
```diff
@@ -58,7 +58,7 @@ public static Uri buildRawResourceUri(int rawResourceId) {
     return Uri.parse(RAW_RESOURCE_SCHEME + ":///" + rawResourceId);
   }
 
-  private static final String RAW_RESOURCE_SCHEME = "rawresource";
+  /* package */ static final String RAW_RESOURCE_SCHEME = "rawresource";
```
ojw28, line 61: This may as well be public (e.g. If someone wants to implement their own variant of a default data source outside of this package, it would be useful to be able to access this).

library/core/src/main/java/com/google/android/exoplayer2/upstream/DefaultDataSource.java
```diff
@@ -60,6 +61,7 @@
   private DataSource contentDataSource;
   private DataSource rtmpDataSource;
   private DataSource dataSchemeDataSource;
+  private DataSource rawResourceDataSource;
```
ojw28, line 64: Please update the list in the Javadoc at the top of this file to include an entry:

<li>rawresource: For fetching data from a raw resource in the applications' apk (e.g. rawresource:///resourceId, where rawResourceId is the integer identifier of the raw resource).</li>

Probably inserted under the asset one, since they seem related.

ojw28 (commented):
This looks good! It needs sending to the `dev-v2` branch though, rather than `dev-v2-r2.6.1`. Happy to merge if resent to that branch with the couple of comments I've added inline addressed. Thanks!

eneim:
@ojw28 Thanks, let me create new one with the update regards your comments. My PR is cut from release-v2 so point to dev-v2 is conflict.

constants.gradle
eneim: Sure.

library/core/src/main/java/com/google/android/exoplayer2/upstream/RawResourceDataSource.java
eneim: Tried to keep it as close as original source, but yes public static will be happy for everyone ;D.

PR #3746, opened 2018-01-24
Upgrade library version

mekya:
There was a bug in the lib. Exoplayer crashes if rtmp server is down. It is fixed.


ojw28:
This change needs to go into the `dev-v2` branch. I'll close this and submit an equivalent change there directly. Thanks!

PR #3771, opened 2018-01-30
Use the pattern "string literal".equals(something) to prevent NPE …

reudismam:
… as it is done in many other locations in the codebase.


reudismam:
I signed it!

tonihei:
Thanks for pull request. It seems that all of the changed lines are guaranteed to be non-null in obvious ways (either by having a null check immediately before that or by using a static non-null constant). So none of these changes could actually prevent a NPE.

tonihei:
We'll still merge this, of course, to keep consistency within our code.

PR #3783, opened 2018-02-01
More permissive SegmentTemplate format string handling

tedjp:
Permit all IEEE 1003.1 integer format specifiers; %d, %i, %o, %u, %x, %X.

Don't require zero-padding width to be specified.

Fixes erroneous handling of `$Number%08x$` or `$Number%u$` (which had an extra "d" suffix in the generated URL), and no longer accepts `$Number%0$` which is malformed.

ISO/IEC 23009-1:2014 § 5.3.9.4.4 is the relevant spec. The wording is a bit ambiguous but it appears to me that the intention is that all IEEE 1003.1 integer formats be accepted (otherwise why cite the IEEE spec just for zero-padded decimal?).

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
@@ -135,13 +135,10 @@ public final class UrlTemplate {
         if (identifier.equals(REPRESENTATION)) {
           identifiers[identifierCount] = REPRESENTATION_ID;
         } else {
-          int formatTagIndex = identifier.indexOf("%0");
+          int formatTagIndex = identifier.indexOf("%");
           String formatTag = DEFAULT_FORMAT_TAG;
           if (formatTagIndex != -1) {
             formatTag = identifier.substring(formatTagIndex);
-            if (!formatTag.endsWith("d")) {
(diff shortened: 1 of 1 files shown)


ojw28:
I'm not sure about this. 5.3.9.4.4 specifically states:

> following this prototype: `%0[width]d`

That doesn't seem in any way ambiguous. Have I misunderstood something?

tedjp:
> Each identifier may be suffixed, within the enclosing ‘$’ characters, with an additional **format tag aligned with the printf format tag** as defined in IEEE 1003.1-2008

Bolded the ambiguous part. `%08x` is aligned with the standard by my understanding.

Why even mention IEEE 1003.1-2008 if the only acceptable format is `%0d` with optional width?

tedjp:
I guess I misinterpreted it. Interesting that `$Number%d$` is then invalid along with `$Number%0$`, although latter is specifically accepted by ExoPlayer.

ojw28:
I think the reference to IEEE 1003.1-2008 is just so that the reader knows where to find the definition of `printf`?

Closing, since it appears we accept everything allowed by the spec. I'm unsure why our implementation also accepts `$Number%0$`. It looks like this has always been the case ever since the code block was committed, so my best guess is that some early examples of DASH manifests that we were working with used this format. I think my preference would be to leave this as is, since it doesn't appear to do any harm.

Issue #2176, opened 2016-12-10
Support EMSG events during DASH playbacks


ChernyshovYuriy:
Hi,
In one of the closed issue (https://github.com/google/ExoPlayer/issues/1414) you mentioned that:
```
For DASH and SmoothStreaming custom data can be embedded directly into the media in 
EMSG boxes, which we'll be adding support for in 2.x.
```
Is there any updates on this case so far for the version 2.0.4?

ojw28:
This is not supported yet. Let's use this issue to track the enhancement.

ChernyshovYuriy:
Thanks for feedback!
Can you spot a class to start investigate from?
Perhaps, if you have not enough resources for this issue, I can start to implement and submit pull request later?

Rohit-Varma1:
Doesn't seem to work. Tested with 2 DASH streams using the DEMO app provided.  Doesn't prints out the EMSG in the EventLogger. The same streams work in SHAKA player when tested on web.

ojw28:
The work is not finished (which is why the issue hasn't been closed). It's not expected to work at this point.

ChernyshovYuriy:
@rohitvarma1986 , there is dedicated branch that works fine:
https://github.com/google/ExoPlayer/tree/dev-v2-emsg-experimental

Rohit-Varma1:
@ojw28 : Thank you for the prompt reply. Is there any estimate at this point , regarding when it would be done ?

jlacivita:
Hi @ojw28. For context, this feature is very important for ExoPlayer/DASH adoption in North America, particularly for OTT apps streaming content originally aired on TV. Nielsen provides a critical viewership measurement SDK that requires access to eMSG data during playback. Any OTT app considers Nielsen integration to be a requirement at some point, and they're all waiting for this.

Until we have eMSG data, north american OTT apps are forced to stick with HLS, which means they can't use DRM on Android.

ojw28:
Note from ref'd issue: An additional test stream is available at http://se-mashup.fokus.fraunhofer.de:8080/dash/assets/adinsertion-samples/events/inband/dash.mpd

ojw28:
This functionality should be working now in `dev-v2`. Please give it a try.

There's still one technical issue where events might go missing when seeking short distances and/or toggling the event tracks on and off, so the issue is kept open to track fixing that. It's unlikely to be much of an issue in practice, however.

ojw28:
[...]

PR #3842, opened 2018-02-13
 Added parsing of optional human-readable title #EXTINF:<duration>,[<title>]

JungleGenius:
Hi Googlers, 

We added parsing of the #EXTINF tag's optional human-readable titles per:
[https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1) and I'm hoping it can be incorporated back into your upstream branch.

I know the <title> tag in HLS M3U8 playlists is kind of a left over from music playlist days but it is still a valid part of the HLS spec. In most if not all HLS implementations the value of the tag  it is typically ignored or discarded. 

However, As it is a valid attribute for a valid tag and it helps us solve for use cases like this:  https://github.com/google/ExoPlayer/issues/1414 , this: https://github.com/google/ExoPlayer/pull/1847 and this https://github.com/google/ExoPlayer/issues/2176 without inserting a bunch of proprietary tags in the process. This way we can access the parsed value in the mediaplaylist and use that value for tracking various things in our Apps.

If you have any questions please let me know. We made our changes against the release-v2 branch not the dev-v2 so I'm not sure how hard it is to merge.

-Daniel


ojw28:
Closing for now. The pull request needs to be into the `dev-v2` branch and not pick up other changes. Thanks!

PR #3851, opened 2018-02-14
Resubmit: Added parsing of optional human-readable title #EXTINF:<duration>,[<title>]

JungleGenius:
Hi Googlers,

This is a resubmit of https://github.com/google/ExoPlayer/pull/3844 and https://github.com/google/ExoPlayer/pull/3842.  It looks like the unit tests were updated to use truth instead of framework asserts between release-v2 / dev-2 and that created the merge conflict. I went back and did a cherry-pick to work our updates into dev-2.  Our original pull request text is below.

=======  

We added parsing of the #EXTINF tag's optional human-readable titles per:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1 and I'm hoping it can be incorporated back into your upstream branch.

I know the <title> tag in HLS M3U8 playlists is kind of a left over from music playlist days but it is still a valid part of the HLS spec. In most if not all HLS implementations the value of the tag it is typically ignored or discarded.

However, As it is a valid attribute for a valid tag and it helps us solve for use cases like this: #1414 , this: #1847 and this #2176 without inserting a bunch of proprietary tags in the process. This way we can access the parsed value in the mediaplaylist and use that value for tracking various things in our Apps.

If you have any questions please let me know. We made our changes against the release-v2 branch not the dev-v2 so I'm not sure how hard it is to merge.

-Daniel

PR #3853, opened 2018-02-15
Resubmit: Added parsing of optional human-readable title #EXTINF:<duration>,[<title>]

JungleGenius:
Hi Googlers,

This is a resubmit of https://github.com/google/ExoPlayer/pull/3844, https://github.com/google/ExoPlayer/pull/3842 and https://github.com/google/ExoPlayer/pull/3851 . 

Per the comment "Your change includes 3 commits, and your main commit seems to have all the CLA requirements. Can you try removing the other two?"

So I went to do that, but I having issues rebasing my commit so I ended up deleting and re-forking my tracking repo.  So I'm having to create a new pull request. My apologies for all the clutter, but this should give you a clean commit as part of our pull request.

------

We added parsing of the #EXTINF tag's optional human-readable titles per:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1 and I'm hoping it can be incorporated back into your upstream branch.

I know the <title> tag in HLS M3U8 playlists is kind of a left over from music playlist days but it is still a valid part of the HLS spec. In most if not all HLS implementations the value of the tag it is typically ignored or discarded.

However, As it is a valid attribute for a valid tag and it helps us solve for use cases like this: #1414 , this: #1847 and this #2176 without inserting a bunch of proprietary tags in the process. This way we can access the parsed value in the mediaplaylist and use that value for tracking various things in our Apps.

If you have any questions please let me know. We made our changes against the release-v2 branch not the dev-v2 so I'm not sure how hard it is to merge.

-Daniel


tonihei:
Seems to have worked now :) 
Assigning @AquilesCanta to take a look - he's our HLS expert.

AquilesCanta:
Hi @JungleGenius, thanks for your pull request! It looks good for merging, but before:

#1414 requires support for custom tags (this can already be access via `HlsPlaylist#tags`).
#1847 asks for support for a master playlist attribute (EXTINF is a media playlist tag). Note: Can also be obtained via tags.
#2176 is about emsg, so I suppose it is not strictly related.

Would you kindly explain the exact usecase for `title`? How exactly does it get used by an app. If possible, It would also be great to explain why `HlsPlaylist#tags` is not enough.

Again, thank you very much!

JungleGenius:
Hi AquilesCanta,

We are using it to access the Title of the individual media playlist EXTINF segments per the HLS spec here: https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1

Currently only the Duration and the URL are getting parsed, but Title is also part of the spec so it makes sense (to us) that it should be exposed as a segment attribute just like the Duration is.

My understanding is that HlsPlaylist#tags is for inspecting third-party tags that are not part of the actual HLS spec but are still present in the manifest. Mostly to give access to proprietary tags at the master playlist level.

We use the media playlist segment Title to determine what parts (e.g. group of TS segments) in a live Stitched media playlist belong to a given show title.  If there are multiple discontinuities in a playlist window setting the Title on each segment allows us to know what group of segments where are a part of each show in our programming.

I hope this makes sense.    

-Daniel

PR #3863, opened 2018-02-17
add h263 format support in fourcc codecID

drhill:


library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1711,9 +1713,22 @@ public void initializeOutput(ExtractorOutput output, int trackId) throws ParserE
           nalUnitLengthFieldLength = hevcConfig.nalUnitLengthFieldLength;
           break;
         case CODEC_ID_FOURCC:
-          initializationData = parseFourCcVc1Private(new ParsableByteArray(codecPrivate));
-          if (initializationData != null) {
-            mimeType = MimeTypes.VIDEO_VC1;
+          initializationData = null;
```
ojw28: Can we remove this line? It doesn't do anything.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1711,9 +1713,22 @@ public void initializeOutput(ExtractorOutput output, int trackId) throws ParserE
...
+          Pair<Integer, List<byte[]>> pair = parseFourCcVc1Private(new ParsableByteArray(codecPrivate));
```
ojw28: Feels like things would be much simpler if the method returned a non-null Pair<String, List<byte[]>>, where the first part is the mimeType and can be set to `VIDEO_UNKNOWN`. You could then do something much more concise here, like:
```
Pair<String, List<byte[]>> mimeTypeAndInitializationData = ...
mimeType = mimeTypeAndInitializationData.first;
initializationData = mimeTypeAndInitializationData.second;
```
The `Log.w` call could be moved inside the method that's called here.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1938,12 +1953,15 @@ public void reset() {
      *     not VC1.
      * @throws ParserException If the initialization data could not be built.
      */
-    private static List<byte[]> parseFourCcVc1Private(ParsableByteArray buffer)
+    private static Pair<Integer, List<byte[]>> parseFourCcVc1Private(ParsableByteArray buffer)
```
ojw28: The Javadoc for this method is no longer accurate. Neither is the method name.

ojw28 (commented):
The idea of the change looks fine, but it looks like it needs cleaning up a bit. Please take a look at the comments. Thanks!

drhill:
Changes made.  How is that?

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1932,38 +1930,42 @@ public void reset() {
...
      * Builds initialization data for a {@link Format} from FourCC codec private data.
      * <p>
-     * VC1 is the only supported compression type.
+     * VC1 and H263 are the only supported compression types.
      *
-     * @return The initialization data for the {@link Format}, or null if the compression type is
-     *     not VC1.
+     * @return A pair object with the first object being the codec mime type
+     * and the second object the initialization data for the {@link Format},
+     * or null if the compression type is not VC1.
```
ojw28: This is still incorrect.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1932,38 +1930,42 @@ public void reset() {
...
-        // bytes of which are: sizeImage(4), xPel/m (4), yPel/m (4), clrUsed(4), clrImportant(4).
-        int startOffset = buffer.getPosition() + 20;
-        byte[] bufferData = buffer.data;
-        for (int offset = startOffset; offset < bufferData.length - 4; offset++) {
-          if (bufferData[offset] == 0x00 && bufferData[offset + 1] == 0x00
-              && bufferData[offset + 2] == 0x01 && bufferData[offset + 3] == 0x0F) {
-            // We've found the initialization data.
-            byte[] initializationData = Arrays.copyOfRange(bufferData, offset, bufferData.length);
-            return Collections.singletonList(initializationData);
+        else if (compression == FOURCC_COMPRESSION_VC1)  {
```
ojw28: nit - Don't line break before "else if"

ojw28 (commented):
Looks much nicer, thanks! There's still one part of the documentation that looks wrong. Can merge once that's addressed.

drhill:
Sorry.  Fixed.

andrewlewis:
@drhill Is the mapping added here from FOURCC_COMPRESSION_DIVX onto H.263 correct? I wonder if we should instead have a separate MIME type (video/divx, video/divx311 or video/divx4). Thanks.

drhill:
Yes that is correct.  Using those mimetypes would makes more sense since as far as I know 3gpp is just a container.

PR #3887, opened 2018-02-22
Add abstract logging methods to EventLogger

CoryCharlton:
Move logging calls to protected methods so consumers can override with logging method of their choosing

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/util/EventLogger.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/util/EventLogger.java
@@ -89,33 +89,33 @@ public class EventLogger
 
   @Override
   public void onLoadingChanged(boolean isLoading) {
-    Log.d(TAG, "loading [" + isLoading + "]");
+    logd(TAG, "loading [" + isLoading + "]");
   }
 
   @Override
   public void onPlayerStateChanged(boolean playWhenReady, int state) {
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/util/EventLogger.java
```diff
@@ -555,4 +555,11 @@ private static String getTimelineChangeReasonString(@Player.TimelineChangeReason
     }
   }
 
+  protected void logd(String tag, String msg) {
+	Log.d(tag, msg);
+  }
+
+  protected void loge(String tag, String msg, Throwable tr) {
+    Log.d(tag, msg, tr);
```
ojw28: Log.e

library/core/src/main/java/com/google/android/exoplayer2/util/EventLogger.java
```diff
@@ -555,4 +555,11 @@ private static String getTimelineChangeReasonString(@Player.TimelineChangeReason
     }
   }
 
+  protected void logd(String tag, String msg) {
+	Log.d(tag, msg);
```
ojw28: nit - Whilst you're fixing the one below, can you fix the indent here (to 2 char like rest of file)? Thanks!

CoryCharlton:
Sorry about that. Fixed with https://github.com/google/ExoPlayer/pull/3887/commits/9f2d53dc218f2c841927c025d0e03379b80e1fb9

PR #3900, opened 2018-02-26
replace 'compile' with 'implementation'

Aditya94A:


ojw28:
Thanks. We will also be updating our `build.gradle` files to use implementation/api where appropriate (no need to send a pull request for this, since we're already on the case).

PR #3921, opened 2018-02-28
support zlib compressed PGS subtitles

wischnow:
PGS subtitles may be compressed with zlib compression. For an example see the MKV-with-PGS sample file from Kodi available here: https://drive.google.com/file/d/0BwxFVkl63-lEVEluSFZ4NlhpLUk/view

The changes in this pull request add code to check if the data buffer passed to PgsDecoder:decode() starts with a zlib header byte and if so, attempts to decompress the subtitle data. With this, the subtitles in the above sample file are displayed correctly.


wischnow:
I signed it.

Issue #3935, opened 2018-03-03
Improve customisability of PlayerView


FD-:
I really appreciate we now have a non-final `PlayerView` class for subclassing! However, I'd love to see a minor improvement:

At the moment, subclasses are unable to override behaviour that occurs in reaction to (player) listener events, because the `PlayerView` doesn't directly implement those listeners but only via a private inner class (ComponentListener). This means subclasses cannot simply (eg) override `onPlayerStateChanged`.

I suggest to change that by removing the `ComponentListener` class and changing the `PlayerView` itself to implement the listeners currently implemented by `ComponentListener`. I know this is going to remove the comfort of using the `DefaultEventListener` as a base, so there would be a few empty methods, but I wouldn't consider that a big issue.

I'm willing to submit a pull request, but wanted to hear your opinion first!

ojw28:
Having `PlayerView` implement the listeners directly may confuse matters. We might end up with developers thinking they need to explicitly add their `PlayerView` instance as a listener to the player. It also adds lots of public methods to the class that developers should never be calling themselves.

Instead, how about we keep `ComponentListener`, but have it forward calls to protected methods in the outer class. Like (not full code):

```
class PlayerView
  protected void onVideoSizeChanged(....) {
    // implementation
  }
  private final class ComponentListener extends .... {
    @Override
    public void onVideoSizeChanged(....) {
      PlayerView.this.onVideoSizeChanged(....);
    }
  }
}
```

This lets you override behavior, which avoiding both of the issues mentioned above. Happy to merge a pull request that makes this change, if you send one.

FD-:
Sounds like a good idea!

ojw28:
We will merge https://github.com/google/ExoPlayer/pull/3942 to support this.

PR #3942, opened 2018-03-04
Allow overriding ComponentListener method functionality

FD-:
Make PlayerView and PlayerControlView more customisable as discussed in issue #3935.


FD-:
I signed it!

PR #3975, opened 2018-03-12
Remove duplicate code MediaSessionConnector.java

Khang-NT:
Fix updateMediaSessionMetadata() called twice;


ojw28:
All changes must go to the `dev-v2` branch. Please could you send this change to that branch instead, where we will happily merge it? Thanks.

Issue #3453, opened 2017-11-13
Consider not calling getFD().sync() in CacheDataSink.closeCurrentOutputStream


ojw28:
This call can be slow, and only guards against scenarios such as power loss. We should look at whether we can just remove it, or make it conditional (e.g. call it for proper offline use cases, but not for opportunistic caching).

ojw28:
A fix was merged for this. We may flip the default to not calling sync() at some point in the future, based on experimental results.

PR #4000, opened 2018-03-19
Provide an option to skip file descriptor sync in CacheDataSink. closeCurrentOutputStream

jianx9:
Pull request for https://github.com/google/ExoPlayer/issues/3453. Make this optional for now so we don't break the download case.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSink.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSink.java
@@ -43,6 +43,7 @@ public final class CacheDataSink implements DataSink {
   private final Cache cache;
   private final long maxCacheFileSize;
   private final int bufferSize;
+  private final boolean skipFDSync;
 
   private DataSpec dataSpec;
   private File file;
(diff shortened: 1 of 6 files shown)


jianx9:
I signed it!

jianx9:
I updated CLA.

library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSink.java
```diff
@@ -70,9 +71,10 @@ public CacheDataSinkException(IOException cause) {
    * @param maxCacheFileSize The maximum size of a cache file, in bytes. If the sink is opened for
    *    a {@link DataSpec} whose size exceeds this value, then the data will be fragmented into
    *    multiple cache files.
+   * @param skipFDSync Skip file descriptor sync when closing current output stream.
```
ojw28: Could we add keep a 2 argument constructor, to avoid having to update all the calling sites, which is most of the rest of this change?

Also, the more I think about this the more I think it'll be fine to eventually default to not doing the sync. What I'd propose is that the two arg constructor defaults to doing to sync for now, we'll experiment with turning the sync off for some internal apps we have, and flip it at some point in the future if we don't see any increase in corruption issues.

Given the plan to eventually default to not doing the sync, it's IMO preferable if the argument is flipped from what it is now (i.e. called `syncFileDescriptors`), so the eventual default will be `false`.

jianx9: Thanks! I will update the change accordingly.

jianx9:
@ojw28 I've updated the changes accordingly, let me know if you see other issues. Thanks!

jianx9:
@ojw28 I see it is still open, anything I need to do from my side for this one?

ojw28:
No action needed. We'll get this merged in due course; thanks!

PR #4008, opened 2018-03-20
Improved minimumUpdatePeriod when not present

tidoemanuele:
when minimumUpdatePeriod is missed the default values is not zero but -9223372036854775807

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaSource.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaSource.java
@@ -878,7 +878,7 @@ public final class DashMediaSource extends BaseMediaSource {
       } else if (scheduleRefresh && manifest.dynamic) {
         // Schedule an explicit refresh if needed.
         long minUpdatePeriodMs = manifest.minUpdatePeriodMs;
-        if (minUpdatePeriodMs == 0) {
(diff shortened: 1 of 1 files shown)


ojw28:
The line of code you're changing is not intended to catch the case in which minimumUpdatePeriod is missing. It's intended to catch the case where the minimum update period is explicitly set to 0.

What behavior are you seeing when minimumUpdatePeriod is missing? Do you see constant refreshes? I think the correct fix would be to never refresh in this case, since as per the spec:

> If this attribute is not present it indicates that the MPD does not change.

Please clarify. Thanks!

tidoemanuele:
yes I see constant mpd requests when minimumUpdatePeriod is missing

ojw28:
Ok. I think the correct fix is to ensure the manifest refresh isn't scheduled if minimumUpdatePeriod is missing.

tidoemanuele:
I got your point but in the case of dash live stream I'd like to refresh the manifest periodically, any suggestions?

ojw28:
Fix your stream to comply with the DASH specification?

ojw28:
I don't think we can merge this. It explicitly goes against the DASH specification, since omitting minimumUpdatePeriod is defined to mean that the manifest does not change, which implies that we should not be re-requesting it. I'll merge a separate fix to not re-request in this case.

PR #4012, opened 2018-03-21
Escaped colons to avoid displaying of emoji

orlandocr:
Changed "urn:mpeg💨event:2012"
To "urn:mpeg:dash:event:2012"


AquilesCanta:
All changes must go to the dev-v2 branch, could you send the pull request there? Also, the CLA needs to be signed before we take in any changes.

PR #4024, opened 2018-03-23
Escaped colons to avoid displaying of emoji

orlandocr:
Changed "urn:mpeg💨event:2012"
To "urn:mpeg: dash :event:2012" (without the spaces inbetween)

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -94,7 +94,7 @@
   ([#3630](https://github.com/google/ExoPlayer/issues/3630)).
 * DASH:
   * Support in-band Emsg events targeting the player with scheme id
-    "urn:mpeg:dash:event:2012" and scheme values "1", "2" and "3".
+    "urn:mpeg`:dash:`event:2012" and scheme values "1", "2" and "3".
   * Support EventStream elements in DASH manifests.
 * HLS:
     * Add opt-in support for chunkless preparation in HLS. This allows an


RELEASENOTES.md
```diff
@@ -94,7 +94,7 @@
   ([#3630](https://github.com/google/ExoPlayer/issues/3630)).
 * DASH:
   * Support in-band Emsg events targeting the player with scheme id
-    "urn:mpeg:dash:event:2012" and scheme values "1", "2" and "3".
+    "urn:mpeg`:dash:`event:2012" and scheme values "1", "2" and "3".
```
AquilesCanta, line 97: Won't this result in "urn:mpeg`:dash:`event:2012"?

If so, could we just replace the double-quotes with `, making the final result be

Support in-band Emsg events targeting the player with scheme id
`urn:mpeg:dash:event:2012` and scheme values "1", "2" and "3".

Sorry for being annoying, it will look weird if only the dash piece has a different style.

orlandocr:
It seemed to me that escaping `:dash:` didn't result on a highlight. The highlight I saw, on Github's preview, was simply marking the content before and after the changes. Can you please confirm this?
Please proceed with either escaping the colons, or replacing the double-quotes with `, as long as no emoji or special highlight happens.
Thanks.

ojw28:
Looks like it does result in highlight just on `:dash:`. I agree with @AquilesCanta that it would be better just to replace to " characters with `.

orlandocr:
Oh ok, my bad.

Issue #1808, opened 2016-09-08
Support seeking for FLAC streams without SEEKTABLE.


nezen:
I want to know is flac seekable or not? Each time I seek it from the demo-ext,  the progress returns to zero immediately.

here is the sample I tried:

```
sampleAdapter.add(new Header("FALC"));
    sampleAdapter.add(new Sample("a flac sample",
            "http://www.zaishuxia.com/music/files/pingfanzhilu.flac",
            Util.TYPE_OTHER));

```

ojw28:
@erdemguven - Please take a look. Thanks!

erdemguven:
Hi nezen, currently flac files are seekable only if they include a SEEKTABLE metadata block. In this case your flac doesn't have one. 
You can check flac files to see if they have a seektable:
`metaflac --list sample.flac | grep SEEKTABLE`
To add 100 seek points to a file
`flac -S 100x sample.flac -o new_sample.flac`
Please see `flac` documentation for other options.

nezen:
erdemguven, thanks! I checked the flac with metaflac and there is no SEEKTABLE. But when I try to  seek it with MediaPlayer.seekTo() , it worked.  So what's going on?

erdemguven:
MediaPlayer uses libflac for seeking which just do a search in flac file for the correct position if there isn't any seektable. Currently ExoPlayer doesn't do that for flac files.

burix:
Hi, is such a feature on the roadmap for the exoplayer?

erdemguven:
@burix, sorry, it isn't.

PaulWoitaschek:
And why not? I receive a lot of user complaints about media the user cannot seek.

ojw28:
As above, seeking in FLAC already works if there's a SEEKTABLE metadata block. We may support seeking (by searching) in other FLAC files at some point, but it's a long way down our list of priorities right now. If you're able to contribute a pull request that adds support, please feel free to do so.

PaulWoitaschek:
Okay it sounded as if the feature was not wanted for some reason. I barely understand all this audio stuff but would really welcome if someone tackles this :)

botaydotcom:
This is finished and will be pushed to the dev branch within the next few days.

andrewlewis:
Fixed in 972304f16b0f5aba21439785b4c9e2b831e364f6.

PR #4032, opened 2018-03-24
Enable flac to seek without SEEKTABLE

sjnyag:
Pull request for #1808.
In the case flac has no SEEKTABLE, search seek position by libflac function.


botaydotcom:
Unfortunately this pull request cannot be merged because it's not quite correct. When I used it locally to seek around the sample provided on #1808, it usually ended up at a different place than the seek position (some time very far from the correct position).

I have implemented this feature anyway, so this is not needed anymore. Thank you very much for the effort.

sjnyag:
Thank you very much for review.
Sorry. I agree with your opinion. I did not verify with streams.
I am looking forward to the release of this feature.

ojw28:
Support was added in https://github.com/google/ExoPlayer/commit/972304f16b0f5aba21439785b4c9e2b831e364f6.

sjnyag:
Many thanks for your support.

PR #4039, opened 2018-03-27
Typo fixed

gaurangmhatre:


gaurangmhatre:
I signed it!

Gratitude!

Regards,
Gaurang Mhatre
Email: gaurang.mhatre@sjsu.edu
Mobile: +1 408-831-8859

tonihei:
Thanks for the fixes. I added one comment.

tonihei:
Ignore that. I didn't add a comment :) It's fine as it is.

README.md
```diff
@@ -38,16 +38,16 @@ repositories {
...
 
 ```gradle
 implementation 'com.google.android.exoplayer:exoplayer:2.X.X'
 ```
 
 where `2.X.X` is your preferred version. Alternatively, you can depend on only
-the library modules that you actually need. For example the following will add
-dependencies on the Core, DASH and UI library modules, as might be required for
+the library modules that you actually need. For example, the following will add
+dependencies to the Core, DASH and UI library modules, as might be required for
```
ojw28, line 50: I'm not convinced about this change. You "depend on" rather than "depend to", don't you? The way it's phrased after this change suggests to me that the Core, DASH and UI modules are actually being modified with new dependencies. Not sure the comma changes are really necessary either.

ojw28:
Closing this. The benefit is probably lower than the cost of merging (bearing in mind we'd also need to merge the change into our internal repository).

PR #4040, opened 2018-03-27
Fixed type

gaurangmhatre:


ojw28:
What is this change supposed to be doing?

tonihei:
That's most likely a failed copy of #4039. Closing under this assumption.

Issue #966, opened 2015-11-17
MPEGTS seeking capability


abdallah95:
Hello...why isn't possible to seek a MPEGTS stream? MPEGTS is providing a great advantage for my application but the drawback is that I'm not able to seek. My case is that I'm streaming MP4 through UDP, and some packets might get dropped so instead of re-transmission I replace the dropped frame by an old frame but doing this is causing problem with Exoplayer(the video doesn't even play) unless I used MPEGTS container format where it distorts as expected but at least it continues playing.

abdallah95:
I will take a look into HLS if I could do a work around. 
Well it will only be possible to seek backward and not forward using a cache.
I'm replacing frames because I cannot skip frames, and I guess this should be the case when live streaming, and my case is that I'm not using a media source from which to request the data at anytime, request happens only once and then data transmission starts through UDP until end of stream is reached.

Avetri:
Indexing of TS files there are In solutions like DVR. So, if somebody would be so interested, an indexer might help.

abdallah95:
Actually I already worked on it and used PCR value in the TS packet header for indexing. You can take a look at it on pull request #1274

ojw28:
Now supported in `dev-v2`.

PR #4076, opened 2018-04-03
Add duration calculation and seeking support for TS files

FD-:
This adds seek capabilities and duration calculation for ts files, based on pcr values.

The original idea came from Pull Request #1274, but:

- Updated to newest ExoPlayer structure
- Refactored to match ExoPlayer's code style
- Improved seek speed by skipping irrelevant payloads while seeking
- Added ts duration calculation (By comparing pcr values from the start and end of the stream)

**Please Note:**

- As ts files do not include a real index structure, all seek information has to be collected as it passes the extractor. This means seeking to a stream section that has not been visited yet is slow. I have improved the performance by skipping irrelevant ts packet payloads while seeking, but it's still slower than seeking in a format that has an index structure (and it's still slower than ts seeking in VLC).

- For calculating the duration of a ts file, this implementation reads a few packets from the start and from the end of the stream to get the maximal difference between the pcr time stamps in the stream. In order to read packets from the end of the stream, the `TsExtractor` gets access to the `DataSource` and `DataSpec`, which it uses to create a new `DataSource` from the end of the stream. This should work faster than skipping through the `ExtractorInput`. The original `DataSource` is then restored.

- The implementation is thought and has been tested for local ts files. It might not work for other use cases.

I have read your stance on supporting ts seeking (Issue #966), so don't know if you'll merge this into the library. I personally consider it a far better user experience to support a (somewhat slow) seeking capability than to not support seeking at all. This is particularly annoying when you wish to resume playback from where you stopped previously.
Additionally,  it would be great to get rid of all these local modifications that make it more effort to migrate whenever there are changes to the ExoPlayer library.

Anyway, this should provide a good starting point for anyone who'd like to use ts seeking in a recent version of the ExoPlayer library.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
@@ -47,6 +47,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


FD-:
The seeking performance can be further improved by using the duration and file size to interpolate `SeekPoint`s for positions that have not yet been visited. I've not committed that for now, as it _might_ lead to imprecise seeking when the stream has a _highly_ varying bitrate. Let me know if I should commit it anyway. For my tests, it improved seeking performance _significantly_ (about 10x faster).

FD-:
I've now committed fast seeking. As I mentioned, it greatly improves seek performance. The worst that can happen if the bitrate varies too much is that seeking jumps to a point that is a bit after the actual position for the requested time, but I wouldn't actually consider that too much of an issue.

ojw28:
We have merged a different sequence of changes that should add this functionality. The last required change was 077f2c3036f417d641f8d4181aeeef97818096b2. Please give it a try on the `dev-v2` branch and let us know if you encounter any problems. Thanks!

FD-:
Looks great! What performance are you getting for seeking? Are you interpolating based on duration and file size?

ojw28:
It's using a type of binary search. Performance should be relatively good, but the best way to check is to try it for yourself!

FD-:
Tested now. Works great and seems faster than my implementation. Thanks!

Issue #4143, opened 2018-04-18
Pause player while playing VMAP 


Veeyikpong:
Hi,
I am implementing ExoPlayer and its ImaAdsLoader to play VMAP.
There are 4 ads in my playlist, 3 are 10 seconds and one is 15 seconds. 

When the ads are playing the 10 seconds ads, we pause and resume the player, everything works fine.

However, when playing the 15 seconds ad, if we pause BEFORE 10 seconds remaining, and resume the ad, after this ad has completed, the remaining ads will be skipped automatically. Then the content resume.

Please help on this issue. We tried a lot of ways but it still didnt work.

andrewlewis:
Please could you provide your ad tag here or via email to `dev.exoplayer@gmail.com` (with subject line "Issue #4143")? Thanks.

Veeyikpong:
Hi @andrewlewis 
I might have found the fix. I noticed that in ImaAdsLoader, when we pause the ad, the state is set to  IMA_AD_STATE_PAUSED;

` imaAdState = IMA_AD_STATE_PAUSED;`

and when RESUMED is triggered, we need to set the state back to IMA_AD_STATE_PLAYING to ensure the pause and resume works correctly for VMAP ads.

In handleAdEvent(AdEvent adEvent) function,
```
 case RESUMED:
                imaAdState = IMA_AD_STATE_PLAYING;
                break;
```
Please help to verify that this is correct and might be able to help others.
Thank you.

andrewlewis:
Thanks for providing the ad tag via email. I can reproduce the issue. If I background the demo app two seconds into the first ad I find that the remaining ads in the group are skipped even with the proposed fix in #4153. Is that consistent with what you're seeing, or did this fix work for you?

After background/resume we get an unexpected PAUSED event from IMA when there are still five seconds of the ad to play. After that we get the URI for the third ad in the pod, which suggests that IMA thinks we are already playing the second ad while we are still on the first. I'll investigate further.

andrewlewis:
@Veeyikpong Is your proposed fix working? I tried integrating it into the demo app and found that the issue was still reproducible.

andrewlewis:
This is likely to be a duplicate of #4140, which seems to be an IMA bug: in both cases, ads are skipped after resuming the ads manager while in a midroll.

Marking as a duplicate for now, but we can reopen if this turned out to be different once that issue is fixed.

andrewlewis:
[...]

PR #4153, opened 2018-04-19
Fiximaextension

Veeyikpong:
Set the **imaAdState** back to **IMA_AD_STATE_PLAYING** when the ads resumed, This can prevent some playback issue when pause and resume the player.
I encountered this issue when i have three 10 seconds ads and one 15 seconds ads in my playlist. When i pause the 15 seconds ads and resume it, the remaining ads will be skipped.
This commit can fix the issue.


Veeyikpong:
I signed it!

andrewlewis:
As discussed in #4143, this fix didn't work when I tried it. A recent IMA server-side fix may also address this issue, so I will close this for now. We can reopen it if this turns out to be the right way to work around the issue.

PR #4156, opened 2018-04-20
OkHttp extension - response headers instead of request in InvalidResponseCodeException

bangfalse:
InvalidResponseCodeException states that the "headerFields" should be the response headers and DefaultHttpDataSource appears to work this way, but OkHttp currently parrots back the request headers. This makes troubleshooting this exception very difficult when using OkHttp.


ojw28:
Thanks for spotting this :).

PR #4178, opened 2018-04-24
Support Webvtt subtitles class colors

### LINKED ISSUE #6581: Webvtt subtitles and embedded colouring.
Hi , I'm using exoplayer 2.10.6

Is there a way to enable embedded subtitle colouring in Webvtt subtitles? For example:

```
00:10:21.920 --> 00:10:23.960 line:83%
<c.white.bg_black>Test subtitle white foreground, black background</c>

00:10:56.840 --> 00:11:00.080 line:75%
<c.lime.bg_black>Test subtitle green, black background</c>
```

Bold, italics and underline Tags works but dynamic colouring doesn't.

danybony:
## Problem

Colors defined in Webvtt cue Class tags were ignored.
For example in the screenshots below the cue text was
```
At the <c.red.caps>left</c> we can see...
```
Example asset: http://dash.edgesuite.net/akamai/test/caption_test/ElephantsDream/elephants_dream_480p_heaac5_1.mpd

## Solution

`WebvttCueParser` already knows about the cue classes, but these were not used. So in case of a color class, we now apply a `ForegroundColorSpan` to the subtitle text.

### Screenshots

| Before | After |
| ------ | ----- |
| pasted image at 2018_03_29 11_46 am |  pasted image at 2018_03_29 11_43 am |


danybony:
I signed it!

AquilesCanta:
I thought VTT required the red class to be defined for the style to apply. Would you kindly provide a specification reference for this behavior change?

AquilesCanta:
Closing due to lack of activity.

icbaker:
Relevant specs:
https://www.w3.org/TR/webvtt1/#default-text-color
https://www.w3.org/TR/webvtt1/#default-text-background

Once this has a signed CLA and has been rebased/merged into current dev-v2 I can take a more in-depth look at the code.

icbaker:
I think this will likely also fix https://github.com/google/ExoPlayer/issues/6581

danybony:
I signed it!

icbaker:
@googlebot rescan

icbaker:
I can't find a record that either @danybony or dany.bony@gmail.com (the email on the commits in this PR) has signed the CLA.

Can you check what details you entered at https://cla.developers.google.com?

danybony:
Signed a newer one, just in case
image

icbaker:
@googlebot rescan

library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java
```diff
@@ -529,6 +532,16 @@ private static void applySpansForTag(
     }
   }
 
+  private static void applySupportedClasses(SpannableStringBuilder text, String[] classes,
+      int start, int end) {
+    for (String className : classes) {
+      if (ColorParser.isNamedColor(className)) {
+        int color = ColorParser.parseCssColor(className);
```
icbaker: This is neat code re-use, but because the WebVTT list is so much shorter than the list of CSS colors I think it would be clearer to explicitly list the WebVTT default colors. Otherwise if someone defines a custom class called [`papayawhip`](https://www.htmlcsscolor.com/hex/FFEFD5) in their style section that maybe has nothing to do with that color, we'll color the cue anyway (which is not in-line with the WebVTT spec).

So I think a `private static final Map<String, Integer> DEFAULT_COLORS` inside `WebvttCueParser` would be the best approach here. You'll probably have to populate it in a static initializer block the same as in `ColorParser`.

That will also make adding background color support less 'different' - we can just add a `DEFAULT_BACKGROUND_COLORS` map alongside. If you have the time/inclination, feel free to also add that in this same PR! :)

library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java
```diff
@@ -514,6 +515,8 @@ private static void applySpansForTag(
         text.setSpan(new UnderlineSpan(), start, end, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
         break;
       case TAG_CLASS:
+        applySupportedClasses(text, startTag.classes, start, end);
```
icbaker, line 544: Can you add some tests to `WebvttCueParserTest` for this?

icbaker (changes requested):

library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java
danybony: Sounds good, I'll take a look and update this soon

danybony: Updated the strategy by adding the two maps (adds support for background too) in `WebvttCueParser` and removing the dependency from the color parser.
Added relevant tests too

icbaker:
Thanks! I'll get this merged shortly.

icbaker (approved):

zegnus:
hi @icbaker what are the next steps on this PR?

PR #4164, opened 2018-04-22
support multiple #EXT-X-MAP tags

TakuSemba:
I'm trying to play HLS + fmp4 contents, and it didn't work when the HLS playlist has multiple contents separated by `#EXT-X-DISCONTINUITY`. 

It looks like `HlsPlaylistParser.java` captures the only one initialSegment the parser found at last eventually, but I think there is a situation where multiple init files exist separated by `#EXT-X-DISCONTINUITY`.

According to the [spec](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.5), it says
> The EXT-X-MAP tag specifies how to obtain the Media Initialization
   Section (Section 3) required to parse the applicable Media Segments.
   It applies to every Media Segment that appears after it in the
   Playlist until the next EXT-X-MAP tag or until the end of the
   playlist.

what shown below is what i tested for this pr.

### content description

this HLS [url](http://160.16.54.236:8081/exoplayer/hls-drm-aes/playlist.m3u8) is available and it can be reproduced, this is what I prepared.

#### playlist.m3u8
```playlist.m3u8
#EXTM3U
#EXT-X-VERSION:6
## Generated with https://github.com/google/shaka-packager version 2453c93f91-release
#EXT-X-TARGETDURATION:12
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MAP:URI="video1/init.mp4"
#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI="data:text/plain;base64,AAAAN3Bzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABciD3Rlc3QgY29udGVudCBpZEjj3JWbBg==",KEYID=0x6D76F25CB17F5E16B8EAEF6BBF582D8E,KEYFORMATVERSIONS="1",KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed"
#EXTINF:10.000,
video1/1.m4s
#EXTINF:10.792,
video1/2.m4s
#EXTINF:9.792,
...
#EXTINF:10.000,
video1/59.m4s
#EXTINF:5.500,
video1/60.m4s
#EXT-X-DISCONTINUITY
#EXT-X-KEY:METHOD=AES-128,URI="http://160.16.54.236:8081/nondrm/hls-two-encryption/test2.key",IV=0x0123456789ABCDEF0123456789ABCDEF
#EXT-X-MAP:URI="video2/init.mp4"
[...]

PR #4145, opened 2018-04-18
Fix order of REGEX_METHOD pattern

TakuSemba:
I'm trying to play a content of `HLS + WideVine + cenc` , and I found that the content is not playing properly on pre-Nougat devices.

`HLS + WideVine + cenc` should be playable according to [this](https://google.github.io/ExoPlayer/supported-formats.html)

After I debugging the code, it seems like ExoPlayer judges the content `cbcs` mode, even if the content is `cenc` mode at [this](https://github.com/google/ExoPlayer/blob/release-v2/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java#L428) point.  

Shouldn't `METHOD_SAMPLE_AES ` be put after `METHOD_SAMPLE_AES_CENC `, `METHOD_SAMPLE_AES_CTR `??, otherwise `METHOD_SAMPLE_AES_CENC`, `METHOD_SAMPLE_AES_CTR` would never match.

```
#EXTM3U
#EXT-X-VERSION:6
## Generated with https://github.com/google/shaka-packager version 2453c93f91-release
#EXT-X-TARGETDURATION:12
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MAP:URI="init.mp4"
#EXTINF:10.000,
1.m4s
#EXT-X-DISCONTINUITY
#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI="data:text/plain;base64,AAAAN3Bzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABciD3Rlc3QgY29udGVudCBpZEjj3JWbBg==",KEYID=0x6D76F25CB17F5E16B8EAEF6BBF582D8E,KEYFORMATVERSIONS="1",KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed"
#EXTINF:10.792,
2.m4s
#EXTINF:9.792,
3.m4s
#EXTINF:10.000,
```

```java
  private static final String METHOD_NONE = "NONE";
  private static final String METHOD_AES_128 = "AES-128";
  private static final String METHOD_SAMPLE_AES = "SAMPLE-AES";
  // Replaced by METHOD_SAMPLE_AES_CTR. Keep for backward compatibility.
  private static final String METHOD_SAMPLE_AES_CENC = "SAMPLE-AES-CENC";
[...]

PR #4180, opened 2018-04-25
support drm init data per discontinuity

TakuSemba:
Im trying to play a HLS content, which has a DRM content and a non-DRM content separated by `#EXT-X-DISCONTINUITY`, it seems like ExoPlayer uses secure decoder even when playing non-DRM content.

I think it is because ExoPlayer holds a single `drmInitData`.

I fixed to let ExoPlayer have `drmInitData` per  `#EXT-X-DISCONTINUITY`, and I confirmed ExoPlayer switches decoder between a DRM content and a non-DRM content.

You can use the [hls url](http://160.16.54.236:8081/exoplayer/hls-drm-aes/playlist.m3u8) that I prepared, but this would only work after successfully these my pull requests merged.

 - https://github.com/google/ExoPlayer/pull/4164
 - https://github.com/google/ExoPlayer/pull/4145

I hope this pr would be acceptable.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -360,7 +360,7 @@ import java.util.List;
             isTimestampMaster,
             timestampAdjuster,
             previous,
-            mediaPlaylist.drmInitData,
+            segment.drmInitData,
             encryptionKey,
             encryptionIv);
   }
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
@@ -42,6 +42,11 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
      * used for all segments that share an EXT-X-MAP tag.
      */
     @Nullable public final Segment initializationSegment;
+    /**
+     * DRM initialization data for sample decryption, or null if none of the segment uses sample
+     * encryption.
+     */
+    @Nullable public final DrmInitData drmInitData;
     /** The duration of the segment in microseconds, as defined by #EXTINF. */
     public final long durationUs;
     /**
@@ -81,7 +86,7 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
      * @param byterangeLength See {@link #byterangeLength}.
      */
     public Segment(String uri, long byterangeOffset, long byterangeLength) {
-      this(uri, null, 0, -1, C.TIME_UNSET, null, null, byterangeOffset, byterangeLength, false);
+      this(uri, null, null, 0, -1, C.TIME_UNSET, null, null, byterangeOffset, byterangeLength, false);
     }
 
     /**
@@ -99,6 +104,7 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
     public Segment(
         String url,
         Segment initializationSegment,
+        @Nullable DrmInitData drmInitData,
         long durationUs,
         int relativeDiscontinuitySequence,
         long relativeStartTimeUs,
@@ -109,6 +115,7 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
         boolean hasGapTag) {
       this.url = url;
       this.initializationSegment = initializationSegment;
+      this.drmInitData = drmInitData;
       this.durationUs = durationUs;
       this.relativeDiscontinuitySequence = relativeDiscontinuitySequence;
       this.relativeStartTimeUs = relativeStartTimeUs;
@@ -184,11 +191,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
    */
   public final boolean hasProgramDateTime;
   /**
-   * DRM initialization data for sample decryption, or null if none of the segment uses sample
-   * encryption.
-   */
-  public final DrmInitData drmInitData;
-  /**
    * The list of segments in the playlist.
    */
   public final List<Segment> segments;
@@ -211,7 +213,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
    * @param hasIndependentSegmentsTag See {@link #hasIndependentSegmentsTag}.
    * @param hasEndTag See {@link #hasEndTag}.
    * @param hasProgramDateTime See {@link #hasProgramDateTime}.
-   * @param drmInitData See {@link #drmInitData}.
    * @param segments See {@link #segments}.
    */
   public HlsMediaPlaylist(
@@ -228,7 +229,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
       boolean hasIndependentSegmentsTag,
       boolean hasEndTag,
       boolean hasProgramDateTime,
-      DrmInitData drmInitData,
       List<Segment> segments) {
     super(baseUri, tags);
     this.playlistType = playlistType;
@@ -241,7 +241,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
     this.hasIndependentSegmentsTag = hasIndependentSegmentsTag;
     this.hasEndTag = hasEndTag;
     this.hasProgramDateTime = hasProgramDateTime;
-    this.drmInitData = drmInitData;
     this.segments = Collections.unmodifiableList(segments);
     if (!segments.isEmpty()) {
       Segment last = segments.get(segments.size() - 1);
@@ -309,7 +308,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
         hasIndependentSegmentsTag,
         hasEndTag,
         hasProgramDateTime,
-        drmInitData,
         segments);
   }
 
@@ -337,7 +335,6 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
         hasIndependentSegmentsTag,
         /* hasEndTag= */ true,
         hasProgramDateTime,
-        drmInitData,
         segments);
   }
 
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
@@ -43,8 +43,8 @@ public final class HlsMediaPlaylist extends HlsPlaylist {
      */
     @Nullable public final Segment initializationSegment;
     /**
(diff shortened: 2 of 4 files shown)


library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -358,7 +358,7 @@ public void getNextChunk(HlsMediaChunk previous, long playbackPositionUs, long l
             isTimestampMaster,
             timestampAdjuster,
             previous,
-            mediaPlaylist.drmInitData,
+            mediaPlaylist.drmInitData.get(segment.relativeDiscontinuitySequence),
```
AquilesCanta: I think more than one drm init data can map to a single discontinuity sequence.

TakuSemba: Thank you for taking a look at this pr.

> more than one drm init data can map to a single discontinuity sequence

I coundn’t think of the situation of multiple init data in a single discontinuity sequence
Could you tell me how the playlist look like?
Do you mean we have to apply drmInitData until the next `#EXT-X-KEY` or the end of the playlist?

And one more thing.
As you explained [here](https://github.com/google/ExoPlayer/pull/4164#issuecomment-384239441), you don’t want to use SparseArray?
If so, how do you want me to implement it?

TakuSemba:
@AquilesCanta 
seeing you implemented multiple EXT-X-MAP tags here (https://github.com/google/ExoPlayer/commit/ddef32c9e2f8d754d46af58566229f179bb00bc8), I also implemented this in the same way as each segment has each drmInitData.

AquilesCanta:
Hi @TakuSemba, thanks for the PR. Will look into this as soon as I have some time, I promise. Hopefully we can get it pulled from. Sorry for the delay.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
```diff
@@ -42,6 +42,11 @@
      * used for all segments that share an EXT-X-MAP tag.
      */
     @Nullable public final Segment initializationSegment;
+    /**
+     * DRM initialization data for sample decryption, or null if none of the segment uses sample
```
AquilesCanta: Should probably be changed to

{@link DrmInitData} for sample decryption, or null if the segment's samples are not DRM-protected.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMediaPlaylist.java
```diff
@@ -81,7 +86,7 @@
      * @param byterangeLength See {@link #byterangeLength}.
      */
     public Segment(String uri, long byterangeOffset, long byterangeLength) {
-      this(uri, null, 0, -1, C.TIME_UNSET, null, null, byterangeOffset, byterangeLength, false);
+      this(uri, null, null, 0, -1, C.TIME_UNSET, null, null, byterangeOffset, byterangeLength, false);
```
AquilesCanta: Nit (not sure it applies): Lines should not exceed 100 chars.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -17,6 +17,7 @@
 
 import android.net.Uri;
 import android.util.Base64;
+import android.util.SparseArray;
```
AquilesCanta: Is this a stray import?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -424,15 +425,16 @@ private static HlsMediaPlaylist parseMediaPlaylist(LineIterator iterator, String
           } else if (method != null) {
             SchemeData schemeData = parseWidevineSchemeData(line, keyFormat);
             if (schemeData != null) {
-              drmInitData =
-                  new DrmInitData(
+              drmInitData = new DrmInitData(
```
AquilesCanta: Can we leave this as before?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -360,7 +360,7 @@ public void getNextChunk(HlsMediaChunk previous, long playbackPositionUs, long l
             isTimestampMaster,
             timestampAdjuster,
             previous,
-            mediaPlaylist.drmInitData,
+            segment.drmInitData,
```
AquilesCanta, line 363: There is an issue with this: If the first segment is not encrypted, then the media chunk will expose no DRM init data, and the track selector will think the segment is not encrypted. I'll try to think about a solution for this.

AquilesCanta (changes requested):

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
TakuSemba: Sorry, I coundn't understand your concern.
I know the fact that shaka-packager creates a playlist like ↓ for a Widevine-encrypted content.
In that case, I believe that video_1.m4s is not (and should not be) encrypted and video_2.m4s ~ are encrypted with Widevine.
I think the track selector should think the segment is not encrypted, because it's not encrypted. 🤔 


```
#EXTM3U
#EXT-X-VERSION:6
#EXT-X-TARGETDURATION:18
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MAP:URI="video_init.mp4"
#EXTINF:13.583,
video_1.m4s
#EXT-X-DISCONTINUITY
#EXT-X-KEY:METHOD=SAMPLE-AES-CTR,URI="data:text/…
#EXTINF:10.417,
video_2.m4s
#EXTINF:7.417,
video_3.m4s
…
```

AquilesCanta: The track selector does not know about specific segments, but rather about tracks. The issue is that the track is indeed encrypted (even if some of the segments are clear), but the first segment's format (unencrypted), as things are now, will be propagated as the track's format.

TakuSemba: Thanks for the replay! I understood what you mean.

> will be propagated as the track's format.

but, I could not find the exact place where the track selector thinks the segment is not encrypted or a wrong track's format is propagated. 

you want to me to fix it on this PR? 
I'm sorry to bother you, but it would be really appreciated if you could tell me the specific point where it's happening.

AquilesCanta: What I suspect might be a solution for this is keeping a DrmInitData in the media playlist which contains SchemeData's with null SchemeData#data, so as to signal that encryption of a specific CDM is eventually there, without tying it to a specific key/pssh. This DrmInitData would then be propagated by the period for the track selector to consume. This field would be populated by the playlist parser after parsing all EXT-X-KEYS present in the playlist.

TakuSemba: I'm not sure about what you suggest.
Does your suggestion mean HlsMediaPlaylist should hold a single DrmInitData with null SchemeData#data? (Instead of letting each segment hold a DrmInitData)

and again, Im playing the content, which has the first non-encrypted segment and the other widevine-encrypted segments, and It plays without any problems, and could not find the place where wrong track's formats are created or propagated while breakpointing inside exoplayer.

AquilesCanta: > Does your suggestion mean HlsMediaPlaylist should hold a single DrmInitData with null SchemeData#data

No, it should contain as many scheme datas as encryption schemes there are in the content. For example, if the media is encrypted using both Widevine and Playready, in would contain two SchemeDatas, one for each.

> Im playing the content, which has the first non-encrypted segment and the other widevine-encrypted segments

This issue does not affect playback. We have a similar problem to the one introduced here in chunkless preparation (but unlike here, we cannot do much without downloading extra stuff, which defeats the purpose of chunkless preparation). 

The issue is track selection. The tracks exposed by the MediaPeriod will not contain DRM data, so the track selector will assume the content is unencrypted. The resulting behavior is a crash instead of graceful handling of unsupported DRM schemes. In this case, you are playing on a device that supports the used DRM scheme. If it didn't, the player would crash instead (potentially -though rare- with a native crash). Hopefully this is clear.

TakuSemba: Thank you for the explanation!.

> The tracks exposed by the MediaPeriod will not contain DRM data

I think the place where the MediaPeriod exposes the tracks is [here](https://github.com/google/ExoPlayer/blob/release-v2/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java#L118).

But, it seems like the MediaPeriod has the track groups with non-null drmInitData in it.



http://160.16.54.236:8081/drmtest/media/drm/180p/master.m3u8 is the stream that I tested .

```json
{
  "name": "HLS Plain (180p)",
  "uri": "http://160.16.54.236:8081/drmtest/media/plain/180p/master.m3u8"
},
{
  "name": "HLS DRM (180p)",
  "uri": "http://160.16.54.236:8081/drmtest/media/drm/180p/master.m3u8",
  "drm_scheme": "widevine",
  "drm_license_url": "https://proxy.uat.widevine.com/proxy"
},
{
  "name": "HLS Plain (180p) -> DRM (180p)",
  "uri": "http://160.16.54.236:8081/drmtest/playlist/plain180p-drm180p/master.m3u8",
  "drm_scheme": "widevine",
  "drm_license_url": "https://proxy.uat.widevine.com/proxy"
},
```

And your solution is not clear to me yet. 

> keeping a DrmInitData in the media playlist which contains SchemeData's with null SchemeData#data

does your suggestion mean keeping an another DrmInitData only for the first segment like ↓?

```java
  private static HlsMediaPlaylist parseMediaPlaylist(LineIterator iterator, String baseUri)
      throws IOException {
    ...
    String encryptionKeyUri = null;
    String encryptionIV = null;
    DrmInitData drmInitData = null;
    DrmInitData drmInitDataforFirstSegment = // drmInitData which contains SchemeData's with null SchemeData#data;

    String line;
    while (iterator.hasNext()) {
      line = iterator.next();
        ...
    }
  }
```

Issue #4185, opened 2018-04-26
DASH - Clip playback at period end times (i.e. once period duration is reached)


kevleyski:
### Issue description
A static MPD with multiple Periods can play out differently between Shaka (e.g. PC/Chromecast) vs Android and AndroidTV devices.

Looking further it seems that the logic in library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java::updatePeriods() is the reason for this because Periods will not advance to the next Period when there might still be some audio/video data remaining to be rendered.

Example work-around...

    // If end of Period is reached then allow advancing to next Period
    boolean isAtEndOfPeriod = (rendererPositionUs >= playingPeriodHolder.next.rendererPositionOffsetUs);

    for (int i = 0; i < renderers.length; i++) {
      Renderer renderer = renderers[i];
      SampleStream sampleStream = readingPeriodHolder.sampleStreams[i];
      if (renderer.getStream() != sampleStream
          || (sampleStream != null && !renderer.hasReadStreamToEnd() && !isAtEndOfPeriod)) {
        // The current reading period is still being read by at least one renderer.
        return;
      }
    }

### Reproduction steps
Posted privately is an example MPD that has correct Periods
However the segments between the audio and video adaptation do not quite align, that is they do not all finish exactly at the point of the end of the Period
The second Period then does not play out until all renderers have finished.
In contrast, Shaka (v2.3.0) and DASH.IF (v2.6.8) advance correctly based on Period duration/start of next period.

### Link to test content
(private post)

### Version of ExoPlayer being used
2.7.3 (release-v2 branch)

### Device(s) and version(s) of Android being used
Google Pixel 2 'phone: Oreo
Emulator: Android P

### A full bug report captured from the device
(private post)

ojw28:
I think this needs fixing somewhere in the DASH module, rather than in `ExoPlayerImplInternal` as in the pull request. In particular this is because:

* I don't think we want this kind of clipping for other media types (e.g. a playlist of mp3 files, where durations may be inaccurate but we really don't want to start clipping before moving on to the next one).
* From DASH IF discussions, my understanding is that period duration is allowed to be much shorter the underlying media. In this case the player shouldn't buffer media beyond the period duration, because it's inefficient to do so. A fix in the DASH module would be necessary to avoid this.

kevleyski:
Let me know if I can help further, happy to try/test

kevleyski:
[...]

PR #4186, opened 2018-04-26
Issue #4185: Shaka vs ExoPlayer behaviour difference. Multi-Periods d…

kevleyski:
…o not advance whilst media renderer has more data

A static MPD with multiple Periods can play out differently between Shaka (e.g. PC/Chromecast) vs Android and AndroidTV devices.

Looking further it seems that the logic in updatePeriods() is the reason for this because Periods will not advance to the next Period when there might still be some audio/video data remaining to be rendered.

### DIFF
--- a/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -1013,8 +1013,14 @@ import java.io.IOException;
       return;
     }
 
+    // Issue #4185: If end of Period is reached then this triggers next Period also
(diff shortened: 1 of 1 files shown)


kevleyski:
(have signed CLA for personal contribution)

kevleyski:
User name and email address were correct

kevleyski:
Corrected email, resigned CLA for personal contribution

ojw28:
Closing as this needs fixing in the DASH module, rather than ExoPlayerImplInternal. Let's use the bug for tracking. Thanks!

kevleyski:
No worries, thanks

Issue #4168, opened 2018-04-23
All CastPlayer bugs I found and possiblie solutions


mkaflowski:
I work for some chromecast-ready apps and CastPlayer make it a lot of easier but I found a lot of troubles with it too:

**1. Wrong getCurrentWindowIndex after setting media sources when CastPlayer is connected to Chromecast device.**

There is use case where you connect to your chromecast device and after that you select media source.

How to recreate:
- connect to Chromecast device by tapping MediaRouteButton button
- create CastPlayer and set sources (castPlayer.loadItems(mediaItems, window, time, Player.REPEAT_MODE_OFF)) after some time (you can do it with some button)
- get currentWindowIndex (it's always 0)

Cause:
StatusListener don't invoke onSessionStarted which don't invoke setRemoteMediaClient when we set new sources so statusListener is no added to remoteMediaClient.

Solution:
Add statusListener to remoteMediaClient in contructor:
```java
	if (remoteMediaClient != null) {
            try {
                remoteMediaClient.addListener(statusListener);
                remoteMediaClient.addProgressListener(statusListener, PROGRESS_REPORT_PERIOD_MS);
                updateInternalState();
            } catch (NullPointerException ignored) {
            }
        }
```
		
**2. I get 0 from getCurrentWindowIndex() which is not true.**

How to recreate:
- invoke getCurrentWindowIndex() in onPositionDiscontinuity

Cause:
We get the value from fetchCurrentWindowIndex() when there is no media session. So current window is not actually 0.

Solution:
Change 0 to other value in fetchCurrentWindowIndex() when there is no media session. You will be able to recognize the real current window index from "no media session":
```java	
     /**
     * Retrieves the current item index from {@code mediaStatus} and maps it into a window index. If
     * there is no media session, returns -1.
     */
    private static int fetchCurrentWindowIndex(@Nullable MediaStatus mediaStatus) {
        Integer currentItemId = mediaStatus != null
                ? mediaStatus.getIndexById(mediaStatus.getCurrentItemId()) : null;
        return currentItemId != null ? currentItemId : -1;
    }
```

You should change getDuration and some other methods (isCurrentWindowDynamic(), isCurrentWindowSeekable()) in this case too to avoid some exceptions:
```java	
@Override
    public long getDuration() {
               return currentTimeline.isEmpty() ? C.TIME_UNSET
                : currentTimeline.getWindow(getCurrentWindowIndex() < 0 ? 0 : getCurrentWindowIndex(),
[...]

PR #4223, opened 2018-05-08
CastPlayer bug fixes for issues/4168

mkaflowski:
Couple bug fixes for issue:
[https://github.com/google/ExoPlayer/issues/4168](https://github.com/google/ExoPlayer/issues/4168)
Mostly wrong currentWindowIndex.

### DIFF
--- a/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
+++ b/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
@@ -18,6 +18,7 @@ package com.google.android.exoplayer2.ext.cast;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
+
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player;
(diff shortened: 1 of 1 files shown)


dachev:
Are there plans to merge this one? The extension has limited utility without these fixes.

AquilesCanta:
@dachev, as mentioned in #4168, there are some violations of the player interface which means we cannot pull as it is. I'll check if there is any specific change that may be worth picking up.

marcin-adamczewski:
@AquilesCanta What are the plans for fixing it? Unfortunately, the Cast extension is really useless with these bugs. I think you should prioritise it.

AquilesCanta:
We are working on a custom receiver app using our own communication protocol, so we'll be able to ensure close to end-to-end control. The new app should replace the existing components. Please be patient. I cannot provide any timelines.

mkaflowski:
> @AquilesCanta What are the plans for fixing it? Unfortunately, the Cast extension is really useless with these bugs. I think you should prioritise it.

You can use fixed class for now. I have implemented them in HybridMediaPlayer lib if you need example.

ojw28:
We're working on revamping the cast extension to resolve the issues that exist with the current iteration. This should happen relatively soon, so I don't think it's a good use of time to merge fixes to the current version in the meantime. Marking as wontmerge, but will leave the pull request around until the revamped version lands.

tonihei:
Closing because it was marked as "won't merge"

PR #4243, opened 2018-05-11
Updating TextRenderer output if last subtitle is not fully played

pedromfmachado:
When the `TextRenderer` treats EOS but the last subtitle has not been fully played out, the renderer should update the output by setting `textRendererNeedsUpdate` to `true`. Not updating it causes situations where the current subtitle will not be show (e.g.:  attaching the player to another `PlayerView`)

- ### Here's a GIF of the problematic situation:
not-fixed

- ### And here's a GIF with the implemented fix:
fixed

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/TextRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/TextRenderer.java
@@ -189,6 +189,8 @@ public final class TextRenderer extends BaseRenderer implements Callback {
             releaseBuffers();
             outputStreamEnded = true;
           }
+        } else {
+          textRendererNeedsUpdate = true;
         }
       } else if (nextSubtitle.timeUs <= positionUs) {
         // Advance to the next subtitle. Sync the next event index and trigger an update.


pedromfmachado:
Hello

ojw28:
I see the problem, but I'm confused about the fix. Isn't this a general problem when switching from one `PlayerView` to another, even if you're a long way away from EOS? It looks like when a `PlayerView` is bound to a player, it doesn't have a way of obtaining the current caption to display. Hence it wont display anything until the next caption is reached.

pedromfmachado:
Thanks for your comment @ojw28.

I've been trying to understand how the `TextRenderer` works and considering the usage of `SingleSampleMediaSource` this fix would make sense, since the whole file is loaded and the EOS is set after that same loading.
However I agree that it's not the most straightforward fix (as it is only based on the usage of `SingleSampleMediaSource` and because it could also deteriorate performance).

As a workaround I've been disabling/enabling `TextRenderer` when the player is detached/attached to the `PlayerView`. Could disabling/enabling `TextRenderer` during `setPlayer` be a solution? Do you have any other suggestion that I could implement and contribute to ExoPlayer?

tonihei:
It feels the right solution would be to handle the issue in `SimpleExoPlayer`. You could save the cues received by `onCues(List<Cue)` and forward the currently active cues to new listeners when they are added in `addTextOutput`. 

Besides that, please make sure to merge your commits into the dev-v2 branch not the release branch.
Thanks!

[Closing because it needs to reupload in the dev branch]

PR #4252, opened 2018-05-14
Saving current subtitle cues on SimpleExoPlayer

pedromfmachado:
When attaching a player to a different `PlayerView` the current subtitles are not shown, as it's possible to see bellow:
not-fixed

This PR aims at solving that problem by saving the current subtitles cues to memory and restore them when a `TextOutput` is added, therefore solving the problem.

(Thanks @tonihei for the feedback on #4243)

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
@@ -92,7 +92,7 @@ public class SimpleExoPlayer implements ExoPlayer, Player.VideoComponent, Player
   private AudioAttributes audioAttributes;
   private float audioVolume;
   private MediaSource mediaSource;
-  private List<Cue> currentCues;
+  private @Nullable List<Cue> currentCues;
 
   /**
    * @param renderersFactory A factory for creating {@link Renderer}s to be used by the instance.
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
```diff
@@ -92,6 +92,7 @@
   private AudioAttributes audioAttributes;
   private float audioVolume;
   private MediaSource mediaSource;
+  private List<Cue> currentCues;
```
tonihei: Can mark this as `private @Nullable`?

library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
```diff
@@ -502,6 +503,7 @@ public void clearVideoListener(VideoListener listener) {
 
   @Override
   public void addTextOutput(TextOutput listener) {
+    listener.onCues(currentCues);
```
tonihei: And check for `null` here before sending the cues to the listener.

tonihei (commented):
Thanks, looks good. I added two comments concerning nullability.

tonihei:
Thanks for updating!

PR #4257, opened 2018-05-14
Convert I/O 2018 demos to Kotlin.

nic0lette:
I created 2 additional modules in the project, "audio-app-kotlin" and "video-app-kotlin" which (mostly) implement the same logic as in the Java version of their respective projects.


0xcaff:
Any reason why this is tagged wontmerge?

ojw28:
Because we don't want to maintain multiple versions of our demo apps.

tonihei:
Closing because it was marked as "won't merge"

PR #4269, opened 2018-05-18
Public notification manager createnotification

dulmandakh:
PlayerNotificationManager is saving me a ton of work, thank you.

There is a case when we pause a player, we need to stop foreground service. But when we want to resume the playback and start foreground service, there is no way to obtain notification from player state.

This PR will create public PlayerNotificationManager.createNotification which will return notification instance to be used with startForeground method. Sorry, I'm not good at Java programming.

Please review @ojw28


andrewlewis:
@dulmandakh Thank you for the pull request. Please could you move this to the `dev-v2` branch? I will close this for now because we can only accept changes there.

Issue #3906, opened 2018-02-26
CEA608 Missing Whitespaces


zsmatyas:
I see lots of content having captions seemingly missing space characters around italic words (when Mid-Row command is used):
"He saidHELLOimmediately. (where the word HELLO is italic in the caption) 

The CEA608 Standard seems to be not too clear about handling of Mid-Row Commands. Some parts mention adding a white space when processing a code, but it is quite ambiguous:

**Part 7.4 Real-Time Scrolling Display:**

> "In Text Mode"... "The cursor moves automatically one column to the right after each character or Mid-Row Code received."

This seems to describe Text Mode only.

**Part B.12 Backspacing:**

> "Backspace shall move the cursor one column to the left, erasing the character or Mid-Row Code occupying that location..."

This suggests that the Mid-Row code occupies the space for a character on the screen.

If we can rely on https://www.law.cornell.edu/cfr/text/47/79.101

> "The cursor moves automatically one column to the right after each character or Mid-Row Code received."

So all in all, this function probably needs to add a whitespace to the text to handle Mid-Row Codes appropriately:

https://github.com/google/ExoPlayer/blob/2b20780482a9c6b07416bcbf4de829532859d10a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java#L374

AquilesCanta:
I don't have full grasp of the 608 spec, so please correct me if I am wrong. As I understand it, our implementation only supports Caption Mode (Text mode is unsupported). 

> This suggests that the Mid-Row code occupies the space for a character on the screen.

I agree this is the case in Text mode. But I am not convinced that

> this function probably needs to add a whitespace to the text

won't break other captions that assume Caption mode. It would be great to receive some insight from someone experienced in 608.

zsmatyas:
After having met this bug again and again, when live channels seems to miss a whitespace around single words in italic, I have checked the 608 spec multiple times. Unfortunately, I have only found implicit signs that the Mid-Row command also requires the whitespace to be added at the current location. 

Sentences quoted above and:
> C.2 Processing Color in a Monochrome-Display Decoder (Regulatory)
 The decoder shall ignore the color selection designated in mid-row codes and non-indenting PACs, but it shall recognize the spacing and other attribute assignments these codes may designate.

> C.8 Solid Spaces Added for Legibility (Regulatory/Preferred)
[...]

PR #4308, opened 2018-05-28
Support 608/708 in H262 + parse isWideAspectRatio flag

peddisri:
Major re-design of CEA708 Cues decoder and subtitle rendering to make it compliant with Sarnoff test streams. Added support for
missing features required to pass Sarnoff tests. Minor fixes in CEA608 decoder too.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/C.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/C.java
@@ -80,6 +80,12 @@ public final class C {
   /** The number of bits per byte. */
   public static final int BITS_PER_BYTE = 8;
 
+  /** non-Wide aspect ratio */
+  public static final int NON_WIDE_ASPECT_RATIO_TYPE = 1;
+
+  /** Wide aspect ratio */
+  public static final int WIDE_ASPECT_RATIO_TYPE = 2;
+
   /**
    * The name of the ASCII charset.
    */
--- a/library/core/src/main/java/com/google/android/exoplayer2/Format.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/Format.java
@@ -15,9 +15,11 @@
  */
 package com.google.android.exoplayer2;
 
+import android.os.Bundle;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.support.annotation.Nullable;
+
 import com.google.android.exoplayer2.drm.DrmInitData;
 import com.google.android.exoplayer2.metadata.Metadata;
 import com.google.android.exoplayer2.util.MimeTypes;
@@ -43,6 +45,7 @@ public final class Format implements Parcelable {
    * the timestamps of their parent samples.
    */
   public static final long OFFSET_SAMPLE_RELATIVE = Long.MAX_VALUE;
+
 
   /** An identifier for the format, or null if unknown or not applicable. */
   public final @Nullable String id;
@@ -148,7 +151,16 @@ public final class Format implements Parcelable {
    */
   public final int encoderPadding;
 
+  /** A Bundle of custom parameters. Could be null or empty */
+  public final Bundle params;
+
   // Audio and text specific.
+  /**
+   * Indicates if the video is wide aspect ratio (16:9) or not (4:3)
+   * Only influences captions if the lines are not middle aligned.
+   * Values are {@link C#NON_WIDE_ASPECT_RATIO} or {@link C#WIDE_ASPECT_RATIO},
+   */
+  public static final String KEY_ASPECT_RATIO_TYPE = "aspect-ratio-type";
 
   /**
    * Track selection flags.
@@ -183,7 +195,7 @@ public final class Format implements Parcelable {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, width,
         height, frameRate, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, selectionFlags, null, NO_VALUE, OFFSET_SAMPLE_RELATIVE,
-        initializationData, null, null);
+        initializationData, null, null, null);
   }
 
   public static Format createVideoSampleFormat(
@@ -238,7 +250,7 @@ public final class Format implements Parcelable {
     return new Format(id, null, sampleMimeType, codecs, bitrate, maxInputSize, width, height,
         frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, 0, null, NO_VALUE,
-        OFFSET_SAMPLE_RELATIVE, initializationData, drmInitData, null);
+        OFFSET_SAMPLE_RELATIVE, initializationData, drmInitData, null, null);
   }
 
   // Audio.
@@ -257,7 +269,7 @@ public final class Format implements Parcelable {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, channelCount, sampleRate,
         NO_VALUE, NO_VALUE, NO_VALUE, selectionFlags, language, NO_VALUE, OFFSET_SAMPLE_RELATIVE,
-        initializationData, null, null);
+        initializationData, null, null, null);
   }
 
   public static Format createAudioSampleFormat(
@@ -313,7 +325,7 @@ public final class Format implements Parcelable {
     return new Format(id, null, sampleMimeType, codecs, bitrate, maxInputSize, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, channelCount, sampleRate, pcmEncoding,
         encoderDelay, encoderPadding, selectionFlags, language, NO_VALUE, OFFSET_SAMPLE_RELATIVE,
-        initializationData, drmInitData, metadata);
+        initializationData, drmInitData, metadata, null);
   }
 
   // Text.
@@ -342,7 +354,7 @@ public final class Format implements Parcelable {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, selectionFlags, language, accessibilityChannel,
-        OFFSET_SAMPLE_RELATIVE, null, null, null);
+        OFFSET_SAMPLE_RELATIVE, null, null, null, null);
   }
 
   public static Format createTextSampleFormat(
@@ -390,6 +402,24 @@ public final class Format implements Parcelable {
   }
 
   public static Format createTextSampleFormat(
+       @Nullable String id,
+	   @Nullable String sampleMimeType,
+	   @Nullable String codecs,
+       int bitrate,
+	   @C.SelectionFlags int selectionFlags,
+	   String language,
+	   int accessibilityChannel,
+       DrmInitData drmInitData,
+       Bundle params) {
+    return new Format(id, null, sampleMimeType, codecs, bitrate,
+            NO_VALUE, NO_VALUE, NO_VALUE,
+        NO_VALUE, NO_VALUE, NO_VALUE,
+            null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE,
+        NO_VALUE, NO_VALUE, selectionFlags, language, accessibilityChannel, OFFSET_SAMPLE_RELATIVE,
+        null, drmInitData, null, params);
+  }
+
+  public static Format createTextSampleFormat(
       @Nullable String id,
       @Nullable String sampleMimeType,
       @Nullable String codecs,
@@ -399,11 +429,12 @@ public final class Format implements Parcelable {
       int accessibilityChannel,
       @Nullable DrmInitData drmInitData,
       long subsampleOffsetUs,
-      List<byte[]> initializationData) {
+      List<byte[]> initializationData
+      ) {
     return new Format(id, null, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, selectionFlags, language, accessibilityChannel, subsampleOffsetUs,
-        initializationData, drmInitData, null);
+        initializationData, drmInitData, null, null);
   }
 
   // Image.
@@ -443,7 +474,7 @@ public final class Format implements Parcelable {
         OFFSET_SAMPLE_RELATIVE,
         initializationData,
         drmInitData,
-        null);
+        null, null);
   }
 
   // Generic.
@@ -459,14 +490,14 @@ public final class Format implements Parcelable {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, selectionFlags, language, NO_VALUE, OFFSET_SAMPLE_RELATIVE, null, null,
-        null);
+        null, null);
   }
 
   public static Format createSampleFormat(
       @Nullable String id, @Nullable String sampleMimeType, long subsampleOffsetUs) {
     return new Format(id, null, sampleMimeType, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, 0, null, NO_VALUE, subsampleOffsetUs, null, null, null);
+        NO_VALUE, 0, null, NO_VALUE, subsampleOffsetUs, null, null, null, null);
   }
 
   public static Format createSampleFormat(
@@ -477,7 +508,7 @@ public final class Format implements Parcelable {
       @Nullable DrmInitData drmInitData) {
     return new Format(id, null, sampleMimeType, codecs, bitrate, NO_VALUE, NO_VALUE, NO_VALUE,
         NO_VALUE, NO_VALUE, NO_VALUE, null, NO_VALUE, null, NO_VALUE, NO_VALUE, NO_VALUE, NO_VALUE,
-        NO_VALUE, 0, null, NO_VALUE, OFFSET_SAMPLE_RELATIVE, null, drmInitData, null);
+        NO_VALUE, 0, null, NO_VALUE, OFFSET_SAMPLE_RELATIVE, null, drmInitData, null, null);
   }
 
   /* package */ Format(
@@ -506,7 +537,8 @@ public final class Format implements Parcelable {
       long subsampleOffsetUs,
       @Nullable List<byte[]> initializationData,
       @Nullable DrmInitData drmInitData,
-      @Nullable Metadata metadata) {
+      @Nullable Metadata metadata,
+      Bundle params) {
     this.id = id;
     this.containerMimeType = containerMimeType;
     this.sampleMimeType = sampleMimeType;
@@ -535,6 +567,7 @@ public final class Format implements Parcelable {
         : initializationData;
     this.drmInitData = drmInitData;
     this.metadata = metadata;
+    this.params = params;
   }
 
   @SuppressWarnings("ResourceType")
@@ -570,6 +603,7 @@ public final class Format implements Parcelable {
     }
     drmInitData = in.readParcelable(DrmInitData.class.getClassLoader());
     metadata = in.readParcelable(Metadata.class.getClassLoader());
+    params = in.readBundle(getClass().getClassLoader());
   }
 
   public Format copyWithMaxInputSize(int maxInputSize) {
@@ -577,7 +611,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithSubsampleOffsetUs(long subsampleOffsetUs) {
@@ -585,7 +619,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithContainerInfo(
@@ -601,7 +635,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   @SuppressWarnings("ReferenceEquality")
@@ -622,7 +656,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithGaplessInfo(int encoderDelay, int encoderPadding) {
@@ -630,7 +664,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithDrmInitData(@Nullable DrmInitData drmInitData) {
@@ -638,7 +672,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithMetadata(@Nullable Metadata metadata) {
@@ -646,7 +680,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   public Format copyWithRotationDegrees(int rotationDegrees) {
@@ -654,7 +688,7 @@ public final class Format implements Parcelable {
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, params);
   }
 
   /**
@@ -825,6 +859,8 @@ public final class Format implements Parcelable {
     }
     dest.writeParcelable(drmInitData, 0);
     dest.writeParcelable(metadata, 0);
+    dest.writeBundle(params);
+
   }
 
   public static final Creator<Format> CREATOR = new Creator<Format>() {
--- a/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
@@ -15,8 +15,11 @@
  */
 package com.google.android.exoplayer2.extractor.ts;
 
+import android.os.Bundle;
 import android.support.annotation.IntDef;
 import android.util.SparseArray;
+
+import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.EsInfo;
 import com.google.android.exoplayer2.util.MimeTypes;
@@ -79,6 +82,13 @@ public final class DefaultTsPayloadReaderFactory implements TsPayloadReader.Fact
    */
   public DefaultTsPayloadReaderFactory(@Flags int flags, List<Format> closedCaptionFormats) {
     this.flags = flags;
+    if (!isSet(FLAG_OVERRIDE_CAPTION_DESCRIPTORS) && closedCaptionFormats.isEmpty()) {
+      closedCaptionFormats = new ArrayList();
+      closedCaptionFormats.add(Format.createTextSampleFormat(null,
+              MimeTypes.APPLICATION_CEA608,  0, null));
+      closedCaptionFormats.add(Format.createTextSampleFormat(null,
+              MimeTypes.APPLICATION_CEA708,  0, null));
+    }
     this.closedCaptionFormats = closedCaptionFormats;
   }
 
@@ -106,7 +116,7 @@ public final class DefaultTsPayloadReaderFactory implements TsPayloadReader.Fact
       case TsExtractor.TS_STREAM_TYPE_HDMV_DTS:
         return new PesReader(new DtsReader(esInfo.language));
       case TsExtractor.TS_STREAM_TYPE_H262:
-        return new PesReader(new H262Reader());
+        return new PesReader(new H262Reader(buildUserDataReader(esInfo)));
       case TsExtractor.TS_STREAM_TYPE_H264:
         return isSet(FLAG_IGNORE_H264_STREAM) ? null
             : new PesReader(new H264Reader(buildSeiReader(esInfo),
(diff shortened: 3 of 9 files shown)


peddisri:
@ojw28 Hi. Can you review this please?

ojw28:
Please can you provide a more detailed change description, so we don't have to reverse engineer what the description should be from the code :)? Please be specific about where the test streams are, how we can run the tests, and what the "missing features" are. Thanks!

peddisri:
Sarnoff compliance test streams are to be licensed. https://www.sri.com/engage/products-solutions/sarnoff-video-test-tools 
 So I can't share it here. They have exhaustive test streams to test CEA 708 spec compatibility.  The original code failed many tests. And I think FCC compliance guidelines requires compatibility with Sarnoff tests

1. Firstly, the streams are MPEG2. So we had to add support to parse CC data out of it. Hence the changes in H262Reader.java and added UserDataReader.java file.

2. Sometimes the CC text is all jumbled up for H.264 content.

[Root Cause] 
There is a timing issue in rendering CC.
Due to H264 B frames decoding and presentation order mismatch, sometimes,
older timestamps CC frames are being decoded that resulted into
jumbled CC text.

This happens due to several reasons.
a. Input buffers are not pre-buffered. This does not allow input
buffers to be pre-parsed and generate subtiles ready for rendering.
Sometimes, input frames timestamps are more than the current playback
position. So by allowing pre-buffering we hope this situation does not
occur.

b. There is no direct rate control mechanism at input buffers. It indirectly
depends on output buffers, but that depends on the content.
i.e depends on how many input buffers constitue one renderable cue.

c.. Input buffers whose timestamp is greater than the last decoded buffers
are being queued to Decoder.

[Solution]
a. We reject input frames that are older than the last decoded frames.
b. We do not start processing input frames till we have received at least
   half of the input frames that can be held in the input queue.

3. Added support for HCR and FF command.  Test using C67GSETS.TRP of Sarnoff test streams

4. To support DLY and DLC commands, we need to have a service input buffer.  Also the current design of CEA708 Decoder is not according to the spec. As per the spec CCP and service blocks are asynchronous. So we can't wait to start parsing service blocks until after generating a complete CCP. The parsing of various layers much be chained and be executed in parallel.

[Solution] We removed the DTVCC packet usage and added functions to parse
incoming data inline. We create a service input buffer which is a circular buffer
that stores all the incoming CC data.


5.Fix for handling multiple service blocks in one CCP
As per the spec, a CCP packet can contain more than one Service Blocks with same service number. Current implementation does not handle this.

 6. Styling such as color, size, location, font, superscript, subscript, edge style etc were not supported. These features were added.  

7. ,  Curently, CEA708Decoder is not aware of aspect ratio. The spec has some differences between the way CC is rendered in 4:3 and 16:9 content, specifically about number of columns it can have 

8. Black window color not rendered.
[Problem] When the window fill color is black, it is not rendered.
The window is transparent.
[Root Cause] Default implementation ignores the window color when it
is set to black. Instead it picks the color from CaptioningManager.
Unfortunately, the default value of window color is set to transparent in captioning manager.  

[Solution] Even if the window color is set to black, explicitly flag the subtitle painter to fill the window with black color. This will make sure that the default window color is drawn black.

9. Adding Parity bit and Valid bit check for Closed Captions. We failed C40UDTAb test file of Sarnoff. 

10. Support for column count and column lock added. 

11. Fixes in supporting user configured styling in captioning manager. 
[Problem]
a. By default, embedded styling is preferred over user styling. But as per the spec, user styling should be preferred over embedded styling.
b. User styling should not override italics, underline and offsets of embedded styling.There is no user style setting for these parameters.

[Root cause]
a. Application of Embedded styling is enabled by default.
b. If user styling is applied, all overridden spans are removed.

**12.Fixed Window Map issues**
Clear Window tests failed. Clearing window removes the window from display, it should just clear the text..
Test Content: c68winda.trp. Video Id:30 and CC Id: eng,id:30

2. CC not displayed. Pen style 7 test failed.
Test Content: c68winda.trp. Video Id: 32 and CC id por, id:32

[Root cause]
1. Currently, clear window clears the cues and we don't allow empty cues to be rendered.
The spec however allows rendering empty window.

2. Extended Service number should not be added to the service number
read from the standard service block header.
Instead, it should be read from the extended service block header.

[Solution]
a. We allow rendering empty captions. Additionally, we calculate the window width independent of the text width by using max column count set in Define Window Command (DFW) command and the width of the font. 

b. Extended service number is NOT added to service number, but just read.

c. One minor fix in Window Style Justification. Incorrectly set to
left instead of center.

[TEST]
   Verification tests of c68winda.trp of Sarnoff streams
   a. Video Id:30 and CC Id: eng,id:30.
   b. Video Id: 32 and CC id por, id:32

13. Added support for top to bottom scroll direction

14. Honor process_cc_data_flag - i.e skip if not set.

15. For CEA608, support Paint ON caption mode

16. To support SetPenLocation (row and column) we need to modify the Cue structure and the way we paint cues in subtitlepainter. This required a major re-design in this area.  If we just insert spaces in the text to start writing from specified column, we will have black background due to "space" character. Ideally, if the window background color is transparent, it should not render anything between 0 to column offset.  Some sarnoff tests failed due to this.

17. If we start decoding CC in between the live content, and the defineWindow command was in the past, we should not render Cues to a window until the window is defined. Rendering cues on undefined window causes unintended CC displays

**18.  Cea608 channels are not handled properly**
[Problem]
CC1 is shown even when CC2-3-4 are selected

[Solution]
There are multiple issues present:
* Incoming channel and frame information is not saved in Cea608Decoder
* Misc Code handling does not use Frame information
* Control code handling has TODOs for multiple channels

[Test]
Sarnoff test C32FR29b.trp:
* has CC1 and CC3 must be rendered with corresponding texts
* CC2 and CC4 must not show up

**19. Handle different video & caption aspect ratio case**

[Problem] If the captions was authored for a aspect ratio that is different than the video aspect ratio, we ignore that and instead consider video aspect ratio to render the captions. This use case typically happens if the video is 16:9 ( wide aspect ratio ) but the captions are authored considering 4:3 content.

[Solution]
Pass the caption aspect ratio to subtitleview and there adjust the width of the cue box if the caption aspect ratio is 4:3 but video aspect ratio is 16:9

[Tests] CS3 caption of C37SZE1 in Sarnoff test suite

**20. Other CEA608 Fixes:**

   -- ERASE command is not processed immediately
   -- Midrow code whitespaces. We have missing whitespaces when italic or other mid-row formattingis used

21. CEA608 NTSC deleted tests
Cea608 has dedicated bandwidth, the bytes should always be present in the stream - at least invalid or null bytes. If they are not present the stream is probably modified and the previously recevied data (already on the screen) is invalid. Sarnoff tests expect the screen being cleared

[Solution]
Clear the screen when there is no 608 data in the stream. It should be a rare
event

[Test]
* run c40udtab: UDTESTS. Check the "NTSC Deleted test"
* run the various frame rate tests: c30-c35

22. CEA608 - Alignment issues

[Problem]
* Switching to Paint-on style is not reflected immediately in the CueBuilders * We right align captions immediately when there is more whitespace on the left side of the text than on the right side

[Solution]
* update CueBuilders immediately when captionMode changes
* Do not right align text if there is more then 3 characters could be added
to the same row.

[Test]
* Check lots of live streams.
* Sarnoff tests should probably not show any difference, these are rare use cases
* Beginning of commercials tend to have paint on captions (as there are no video
frames shown before the beginning of the commercials that could transport the
caption data to pre-load them. They want to show the characters as soon as they
arrive - paint on mode is doing that.
  
22. Window Size fixes:
Window size depends on max column and effective font size, which depends on embedded font scale.

peddisri:
Features not implemented and known issues:

1. Other Scroll and Print directions
2. Window Style 7
3. Full justification
4.  c68winda.trp : Video Track ID 32 and CC Track index 8 and 9 does not render text because of the pen style has uniform black edge. The EdgeStyleSpan sets  the color of the pen as black which also changes the text color. This is a limitation of the way span works. We can't implement uniform style correctly without causing this regression.  The only way is to implement a ReplaceamentSpan, which means the derived class needs to implement all style features. Too complicated !

peddisri:
@ojw28  is this information sufficient?

ojw28:
Yes, although when fixing a lot of separate issues it's *much* easier to review the changes if they're split into separate commits (one for each issue, with a corresponding commit message that accurately describes it). And perhaps separate pull requests if it's likely some parts of the change are less controversial and/or simpler than others. Handling a single large commit requires the reviewer to manually untangle it to figure out which parts correspond to which feature or fix.

We will look at this when we have a chance. It's likely we'll look at parts of it far more quickly if it's split into smaller changes. Up to you whether to split or wait :).

ojw28:
Looking at this a bit more, it would be really helpful if it could be split up into smaller pieces. For example could we have a pull request that just fixes https://github.com/google/ExoPlayer/issues/3906, and one that just fixes https://github.com/google/ExoPlayer/issues/4321? I'd expect the first to be very small, and the second to be a bit larger. Then we can continue picking off medium size pieces until we get everything landed.

peddisri:
It took many patches to get to this state. There were incremental improvements during development, i.e some change that happened to fix an issue was later reworked. And there are too many patches. So it would be convenient if all of them are reviewed under one patch and pull request. During review, feel free to ask clarification about a specific change and I'll be happy to answer. Most of the change is self explanatory.

ojw28:
I don't think a 2500 line diff that corresponds to 22 features/issues, collapsed together, can ever be classed as self explanatory :). Or easy to review for that matter. Sending the change as a single pull request also means that we can't fix any of the 22 features/issues until we're happy with everything, which isn't ideal.

I think we can do one of two things:

1. You split the change up into more manageable pieces, preferably with the easy bits first.
1. We'll merge the whole change into a new branch, then do the work of teasing out the individual features/issues ourselves, merging them into `dev-v2` until we've ticked everything off.

Please let us know which option suits you best. I'm happy to have a stab at the second approach, if you prefer, in which case no action is required from your side. Although for future changes please keep in mind that it's much easier to handle pull requests that do not conflate many individual changes together.

Thanks!

ojw28:
p.s. We are looking into getting hold of the test streams for internal testing going forward.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
```diff
@@ -70,8 +71,10 @@ public SubtitleView(Context context, AttributeSet attrs) {
     painters = new ArrayList<>();
     textSizeType = Cue.TEXT_SIZE_TYPE_FRACTIONAL;
     textSize = DEFAULT_TEXT_SIZE_FRACTION;
-    applyEmbeddedStyles = true;
-    applyEmbeddedFontSizes = true;
+    // As per the CEA specs user styling should take priority over embedded styling.
```
ojw28: This view is not CEA specific, so it shouldn't be defaulting to whatever happens to be correct for CEA. If you know you're dealing with 608 captions then you can call the corresponding methods to achieve the same thing. Alternatively, the preference would need baking into the cues themselves, or equivalent.

I think it's probably best just to let the app configure things?

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
```diff
@@ -84,17 +87,18 @@ public void onCues(List<Cue> cues) {
   /**
    * Sets the cues to be displayed by the view.
    *
-   * @param cues The cues to display, or null to clear the cues.
+   * @param cues The cues to display.
```
ojw28: Revert? (and restore the nullable annotation).

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
```diff
@@ -126,10 +130,15 @@ public void setFixedTextSize(int unit, float size) {
    * default size before API level 19.
    */
   public void setUserDefaultTextSize() {
-    float fontScale = Util.SDK_INT >= 19 && !isInEditMode() ? getUserCaptionFontScaleV19() : 1f;
+    fontScale = Util.SDK_INT >= 19 && !isInEditMode() ?
```
ojw28: I'm not sure I understand what's happening here. Doesn't fontScale end up being applied twice (via setFractionalTextSize and when passed directly to painter.draw)?

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
```diff
@@ -187,7 +196,7 @@ public void setApplyEmbeddedStyles(boolean applyEmbeddedStyles) {
   }
 
   /**
-   * Sets whether font sizes embedded within the cues should be applied. Enabled by default.
+   * Sets whether font sizes embedded within the cues should be applied. Disabled by default.
```
ojw28: Revert

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitlePainter.java
```diff
@@ -150,19 +289,88 @@ public SubtitlePainter(Context context) {
...
-      CaptionStyleCompat style, float textSizePx, float bottomPaddingFraction, Canvas canvas,
-      int cueBoxLeft, int cueBoxTop, int cueBoxRight, int cueBoxBottom) {
+                   CaptionStyleCompat style, float fontScale,
+                   float textSizePx,
+                   float bottomPaddingFraction, Canvas canvas,
+                   int cueBoxLeft, int cueBoxTop, int cueBoxRight, int cueBoxBottom) {
     boolean isTextCue = cue.bitmap == null;
     int windowColor = Color.BLACK;
+    int windowBorderColor = Color.BLACK;
+    // Spec: 8.4.6 Window Size
```
ojw28: This class is not CEA specific, and shouldn't contain CEA specific comments or code.

peddisri:
I prefer the 2nd option too.  The thing is I had some 25-30 changes, and It took quite an effort to squash them into one for this pull request. (up-merge to 2.8.1 , remove our change markers in the code etc). 

I do understand and agree that incremental changes are easier to handle and review. But this specific one is like  a complete re-write./re-design and I felt it would be a better approach to review it in entirety. 

In future it'll be incremental specific changes, for sure.

library/core/src/main/java/com/google/android/exoplayer2/text/CaptionStyleCompat.java
```diff
@@ -67,7 +67,9 @@
    * Default caption style.
    */
   public static final CaptionStyleCompat DEFAULT = new CaptionStyleCompat(
-      Color.WHITE, Color.BLACK, Color.TRANSPARENT, EDGE_TYPE_NONE, Color.WHITE, null);
+          Color.WHITE, Color.BLACK,
+          /* Transparent Blue */Color.BLUE & (0x00FFFFFF),
```
ojw28: Why has this been changed to transparent blue? Note this class is not CEA specific. Ditto for changing the default to MONOSPACE below.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
```diff
@@ -137,8 +141,34 @@ public TsPayloadReader createPayloadReader(int streamType, EsInfo esInfo) {
...
+  /**
+   * If {@link #FLAG_OVERRIDE_CAPTION_DESCRIPTORS} is set, returns a {@link List<Format>} of
+   * {@link #closedCaptionFormats}. If unset, parses the PMT descriptor information and returns a
+   * {@link List<Format>} for the declared formats, or {@link #closedCaptionFormats} if the descriptor
+   * is not present.
+   *
+   * @param esInfo The {@link EsInfo} passed to {@link #createPayloadReader(int, EsInfo)}.
+   * @return A {@link List<Format>} containing list of closed caption formats.
+   */
+  private List<Format>  getCCformats(EsInfo esInfo) {
```
ojw28, line 174: This method appears to still be skipping the wide_aspect_ratio bit. In practice I think this means `Format.isWideAspectRatio` is being set to true along all code paths, which means that it's not currently doing anything useful. Did you mean to parse the bit in this method and use it when creating the corresponding `Format`?

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -150,6 +150,12 @@
    */
   public final long subsampleOffsetUs;
 
+  /**
+   * Boolean to indicate if the video is wide aspect ratio (16:9) or not (4:3)
+   * Only influences captions if the lines are not middle aligned.
+   * Defaulting to true as sources not having any flags are most probably wide screen
```
ojw28: At the `Format` level it feels like defaulting to false is more appropriate. If you want to default to true from specific CEA calling sites where `Format` instances are constructed, that would probably be a better place to do it (and put this comment).

I also wonder whether this boolean is the right thing to be adding, or whether something more general would be more flexible (e.g. `float textAspectRatio`). Such a value could also have an unset value (e.g. `-1`), which would allow the decision about what default to assume to be made much later (e.g. in the CEA decoders).

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -572,22 +599,23 @@ public static Format createSampleFormat(
...
+    isWideAspectRatio = Util.readBoolean(in);
   }
 
   public Format copyWithMaxInputSize(int maxInputSize) {
     return new Format(id, containerMimeType, sampleMimeType, codecs, bitrate, maxInputSize, width,
         height, frameRate, rotationDegrees, pixelWidthHeightRatio, projectionData, stereoMode,
         colorInfo, channelCount, sampleRate, pcmEncoding, encoderDelay, encoderPadding,
         selectionFlags, language, accessibilityChannel, subsampleOffsetUs, initializationData,
-        drmInitData, metadata);
+        drmInitData, metadata, true);
```
ojw28: All of these copy methods need to copy whatever the value of the field is in this instance, rather than hardcoding it to true as part of the copy step.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
peddisri: Well, the basic principle behind this recommendation from CEA is that the user preference has higher priority than content provider. I'm okay if this configuration is done by app.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
peddisri: The fontScale passed to painter.draw is only used to decide the effectiveFontScale that is required to calculate the window size. It is not directly used in the calculation of effectiveFontScale, but to decide what it should be

library/core/src/main/java/com/google/android/exoplayer2/Format.java
peddisri: CEA spec only indicates if the aspect ratio is for wide screen content or for letter boxed content, does not specify the actual ratio. A widescreen could mean any ratio greater than 1.55 i.e mid way of 1.33 (4:3) to 1.77 (16:9). 
floating point comparisons are tricky. 

https://stackoverflow.com/questions/1088216/whats-wrong-with-using-to-compare-floats-in-java

library/core/src/main/java/com/google/android/exoplayer2/Format.java
peddisri: agreed.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
peddisri: This is a mistake in the patch. The original change is as per below. Will fix this.

byte misc = (byte)scratchDescriptorData.readUnsignedByte();
          boolean isWideAspectRatio = ((misc & 0x60) == 0x60);
          closedCaptionFormats.add(Format.createTextSampleFormat(null, mimeType, null,
              Format.NO_VALUE, 0, language, accessibilityChannel, null,
              isWideAspectRatio));
          scratchDescriptorData.skipBytes(1);

library/core/src/main/java/com/google/android/exoplayer2/text/CaptionStyleCompat.java
peddisri: The Default Window Color from caption manager in some devices that i tested was this color. If we need to apply user styling, we need to know if the user has modified this value from the default or not. So , we need to change the default value.  LIne 325 of SubtitlePainer.java
I'm not sure if this is a bug in some devices or in AOSP code. i.e instead of opaque black. its  transparent. 

As for Monospace, the calculation of Window Size (specifically width) is a function of max window columns and font size. i.e window width = maxWindowColumns * fontSize;
Now if the text is not monochrome, the calculated window size and the actual text size results into a visual where window is disproportionately larger or smaller than the text. By using MONOSPACE the window size and text size are calculated correctly and some of the sarnoff tests visual output is comparable to the reference output.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitlePainter.java
peddisri: The concept of Window May be unique to CEA, but it can be generically implemented in SubtitlePainter, and other caption types can set the default values. 

I can remove the CEA specific  comment though.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
peddisri: Ok.

peddisri (commented):
Responded to few comments.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitlePainter.java
ojw28: * I don't think the concept of a window is unique to CEA
* If there are any assumptions around the font being monospace in the code below, then I don't think it's generic enough. I think we'd need something more like a `Cue.sizeType` variable that can indicate `Cue.size` is something other than a fraction.

library/core/src/main/java/com/google/android/exoplayer2/text/CaptionStyleCompat.java
ojw28: As above, this class is not CEA specific. You can't change the default to MONOSPACE just because it fixes CEA-608, because it will affect how other caption types are rendered. If we need to be able to specify that CEA-608 uses monospace fonts, then that information needs to be hardcoded somewhere in CEA-608 specific code. For example by setting styling information on each `Cue`. The text package then needs to be able to handle that information in the correct way.

ojw28:
Responded to a couple. I think we need the comments addressed, and in particular all CEA-608 specific changes removed from the text package, before we can make progress here. It would still be preferable if the change could be split into smaller pieces.

library/core/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -98,11 +101,33 @@
...
+  public static final int BORDER_AND_EDGE_TYPE_UNIFORM = 3;
+  public static final int BORDER_AND_EDGE_TYPE_SHADOW_LEFT = 4;
+  public static final int BORDER_AND_EDGE_TYPE_SHADOW_RIGHT = 5;
+
+
+  /**
+   * The list of pair of cue text and column offsets. Note the {@link CharSequence} may be decorated
+   * with styling spans.
+   */
+  public final List<Pair<Integer, CharSequence>> textRows;
```
ojw28: Does this effectively duplicate the text in every `Cue`? That doesn't seem ideal from a modelling point of view. Given CEA uses monospace fonts, why can't columns be handled using spaces at the start of each line in `text`?

peddisri: It does not duplicate the text.  Either text field is set or textRows is set, not both. 
Earlier text field used to contain a 4 line text. But now textRows will have 4 elements, each having one line of text and associated column offset. 

In fact, initially we did implement column offset by inserting a simple space. The problem with this approach is that space draws the text background, which does not match the expected reference output when the window color is transparent.

peddisri:
Ok I'll address the feedback so far. For the splitting part, let me explore if I can split this into three parts :
1. CEA 608 specific.
2. CEA 708 Decoding/parsing specific
3. CEA 708 Styling Specific (including painter)

peddisri:
@ojw28  Let's do this pull request in phases. It's hard to split the code into three and I'd like to take baby steps here. The latest pull request incluedes
1. CEA708 decoder changes 
2. Minor chane in CEA608 to support this change.

One you review this, I'll put up another pull request for 608 decoder only, and then lastly styling related.

peddisri:
@ojw28  hope you can find some time to review it this week

library/core/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -77,6 +77,11 @@ private C() {}
    */
   public static final long NANOS_PER_SECOND = 1000000000L;
 
+  /**
+   * The number of 10th of a second per microseconds.
+   */
+  public static final long TENTH_OF_SECOND_PER_MICROSECOND = 100000L;
```
ojw28: Constant name is wrong way around (it's MICROS_PER_TENTH_OF_SECOND, right?). That aside, this is a pretty weird constant, and it's not likely to be used in many places. Can you just inline the value in the one place it's currently used, or use `C.MICROS_PER_SECOND / 10` there?

library/core/src/main/java/com/google/android/exoplayer2/Format.java
ojw28: This class is not specific to CEA, so it's more important to come up with a sufficiently flexible model than to do something that maps directly onto the specific format you're interested in. If the CEA spec only provides "wide" or "not wide" then define two constants somewhere (one for each case), and use those for your comparisons. It would also make sense to default it to `Format.NO_VALUE`.

Alternatively, if you don't think this is ever going to be useful for non-CEA, use the existing `initializationData` field, which is decoder/format specific, and appears to be unused for CEA at the moment. You could define it to contain a single byte[] containing a single byte, which is 1 or 0 based on this flag. That's a much less invasive change, since it lets you revert all changes to this file, and is probably the way to go as I think other caption formats put windowing information in `initializationData` or in the samples already.

As an aside, `accessibilityChannel` should probably be in `initializationData` as well. We might make that change at some point.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -508,7 +533,8 @@ public static Format createSampleFormat(
       long subsampleOffsetUs,
       @Nullable List<byte[]> initializationData,
       @Nullable DrmInitData drmInitData,
-      @Nullable Metadata metadata) {
+      @Nullable Metadata metadata,
+      boolean isWideAspectRatio) {
```
ojw28: `FormatTest` does not compile after this change

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
```diff
@@ -77,8 +77,12 @@ public DefaultTsPayloadReaderFactory(@Flags int flags) {
   public DefaultTsPayloadReaderFactory(@Flags int flags, List<Format> closedCaptionFormats) {
     this.flags = flags;
     if (!isSet(FLAG_OVERRIDE_CAPTION_DESCRIPTORS) && closedCaptionFormats.isEmpty()) {
-      closedCaptionFormats = Collections.singletonList(Format.createTextSampleFormat(null,
-          MimeTypes.APPLICATION_CEA608, 0, null));
+      closedCaptionFormats = new ArrayList();
+      closedCaptionFormats.add(Format.createTextSampleFormat(null,
+              MimeTypes.APPLICATION_CEA608,  0, null));
+      closedCaptionFormats.add(Format.createTextSampleFormat(null,
```
ojw28: This change is causing `TsExtractorTest` to fail. You need to update the sample dump files with the new expected extractor output.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -730,7 +758,8 @@ public boolean equals(@Nullable Object obj) {
         && Util.areEqual(metadata, other.metadata)
         && Util.areEqual(colorInfo, other.colorInfo)
         && Arrays.equals(projectionData, other.projectionData)
-        && initializationDataEquals(other);
+        && initializationDataEquals(other)
+        && isWideAspectRatio == other.isWideAspectRatio    ;
```
ojw28: nit - We have an cheap-to-expensive ordering of comparisons in this method. Please move this one up to the primitive comparisons (also rm whitespace before the semi-colon).

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H262Reader.java
```diff
@@ -130,7 +148,25 @@ public void consume(ParsableByteArray data) {
...
+      if (userDataReader != null && userData != null) {
+        int lengthToStartCode = startCodeOffset - offset;
+        int bytesAlreadyPassed = 0;
+        if (lengthToStartCode > 0) {
+          userData.appendToNalUnit(dataArray, offset, startCodeOffset);
+        } else {
+          bytesAlreadyPassed = -lengthToStartCode;
+        }
+
+        if (userData.endNalUnit(bytesAlreadyPassed)) {
```
ojw28: In the H264 case when dealing with SEI, we call `NalUnitUtil.unescapeStream` here. Are you certain it's not necessary to do unescaping here also? The equivalent would be something like:
```
int unescapedLength = NalUnitUtil.unescapeStream(userData.nalData, userData.nalLength);
userDataParsable.reset(sei.nalData, unescapedLength);
userDataParsable.setPosition(4); // NAL prefix and nal_unit() header.
userDataReader.consume(sampleTimeUs, userDataParsable);
```

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H262Reader.java
```diff
@@ -130,7 +148,25 @@ public void consume(ParsableByteArray data) {
...
+          //send media format stuff here.
+          userDataParsable.reset(userData.nalData, userData.nalLength);
+          userDataReader.consume(sampleTimeUs, userDataParsable);
+        }
 
+        if (startCodeValue == START_USER_DATA && data.data[startCodeOffset + 2] == 0x1) {
```
ojw28: I think you can do `userData.startNalUnit(startCodeValue)` unconditionally. `startNalUnit` already checks the type internally, and I think the data.data check is guaranteed to be true due to `NalUnitUtil.findNalUnit` having found a NAL unit further up in this method.

library/core/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java
```diff
@@ -107,7 +107,7 @@ public SubtitleDecoder createDecoder(Format format) {
             case MimeTypes.APPLICATION_MP4CEA608:
               return new Cea608Decoder(format.sampleMimeType, format.accessibilityChannel);
             case MimeTypes.APPLICATION_CEA708:
-              return new Cea708Decoder(format.accessibilityChannel);
+              return new Cea708Decoder(format.accessibilityChannel, format.isWideAspectRatio);
```
ojw28: Feel free to just put `isWideAspectRatio` into `initializationData` for now, as suggested in `Format`. However the eventual goal would be to have these decoders just take `initializationData` like the others. We'll probably move `accessibilityChannel` and `packetLength` into `initializationData` in a subsequent change.

(`packetLength` is derived from `sampleMimeType`, which is why `sampleMimeType` is currently being passed to the 608 decoder)

ojw28 (commented):
I've reviewed the extractor side changes. They look good, thanks, although I've posted some comments. The main one being to revert the `Format` change. The rest are pretty minor.

It would be preferable if you could split this change into the extractor side changes, and the decoder side changes. They're almost completely decoupled, and so there's no reason for them to exist in the same change. We can then get the extractor side changes submitted, and then continue with the decoder side changes as the next step.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H262Reader.java
ojw28: [I have no idea what the answer to this is; I was just comparing the two whilst reviewing]

library/core/src/main/java/com/google/android/exoplayer2/Format.java
peddisri: Ideally, Format in its current form needs to be re-designed anyways because there are many fields that are applicable only to one type of format like Video or audio. In fact even language is applicable only to either audio or text.  

We should re-define Format that just has mimeType and perhaps Id,  and then have derived classes like VideoFormat, AudioFormat, TextFormat etc and have specific fields in each of them. Alternatively, add a list of key/value pairs to make it easily extendable, i.e hash map

I don't know if wide and non-wide captions would be applicable for other subtitle formats or not. I assumed usage of initializationData is like a binary blob that is sent to decoder which is not interpreted at app layer. But by using it in the way you advised, we are now giving meaning to the data elements of the blob. 

So here's what I'm gonna try to do. Introduce a generic hashtable to pass custom parameters. In future you may want to use it to remove accessibilityChannel and any other fields and make the Format Constructor light weight and common for all media types. This will now be similar to the design paradigm of Android's MediaFormat

ojw28: > I assumed usage of initializationData is like a binary blob that is sent to decoder which is not interpreted at app layer.

It is a binary blob that's sent to the decoder. Where the decoder lives (app layer or platform) is not well defined. Even for audio and video, you can have decoders in both the platform layer (MediaCodec) and the app layer (e.g. the VP9 and FFMPEG extensions). For text the decoders all happen to be in the app layer.

> But by using it in the way you advised, we are now giving meaning to the data elements of the blob.

The blobs have a well defined structure in all cases (else whatever receives it wont know what to do with it). The only distinction is whether the app layer needs to know what the structure is, which it doesn't if it can transparently pass through the blob from the container to the platform in all cases, and which it does if it ever needs to construct or parse one itself.

Defining what we want the blob to look like for CEA formats seems reasonable to me. If it makes it better you can add a `CeaInitializationData` class in the `text.cea` package with methods for converting to/from a `byte[]`, and use that so that the structure of the blob is encapsulated within that class.

> So here's what I'm gonna try to do. Introduce a generic hashtable to pass custom parameters. In future you may want to use it to remove accessibilityChannel and any other fields and make the Format Constructor light weight.

Please don't do this. I agree that `Format` probably needs to be redesigned at some point, but for now `initializationData` is the mechanism for carrying this type of data. Adding a new alternative way of doing things makes things even more confusing.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H262Reader.java
peddisri: We need to check two bytes : 0x01B2.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/H262Reader.java
peddisri: done

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
peddisri: done

peddisri:
@ojw28 addressed your comments. split patch into extractor changes first.

library/core/src/main/java/com/google/android/exoplayer2/C.java
peddisri: done

ojw28:
Thanks! We'll get this merged next week (most likely finding another way to pass the information through `Format`, but no need to make any changes your side).

peddisri:
Thanks. Once you merge it to dev, I'll work on the remaining part.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
```diff
@@ -162,17 +198,24 @@ private SeiReader buildSeiReader(EsInfo esInfo) {
             mimeType = MimeTypes.APPLICATION_CEA608;
             accessibilityChannel = 1;
           }
-          closedCaptionFormats.add(Format.createTextSampleFormat(null, mimeType, null,
-              Format.NO_VALUE, 0, language, accessibilityChannel, null));
           // Skip easy_reader(1), wide_aspect_ratio(1), reserved(14).
-          scratchDescriptorData.skipBytes(2);
+          byte misc = (byte)scratchDescriptorData.readUnsignedByte();
+          boolean isWideAspectRatio = ((misc & 0x60) == 0x60);
```
ojw28, line 203: `0x60` should be `0x40`, right? (no need to update it here, we can do that as part of our merge).

peddisri: Either of them is okay, because the remaining 6 bits are reserved as 1s anyways as per the spec. But yes, more correct would be 0x40.

PR #4342, opened 2018-06-04
Support emsg box for HLS + fmp4 playback

TakuSemba:
It seems like emsg box is disabled when playing HLS + fmp4.

When I enabled and test it, it worked without any problems.

Do you have any reasons of disabling emsg box for HLS playback?

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/DefaultHlsExtractorFactory.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/DefaultHlsExtractorFactory.java
@@ -74,7 +74,8 @@ public final class DefaultHlsExtractorFactory implements HlsExtractorFactory {
     } else if (lastPathSegment.endsWith(MP4_FILE_EXTENSION)
         || lastPathSegment.startsWith(M4_FILE_EXTENSION_PREFIX, lastPathSegment.length() - 4)
         || lastPathSegment.startsWith(MP4_FILE_EXTENSION_PREFIX, lastPathSegment.length() - 5)) {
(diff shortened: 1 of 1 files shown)


AquilesCanta:
The way I see it, on one hand:

- EMSG needs manual handling. Even if enabled by default.
- Users only need to change a flag. Arguably, this could be made easier by the default extractor factory.

but also:

- DashMediaSource enables it by default.

> Do you have any reasons of disabling emsg box for HLS playback?

In general, I tend to prefer phrasing the question as "Is there a reason for enabling it by default?".

I'll try to get some sample content and have a look.

TakuSemba:
> In general, I tend to prefer phrasing the question as "Is there a reason for enabling it by default?".

I see, I think so too. 
and it would be better if we could control the flags without changing the source code. 

> I'll try to get some sample content and have a look.

thank you!

VidaSanauk (approved):

AquilesCanta:
Hi @TakuSemba, I'll look into adding this internally. There are a few things that this pull request is not handling and it will be easier to implement on my side:

+ Support EMSG in chunkless preparation. In this PR, the EMSG track will not be visible when using chunkless preparation.
+ Avoid duplicating an EMSG track per demuxed track. In this PR, each demuxed track will produce an EMSG track, which be pretty confusing for app developers. 

We honestly appreciate your pull request. I apologize for not looking into this in detail before.

icbaker:
emsg support for HLS has been added to the dev-v2 branch:
https://github.com/google/ExoPlayer/commit/60566721d4b173eb767a083d5e91e71fd3b46c0c

By default ExoPlayer unwraps ID3-in-EMSG messages and drops all others. This behaviour can be configured with `HlsMediaSource#setMetadataType(int)`.

PR #4365, opened 2018-06-10
fix datasource not close when we don't need

michalliu:
We should close the datasource when cancel is requested


ojw28:
These changes don't look valid to me. If you still believe there are problems, please describe the symptoms you're seeing in more detail. Thanks.

Issue #4419, opened 2018-06-25
ANR on Fire TV when changing surface holder


murdly:
### Issue description
ANR occurs when setting video surface holder to null after it got destroyed.

### Reproduction steps
Trigger replacing surface.

### Link to test content
Any video.

### Version of ExoPlayer being used
2.7.0, 2.7.2 and 2.8.1

### Device(s) and version(s) of Android being used
Fire TV 6.0.1.4-012, OS 6.2.1.3

Codec: OMX.amlogic.avc.decoder.awesome
Device: needle
Model: AFTN

### A full bug report captured from the device
Will be sent by email.

ojw28:
A question before we merge the change: Does the ANR occur when switching between any surfaces, or only when `DummySurface` is used? Can you try returning false from `shouldUseDummySurface`, with the change to `codecNeedsSetOutputSurfaceWorkaround` reverted?

murdly:
When `shouldUseDummySurface` is returning false the ANR doesn't occur.

ojw28:
My question probably wasn't precise enough. What I'm really interested in is whether `setOutputSurface` is broken on this device, or whether it's broken *only* when using `DummySurface`. The test I asked you to do was unfortunately not sufficient to disambiguate between the two, since your app might not ever be switching directly from one non-dummy surface to another. Can you test whether switching directly from one valid non-dummy surface to another works correctly?

We'll probably just merge the PR anyway to speed things along, but in the long run I'd like to start disambiguating between the two cases and making these workarounds more targeted. Doing this will help reduce the amount by which we're degrading the user experience.

murdly:
I can confirm that switching between non-dummy surfaces works as expected. 

I have also tested on several other Fire TV devices plus Android TV emulator and it works correctly when using `DummySurface`.

Is there anything else I could do?

ojw28:
In that case can we use a more targeted workaround? It should work to change `shouldUseDummySurface` to:
```
  private boolean shouldUseDummySurface(MediaCodecInfo codecInfo) {
    return Util.SDK_INT >= 23
        && !tunneling
        && !codecNeedsDummySurfaceWorkaround(codecInfo.name)
        && !codecNeedsSetOutputSurfaceWorkaround(codecInfo.name)
        && (!codecInfo.secure || DummySurface.isSecureSupported(context));
  }
```

[...]

PR #4421, opened 2018-06-25
Add Fire TV to codecNeedsSetOutputSurfaceWorkaround

murdly:
Solves issue #4419


ojw28:
I tried to do it the way you probably just tried, but it's picking up a bunch of other commits. It needs to be just the single commit into `dev-v2`.

murdly:
I see. Sure

PR #4441, opened 2018-06-27
opt out of using DummySurface on specific Fire TV device

murdly:
Solves #4419


peddisri:
There are more than one versions of FTV that needs this. 

https://github.com/amzn/exoplayer-amazon-port/commit/3ed246771f3ab5a67be105650b4630162d66f0c1

PR #4442, opened 2018-06-27
Add withSkippedAd method to AdPlaybackState. 

keiththompson:
Currently there is no way to "skip" an individual ad within an ad group. We can only skip entire ad groups with `withSkippedAdGroup`.

My use case here is that I have multiple pre rolls of unknown length in which some may or may not be skippable. Making multiple ad groups with a single ad each is cumbersome here. 

As a workaround I'm using `withPlayedAd` which pretty much has the effect I'm after but seems slightly incorrect to me when in truth I'm skipping the ad. 

Feel free to close if you think `withPlayedAd` is still a preferable way to achieve this (and possibly explain why).


keiththompson:
I've signed the CLA.

andrewlewis:
Looks good. Thanks!

PR #4534, opened 2018-07-19
Allow specifying the default live presentation delay

NicolaVerbeeck:
When the suggestedPresentationDelay is missing from the manifest, a fallback value is used (DEFAULT_LIVE_PRESENTATION_DELAY_FIXED_MS). This PR adds a method to the factory to override this value


ojw28:
Thanks for the change! It looked a bit complicated having default and non-default delays, so we simplified this a little in https://github.com/google/ExoPlayer/commit/41345dcb8353ffe9cc78ff11ec030696d95d00e1, to have a single delay and a boolean specifying whether it should override the manifest provided value. I think this should be equivalent in terms of expressiveness, but if not and we've done something bad, then (a) sorry, and (b) let us know so we can fix it :).

NicolaVerbeeck:
Looks good!

PR #4556, opened 2018-07-23
r2.8.3

ojw28:


oleksandr-yefremov:
Is there any estimate on when this release might land? We are specifically interested in "Widen setOutputSurface workaround", because we suspect we have negative feedback growing due to it.

ojw28:
We have two more fixes to cherry-pick into the release, both of which are in flight so it should happen soon (i.e. this week).

PR #4557, opened 2018-07-23
Fix pending Metadata timestamp relative to subsample offset

phhusson:
Since MetadataRenderer queues events until event times are current
position, this leads to metadata event like SCTE-35 not being delivered.

To compare to player timestamp, we need to include subsample offset.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/metadata/MetadataRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/metadata/MetadataRenderer.java
@@ -128,7 +128,7 @@ public final class MetadataRenderer extends BaseRenderer implements Callback {
           try {
             int index = (pendingMetadataIndex + pendingMetadataCount) % MAX_PENDING_METADATA_COUNT;
             pendingMetadata[index] = decoder.decode(buffer);
-            pendingMetadataTimestamps[index] = buffer.timeUs;
+            pendingMetadataTimestamps[index] = buffer.timeUs + buffer.subsampleOffsetUs;
             pendingMetadataCount++;
           } catch (MetadataDecoderException e) {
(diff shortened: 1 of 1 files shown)


ojw28:
I don't think this fix is correct. Could you please provide some reproduction steps and sample content?

phhusson:
I'm debugging and reproducing this by adding:
+    if(pendingMetadataCount > 0 && pendingMetadataTimestamps[pendingMetadataIndex] > positionUs) {
+      android.util.Log.d("SAH", "Delaying metadata sample " + positionUs + " : " + pendingMetadataTimestamps[pendingMetadataIndex]);
+    }
in MetadataRender's render function
And by adding a MetadataOutput to the player, and listening for events.

On my internal test streams, without this fix, pendingMetadataTimestamps[pendingMetadataIndex] is in the order of 600 000s, and is never received.
With it, few seconds. It is de-queued few seconds after reception, and is received by the MetadataOutput.

I'll search for a public scte-35 stream, or see if I can publish mine.

ojw28:
You can also send your test stream to `dev.exoplayer@gmail.com`. It's somewhat hard to tell whether the issue is with the media itself, or with ExoPlayer (and if so, where), without being able to take a look. I'm pretty sure we have some SCTE-35 test streams internally that work correctly with the current implementation.

phhusson:
There is a sample:
https://drive.google.com/open?id=1My5PYYLOHQkUPTKDjZbEqN3wdtWsqA4b

ojw28:
Thanks for the sample! I'm pretty sure the issue is elsewhere, so I'll go ahead and close this pull request. I've filed an issue [here](https://github.com/google/ExoPlayer/issues/4573).

PR #4561, opened 2018-07-24
Support vector drawables to be used as default artwork

mitrejcevski:
Currently references to vector drawables for the default artwork of the PlayerView are not supported (the call to `BitmapFactory.decodeResource(context.getResources(), defaultArtworkId);` returns `null` if vector drawable is referenced), and this PR will allow vector drawables to be used as well.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
@@ -25,6 +25,7 @@ import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
 import android.graphics.Matrix;
 import android.graphics.RectF;
+import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
 import android.support.annotation.NonNull;
@@ -251,7 +252,7 @@ public class PlayerView extends FrameLayout {
   private Player player;
   private boolean useController;
   private boolean useArtwork;
-  private Bitmap defaultArtwork;
+  private Drawable defaultArtwork;
   private boolean showBuffering;
   private boolean keepContentOnPlayerReset;
   private @Nullable ErrorMessageProvider<? super ExoPlaybackException> errorMessageProvider;
@@ -373,7 +374,7 @@ public class PlayerView extends FrameLayout {
     artworkView = findViewById(R.id.exo_artwork);
     this.useArtwork = useArtwork && artworkView != null;
     if (defaultArtworkId != 0) {
-      defaultArtwork = loadFromResource(defaultArtworkId);
+      defaultArtwork = ContextCompat.getDrawable(getContext(), defaultArtworkId);
     }
 
     // Subtitle view.
@@ -419,21 +420,6 @@ public class PlayerView extends FrameLayout {
     this.controllerHideDuringAds = controllerHideDuringAds;
     this.useController = useController && controller != null;
     hideController();
-  }
-
-  private Bitmap loadFromResource(int defaultArtworkId) {
-    Drawable drawable = ContextCompat.getDrawable(getContext(), defaultArtworkId);
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
-      drawable = (DrawableCompat.wrap(drawable)).mutate();
-    }
-
-    Bitmap bitmap = Bitmap.createBitmap(drawable.getIntrinsicWidth(),
-            drawable.getIntrinsicHeight(), Bitmap.Config.ARGB_8888);
-    Canvas canvas = new Canvas(bitmap);
-    drawable.setBounds(0, 0, canvas.getWidth(), canvas.getHeight());
-    drawable.draw(canvas);
-
-    return bitmap;
   }
 
   /**
@@ -573,7 +559,7 @@ public class PlayerView extends FrameLayout {
   }
 
   /** Returns the default artwork to display. */
-  public Bitmap getDefaultArtwork() {
+  public Drawable getDefaultArtwork() {
     return defaultArtwork;
   }
 
@@ -584,6 +570,16 @@ public class PlayerView extends FrameLayout {
    * @param defaultArtwork the default artwork to display.
    */
   public void setDefaultArtwork(Bitmap defaultArtwork) {
+    setDefaultArtwork(new BitmapDrawable(getResources(), defaultArtwork));
+  }
+
+  /**
+   * Sets the default artwork to display if {@code useArtwork} is {@code true} and no artwork is
+   * present in the media.
+   *
+   * @param defaultArtwork the default artwork to display
+   */
+  public void setDefaultArtwork(Drawable defaultArtwork) {
     if (this.defaultArtwork != defaultArtwork) {
       this.defaultArtwork = defaultArtwork;
       updateForCurrentTrackSelections(/* isNewPlayer= */ false);
@@ -1061,7 +1057,7 @@ public class PlayerView extends FrameLayout {
           }
         }
       }
-      if (setArtworkFromBitmap(defaultArtwork)) {
+      if (setDrawableArtwork(defaultArtwork)) {
         return;
       }
     }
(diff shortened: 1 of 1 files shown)


ojw28:
Thanks for the contribution. It doesn't seem ideal to add support by rendering the vector drawable into a bitmap, however. Doesn't that negate several of the benefits of using vector drawables in the first place.

Why not convert `PlayerView` to use `Drawable` by default (i.e. change the type of `defaultArtwork` to be a `Drawable`, and use `artworkView.setImageDrawable` rather than `artworkView.setImageBitmap`)?

You can support `setDefaultArtwork(Bitmap)` and bitmaps provided in metadata using `BitmapDrawable`, and add a new `setDefaultArtwork(Drawable)` method. We'll have to change the return type of `getDefaultArtwork` to be `Drawable` rather than `Bitmap`, but I think that's an acceptable price to pay.

What do you think?

mitrejcevski:
I absolutely agree with your points and I am very thankful to you for being that constructive. I am more than happy to do the changes and update this PR.

I was a little bit concerned about the backward compatibility of the API initially, and I thought I would do with as little changes as possible, but sure it ended up being less than ideal.

Thank you for the feedback, I will update the PR.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -1055,21 +1071,30 @@ private boolean setArtworkFromMetadata(Metadata metadata) {
...
     }
     return false;
   }
 
-  private boolean setArtworkFromBitmap(Bitmap bitmap) {
-    if (bitmap != null) {
-      int bitmapWidth = bitmap.getWidth();
-      int bitmapHeight = bitmap.getHeight();
+  private boolean setDrawableArtwork(Drawable drawable) {
+    if(drawable instanceof BitmapDrawable) {
```
ojw28: Is this code path necessary? I was thinking you could just pass the `BitmapDrawable` drawable to `setImageDrawable` and the right thing would happen, rather than having to special case this to get the bitmap back out again.

ojw28: Perhaps this is related to calling `contentFrame.setAspectRatio`. Although looking at that, isn't it problematic to be calling it for the bitmap case only? Setting a bitmap will cause the aspect ratio to become to a particular value, which will then still be applied should another type of `Drawable` be set later on. Do `getIntrinsic[Width/Height]` on a `Drawable` do anything useful for this (I've no idea)?

I think one way or another we need to get to a single code path for this to fall out nicely; I'm just not quite sure how to do that.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -564,6 +570,16 @@ public Bitmap getDefaultArtwork() {
...
+    setDefaultArtwork(new BitmapDrawable(getResources(), defaultArtwork));
+  }
+
+  /**
+   * Sets the default artwork to display if {@code useArtwork} is {@code true} and no artwork is
+   * present in the media.
+   *
+   * @param defaultArtwork the default artwork to display
+   */
+  public void setDefaultArtwork(Drawable defaultArtwork) {
```
ojw28, line 581: Minor comment - Let's deprecate the method above. We should leave it there to avoid an unnecessary breaking change, but it becomes a little confusing because getDefaultArtwork will no longer pass back what the app has passed in.

ojw28 (commented):
This looks much better; thanks! Posted a few comments, the main one being trying to get down to a single code path.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
mitrejcevski: Right, I mainly left it because of the aspect ratio, but as you said, calling `setImageDrawable()` should do it, and since the `artWork` container is an `ImageView`, the user has the flexibility to apply his own scale type and/or `adjustViewBounds`. So I think we can remove that logic from here. What do you think?

ojw28: I think that sounds right. If the content frame aspect ratio is unnecessary then you probably want to explicitly set it to 0, in case it was set previously (e.g. due to a video being played).

mitrejcevski: Ok, just updated, let's check if that makes more sense. Thanks a lot!

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -1055,26 +1070,17 @@ private boolean setArtworkFromMetadata(Metadata metadata) {
...
-        if (contentFrame != null) {
-          contentFrame.setAspectRatio((float) bitmapWidth / bitmapHeight);
-        }
-        artworkView.setImageBitmap(bitmap);
-        artworkView.setVisibility(VISIBLE);
-        return true;
-      }
+  private boolean setDrawableArtwork(Drawable drawable) {
+    if(drawable != null) {
+      artworkView.setImageDrawable(drawable);
```
ojw28: I suspect you still want to explicitly clear any previously set aspect ratio here, like:

```
if (contentFrame != null) {	
  contentFrame.setAspectRatio(0);	
}
```

mitrejcevski: oh, correct. My apology I missed it.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -1055,26 +1070,20 @@ private boolean setArtworkFromMetadata(Metadata metadata) {
...
-      if (bitmapWidth > 0 && bitmapHeight > 0) {
-        if (contentFrame != null) {
-          contentFrame.setAspectRatio((float) bitmapWidth / bitmapHeight);
-        }
-        artworkView.setImageBitmap(bitmap);
-        artworkView.setVisibility(VISIBLE);
-        return true;
+  private boolean setDrawableArtwork(Drawable drawable) {
+    if(drawable != null) {
+      artworkView.setImageDrawable(drawable);
```
ojw28, line 1081: We need to add `artworkView.setVisibility(VISIBLE)` here for the artwork to actually be displayed.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -1055,26 +1070,20 @@ private boolean setArtworkFromMetadata(Metadata metadata) {
...
+      if(contentFrame != null) {
+        contentFrame.setAspectRatio(0);
```
ojw28, line 1083: Removing the logic for setting the aspect ratio seems to break displaying bitmap artwork (artwork is now stretched). It appears it's possible to fix this either by:

1. Bringing back setting the aspect ratio correctly, but using `getIntrinsicWidth` and `getIntrinsicHeight` on the `Drawable`.
1. In `exo_simple_player_view.xml`, changing `exo_artwork`'s `scaleType` to `fitCenter`.

I think my preference would be the first one, since it's most similar to what we had before, and use of the `contentFrame` for controlling aspect ratio is more consistent with what happens for video.

I'm unsure whether it breaks your use case for non-bitmap drawables though. Could you take a look at this?

No need to update the pull request, but let me know whether use of `getIntrinsicWidth` and `getIntrinsicHeight` works for you. Thanks!

ojw28: We'll go with using the intrinsic dimensions for now, but feel free to send another pull request changing that if it doesn't work for some use cases.

mitrejcevski: Hi, sorry for the late reply. It won't break my case, and I think it would be fine, but my idea was to indeed leave this to the developer to set the `scaleType` in the XML. Anyway, the 1st solution works for my case. Thanks!

Issue #4542, opened 2018-07-20
[PR offer] Remove an index range in ConcatenatingMediaSource


BrainCrumbz:
### Issue description
We found ourselves in need of removing a contiguous set of tracks (i.e. playlist items) from a `ConcatenatingMediaSource`. Basically that would result in extending its public API with two methods like:

```java
public final synchronized void removeMediaSourceRange(int fromIndex, int toIndex) { ... }

public final synchronized void removeMediaSourceRange(
  int fromIndex, int toIndex, @Nullable Runnable actionOnCompletion) { ... }
```

Involved indices go from `fromIndex` (included) to `toIndex` (excluded), following same convention as e.g. `List<E>.subList()`.

We put together a working implementation and started using that: things seems to work fine. If you're interested already in details here, we:

* created a new `MSG_REMOVE_RANGE` message, 
* created a `MessageData<Integer>` passing 2nd index in `customData`, 
* implemented checks on both indices and sequence length, to avoid empty ranges and overflows,
* handled message by invoking `shuffleOrder.cloneAndRemove()` and `removeMediaSourceInternal()` within loops

Would you consider a PR to add this feature (of course with all required tests etc)? 

### Version of ExoPlayer being used
2.8.2

Please do not hesitate to ask for further details. Regards

ojw28:
Sounds like something that would be useful. @tonihei - Any thoughts?

tonihei:
Yes, definitely! PRs are always welcome. The general approach you describe sounds correct and such a method could be useful for other people as well.

BrainCrumbz:
Great. Could you share some info on the testing side? We had no chance to look at that yet, it was an experimental implementation after all. 

We noticed [ConcatenatingMediaSourceTest.java](https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/test/java/com/google/android/exoplayer2/source/ConcatenatingMediaSourceTest.java) is quite a long one. Have you got any suggestion on what/where to add to test such a feature?

Thanks

tonihei:
Yes, tests would need to be added to this file. The file is already quite long as we added a lot of tests for standard and edge case scenarios. If you find that too confusing we can easily add the test ourselves.

BrainCrumbz:
Have you got any preference on behaviour w.r.t. indices and overflows? We'd prefer a more *tolerant* approach, like the following:

1. when range is empty (i.e. beginning === end), no track is removed, no error is thrown, callback is invoked
[...]

PR #4564, opened 2018-07-24
feat(playlist): remove MediaSource range

BrainCrumbz:
See #4542 for related discussion

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
@@ -50,11 +50,11 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
   private static final int MSG_ADD = 0;
   private static final int MSG_ADD_MULTIPLE = 1;
   private static final int MSG_REMOVE = 2;
-  private static final int MSG_REMOVE_RANGE = 200;
-  private static final int MSG_MOVE = 3;
-  private static final int MSG_CLEAR = 4;
-  private static final int MSG_NOTIFY_LISTENER = 5;
-  private static final int MSG_ON_COMPLETION = 6;
+  private static final int MSG_REMOVE_RANGE = 3;
+  private static final int MSG_MOVE = 4;
+  private static final int MSG_CLEAR = 5;
+  private static final int MSG_NOTIFY_LISTENER = 6;
+  private static final int MSG_ON_COMPLETION = 7;
 
   // Accessed on the app thread.
   private final List<MediaSourceHolder> mediaSourcesPublic;
@@ -324,7 +324,7 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
 
   /**
    * Removes a range of {@link MediaSource}s from the playlist, by specifying an initial index
-   * (included) and a final index (excluded), and executes a custom action on completion..
+   * (included) and a final index (excluded), and executes a custom action on completion.
    *
    * <p>Note: when specified range is empty, no actual media source is removed and no exception
    * is thrown.
@@ -349,10 +349,10 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
     }
     if (player != null) {
       player
-              .createMessage(this)
-              .setType(MSG_REMOVE_RANGE)
-              .setPayload(new MessageData<>(fromIndex, toIndex, actionOnCompletion))
-              .send();
+          .createMessage(this)
+          .setType(MSG_REMOVE_RANGE)
+          .setPayload(new MessageData<>(fromIndex, toIndex, actionOnCompletion))
+          .send();
     } else if (actionOnCompletion != null) {
       actionOnCompletion.run();
     }
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
@@ -312,16 +312,11 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
    * is thrown.
    *
    * @param fromIndex The initial range index, pointing to the first media source that will be
-   *     removed. This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   *     removed. This index must be in the range of 0 &lt;= index &lt;= {@link #getSize()}.
    * @param toIndex The final range index, pointing to the first media source that will be left
-   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
-   *     This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
-   *
-   * @throws IndexOutOfBoundsException when either index is out of playlist bounds, i.e. {@code
-   *      fromIndex} &lt; 0 or &gt;= {@link #getSize()}, {@code toIndex} &lt; 0 or &gt; {@link
-   *      #getSize()}
-   * @throws IndexOutOfBoundsException when range is malformed, i.e. {@code fromIndex} &gt;
-   *      {@code toIndex}
+   *     untouched. This index must be in the range of 0 &lt;= index &lt;= {@link #getSize()}.
+   * @throws IndexOutOfBoundsException When the range is malformed, i.e. {@code fromIndex} &lt;
+   *       0, {@code toIndex} &gt; {@link #getSize()}, {@code fromIndex} &gt; {@code toIndex}
    */
   public final synchronized void removeMediaSourceRange(int fromIndex, int toIndex) {
     removeMediaSourceRange(fromIndex, toIndex, null);
@@ -335,37 +330,23 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
    * is thrown.
    *
    * @param fromIndex The initial range index, pointing to the first media source that will be
-   *     removed. This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   *     removed. This index must be in the range of 0 &lt;= index &lt;= {@link #getSize()}.
    * @param toIndex The final range index, pointing to the first media source that will be left
-   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
-   *     This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
-   *
-   * @throws IndexOutOfBoundsException when either index is out of playlist bounds, i.e. {@code
-   *      fromIndex} &lt; 0 or &gt;= {@link #getSize()}, {@code toIndex} &lt; 0 or &gt; {@link
-   *      #getSize()}
-   * @throws IndexOutOfBoundsException when range is malformed, i.e. {@code fromIndex} &gt;
-   *      {@code toIndex}
+   *     untouched. This index must be in the range of 0 &lt;= index &lt;= {@link #getSize()}.
    * @param actionOnCompletion A {@link Runnable} which is executed immediately after the media
    *     source range has been removed from the playlist.
+   * @throws IndexOutOfBoundsException When the range is malformed, i.e. {@code fromIndex} &lt;
+   *       0, {@code toIndex} &gt; {@link #getSize()}, {@code fromIndex} &gt; {@code toIndex}
    */
   public final synchronized void removeMediaSourceRange(
           int fromIndex, int toIndex, @Nullable Runnable actionOnCompletion) {
-    if (fromIndex < 0 || fromIndex >= mediaSourcesPublic.size()) {
-      throw new IndexOutOfBoundsException(String.format("Cannot remove source range: initial index (%d) out of bounds", fromIndex));
-    }
-    if (toIndex < 0 || toIndex > mediaSourcesPublic.size()) {
-      throw new IndexOutOfBoundsException(String.format("Cannot remove source range: final index (%d) out of bounds", toIndex));
-    }
-    if (fromIndex > toIndex) {
-      throw new IndexOutOfBoundsException(String.format("Cannot remove source range: range malformed (%d, %d)", fromIndex, toIndex));
-    }
+    Util.removeRange(mediaSourcesPublic, fromIndex, toIndex);
     if (fromIndex == toIndex) {
       if (actionOnCompletion != null) {
         actionOnCompletion.run();
       }
       return;
     }
-    mediaSourcesPublic.subList(fromIndex, toIndex).clear();
     if (player != null) {
       player
               .createMessage(this)
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
@@ -50,6 +50,7 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
   private static final int MSG_ADD = 0;
   private static final int MSG_ADD_MULTIPLE = 1;
   private static final int MSG_REMOVE = 2;
+  private static final int MSG_REMOVE_RANGE = 200;
   private static final int MSG_MOVE = 3;
   private static final int MSG_CLEAR = 4;
   private static final int MSG_NOTIFY_LISTENER = 5;
@@ -265,6 +266,9 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
    * <p>Note: If you want to move the instance, it's preferable to use {@link #moveMediaSource(int,
    * int)} instead.
    *
+   * <p>Note: If you want to remove a set of contiguous sources, it's preferable to use
+   * {@link #removeMediaSourceRange(int, int)} instead.
+   *
    * @param index The index at which the media source will be removed. This index must be in the
    *     range of 0 &lt;= index &lt; {@link #getSize()}.
    */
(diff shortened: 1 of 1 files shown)


BrainCrumbz:
CLA added for all commit authors

GiuseppePiscopo:
I confirm I'm ok with that.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -50,6 +50,7 @@
   private static final int MSG_ADD = 0;
   private static final int MSG_ADD_MULTIPLE = 1;
   private static final int MSG_REMOVE = 2;
+  private static final int MSG_REMOVE_RANGE = 200;
```
tonihei: Can you renumber this to be 3 and change the numbers below accordingly?

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   * Removes a range of {@link MediaSource}s from the playlist, by specifying an initial index
+   * (included) and a final index (excluded).
+   *
+   * <p>Note: when specified range is empty, no actual media source is removed and no exception
+   * is thrown.
+   *
+   * @param fromIndex The initial range index, pointing to the first media source that will be
+   *     removed. This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   * @param toIndex The final range index, pointing to the first media source that will be left
+   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
```
tonihei: "The last media source to be removed is at index {@code toIndex} - 1."  - That seems redundant and can be removed.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
 
+  /**
+   * Removes a range of {@link MediaSource}s from the playlist, by specifying an initial index
+   * (included) and a final index (excluded).
+   *
+   * <p>Note: when specified range is empty, no actual media source is removed and no exception
+   * is thrown.
+   *
+   * @param fromIndex The initial range index, pointing to the first media source that will be
+   *     removed. This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
```
tonihei: Both ranges should be "in the range of 0 &lt;= index &lt;= {@link #getSize()}."  That is, including the getSize index. 
Otherwise you wouldn't be able to do removeMediaSourceRange(getSize(), getSize()) as an edge case.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   * @param toIndex The final range index, pointing to the first media source that will be left
+   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
+   *     This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   *
+   * @throws IndexOutOfBoundsException when either index is out of playlist bounds, i.e. {@code
+   *      fromIndex} &lt; 0 or &gt;= {@link #getSize()}, {@code toIndex} &lt; 0 or &gt; {@link
+   *      #getSize()}
+   * @throws IndexOutOfBoundsException when range is malformed, i.e. {@code fromIndex} &gt;
```
tonihei: Can you combine them in one @throws doc? Maybe just as "When the range is malformed, i.e. {@code fromIndex} &lt; 0, {@code toIndex} &gt; {@link  #getSize()}, or {@code fromIndex} &gt; {@code toIndex}."

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   *      {@code toIndex}
+   */
+  public final synchronized void removeMediaSourceRange(int fromIndex, int toIndex) {
+    removeMediaSourceRange(fromIndex, toIndex, null);
+  }
+
+  /**
+   * Removes a range of {@link MediaSource}s from the playlist, by specifying an initial index
+   * (included) and a final index (excluded), and executes a custom action on completion..
```
tonihei: Remove second full stop.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   * @param toIndex The final range index, pointing to the first media source that will be left
+   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
+   *     This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   *
+   * @throws IndexOutOfBoundsException when either index is out of playlist bounds, i.e. {@code
+   *      fromIndex} &lt; 0 or &gt;= {@link #getSize()}, {@code toIndex} &lt; 0 or &gt; {@link
+   *      #getSize()}
+   * @throws IndexOutOfBoundsException when range is malformed, i.e. {@code fromIndex} &gt;
+   *      {@code toIndex}
+   * @param actionOnCompletion A {@link Runnable} which is executed immediately after the media
```
tonihei: This param should be together with the other params above.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   *
+   * <p>Note: when specified range is empty, no actual media source is removed and no exception
+   * is thrown.
+   *
+   * @param fromIndex The initial range index, pointing to the first media source that will be
+   *     removed. This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   * @param toIndex The final range index, pointing to the first media source that will be left
+   *     untouched. The last media source to be removed is at index {@code toIndex} - 1.
+   *     This index must be in the range of 0 &lt;= index &lt; {@link #getSize()}.
+   *
```
tonihei: We usually leave no empty line between @param and @throws.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+    if (fromIndex > toIndex) {
+      throw new IndexOutOfBoundsException(String.format("Cannot remove source range: range malformed (%d, %d)", fromIndex, toIndex));
+    }
+    if (fromIndex == toIndex) {
+      if (actionOnCompletion != null) {
+        actionOnCompletion.run();
+      }
+      return;
+    }
+    mediaSourcesPublic.subList(fromIndex, toIndex).clear();
```
tonihei: Can you replace that with Util.removeRange(mediaSourcesPublic, fromIndex, toIndex)? That calls the same method but is slightly more readable.

Also move it above the if(fromIndex == toIndex) check to ensure the exceptions are thrown first.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+   *      fromIndex} &lt; 0 or &gt;= {@link #getSize()}, {@code toIndex} &lt; 0 or &gt; {@link
+   *      #getSize()}
+   * @throws IndexOutOfBoundsException when range is malformed, i.e. {@code fromIndex} &gt;
+   *      {@code toIndex}
+   * @param actionOnCompletion A {@link Runnable} which is executed immediately after the media
+   *     source range has been removed from the playlist.
+   */
+  public final synchronized void removeMediaSourceRange(
+          int fromIndex, int toIndex, @Nullable Runnable actionOnCompletion) {
+    if (fromIndex < 0 || fromIndex >= mediaSourcesPublic.size()) {
```
tonihei: No need to do the explicit bound checks here. As soon as you call the subList method, the IndexOutOfBoundsException will be thrown anyway.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
```diff
@@ -297,6 +304,79 @@ public final synchronized void removeMediaSource(
...
+    if (fromIndex == toIndex) {
+      if (actionOnCompletion != null) {
+        actionOnCompletion.run();
+      }
+      return;
+    }
+    mediaSourcesPublic.subList(fromIndex, toIndex).clear();
+    if (player != null) {
+      player
+              .createMessage(this)
```
tonihei: These should be indented by 4 spaces (relative to "player")

tonihei (commented):
Looks good in general. Just a couple of formatting issues and minor changes.

ojw28:
No particular need to fix it for this contribution, but for future contributions it's preferable to configure git to use the same email as the one you specified when signing the CLA ([setting user.email is described here](https://help.github.com/articles/setting-your-commit-email-address-in-git/)).

ojw28:
CLA status was manually verified for the committing username (GiuseppePiscopo), as the commits use a  `@users.noreply.github.com` email address.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
BrainCrumbz: At a given time, "occupied" track indices are 0, 1, ... `getSize() - 1`. `fromIndex` should point to an occupied index, as the initial range index is included in the range. How can one remove the track at index `getSize()`, if there's no track at that index? Just trying to understand

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
BrainCrumbz: Totally agree with grouping. Are you also suggesting to slim down the number of (broken) examples provided in JavaDoc?

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
BrainCrumbz: Ok, just not sure what that Util method does w.r.t. exceptions (I think `sublist()` does not throw when indices are the same).

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
tonihei: As the end index is exclusive, the method must allow getSize() as end index. 

And to support the example you gave in the issue, to keep an element at index N only, it would be nice to be able to call
removeMediaSourceRange(0, N) followed by removeMediaSourceRange(1,getSize()). If this is supposed to remove the last element, the second call will result in removeMediaSourceRange(1, 1) which should be a no-op.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
tonihei: Yes. As we already require fromIndex>lastIndex, the other two examples are not needed.

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
BrainCrumbz: We've abandoned the "simplicity" implied by initial issue example when we "gave up" on API tolerance (so to say). Since then, the idea seemed: have a clear API with errors when arguments are not correct. And for `fromIndex` to be equal to `getSize()` seems not correct. Of course we noticed that you cannot call `removeMediaSourceRange(getSize(), getSize())` anymore. And also you cannot call `removeMediaSourceRange(1, getSize())` in issue second call, that would throw too.

Staying consistent with allowed param values in signature seems quite a thing. By allowing passing `getIndex()` in `fromIndex`, it seems to go back to the "tolerant" approach somehow, which is fine anyway. Don't know, your call.

tonihei: I'd say it's best to follow the existing behaviour of [List.subList](https://developer.android.com/reference/java/util/List.html#subList(int,%20int)) and the like. That is, throwing if `startIndex<0`, `endIndex>size` or `startIndex>endIndex`. This ensures the operation is valid, but is also tolerant enough to support common use cases like "remove everything behind element N" with removeRange(N, size) which will work even if N is the last element.

BrainCrumbz: Cool. Just to tease a little bit, in case like "remove everything behind element N" when N is the last element, then `N = size - 1`, and `removeMediaSourceRange(size - 1, size)` already works as is. What instead currently throws (and probably looks not tolerant) is `removeMediaSourceRange(size, size)`. Right?

Anyway, changes are coming

tonihei: `removeMediaSourceRange(size - 1, size)` would remove the element at `N = size - 1` though, right? 

And if I'm not mistaken `removeMediaSourceRange(size, size)` won't throw too because List.subList doesn't throw for this case.

tonihei:
Thanks! Looks good now :)

library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
BrainCrumbz: Q1: correct, current implementation would remove element at `size - 1`, i.e. "the last element"

Q2: `removeMediaSourceRange(size, size)` throws in current (<=> initial version) of this PR. It will not throw after applying your proposed changes, as `List.subList(size, size)` does not throw. I was referring to current implementation.

Cool

BrainCrumbz:
Not sure what's happening re. CLA. The other commit author account (GiuseppePiscopo) has registered google CLA with the same email address also associated to Github profile.

Can that be related to the "keep my address private" option in github profile?

ojw28:
I think it's because the email address the CLA is signed under doesn't match the email address on the commits (which is a `@users.noreply.github.com` address). See my comment above (https://github.com/google/ExoPlayer/pull/4564#issuecomment-407765117) for how to fix that for future commits. It could well be that the "keep my address private" option overrides whatever address you configure, so yes, that could be related!

For this commit we manually verified the username against our CLA records, so it's not a blocking issue.

BrainCrumbz:
Thanks for feedback, yes we read that. Do you know if CLA can be signed under a `@users.noreply.github.com` address?

ojw28:
Unsure. It seems a bit strange to do that :). We should probably just make the googlebot clever enough to look at the username as well. I'll try and get someone to do that internally!

ojw28:
Turns out the issue with the CLA is that the account sending the pull request is different to the account whose commits are included within it. Hence we need to manually confirm that the commit author is OK with their changes being part of the pull request, under their signed CLA. If we didn't do this, it would be possible for someone to try and contribute commits by an author who has signed a CLA, but does not want the commits in question to be contributed.

Please could you confirm with account @GiuseppePiscopo that you're OK with all commits that are part of this pull request being submitted (you'll need to do this again whenever a new commit is added).

In future, using the same account for commits and pull requests, which is an account with a signed CLA, would help streamline this process.

BrainCrumbz:
Hi there. Yes, we were aware of this distinction between *pull request opener* and *commit authors*, thanks for clarifying. Could not find more info on how to get a user confirmation, so we just added [this comment here above](https://github.com/google/ExoPlayer/pull/4564#issuecomment-407367147). Does that count as a confirmation?

Re. CLA, @GiuseppePiscopo also has a signed CLA in google, using the same email address associated with Github. Problem is (we think), email preference on Github is set to private, hence commits appear authored by a *whatever-username@users.noreply.github.com*, and it's not possible to sign a CLA under that "dummy" email address.

If one wants to keep distinct commit authors, it seems there's no way to also keep Github email private. Anyway, for the next time we'll try to switch user so that all commits show up under same account as PR creator.

ojw28:
> Re. CLA, @GiuseppePiscopo also has a signed CLA in google, using the same email address associated with Github. Problem is (we think), email preference on Github is set to private, hence commits appear authored by a whatever-username@users.noreply.github.com, and it's not possible to sign a CLA under that "dummy" email address.

It turned out this was not the problem. The CLA bot is clever enough to look at the username in this case. The problem is solely that the pull request opener is different to the commit author. In this case there is no automated way for the CLA bot to validate that the commit author is OK with their commit being included in the pull request, which is why manual validation is necessary. Them having signed the CLA is not sufficient (since they might have signed it to contribute some completely unrelated change to another Google project).

We need the commit author to manually confirm they are OK with their commits being included in this case. Since additional commits were added after the comment you reference, we need the commit author to comment again on this thread, to cover the additional commits since that comment. Thanks!

GiuseppePiscopo:
I confirm I'm ok with including those additional commits as well

BrainCrumbz:
How can we track when this PR will be included in a public release? Thanks

ojw28:
It will be in 2.9.0 (for which we don't yet have a solid eta).

PR #4569, opened 2018-07-24
Extract tags from SubRip subtitles, add support for alignment

szaboa:
Extracted valid tags from srt subtitles (so those won't be rendered anymore) and added support for different alignments tags (based on SSA v4+ specification) such as ``{\an2}`` for example.


szaboa:
Closing this pull request and sending a new one, where the source branch is a feature branch and not dev-2.

Issue #4306, opened 2018-05-28
SRT subs position tags


ghost:
**### Issue description**
It doesn't recognize the position of the subtitles. I don't know if its possible just ignore and not show the tag, or position in the correct place.

### Link to test content
![Image of subs](http://images2.imagebam.com/f2/a4/9f/6c5c3d878070534.jpg)

### Version of ExoPlayer being used
Plex android app embedded.

ojw28:
According to [this thread](https://forum.videohelp.com/threads/332300-How-to-position-a-subtitle-in-a-srt-file) positional tags like `{\an8}` aren't actually part of the SRT format, although some tools do appear to support such tags.

It's really hard for us to support subtitle formats that aren't properly specified, and in particular where undocumented additions have been made, as appears to be the case here. I'm going to class this as an enhancement rather than a bug, since the tags aren't really part of SRT, and mark it as low priority. We would accept a pull request into the `dev-v2` branch to either ignore such tags or handle them properly, so please feel free to send such a request if you're sufficiently motivated to fix this yourself. Thanks!

szaboa:
After a quick research, I got the same conclusion as mentioned above, those tags are not part of the SRT format. People started to borrow these tags from other places over time and step by step the players started to support them. These tags are taken from ASS (https://www.matroska.org/technical/specs/subtitles/ssa.html). I will try to add support for these ```\an``` tags, as my first contribution to ExoPlayer. What do you think @ojw28 ?

ojw28:
That would be great! A few things to keep in mind to help us merge any changes with minimal effort:

1. I would not envisage you needing to change anything in `Cue`, `SubtitlePainter` or `SubtitleView` to achieve this. `Cue` already supports positioning that's hopefully flexible for what you need to do, and the UI classes already make use of it. If you find yourself needing to edit those classes, I'd suggest we discuss that before you go too far :).
1. Given the tags aren't part of the format, please make sure you're handling of them is really robust (i.e. does not crash if the tag isn't exactly as expected). We can't really have loads of confidence that they'll be formatted correctly, given they've not been properly defined :).

Thanks!

szaboa:
Sure thing! 
Just started to working on it, I will do the pull-request once is done, thank you :)

szaboa:
@ojw28 
I am about to extend ```com.google.android.exoplayer2.text.subrip.SubripDecoderTest```, but it seems all the tests are failing for me:

```
[...]

PR #4582, opened 2018-07-26
#4306 - Extract tags from SubRip subtitles, add support for alignment

szaboa:
Extracted valid tags from srt subtitles (so those won't be rendered anymore) and added support for different alignments tags (based on SSA v4+ specification) such as {\an2} for example.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
@@ -15,7 +15,9 @@
  */
 package com.google.android.exoplayer2.text.subrip;
 
+import android.support.annotation.StringDef;
 import android.text.Html;
+import android.text.Layout;
 import android.text.Spanned;
 import android.text.TextUtils;
 import android.util.Log;
@@ -23,7 +25,11 @@ import com.google.android.exoplayer2.text.Cue;
 import com.google.android.exoplayer2.text.SimpleSubtitleDecoder;
 import com.google.android.exoplayer2.util.LongArray;
 import com.google.android.exoplayer2.util.ParsableByteArray;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -37,6 +43,33 @@ public final class SubripDecoder extends SimpleSubtitleDecoder {
   private static final String SUBRIP_TIMECODE = "(?:(\\d+):)?(\\d+):(\\d+),(\\d+)";
   private static final Pattern SUBRIP_TIMING_LINE =
       Pattern.compile("\\s*(" + SUBRIP_TIMECODE + ")\\s*-->\\s*(" + SUBRIP_TIMECODE + ")?\\s*");
+
+  private static final Pattern SUBRIP_TAG_PATTERN = Pattern.compile("\\{\\\\.*?\\}");
+  private static final String SUBRIP_ALIGNMENT_TAG = "\\{\\\\an[1-9]\\}";
+
+  private static final float DEFAULT_START_FRACTION = 0.08f;
+  private static final float DEFAULT_END_FRACTION = 1 - DEFAULT_START_FRACTION;
+  private static final float DEFAULT_MID_FRACTION = 0.5f;
+
+  @Retention(RetentionPolicy.SOURCE)
+  @StringDef({
+      ALIGN_BOTTOM_LEFT, ALIGN_BOTTOM_MID, ALIGN_BOTTOM_RIGHT,
+      ALIGN_MID_LEFT, ALIGN_MID_MID, ALIGN_MID_RIGHT,
+      ALIGN_TOP_LEFT, ALIGN_TOP_MID, ALIGN_TOP_RIGHT
+  })
+
+  private @interface SubRipTag {}
+
+  // Possible valid alignment tags based on SSA v4+ specs
+  private static final String ALIGN_BOTTOM_LEFT  = "{\\an1}";
+  private static final String ALIGN_BOTTOM_MID   = "{\\an2}";
+  private static final String ALIGN_BOTTOM_RIGHT = "{\\an3}";
+  private static final String ALIGN_MID_LEFT     = "{\\an4}";
+  private static final String ALIGN_MID_MID      = "{\\an5}";
+  private static final String ALIGN_MID_RIGHT    = "{\\an6}";
+  private static final String ALIGN_TOP_LEFT     = "{\\an7}";
+  private static final String ALIGN_TOP_MID      = "{\\an8}";
+  private static final String ALIGN_TOP_RIGHT    = "{\\an9}";
 
   private final StringBuilder textBuilder;
 
@@ -95,8 +128,36 @@ public final class SubripDecoder extends SimpleSubtitleDecoder {
         textBuilder.append(currentLine.trim());
       }
 
-      Spanned text = Html.fromHtml(textBuilder.toString());
-      cues.add(new Cue(text));
+      // Extract tags
+      SubtitleTagResult tagResult = extractTags(textBuilder);
+      Spanned text = Html.fromHtml(tagResult.cue);
+
+      Cue cue = null;
+
+      // Check if tags are present
+      if (tagResult.tags.length > 0) {
+
+        boolean alignTagFound = false;
+
+        // At end of this loop the clue must be created with the applied tags
+        for (String tag : tagResult.tags) {
+
+          // Check if the tag is an alignment tag
+          if (tag.matches(SUBRIP_ALIGNMENT_TAG)) {
+
+            // Based on the specs, in case of the alignment tags only the first appearance counts
+            if (alignTagFound) continue;
+            alignTagFound = true;
+
+            AlignmentResult alignmentResult = getAlignmentValues(tag);
+            cue = new Cue(text, Layout.Alignment.ALIGN_NORMAL, alignmentResult.line, Cue.LINE_TYPE_FRACTION,
+                alignmentResult.lineAnchor, alignmentResult.position, alignmentResult.positionAnchor, Cue.DIMEN_UNSET);
+          }
+        }
+      }
+
+      cues.add(cue == null ? new Cue(text) : cue);
+
       if (haveEndTimecode) {
         cues.add(null);
       }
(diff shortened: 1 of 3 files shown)


library/core/src/test/java/com/google/android/exoplayer2/text/subrip/SubripDecoderTest.java
```diff
@@ -154,6 +155,27 @@ public void testDecodeNoEndTimecodes() throws IOException {
         .isEqualTo("Or to the end of the media.");
   }
 
+  @Test
+  public void testDecodeCueWithTag() throws IOException{
```
ojw28, line 161: This test doesn't appear to be asserting anything other than that the tags were removed (or not) from the text. Shouldn't the test also be asserting that they were applied successfully (e.g. that the subtitles in the cues have the right alignment parameters)

library/core/src/test/java/com/google/android/exoplayer2/text/subrip/SubripDecoderTest.java
```diff
@@ -154,6 +155,27 @@ public void testDecodeNoEndTimecodes() throws IOException {
...
+    byte[] bytes = TestUtil.getByteArray(RuntimeEnvironment.application, TYPICAL_WITH_TAGS);
+    SubripSubtitle subtitle = decoder.decode(bytes, bytes.length, false);
+    assertThat(subtitle.getCues(subtitle.getEventTime(0)).get(0).text.toString())
+        .isEqualTo("This is the first subtitle.");
+    assertThat(subtitle.getCues(subtitle.getEventTime(2)).get(0).text.toString())
+        .isEqualTo("This is the second subtitle.\nSecond subtitle with second line.");
+    assertThat(subtitle.getCues(subtitle.getEventTime(4)).get(0).text.toString())
+        .isEqualTo("This is the third subtitle.");
+
+    // Based on the SSA v4+ specs the curly bracket must be followed by a backslash, so this is
```
ojw28, line 172: Can you link to this doc (just in this thread, not in the source). The doc I found has a bunch of examples where this is always true, but it's also always true that there's no whitespace inside the brackets either. Hence it seems strange to forbid `{ \\an2}` but not `{\ an2}`. Unless you have counter examples, I think we should just require no whitespace in the bracketed region. Which I think you can do by using `[^\s]` instead of `.` in the `Pattern`.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -95,8 +128,36 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
         textBuilder.append(currentLine.trim());
       }
 
-      Spanned text = Html.fromHtml(textBuilder.toString());
-      cues.add(new Cue(text));
+      // Extract tags
+      SubtitleTagResult tagResult = extractTags(textBuilder);
```
ojw28: Given you wouldn't expect a tag to be separated over multiple lines, it seems cleaner to do this in the while loop above. You could replace the last line in the while loop with something like:
```
textBuilder.append(processLine(currentLine, tags));
```
Where `tags` is an `Arraylist` to which any new tags are added, and `processLine` is responsible for trimming the line and extracting tags. Note passing the list in as an argument also avoids having to define `SubtitleTagResult`.

You'd then make the `tags` list before the start of the while loop to pass in (you could even re-use the same single instance if you want, as is done for the `StringBuilder`, as long as you remember to clear it.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -95,8 +128,36 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+      Spanned text = Html.fromHtml(tagResult.cue);
+
+      Cue cue = null;
+
+      // Check if tags are present
+      if (tagResult.tags.length > 0) {
```
ojw28: nit - You can probably remove this. The loop will be a no-op if tags has length zero.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -95,8 +128,36 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+
+        boolean alignTagFound = false;
```
ojw28: Can't you just `break;` from the loop when you find the first align tag, and avoid the need for this?

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -95,8 +128,36 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+        for (String tag : tagResult.tags) {
+
+          // Check if the tag is an alignment tag
+          if (tag.matches(SUBRIP_ALIGNMENT_TAG)) {
+
+            // Based on the specs, in case of the alignment tags only the first appearance counts
+            if (alignTagFound) continue;
+            alignTagFound = true;
+
+            AlignmentResult alignmentResult = getAlignmentValues(tag);
```
ojw28: If you change `getAlignmentValues(tag)` to be `buildCue(Spanned text, String alignmentTag)` then you can get rid of `AlignmentResult`.

library/core/src/test/java/com/google/android/exoplayer2/text/subrip/SubripDecoderTest.java
szaboa: I followed the docs that can be downloaded from here:
moodub.free.fr/video/ass-specs.doc

And it says: "All override codes are always preceded by a backslash \"
So that's why I didn't allow a space after the bracket. Regarding spaces after the backslash I just followed how the VLC handles these cases.

We sure can require no whitespace in the bracketed region, is up to you :)

ojw28: Guess we can allow space, if there's an existing example of doing this (as there seems to be)!

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
szaboa: Sure I could, but I intended to wrote like this to be generic and to be able to extend the loop with other tag's "apply" logic. Note that this for loop isn't only for align tags (for **now** it is), instead goes over **all** the tags and applies them, I suppose it will have support for more tags in the future and it that case breaking after the first align tag would be incorrect behavior.

Taking these into consideration, if you still would like to just "break", then I can change it.

szaboa:
@ojw28 waiting for your opinion on the `break;` thing, meanwhile I will extend the test cases. The rest is resolved.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
ojw28: I think our preference (for small blocks of code; this doesn't really apply to larger scale design choices for obvious reasons) is to write it in the simplest way for current usage, and change it as and when the usage changes. So I think using break is still preferable for now.

ojw28:
Replied to the `break;` thing. Please push the updated revision when ready. Thanks!

ojw28:
@szaboa - I think there's still an open comment on `testDecodeCueWithTag`. Perhaps you're aware of this already; just checking!

szaboa:
Yes, there is that and the break thing. I am on vacation right now, I will resolve those on the next Monday (first of October)  when I am back, sorry for the delays!

szaboa:
@ojw28 
Just pushed the changes for the remaining open comments. Everything should be OK now.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -108,12 +156,122 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+        break;
+      case ALIGN_TOP_RIGHT:
+        line = DEFAULT_START_FRACTION;
+        position = DEFAULT_END_FRACTION;
+        positionAnchor = Cue.ANCHOR_TYPE_END;
+        lineAnchor = Cue.ANCHOR_TYPE_START;
+        break;
+    }
+
+    return new Cue(text, Layout.Alignment.ALIGN_NORMAL, line, Cue.LINE_TYPE_FRACTION, lineAnchor, position, positionAnchor, Cue.DIMEN_UNSET);
```
ojw28: Can you use null for text alignment? `Cue` implies that's what you should use unless you know better.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -108,12 +156,122 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+    float line = DEFAULT_END_FRACTION, position = DEFAULT_MID_FRACTION;
+    @Cue.AnchorType int positionAnchor = Cue.ANCHOR_TYPE_MIDDLE;
+    @Cue.AnchorType int lineAnchor = Cue.ANCHOR_TYPE_END;
+
+    switch (alignmentTag) {
+      case ALIGN_BOTTOM_LEFT:
+        line = DEFAULT_END_FRACTION;
+        position = DEFAULT_START_FRACTION;
+        positionAnchor = Cue.ANCHOR_TYPE_START;
+        lineAnchor = Cue.ANCHOR_TYPE_END;
```
ojw28: Would be nice to order the assignments as `line`,`lineAnchor`,`position`,`positionAnchor` so the line and position ones are grouped. Ignorable, but it might be even clearer if split into two switch blocks with three cases each, like:

```
float position;
@Cue.AnchorType int positionAnchor;
switch (alignmentTag) {
  case ALIGN_BOTTOM_LEFT:
  case ALIGN_MID_LEFT:
  case ALIGN_TOP_LEFT:
    position = DEFAULT_START_FRACTION;
    positionAnchor = Cue.ANCHOR_TYPE_START;
  case ALIGN_BOTTOM_MID:
  case ALIGN_MID_MID:
  case ALIGN_TOP_MID:
    ...
  case ALIGN_BOTTOM_RIGHT:
  case ALIGN_MID_RIGHT:
  case ALIGN_TOP_RIGHT:
    ...
}
... (ditto for line and lineAnchor) ...
```

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -108,12 +156,122 @@ protected SubripSubtitle decode(byte[] bytes, int length, boolean reset) {
...
+   * where the number specifies the direction (based on the numpad layout).
+   * Note. older SSA scripts may contain tags like {\a1[1-9]} but these are based on
+   * other direction rules, but multiple sources says that these are deprecated, so no support here either
+   *
+   * @param alignmentTag Alignment tag
+   * @return Built cue
+   */
+  private Cue buildCue(Spanned text, String alignmentTag) {
+    // Default values used for positioning the subtitle in case of align tags
+    float line = DEFAULT_END_FRACTION, position = DEFAULT_MID_FRACTION;
```
ojw28: No need to assign initial values to any of these (it's better not to, since that'll make the compiler ensure they're initialized along all code paths below). Add `default:` to the last case in the switch statement if needed to make the compiler happy.

ojw28 (commented):
Looks good, thanks. I put a few final comments inline.

library/core/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
szaboa: Yep, much nicer!

szaboa:
Pushed the changes according to the comments.

ojw28:
Thanks! We made some further improvements during internal review, which were merged in https://github.com/google/ExoPlayer/commit/7849a5eb52a434883a0c0d33f0b9a9ab20067679. Let us know if we accidentally broken anything :).

szaboa:
It works fine and looks much better now, actually I am feeling that I did poor job after seeing your changes :)

Anyway, based on this I will try to pick up a better coding style in my next pull request. 
Cheers!

ojw28:
Not at all; this feature wouldn't have happened without you ;). Thanks for the contribution!

Issue #4304, opened 2018-05-27
Add option to show buffering view when setPlayWhenReady=false


Gidex:
private void updateBuffering() {
    if (bufferingView != null) {
      boolean showBufferingSpinner =
          showBuffering
              && player != null
              && player.getPlaybackState() == Player.STATE_BUFFERING
              && `player.getPlayWhenReady();`
      bufferingView.setVisibility(showBufferingSpinner ? View.VISIBLE : View.GONE);
    }
  }

I think PlayerView.getPlayWhenReady() should not interfere with the visibility of the bufferingView, since the player keeps buffering even when paused...

botaydotcom:
Hi @Gidex,
I think the idea is to only show buffering view when user is watching/focusing on the player. If `player.getPlayWhenReady()` is false, the player is paused, and buffering is only done in the background. This is something user probably doesn't care about.
Is this something that's causing problem in your use case? Could you please clarify the problem you are facing in more details?

ojw28:
I can confirm the current behavior is intentional. Displaying a buffering spinner when the user intention isn't to start playback immediately feels like extra UI noise that doesn't provide the user with useful information. We could probably make the behavior configurable, but I'm not convinced it's important.

Gidex:
As a user, I find it common for the user to use the buffering spinner to know when the player is ready to play. When watching and the player starts buffering, if the user pauses and the buffering spinner disappears, he may think the player is ready, but if the buffering spinner appears again when playing, it may seem confusing. In slow or unstable connections it is common for the user to pause to wait for the player to load enough so that he can continue without problems, so if he pauses and leaves his cell phone on the couch to go to drink water, when he returns and does not see the buffering spinner, no doubt he will think that the player is ready to play.
I'm using `onPlayerStateChanged` to do this, but it would be great if `PlayerView.setShowBuffering` was configurable :)

szaboa:
@ojw28 
Can I work on this or it is already in progress?

ojw28:
Feel free to work on this! I think I'd envisage a new `PlayerView.setShowBuffering` method that takes an `IntDef` that can have three values (`SHOW_BUFFERING_NEVER` / `SHOW_BUFFERING_WHEN_PLAYING` / `SHOW_BUFFERING_ALWAYS`). You can then deprecate the old method and update it to call the new one in a way that maintains its current behavior:

```
public void setShowBuffering(boolean showBuffering) {
  setShowBuffering(
[...]

PR #4585, opened 2018-07-26
#4304 - Add option to show buffering view when setPlayWhenReady is false

szaboa:
Added an option to show buffering view when ```playWhenReady``` is set to false. So from now the visibility state can be defined with the following options:

1. ```SHOW_BUFFERING_NEVER```
2. ```SHOW_BUFFERING_ALWAYS```
3. ```SHOW_BUFFERING_WHEN_PLAYING```

Checked on emulator where I tweaked the network type / signal strength to be able to test it and it seemed OK to me in every case. 

Link to the original issue: https://github.com/google/ExoPlayer/issues/4304

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
@@ -666,13 +666,13 @@ public class PlayerView extends FrameLayout {
    * Sets whether a buffering spinner is displayed when the player is in the buffering state. The
    * buffering spinner is not displayed by default.
    *
-   * Deprecated, use {@link #setShowBuffering(int)}
+   * @deprecated Use {@link #setShowBuffering(int)}
    *
    * @param showBuffering Whether the buffering icon is displayed
    */
   @Deprecated
   public void setShowBuffering(boolean showBuffering) {
-    setShowBuffering(showBuffering ? SHOW_BUFFERING_ALWAYS : SHOW_BUFFERING_NEVER);
+    setShowBuffering(showBuffering ? SHOW_BUFFERING_WHEN_PLAYING : SHOW_BUFFERING_NEVER);
   }
 
   /**
@@ -683,7 +683,7 @@ public class PlayerView extends FrameLayout {
    * <li>{@link #SHOW_BUFFERING_ALWAYS} displayed always when buffering
    * <li>{@link #SHOW_BUFFERING_NEVER} not displayed at all
    * <li>{@link #SHOW_BUFFERING_WHEN_PLAYING} displayed only when playing and buffering
-   * </p></ul>
+   * </ul></p>
    *
    * @param showBuffering Buffering strategy that defines when the buffering icon is displayed
    */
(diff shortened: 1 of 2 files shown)


library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -655,9 +666,28 @@ public void setKeepContentOnPlayerReset(boolean keepContentOnPlayerReset) {
    * Sets whether a buffering spinner is displayed when the player is in the buffering state. The
    * buffering spinner is not displayed by default.
    *
+   * Deprecated, use {@link #setShowBuffering(int)}
+   *
    * @param showBuffering Whether the buffering icon is displayed
    */
+  @Deprecated
   public void setShowBuffering(boolean showBuffering) {
+    setShowBuffering(showBuffering ? SHOW_BUFFERING_ALWAYS : SHOW_BUFFERING_NEVER);
```
ojw28: It should use SHOW_BUFFERING_WHEN_PLAYING rather than SHOW_BUFFERING_ALWAYS, so that the behavior of this method doesn't change.

library/ui/src/main/res/values/attrs.xml
```diff
@@ -51,7 +51,11 @@
     <attr name="hide_on_touch" format="boolean"/>
     <attr name="hide_during_ads" format="boolean"/>
     <attr name="auto_show" format="boolean"/>
-    <attr name="show_buffering" format="boolean"/>
+    <attr name="show_buffering" format="enum">
+      <enum name="show_buffering_never" value="0"/>
```
ojw28: Given how these are used, I think the names can probably be shortened to just "never" "always" and "when_playing".

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -655,9 +666,28 @@ public void setKeepContentOnPlayerReset(boolean keepContentOnPlayerReset) {
...
+
+  /**
+   * Sets whether a buffering spinner is displayed when the player is in the buffering state. The
+   * buffering spinner is not displayed by default (initial value {@link #SHOW_BUFFERING_NEVER})
+   *
+   * <p><ul>
+   * <li>{@link #SHOW_BUFFERING_ALWAYS} displayed always when buffering
+   * <li>{@link #SHOW_BUFFERING_NEVER} not displayed at all
+   * <li>{@link #SHOW_BUFFERING_WHEN_PLAYING} displayed only when playing and buffering
+   * </p></ul>
```
ojw28: Are `</p>` and `</ul>` the wrong way round (not sure :)).

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -655,9 +666,28 @@ public void setKeepContentOnPlayerReset(boolean keepContentOnPlayerReset) {
    * Sets whether a buffering spinner is displayed when the player is in the buffering state. The
    * buffering spinner is not displayed by default.
    *
+   * Deprecated, use {@link #setShowBuffering(int)}
```
ojw28: Use the `@deprecated` tag

ojw28:
Looks good. Posted some minor comments that would be good to address, and then we'll get this merged. Thanks!

szaboa:
@ojw28 

~~~> Posted some minor comments that would be good to address~~~

~~~I am not seeing any comments so I am not sure what minor things should I address, or did you posted somewhere internally and my job is done here ? :)~~~

Ok, I am seeing now, not sure why the comments delayed so much. I am going to fix those today.

ojw28:
Oops, posted them now. Thanks!

szaboa:
@ojw28 
The adjustments are done based on the comments

PR #4595, opened 2018-07-30
CEA 708 Decoder Improvements to meet Spec compliance and Sarnoff test…

peddisri:
… pass.

1. To support DLY and DLC commands, we need to have a service input buffer. Also the current design of CEA708 Decoder is not according to the spec. As per the spec CCP and service blocks are asynchronous. So we can't wait to start parsing service blocks until after generating a complete CCP. The parsing of various layers much be chained and be executed in parallel.

2. As per the spec, a CCP packet can contain more than one Service Blocks with same service number. Current implementation does not handle this.

3. Extended Service number should not be added to the service number read from the standard service block header.Instead, it should be read from the extended service block header.

4. Clear Window tests failed. Clearing window removes the window from display, it should just clear the text..

5. Sometimes the CC text is all jumbled up for H.264 content.There is a timing issue in rendering CC. Due to H264 B frames decoding and presentation order mismatch, sometimes, older timestamps CC frames are being decoded that resulted into jumbled CC text.We reject input frames that are older than the last decoded frames and delay processing input frames till we have received at least half of the input frames that can be held in the input queue.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -187,6 +187,7 @@ public final class Cea608Decoder extends CeaDecoder {
   private CueBuilder currentCueBuilder;
   private List<Cue> cues;
   private List<Cue> lastCues;
+  private long inputTimestampUs;
 
   private int captionMode;
   private int captionRowCount;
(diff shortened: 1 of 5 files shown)


peddisri:
@ojw28 The next change in this series

peddisri:
@ojw28  Any update on this?

peddisri:
@ojw28 Did you get a chance to look into this? Thanks!

peddisri:
@ojw28 Ping.

peddisri:
@ojw28 did you get hold of content to test this? awaiting feedback.

BetweenTwoBits:
@peddisri anyway you can pull in the latest from `dev-v2`? I'd like to review and test this branch with some Comcast assets.

zsmatyas:
Unfortunately, we are not going to keep this pull request up-to-date any more. I have created independent pull requests for almost all CEA608 features that are not dependent on other modifications, they should be already part of the latest release branches.
 
708 changes were rejected as there was no clear way to test the code an confirm behavior before. As I recall exo v2.5.4 is the latest code base we had a pull request ready for. It was rather large, so there was no chance of it being accepted as is.

PR #4701, opened 2018-08-21
ClearKey API < 26 fix

TadejZupancic:
Setting mimeType to CENC_SCHEME_MIME_TYPE for APIs prior to 26 in getKeyRequest never executed, because UUID was checked against ClearKey UUID. For APIs less than 27 ClearKey UUID is changed to Common PSSH UUID in the constructor.


TadejZupancic:
wrong branch

PR #4702, opened 2018-08-21
ClearKey API < 26 fix

TadejZupancic:
Setting mimeType to CENC_SCHEME_MIME_TYPE for APIs prior to 26 in getKeyRequest never executed, because UUID was checked against ClearKey UUID. For APIs less than 27 ClearKey UUID is changed to Common PSSH UUID in the constructor.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
@@ -153,8 +153,9 @@ public final class FrameworkMediaDrm implements ExoMediaDrm<FrameworkMediaCrypto
     }
 
     // Prior to API level 26 the ClearKey CDM only accepted "cenc" as the scheme for MP4.
+    // For APIs less than 27 ClearKey UUID is changed to Common PSSH UUID in the constructor
     if (Util.SDK_INT < 26
-        && C.CLEARKEY_UUID.equals(uuid)
+        && C.COMMON_PSSH_UUID.equals(uuid)
         && (MimeTypes.VIDEO_MP4.equals(mimeType) || MimeTypes.AUDIO_MP4.equals(mimeType))) {
       mimeType = CENC_SCHEME_MIME_TYPE;
     }


ojw28:
Thanks!

library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
```diff
@@ -153,8 +153,9 @@ public KeyRequest getKeyRequest(
     }
 
     // Prior to API level 26 the ClearKey CDM only accepted "cenc" as the scheme for MP4.
+    // For APIs less than 27 ClearKey UUID is changed to Common PSSH UUID in the constructor
     if (Util.SDK_INT < 26
-        && C.CLEARKEY_UUID.equals(uuid)
+        && C.COMMON_PSSH_UUID.equals(uuid)
```
ojw28, line 158: I don't think this fix is sufficient in all cases, because there are two other `C.CLEARKEY_UUID.equals(uuid)` checks in this class. I'm going to submit a cleaner fix that makes it so `uuid` is set to `C.CLEARKEY_UUID` in the first place.

ojw28:
Closing as we'll merge an alternative fix in the next day or so.

ojw28:
Should be fixed by https://github.com/google/ExoPlayer/commit/ff1812d3fed7d18baefb778b91f7397c9c61a8bc, but please let us know if that's not the case! Thanks for bringing this to our attention.

TadejZupancic:
I've been using an old version with my fix up till now and now that I've upgraded to the current version I've noticed, that it is not working properly. The uuid passed to MediaCrypto inside createMediaCrypto should be checked and changed as it was before in the constructor of FrameworkMediaDrm. So it should be: new MediaCrypto(Util.SDK_INT < 27 && C.CLEARKEY_UUID.equals(uuid) ? C.COMMON_PSSH_UUID : uuid, initData)

andrewlewis:
@TadejZupancic That sounds right so I'll make a fix. Thanks for reporting it.

andrewlewis:
Fix is in 282cf303a4ad321fa97e73faa8a133c130c8dd95.

Issue #4777, opened 2018-09-05
Frame rate limited track selection


nicklasl:
I am missing the feature to limit/filter the track selection based on the format frame rate. We are currently seeing issues with dropped frames on some devices when using DASH+Widevine combined with a higher framerate (50fps). 
Suggestion is to add frame rate to `TrackSelector` parameters?

tonihei:
That's certainly possible. Marking as an enhancement. 

If you are happy to give it try, we are always happy to accept pull requests. Otherwise, you'd need to wait until we get around to add this.

tonihei:
Closing as the pull request is already merged.

PR #4801, opened 2018-09-11
Trackselector filter on framerate

nicklasl:
This PR is to add the feature of selecting a track by frame rate: https://github.com/google/ExoPlayer/issues/4777

I'm not sure if I should start dabble with track score in the `selectFixedVideoTrack` case, depending frame rate?


nicklasl:
Submitted CLA for the email address used in commits.

tonihei:
Thanks for the pull request! The changes look good and can be merged as they are.

Issue #4757, opened 2018-08-31
How to know Hardware Accelerated Video Decoders ?


manishPatwari:
Hi,

Is there anyway we can get to know, which all Video decoders are Hardware accelerated ? I have seen the [MediaCodeUtil](https://github.com/google/ExoPlayer/blob/09d1e3e35d98fd02c31391ac1c891f0c73e76375/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java#L241) which depend on the framework to give best decoder _"MediaCodecList is sorted by the framework such that the best decoders come first_.". 

I have played Multi-Codec DASH file in [Vivo V9](https://www.devicespecifications.com/en/model/c3ed4935) ["Video file formats/codecs"] mobile which has hardware acceleration for H265 & VP9 but still it played video in H264 codec. 
Is there anyway I can play the video in best codec which is Hardware Accelerated, preference would be "V9" or "H265" and then "H264".

AquilesCanta:
In the case of a multi-codec adaptive stream, the first thing you need to look at is the TrackSelector. The TrackSelector will select the tracks from which the adaptive algorithm will pick chunks. If you want to choose tracks with a specific mime type (e.g: video/hevc), then you can do that by implementing your own TrackSelector. Maybe overriding the default track selector is an option for you.

Second point: To implement the track selection criteria (for instance, prefer HEVC over AVC) you might then want to look into the codec list. MediaCodecUtil might help with that. Also have a [look around](https://stackoverflow.com/questions/37715529/how-to-know-android-decoder-mediacodec-createdecoderbytypetype-is-hardware-or) for information about codecs being HW or SW.

Does this help?

manishPatwari:
@AquilesCanta  Thanks for the quick reply.

As per the Stackoverflow post, `codec.getName().startsWith("OMX.google.")` is Software encoder, but it doesn't seems correct. 

In the below screenShoot, I was doing CPU profiling in Moto E2, and the codec used for HEVC is "le.hevc.decoder" & for VP9 "gle.vp9.decoder" is used. If we go with definition  `codec.getName().startsWith("OMX.google.")`  then this codecs should be Hardware Accelerated. But the CPU usage is huge. 

Is there anyway we can avoid such cases and always use the optimised decoder ?

[...]

PR #4833, opened 2018-09-18
Hardware support flag

manishPatwari:
Added "isHardwareSupported" flag in MediaCodecInfo. This flag I have added here because MediaCodecInfo is cache and its been used multiple time. I have written TrackSelector which will select MediaCodecInfo if its Hardware supproted.


library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -158,6 +164,11 @@ private MediaCodecInfo(
     adaptive = !forceDisableAdaptive && capabilities != null && isAdaptive(capabilities);
     tunneling = capabilities != null && isTunneling(capabilities);
     secure = forceSecure || (capabilities != null && isSecure(capabilities));
+    if (!name.toLowerCase().startsWith("omx.google.")) {
```
ojw28, line 167: This isn't really a reliable test. In particular Google are likely to shortly start shipping software decoders that start with `c2.android`. More fundamentally, it's unclear what it means to be "hardware supported". In practice the line between fully-software and fully-hardware codecs is blurry, and it's unclear at what point a codec becomes "hardware supported".

If you want to do your own local check for `omx.google.` and `c2.android.` then you're probably going to get the answer you want, but it's an ad-hoc, best effort test, and I don't think we want or need to endorse it with an actual API.

ojw28: Note that it's also possible for a SoC or device vendor to implement a purely software codec with a different prefix. You're only really checking for Google implementations here. There is no 100% reliable way to determine whether a codec is hardware supported, for whatever definition of that you want to use :).

manishPatwari: @ojw28 Thanks alot for the comment. I made this changes based on this conversation https://github.com/google/ExoPlayer/issues/4757 . 

One question : When I am playing video in youtube it is selecting VP9. When I am playing Multi-Codec DASH file in the same device, its selecting H264 codec. Youtube also internally used ExoPlayer. Why Exoplayer is not picking the more optimised Codec when it has given the choice of VP9, H265 & H264 ?

ojw28: ExoPlayer is highly extensible, and YouTube have exploited this to customize behavior to suit there needs. So you shouldn't expect them to behave in exactly the same way.

It probably wont always make sense to select VP9 over H264. There are quite a few factors to consider. It's much harder for us to implement a generic track selection algorithm that'll be optimal for all use cases, than it is for YouTube to implement it for just their specific use case, with knowledge of things like the transcodes they generate etc. Hence for large streaming services it often makes sense to provide some custom components for ExoPlayer, such as a custom `TrackSelector`.

That said, there's probably room for improvement in our `DefaultTrackSelector` logic. For example if the not-quite-accurate hardware check indicates both codecs are supported by hardware, and all other things are equal, then it would indeed make sense to default to VP9 (ditto for H265). I've filed #4835 to track this.

PR #4882, opened 2018-09-27
Update flac extension to use more recent libflac and support newer NDK

Tolriq:
Update supported NDK to 17c and more recent
Update instructions and flags for lib flac 1.3.2 that is now out since a very longtime


erdemguven:
Thanks for the pull request. We'll increase libflac version by an internal change. Support for latest NDK will be a separate change but first we need to make sure other extension will work fine with the same NDK version.

Tolriq:
Not that I care about my name but would have been nicer to just ask to remove the NDK increase ;)

PR #4884, opened 2018-09-27
Call rating with extras

hacker1024:
This solves issue #4885.
I haven't tested this, so I might have made a syntax error somewhere, but this should work.


hacker1024:
I signed it!

ojw28:
See comment on the related issue: https://github.com/google/ExoPlayer/issues/4885

PR #4886, opened 2018-09-28
Make DefaultTrackSelector.AudioTrackScore protected

talklittle:
This allows selectAudioTrack() to be overridden.


AquilesCanta:
I am a bit inclined to think it should remain private, and subclasses should fork if necessary. 

Otherwise, when changing the implementation of selectAudioTrack, we might break the contract of this class. So what should have been an implementation change becomes an API change. 

If we don't care about eventually breaking the API, seems reasonable.

talklittle:
`selectAudioTrack` is already protected, and the API already had a breaking change (adding the AudioTrackScore to the return type) going from 2.8.4 to 2.9.0, so I don't think that's a concern. If it's not meant to be overridden, then `selectAudioTrack` should have been marked `protected final` or `private` originally.

AquilesCanta:
Oh right. We now return the score.

PR #4910, opened 2018-10-04
Add support for groupId in Format class

BetweenTwoBits:
In HLS, we can end up with text and audio tracks that have duplicated names, which yield non-unique ids in the format object. This PR adds support for the (already parsed) groupId. Because the spec defines the group as requiring unique names, the hope is that between groupId and id (within format) tracks may then have a unique id.

```
4.4.4.1.1.  Rendition Groups
   A set of one or more EXT-X-MEDIA tags with the same GROUP-ID value
   and the same TYPE value defines a Group of Renditions.  Each member
   of the Group MUST be an alternative Rendition of the same content;
   otherwise, playback errors can occur.
   All EXT-X-MEDIA tags in a Playlist MUST meet the following
   constraints:
   o  All EXT-X-MEDIA tags in the same Group MUST have different NAME
      attributes.
   o  A Group MUST NOT have more than one member with a DEFAULT
      attribute of YES.
   o  Each EXT-X-MEDIA tag with an AUTOSELECT=YES attribute SHOULD have
      a combination of LANGUAGE [RFC5646], ASSOC-LANGUAGE, FORCED, and
      CHARACTERISTICS attributes that is distinct from those of other
      AUTOSELECT=YES members of its Group.
```

### DIFF
--- a/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
+++ b/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
@@ -277,6 +277,7 @@ public final class FlacExtractor implements Extractor {
             /* drmInitData= */ null,
             /* selectionFlags= */ 0,
             /* language= */ null,
+            /* groupId= */ null,
             isId3MetadataDisabled ? null : id3Metadata);
     trackOutput.format(mediaFormat);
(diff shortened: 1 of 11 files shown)


BetweenTwoBits:
Closing in favor of:https://github.com/google/ExoPlayer/pull/4911

PR #4916, opened 2018-10-05
Add test for setForceHighestSupportedBitrate in DefaultTrackSelector

BetweenTwoBits:
There is already a test for setForceLowestBitrate(boolean forceLowestBitrate). This PR adds a test for the opposing method setForceHighestSupportedBitrate(boolean forceHighestSupportedBitrate).

### DIFF
--- a/library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
+++ b/library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
@@ -1061,6 +1061,9 @@ public final class DefaultTrackSelectorTest {
       throws Exception {
     trackSelector.setParameters(new ParametersBuilder().setForceLowestBitrate(true).build());
 
+    Format exceedsBitrateFormat =
+        Format.createAudioSampleFormat("exceedsBitrateFormat", MimeTypes.AUDIO_AAC, null, 5000,
+                    Format.NO_VALUE, 2, 44100, null, null, 0, null);
     Format lowerBitrateFormat =
         Format.createAudioSampleFormat("lowerBitrateFormat", MimeTypes.AUDIO_AAC, null, 15000,
             Format.NO_VALUE, 2, 44100, null, null, 0, null);
@@ -1068,10 +1071,17 @@ public final class DefaultTrackSelectorTest {
         Format.createAudioSampleFormat("higherBitrateFormat", MimeTypes.AUDIO_AAC, null, 30000,
             Format.NO_VALUE, 2, 44100, null, null, 0, null);
 
+    Map<String, Integer> mappedCapabilities = new HashMap<>();
+    mappedCapabilities.put(exceedsBitrateFormat.id, FORMAT_EXCEEDS_CAPABILITIES);
+    mappedCapabilities.put(lowerBitrateFormat.id, FORMAT_HANDLED);
+    mappedCapabilities.put(higherBitrateFormat.id, FORMAT_HANDLED);
+    RendererCapabilities mappedAudioRendererCapabilities =
+            new FakeMappedRendererCapabilities(C.TRACK_TYPE_AUDIO, mappedCapabilities);
+
     TrackSelectorResult result =
         trackSelector.selectTracks(
-            new RendererCapabilities[] {ALL_AUDIO_FORMAT_SUPPORTED_RENDERER_CAPABILITIES},
-            singleTrackGroup(lowerBitrateFormat, higherBitrateFormat),
+            new RendererCapabilities[] {mappedAudioRendererCapabilities},
+            singleTrackGroup(exceedsBitrateFormat, lowerBitrateFormat, higherBitrateFormat),
             periodId,
             TIMELINE);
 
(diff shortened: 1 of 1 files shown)


library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
```diff
@@ -1078,6 +1078,33 @@ public void testSelectTracksWithinCapabilitiesAndForceLowestBitrateSelectLowerBi
...
+                    Format.NO_VALUE, 2, 44100, null, null, 0, null);
+
+    TrackSelectorResult result =
+            trackSelector.selectTracks(
+                    new RendererCapabilities[] {ALL_AUDIO_FORMAT_SUPPORTED_RENDERER_CAPABILITIES},
+                    singleTrackGroup(lowerBitrateFormat, higherBitrateFormat),
+                    periodId,
+                    TIMELINE);
+
+    assertThat(result.selections.get(0).getSelectedFormat()).isEqualTo(higherBitrateFormat);
```
tonihei, line 1126: You should probably add an assertion that only one track is in the selection:
assertThat(result.selections.get(0).length()).isEqualTo(1);

Or to put it the other way round: The test should fail if you comment out the first line (which I think is currently not the case).

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
```diff
@@ -1078,6 +1078,33 @@ public void testSelectTracksWithinCapabilitiesAndForceLowestBitrateSelectLowerBi
     assertThat(result.selections.get(0).getSelectedFormat()).isEqualTo(lowerBitrateFormat);
   }
 
+  /**
+   * Tests that track selector will select audio tracks with higher bitrate when {@link Parameters}
+   * indicate highest bitrate preference, even when tracks are within capabilities.
```
tonihei: The "even when tracks are within capabilities" part doesn't make much sense here. Maybe say:
"Tests that track selector will select the highest supported audio track when {@link Parameters#forceHighestSupportedBitrate} is set."

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
```diff
@@ -1078,6 +1078,33 @@ public void testSelectTracksWithinCapabilitiesAndForceLowestBitrateSelectLowerBi
...
+   */
+  @Test
+  public void testSelectTracksWithinCapabilitiesAndForceHighestBitrateSelectHigherBitrate()
+          throws Exception {
+    trackSelector.setParameters(new ParametersBuilder().setForceHighestSupportedBitrate(true).build());
+
+    Format lowerBitrateFormat =
+            Format.createAudioSampleFormat("audioFormat", MimeTypes.AUDIO_AAC, null, 15000,
+                    Format.NO_VALUE, 2, 44100, null, null, 0, null);
+    Format higherBitrateFormat =
```
tonihei, line 1104: Would be nice to add a third format (with an even higher bitrate) which exceeds the renderer capabilities. See for example testSelectTracksPreferTrackWithinCapabilities for how to configure that.

tonihei (commented):
Thanks for the test! Looks good, just some small improvement ideas.

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
BetweenTwoBits: Hmm, should this change also be made for `testSelectTracksWithinCapabilitiesAndForceLowestBitrateSelectLowerBitrate`?

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
BetweenTwoBits: This change also seems applicable to the other test.

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
tonihei: The tracks will be sorted by decending bitrate, so having the lowest bitrate as the first track means there can be only one track. That's very non-obvious though and I agree we should add that to testSelectTracksWithinCapabilitiesAndForceLowestBitrateSelectLowerBitrate as well.

library/core/src/test/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelectorTest.java
tonihei: Yes, the other comment can be improved too. 

Aside: We generally don't provide javadocs for tests as long as the test name itself is descriptive enough. This class seems to be an exception to that rule :)

BetweenTwoBits:
All changes made(in highest and lowest bitrate scenarios). I can't help but notice the naming is slightly different between these methods:
`setForceLowestBitrate`
`setForceHighestSupportedBitrate`

Is there a reason for that difference or could I also throw in: `setForceLowestBitrate` -> `setForceLowestSupportedBitrate`

tonihei:
Nice, thanks! 

The reason for the different naming is that we had `setForceLowestBitrate` only until recently. And in almost all cases the formats at the lower end won't exceed renderer capabilities - so the distinction between just "lowest bitrate" and "lowest supported bitrate" is not really relevant.

When we added  `setForceHighestSupportedBitrate`, we specifically needed to include the "supported" bit because otherwise it would be unclear if it would force the selection to go beyond the renderer capabilities or not. For consistency, I agree that it would be clearer to rename the other parameter as well. But keep in mind that people are already using it and we would break existing implementation unnecessarily.

BetweenTwoBits:
Ah, that all makes sense. Thanks for the review!

PR #4930, opened 2018-10-10
Add Support for Parsing ProgramInformation

jlieblich2:
This adds support for parsing ProgramInformation, its defined elements of Title/Source/Copyright, and custom elements as defined in the DASH spec.

- There is a `ByteArrayOutputStream` instantiated elsewhere with a parameter of 512 passed in. Should this value be used for parsing this out or is there a method of determining the size?

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -982,10 +982,9 @@ public class DashManifestParser extends DefaultHandler
   }
 
   protected ProgramInformation parseProgramInformation(XmlPullParser xpp) throws IOException, XmlPullParserException {
-    String title = "";
-    String source = "";
-    String copyright = "";
-    List<byte[]> customEvents = new ArrayList<>();
+    String title = null;
+    String source = null;
+    String copyright = null;
     do {
       xpp.next();
       if (XmlPullParserUtil.isStartTag(xpp, "Title")) {
@@ -994,28 +993,9 @@ public class DashManifestParser extends DefaultHandler
         source = xpp.getText();
       } else if (XmlPullParserUtil.isStartTag(xpp, "Copyright")) {
         copyright = xpp.getText();
-      } else {
-        byte[] customElement = parseCustomElement(xpp, new ByteArrayOutputStream(512));
-        if (customElement.length > 0) {
-          customEvents.add(customElement);
-        }
       }
     } while (!XmlPullParserUtil.isEndTag(xpp, "ProgramInformation"));
-    return new ProgramInformation(title, source, copyright, customEvents);
-  }
-
-  private byte[] parseCustomElement(XmlPullParser xpp, ByteArrayOutputStream outputStream) throws IOException, XmlPullParserException {
-    XmlSerializer serializer = Xml.newSerializer();
-    serializer.setOutput(outputStream, C.UTF8_NAME);
-    if (xpp.getEventType() == XmlPullParser.START_TAG) {
-      serializer.startTag(xpp.getNamespace(), xpp.getName());
-      for (int i = 0; i < xpp.getAttributeCount(); i++) {
-        serializer.attribute(xpp.getAttributeNamespace(i), xpp.getAttributeName(i), xpp.getAttributeValue(i));
-      }
-      serializer.endTag(xpp.getNamespace(), xpp.getName());
-    }
-    serializer.flush();
-    return outputStream.toByteArray();
+    return new ProgramInformation(title, source, copyright);
   }
 
   // AudioChannelConfiguration parsing.
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -985,17 +985,19 @@ public class DashManifestParser extends DefaultHandler
     String title = null;
     String source = null;
     String copyright = null;
+    String moreInformationURL = parseString(xpp, "moreInformationURL", null);
+    String lang = parseString(xpp, "lang", null);
     do {
       xpp.next();
       if (XmlPullParserUtil.isStartTag(xpp, "Title")) {
-        title = xpp.getText();
+        title = xpp.nextText();
       } else if (XmlPullParserUtil.isStartTag(xpp, "Source")) {
-        source = xpp.getText();
+        source = xpp.nextText();
       } else if (XmlPullParserUtil.isStartTag(xpp, "Copyright")) {
-        copyright = xpp.getText();
+        copyright = xpp.nextText();
       }
     } while (!XmlPullParserUtil.isEndTag(xpp, "ProgramInformation"));
-    return new ProgramInformation(title, source, copyright);
+    return new ProgramInformation(title, source, copyright, moreInformationURL, lang);
   }
 
   // AudioChannelConfiguration parsing.
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -122,6 +122,7 @@ public class DashManifestParser extends DefaultHandler
     long publishTimeMs = parseDateTime(xpp, "publishTime", C.TIME_UNSET);
     UtcTimingElement utcTiming = null;
     Uri location = null;
+    ProgramInformation programInformation = null;
 
     List<Period> periods = new ArrayList<>();
     long nextPeriodStartMs = dynamic ? C.TIME_UNSET : 0;
@@ -138,6 +139,8 @@ public class DashManifestParser extends DefaultHandler
         utcTiming = parseUtcTiming(xpp);
       } else if (XmlPullParserUtil.isStartTag(xpp, "Location")) {
         location = Uri.parse(xpp.nextText());
+      } else if (XmlPullParserUtil.isStartTag(xpp, "ProgramInformation")) {
+        programInformation = parseProgramInformation(xpp);
       } else if (XmlPullParserUtil.isStartTag(xpp, "Period") && !seenEarlyAccessPeriod) {
         Pair<Period, Long> periodWithDurationMs = parsePeriod(xpp, baseUrl, nextPeriodStartMs);
         Period period = periodWithDurationMs.first;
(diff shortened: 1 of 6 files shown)


library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -978,6 +981,43 @@ protected RangedUri buildRangedUri(String urlText, long rangeStart, long rangeLe
     return new RangedUri(urlText, rangeStart, rangeLength);
   }
 
+  protected ProgramInformation parseProgramInformation(XmlPullParser xpp) throws IOException, XmlPullParserException {
```
ojw28, line 1004: Why not parse the `lang` and `moreInformationURL` attributes defined in the spec, and put those into the `ProgramInformation` as well?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -978,6 +981,43 @@ protected RangedUri buildRangedUri(String urlText, long rangeStart, long rangeLe
...
+      if (XmlPullParserUtil.isStartTag(xpp, "Title")) {
+        title = xpp.getText();
+      } else if (XmlPullParserUtil.isStartTag(xpp, "Source")) {
+        source = xpp.getText();
+      } else if (XmlPullParserUtil.isStartTag(xpp, "Copyright")) {
+        copyright = xpp.getText();
+      } else {
+        byte[] customElement = parseCustomElement(xpp, new ByteArrayOutputStream(512));
+        if (customElement.length > 0) {
+          customEvents.add(customElement);
```
ojw28: Can we consistently use "event" or "element"? I think "element" is the correct one. Ditto in `ProgramInformation.java`, which is using "event" currently.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -978,6 +981,43 @@ protected RangedUri buildRangedUri(String urlText, long rangeStart, long rangeLe
...
+        }
+      }
+    } while (!XmlPullParserUtil.isEndTag(xpp, "ProgramInformation"));
+    return new ProgramInformation(title, source, copyright, customEvents);
+  }
+
+  private byte[] parseCustomElement(XmlPullParser xpp, ByteArrayOutputStream outputStream) throws IOException, XmlPullParserException {
```
ojw28: This implementation seems to assume that a custom element wont contain anything other than attributes. Is this guaranteed anywhere? If not, I'm not sure this is the right approach.

More generally, do you think handling custom elements in a generic way is really necessary? It would be pretty easy for an app to extend `DashManifestParser` and override `parseProgramInformation` to parse the specific the custom elements they want to look at, returning an object that extends `ProgramInformation` to add in any specific additional fields.  The app would need to cast `manifest.programInformation` to get access to the extra fields, but that still seems cleaner than having to have app code retrieve and parse byte arrays. It's also simpler than attempting a generic solution, and naturally extends to custom attributes as well.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
```diff
@@ -0,0 +1,67 @@
+package com.google.android.exoplayer2.source.dash.manifest;
```
ojw28, line 16: Please add the copyright header as found in other source files (with the year set to 2018 obviously :)).

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -978,6 +981,43 @@ protected RangedUri buildRangedUri(String urlText, long rangeStart, long rangeLe
     return new RangedUri(urlText, rangeStart, rangeLength);
   }
 
+  protected ProgramInformation parseProgramInformation(XmlPullParser xpp) throws IOException, XmlPullParserException {
+    String title = "";
```
ojw28: Use null as defaults for these?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
```diff
@@ -0,0 +1,67 @@
...
+    for (int i = 0; i < customEvents.size(); i++) {
+      result = 31 * result + Arrays.hashCode(customEvents.get(i));
+    }
+    return result;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof ProgramInformation
+            && ((ProgramInformation)obj).title.equals(this.title)
```
ojw28: This code seems a bit confused about whether the fields can be null or not. `hashCode` assumes they can be null, but `equals` will crash if they are.  Allowing null conveys slightly more information, since it distinguishes between an element being present in the manifest as an empty string, and it being omitted entirely.

Notes:
* You can use `Util.areEqual(other.title, this.title)`, and it'll handle nulls for you.
* Can we only cast once? It's kinda ugly casting on every line :).

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
```diff
@@ -0,0 +1,67 @@
...
+            && ((ProgramInformation)obj).source.equals(this.source)
+            && ((ProgramInformation)obj).copyright.equals(this.copyright)
+            && validateEvents(((ProgramInformation)obj).customEvents);
+  }
+
+  private boolean validateEvents(List<byte[]> customEvents) {
```
ojw28: This looks like it always returns true if the two lists differ in length, with the shorter one matching the longer one up to its length. That doesn't seem right though. The method should probably compare their lengths as the first thing it does, and immediately return false if they differ.

Although see comment elsewhere about getting rid of generic custom element handling.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
```diff
@@ -0,0 +1,67 @@
...
+    for (int i = 0; i < customEvents.size() && i < this.customEvents.size(); i++) {
```
ojw28: I think you can possibly simplify the whole body of this for loop to:
```
if (!Arrays.equals(customElements.get(i), otherElements.get(i))) {
  return false;
}
```
Although see comment elsewhere about getting rid of generic custom element handling.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
jlieblich2: I completely glazed over the attributes section of the spec, I'll go ahead and add those.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
jlieblich2: The custom element in the spec doesn't explicitly state that there will be no content within the `ProgramInformation` tags, so no it isn't guaranteed. This was supposed to be parsed similarly to how `parseEventObject` works within this class. I do think that you're correct in saying that this is not the best way of doing this and will make a change to only parse elements/attributes explicitly defined in the spec.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
jlieblich2: Yea, I am following your advice from the other comment so this is going away.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
jlieblich2: This is actually going away with custom elements being removed since there is no longer any need to override an object with just Strings.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
jlieblich2: Ah good catch, I'm used to AS adding it for me.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
jlieblich2: With custom elements removed, this change is no longer needed.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
jlieblich2: Scratch that, unit tests revealed why it's still needed.

jlieblich2:
@ojw28 I've made the requested changes, let me know if there's anything else that can be improved.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/ProgramInformation.java
```diff
@@ -0,0 +1,77 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.exoplayer2.source.dash.manifest;
+
+import com.google.android.exoplayer2.util.Util;
+
+public class ProgramInformation {
```
ojw28, line 23: Please add Javadoc

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -978,6 +981,25 @@ protected RangedUri buildRangedUri(String urlText, long rangeStart, long rangeLe
...
+    String lang = parseString(xpp, "lang", null);
+    do {
+      xpp.next();
+      if (XmlPullParserUtil.isStartTag(xpp, "Title")) {
+        title = xpp.nextText();
+      } else if (XmlPullParserUtil.isStartTag(xpp, "Source")) {
+        source = xpp.nextText();
+      } else if (XmlPullParserUtil.isStartTag(xpp, "Copyright")) {
+        copyright = xpp.nextText();
+      }
```
ojw28, line 1020: When reviewing this change I realized there was a general problem in this parser where we'd potentially still process tags nested inside tags we don't recognize. Like:
```
<ProgramInformation>
  <Title>Blah</Title>
  <SomeCustomThing>
    <Title/>                                      <-- We don't want to be parsing this :)
  </SomeCustomThing>
</ProgramInformation>
```
I fixed this in https://github.com/google/ExoPlayer/commit/d511370338415733a1b9a40a23c3e9837dd5ed7d. Please add an
```
else {
  maybeSkipTag(xpp);
}
```
to the end of this block to fix it here as well. Once that's done I think we're good to merge; thanks!

jlieblich2: I didn't even consider that something like that is technically in line with the spec, we should be good to go now.

ojw28: Neither did I until now :). Thanks; we'll get this merged soon.

Issue #4973, opened 2018-10-17
Null pointer exception when used extension ima


ummerakbar45:
### Issue description
Null pointer exception

### Reproduction steps
Added
  implementation 'com.google.android.exoplayer:extension-ima:2.9.0'

### Link to test content
No link available

### Version of ExoPlayer being used
implementation 'com.google.android.exoplayer:exoplayer:2.9.0'
    implementation 'com.google.android.exoplayer:extension-ima:2.9.0'

### Device(s) and version(s) of Android being used
I tested only on mi a5 and version 8

### A full bug report captured from the device
   Process: com.recipe.filmrise, PID: 10340
    java.lang.NullPointerException: Attempt to invoke virtual method 'void com.google.android.exoplayer2.source.ads.AdsMediaSource$ComponentListener.release()' on a null object reference
        at com.google.android.exoplayer2.source.ads.AdsMediaSource.releaseSourceInternal(AdsMediaSource.java:391)
        at com.google.android.exoplayer2.source.BaseMediaSource.releaseSource(BaseMediaSource.java:165)
        at com.google.android.exoplayer2.ext.ima.ImaAdsMediaSource.releaseSourceInternal(ImaAdsMediaSource.java:105)
        at com.google.android.exoplayer2.source.BaseMediaSource.releaseSource(BaseMediaSource.java:165)
        at com.google.android.exoplayer2.source.CompositeMediaSource.releaseSourceInternal(CompositeMediaSource.java:70)
        at com.google.android.exoplayer2.source.ConcatenatingMediaSource.releaseSourceInternal(ConcatenatingMediaSource.java:509)
        at com.google.android.exoplayer2.source.BaseMediaSource.releaseSource(BaseMediaSource.java:165)
        at com.google.android.exoplayer2.ExoPlayerImplInternal.resetInternal(ExoPlayerImplInternal.java:815)
        at com.google.android.exoplayer2.ExoPlayerImplInternal.stopInternal(ExoPlayerImplInternal.java:734)
        at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:358)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:172)
        at android.os.HandlerThread.run(HandlerThread.java:65)

BetweenTwoBits:
Couple of questions on this.
[...]

PR #4974, opened 2018-10-17
[ AdsMediaSource ] Prevent releasing already null listener

BetweenTwoBits:
This is a fix for issue https://github.com/google/ExoPlayer/issues/4973

Currently if release is called before prepare, componentListener will be null. This is probably a state people shouldn't ever find themselves in, but better safe than sorry. 😄

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ads/AdsMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ads/AdsMediaSource.java
@@ -390,8 +390,10 @@ public final class AdsMediaSource extends CompositeMediaSource<MediaPeriodId> {
   @Override
   public void releaseSourceInternal() {
     super.releaseSourceInternal();
-    componentListener.release();
-    componentListener = null;
+    if (componentListener != null) {
+      componentListener.release();
+      componentListener = null;
+    }
     deferredMediaPeriodByAdMediaSource.clear();
     contentTimeline = null;
     contentManifest = null;


andrewlewis:
Is there a case where the source gets released without being prepared first in correct usage? I wonder whether the referenced bug is caused by the application releasing the source, or whether somehow the prepare/release calls inside the player are not matched.

BetweenTwoBits:
Seems like `ExoPlayerImplInternal` has a method called `prepareInternal` which calls `resetInternal` with a flag `releaseMediaSource` (which is set to true), this calls `releaseSource`. This all happens before mediaSource.prepareSource is called in the same `prepareInternal` method.

I think this might be the root of the issue.

<pre>
private void prepareInternal(MediaSource mediaSource, boolean resetPosition, boolean resetState) {
    pendingPrepareCount++;
    <b>resetInternal(/* releaseMediaSource= */ true, resetPosition, resetState);</b>
    loadControl.onPrepared();
    this.mediaSource = mediaSource;
    setState(Player.STATE_BUFFERING);
    <b>mediaSource.prepareSource(
        player,
        /* isTopLevelSource= */ true,
        /* listener= */ this,
        bandwidthMeter.getTransferListener());</b>
    handler.sendEmptyMessage(MSG_DO_SOME_WORK);
  }
</pre>

<pre>
  private void resetInternal(boolean releaseMediaSource, boolean resetPosition, boolean resetState) {
    .
    .
    .
    if (releaseMediaSource) {
      if (mediaSource != null) {
        mediaSource.releaseSource(/* listener= */ this);
        mediaSource = null;
      }
    }
  }
</pre>

BetweenTwoBits:
Link to the line in question: https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java#L397

andrewlewis:
Whenever the `mediaSource` field gets assigned we prepare the media source. So `mediaSource` will be null in `resetInternal` unless it was previously prepared, right?

BetweenTwoBits:
Ah I see what you mean, possibly was just a problem on the application side then.

ojw28:
@andrewlewis - Can we decide what to do with this?

andrewlewis:
It seems likely that the original issue was an application bug rather than an ExoPlayer issue, and the error case addressed in this PR seems useful as a way to catch unexpected behavior (and there's at least one other case (`HlsMediaSource`) where it looks like we assume release only happens after prepare), so I suggest we close this.

PR #4975, opened 2018-10-17
[ ImaAdsLoader ] Prevent acting on null player in detach method

BetweenTwoBits:
Similar to https://github.com/google/ExoPlayer/pull/4974
Currently if detachPlayer is called before attachPlayer, player will be null. This is probably a state people shouldn't ever find themselves in, but better safe than sorry. 😄


ojw28:
@andrewlewis - Can we decide what to do with this?

andrewlewis:
Similar to #4974, I'm not aware of any way we can detach the loader without attaching at the moment and IMO there is value in failing early in this case. Please post a comment if there are cases where this can happen and I'll reopen the PR. Thanks!

PR #4996, opened 2018-10-23
fix checkIdleRequirement

YukiMatsumura:
I think that the conditional expression is reversed if you want to download it in the IDLE state.

https://developer.android.com/reference/android/os/PowerManager#isDeviceIdleMode()


YukiMatsumura:
I signed it!

PR #5004, opened 2018-10-23
Provide http status message to InvalidResponseCodeException

pakerfeldt:
When response code alone is not enough to distinguish error responses, the http status message is helpful.

### DIFF
--- a/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
+++ b/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
@@ -327,7 +327,7 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
     int responseCode = responseInfo.getHttpStatusCode();
     if (responseCode < 200 || responseCode > 299) {
       InvalidResponseCodeException exception = new InvalidResponseCodeException(responseCode,
-          responseInfo.getAllHeaders(), currentDataSpec);
+          responseInfo.getHttpStatusText(), responseInfo.getAllHeaders(), currentDataSpec);
       if (responseCode == 416) {
         exception.initCause(new DataSourceException(DataSourceException.POSITION_OUT_OF_RANGE));
       }
@@ -611,7 +611,8 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
         // The industry standard is to disregard POST redirects when the status code is 307 or 308.
         if (responseCode == 307 || responseCode == 308) {
           exception =
-              new InvalidResponseCodeException(responseCode, info.getAllHeaders(), currentDataSpec);
+              new InvalidResponseCodeException(responseCode, info.getHttpStatusText(),
+                      info.getAllHeaders(), currentDataSpec);
           operation.open();
           return;
         }
--- a/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
+++ b/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
@@ -326,8 +326,12 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
     // Check for a valid response code.
     int responseCode = responseInfo.getHttpStatusCode();
     if (responseCode < 200 || responseCode > 299) {
-      InvalidResponseCodeException exception = new InvalidResponseCodeException(responseCode,
-          responseInfo.getHttpStatusText(), responseInfo.getAllHeaders(), currentDataSpec);
+      InvalidResponseCodeException exception =
+          new InvalidResponseCodeException(
+              responseCode,
+              responseInfo.getHttpStatusText(),
+              responseInfo.getAllHeaders(),
+              currentDataSpec);
       if (responseCode == 416) {
         exception.initCause(new DataSourceException(DataSourceException.POSITION_OUT_OF_RANGE));
       }
@@ -611,8 +615,11 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
         // The industry standard is to disregard POST redirects when the status code is 307 or 308.
         if (responseCode == 307 || responseCode == 308) {
           exception =
-              new InvalidResponseCodeException(responseCode, info.getHttpStatusText(),
-                      info.getAllHeaders(), currentDataSpec);
+              new InvalidResponseCodeException(
+                  responseCode,
+                  info.getHttpStatusText(),
+                  info.getAllHeaders(),
+                  currentDataSpec);
           operation.open();
           return;
         }
--- a/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
+++ b/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
@@ -172,8 +172,8 @@ public class OkHttpDataSource extends BaseDataSource implements HttpDataSource {
     if (!response.isSuccessful()) {
       Map<String, List<String>> headers = response.headers().toMultimap();
       closeConnectionQuietly();
-      InvalidResponseCodeException exception = new InvalidResponseCodeException(
-          responseCode, response.message(), headers, dataSpec);
+      InvalidResponseCodeException exception =
+          new InvalidResponseCodeException(responseCode, response.message(), headers, dataSpec);
       if (responseCode == 416) {
         exception.initCause(new DataSourceException(DataSourceException.POSITION_OUT_OF_RANGE));
       }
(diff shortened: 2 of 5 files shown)


pakerfeldt:
I ran `./gradlew test` and `./gradlew lint`. Please let me know if there's any additional checks/tests I should run.

library/core/src/test/java/com/google/android/exoplayer2/upstream/DefaultLoadErrorHandlingPolicyTest.java
```diff
@@ -34,23 +34,23 @@
...
+        new InvalidResponseCodeException(410, "Gone", Collections.emptyMap(), new DataSpec(Uri.EMPTY));
     assertThat(getDefaultPolicyBlacklistOutputFor(exception))
         .isEqualTo(DefaultLoadErrorHandlingPolicy.DEFAULT_TRACK_BLACKLIST_MS);
   }
 
   @Test
   public void getBlacklistDurationMsFor_dontBlacklistUnexpectedHttpCodes() {
     InvalidResponseCodeException exception =
-        new InvalidResponseCodeException(500, Collections.emptyMap(), new DataSpec(Uri.EMPTY));
+        new InvalidResponseCodeException(500, "Internal Server Error", Collections.emptyMap(), new DataSpec(Uri.EMPTY));
```
tonihei:
```suggestion
        new InvalidResponseCodeException(
            500, "Internal Server Error", Collections.emptyMap(), new DataSpec(Uri.EMPTY));
```

library/core/src/main/java/com/google/android/exoplayer2/upstream/HttpDataSource.java
```diff
@@ -294,15 +295,22 @@ public InvalidContentTypeException(String contentType, DataSpec dataSpec) {
...
+     */
+    public final @Nullable String responseMessage;
+
     /**
      * An unmodifiable map of the response header fields and values.
      */
     public final Map<String, List<String>> headerFields;
 
-    public InvalidResponseCodeException(int responseCode, Map<String, List<String>> headerFields,
+    public InvalidResponseCodeException(int responseCode, @Nullable String responseMessage,
```
tonihei:
```suggestion
    public InvalidResponseCodeException(
        int responseCode,
        @Nullable String responseMessage,
```

library/core/src/main/java/com/google/android/exoplayer2/upstream/HttpDataSource.java
```diff
@@ -294,15 +295,22 @@ public InvalidContentTypeException(String contentType, DataSpec dataSpec) {
...
+                                        Map<String, List<String>> headerFields,
```
tonihei:
```suggestion
        Map<String, List<String>> headerFields,
```

library/core/src/main/java/com/google/android/exoplayer2/upstream/HttpDataSource.java
```diff
@@ -294,15 +295,22 @@ public InvalidContentTypeException(String contentType, DataSpec dataSpec) {
      */
     public final int responseCode;
 
+    /**
+     * The HTTP status message.
+     */
+    public final @Nullable String responseMessage;
```
tonihei:
```suggestion
    @Nullable public final String responseMessage;
```

extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
```diff
@@ -173,7 +173,7 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
       Map<String, List<String>> headers = response.headers().toMultimap();
       closeConnectionQuietly();
       InvalidResponseCodeException exception = new InvalidResponseCodeException(
```
tonihei:
```suggestion
      InvalidResponseCodeException exception =
          new InvalidResponseCodeException(responseCode, response.message(), headers, dataSpec);
```

extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
```diff
@@ -173,7 +173,7 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
...
-          responseCode, headers, dataSpec);
+          responseCode, response.message(), headers, dataSpec);
```
tonihei:
```suggestion
```

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -611,7 +611,8 @@ public synchronized void onRedirectReceived(
         // The industry standard is to disregard POST redirects when the status code is 307 or 308.
         if (responseCode == 307 || responseCode == 308) {
           exception =
-              new InvalidResponseCodeException(responseCode, info.getAllHeaders(), currentDataSpec);
+              new InvalidResponseCodeException(responseCode, info.getHttpStatusText(),
```
tonihei:
```suggestion
              new InvalidResponseCodeException(
                  responseCode,
                  info.getHttpStatusText(),
```

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -611,7 +611,8 @@ public synchronized void onRedirectReceived(
...
+                      info.getAllHeaders(), currentDataSpec);
```
tonihei:
```suggestion
                      info.getAllHeaders(),
                      currentDataSpec);
```

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -327,7 +327,7 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
     int responseCode = responseInfo.getHttpStatusCode();
     if (responseCode < 200 || responseCode > 299) {
       InvalidResponseCodeException exception = new InvalidResponseCodeException(responseCode,
```
tonihei:
```suggestion
      InvalidResponseCodeException exception =
          new InvalidResponseCodeException(
              responseCode,
```

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -327,7 +327,7 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
...
-          responseInfo.getAllHeaders(), currentDataSpec);
+          responseInfo.getHttpStatusText(), responseInfo.getAllHeaders(), currentDataSpec);
```
tonihei:
```suggestion
              responseInfo.getHttpStatusText(),
              responseInfo.getAllHeaders(),
              currentDataSpec);
```

tonihei (commented):
Looks good. I added some formatting changes to adhere to our style guide.

library/core/src/main/java/com/google/android/exoplayer2/upstream/HttpDataSource.java
pakerfeldt: I was actually looking at `OkHttpDataSource` to put the `@Nullable` at the right position :)

pakerfeldt:
Do you want me to squash the changes into one commit?

tonihei:
He. Wasn't expecting that ;) May be easier if it's not too much work. There's also still one wrong indentation at L621/622 of CronetDataSource.

pakerfeldt:
Not at all, already done!

tonihei:
Thanks. Good to be merged now :)

pakerfeldt:
@tonihei Turns out we have some other use cases where the body is actually needed after all. The way we solve this today is by having our own derivative of `OkHttpDataSource` and `InvalidResponseCodeException` which really isn't optimal. While this PR took us almost all the way, we still need the response body accessible when getting an invalid response code. What's your opinion on this?

tonihei:
Sorry for the delay. The approach sounds good to me. I just wonder if it's getting too implementation specific or if that's something people may generally want to have in there. Given that you need to fork OkHttpDataSource just to inject your own implementation here suggests we should just add the body too.

+ @AquilesCanta I'm not really an expert for typical http response bodies. Can you comment on whether you'd think that's a common enough feature to add?

pakerfeldt:
To give some context. We have two different use-cases we need to support.
1) When user has turned their app in offline mode we rely on OkHttp interceptors to force cache-only. We want to figure out that playback failed because app is in offline mode and we're doing this by looking at the http response code & status which says 504 `Unsatisfiable Request (only-if-cached)`.
2) We also use the OkHttpDataSource for downloading our manifest files and this request can fail with various error codes which we need to read out from the body.

It would be really good to get rid of our OkHttpDataSource/Factory derivates.

AquilesCanta:
I don't see any reason to hide a piece of the response from the client.

However, reading the body will probably be harder to implement than just providing what's in the header. We should address this change in all DataSource implementations.

ojw28:
`Util.toByteArray` will probably help with some of the implementations. What would the implementations do if an error occurs whilst reading the body of the non-success response? Swallow the exception and not set the body in the exception at all?

pakerfeldt:
Swallow the exception is what we do in our derivative.

tonihei:
@pakerfeldt Motivated to do another pull request for that? Probably worth giving a quick summary of the planned changes before starting to implement them.

pakerfeldt:
Sure, I'll get back to you when I've had time to give this some thought.

PR #5058, opened 2018-11-05
keep hls chunk source from getting redundant playlists

TakuSemba:
### Overview

when playing **live** hls stream, ExoPlayer periodically getting all playlists which has been loaded at least once. Because of this, ExoPlayer keeps sending network requests more than necessary.

#### since version 2.9.0

[This](https://abclive1-lh.akamaihd.net/i/abc_live10@420897/master.m3u8) is the url that I tested, and it is reproducible on ExoPlayer demo. 

You will see ExoPlayer keeps fetching all variants of mediaPlaylist which ExoPlayer has loaded during the playback.

### Cause 

Inside the HlsChunkSource, HlsChunkSource#createMediaChunkIterators() provides MediaChunkIterator[] that contains information for upcoming media chunks. However, in `createMediaChunkIterators()`, playlistTracker.getPlaylistSnapshot(variantUrl) is called and it sends an network request to fetch an MediaPlaylist. It's called for every variant that has `playlistSnapshot`. 

I think unselected playlists should no be loaded through network calls from the viewpoint of reducing network requests.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -431,7 +431,8 @@ import java.util.List;
     for (int i = 0; i < chunkIterators.length; i++) {
       int variantIndex = trackSelection.getIndexInTrackGroup(i);
       HlsUrl variantUrl = variants[variantIndex];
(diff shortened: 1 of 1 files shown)


AquilesCanta:
This does look like a bug. But it needs to be fixed on the HlsPlaylistTracker side. I have filed #5059 to track its fix. Any pushed commits will be referenced there.

PR #5075, opened 2018-11-07
Cea608 fixes

zsmatyas:
I start refactoring the previously rejected "one big change" into smaller fixes, optimisations, fine tuning etc. 
Please add code review comments as early as possible to avoid complex rebasing and conflicts.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -179,6 +179,41 @@ public final class Cea608Decoder extends CeaDecoder {
     0xC5, 0xE5, 0xD8, 0xF8, 0x250C, 0x2510, 0x2514, 0x2518
   };
 
+  private static final boolean ODD_PARITY_BYTE_TABLE[] = {
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 0 - 7
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 8 - 15
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 16 - 23
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 24 - 31
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 32 - 39
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 40 - 47
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 48 - 55
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 56 - 63
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 64 - 71
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 72 - 79
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 80 - 87
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 88 - 95
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 96 - 103
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 104 - 111
+    true,	false,	false,	true,	false,	true,	true,	false,	// for values 112 - 119
+    false,	true,	true,	false,	true,	false,	false,	true,	// for values 120 - 127
+    false,	false,	true,	false,	false,	false,	false,	true,	// for values 128 - 135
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -284,6 +328,24 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
...
+        resetCueBuilders();
+        continue;
+      }
+
+      // When transmission is digital and reliable (TCP) we should not meet any data corruption.
+      // But any over-the-air content and test streams might give the decoder corrupted data, so we
+      // need to drop those. When corrupt data is met the screen needs to be cleared, so we:
+      // - reset the cue builders
+      // - update the screen (by having this check after having captionDataProcessed set above)
+      if (byte1ParityBit == ODD_PARITY_BYTE_TABLE[ccData1]
```
ojw28, line 343: I think it's equivalent and simpler to do:
```
if (!ODD_PARITY_BYTE_TABLE[ccFullByte1] || !ODD_PARITY_BYTE_TABLE[ccFullByte2])
```
or better:
```
if (EVEN_PARITY_BYTE_TABLE[ccFullByte1] || EVEN_PARITY_BYTE_TABLE[ccFullByte2])
```
and then get rid of `byte1ParityBit` and `byte2ParityBit`?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -284,6 +328,24 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
       // If we've reached this point then there is data to process; flag that work has been done.
       captionDataProcessed = true;
 
+      if ((ccDataHeader & CC_VALID_FLAG) != CC_VALID_FLAG) {
+        // The screen needs to be cleared when the closed captions are turned off
```
ojw28, line 332: Is this defined in a specification somewhere? The information I've managed to find suggests the pair should be ignored / treated as padding. Which is what the implementation did prior to this change.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -284,6 +328,24 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
...
+        // (the encoder flips the validity bit)
+        resetCueBuilders();
+        continue;
+      }
+
+      // When transmission is digital and reliable (TCP) we should not meet any data corruption.
+      // But any over-the-air content and test streams might give the decoder corrupted data, so we
+      // need to drop those. When corrupt data is met the screen needs to be cleared, so we:
+      // - reset the cue builders
```
ojw28, line 341: Is resetting definitely the best thing to do? What happens if we just ignore the corrupt pair? I would suspect the outcome may be preferable.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: * I agree, the standard is quite ambiguous and miss describing requirements.
* Seemingly some caption providers simply set this flag to "not valid" and do not put any other data or command into the stream. This is the case for live (over the air) TV, 608 is not used for current digital on-demand formats. Missing this would cause, for example, the last caption of the previous movie kept continuously on for an entire new movie without captions. As this new movie does not have any valid data, there is no other signs to clear the screen for. 
* The [Sarnoff](https://www.sri.com/engage/products-solutions/sarnoff-bitstreams-decoder-tests) CEA608-708 verification documents clearly say that as soon as the flag is set to "not valid", the screen must be empty. Unfortunately, these test streams are not free.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: good idea!

zsmatyas: Spec describes all valid bytes having odd parity. So I kept that version and added a comment about the requirement.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: Ok, although if this turns out to break anyone we will likely revert it to be in line with what the specification says.

PR #5126, opened 2018-11-20
fix(playlist): always call onCompletion when moving media sources

### LINKED ISSUE #5125: [PR offer] Always call onCompletion in ConcatenatingMediaSource.moveMediaSource
### Issue description
In `moveMediaSource` method in `ConcatenatingMediaSource` class, [at this line](https://github.com/google/ExoPlayer/blob/r2.9.0/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java#L365), when there's actually no source to move (current and new index are the same), the `onCompletion` action is **not** invoked:

```java
if (currentIndex == newIndex) {
    return;
}
```

That is somehow a different behavior than e.g. `removeMediaSourceRange`, where completion callback is invoked right away when there's nothing to remove.

Also, that forces callers to always check for this special scenario, basically requiring a wrapper around this specific operation.

Would you consider a PR changing `moveMediaSource` behavior, so that it always calls completion callback?

We understand this can be considered a breaking change in API, as client code in its own logic may rely on this *missing invocation*. Maybe you could make the new behavior as an opt-in, like an additional method signature on `ConcatenatingMediaSource`, e.g. `moveMediaSource(int currentIndex, int newIndex, @Nullable Runnable actionOnCompletion, bool alwaysComplete)`.

Thanks for your attention

### Reproduction steps
Not Applicable

### Link to test content
Not Applicable

2.9.0

Not Applicable

Not Applicable

BrainCrumbz:
Closes #5125

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ConcatenatingMediaSource.java
@@ -365,6 +365,9 @@ public class ConcatenatingMediaSource extends CompositeMediaSource<MediaSourceHo
   public final synchronized void moveMediaSource(
       int currentIndex, int newIndex, @Nullable Runnable actionOnCompletion) {
     if (currentIndex == newIndex) {
+      if (actionOnCompletion != null) {
+        actionOnCompletion.run();
+      }
       return;
     }
     mediaSourcesPublic.add(newIndex, mediaSourcesPublic.remove(currentIndex));


GiuseppePiscopo:
I'm ok with my commits being contributed to this project.

BrainCrumbz:
Next time we'll have to keep an eye on commit author (it's not enough to use it's public email, the same registered on Google CLA).

tonihei:
Looks good, thanks!

ojw28:
This was merged in the commit ref'd above. We're experimenting with a new way of syncing internal/external changes, and didn't get it quite right on this PR, which is why it's not showing as merged.

PR #5147, opened 2018-11-26
CEA608: PAINT-ON Mode must keep the last shown captions on the screen

zsmatyas:
## Problem
PAINT-ON mode is not implemented. From compliance tests:
* RDC command has no effect except to select paint-on style.
* Next data are written directly to the display upon receipt.
* If other captioning is already on the screen, it stays on
* The four-row limit is still in effect.

## Solution
It is a rare use case, we do not support overriding characters in
existing cueBuilders as PAINT-ON would require. But the screen
must not be cleared when the mode switch happens.
We must keep the old captions when switching to PAINT-ON mode

## Test
- Live Over-the-Air content, beginning of commercials often uses
PAINT-ON mode

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -559,8 +559,19 @@ public final class Cea608Decoder extends CeaDecoder {
     int oldCaptionMode = this.captionMode;
     this.captionMode = captionMode;
 
-    // Clear the working memory.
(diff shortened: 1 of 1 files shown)


ojw28:
Looks fine to merge

zsmatyas:
Shall I add the next change on top of this in a new pull request? Shall I wait till this one gets merged so the branches have no conflicts?

ojw28:
Does it need to be on top at all? Unless the next change happens to be touching the same lines of code, I'd expect it to be possible to prepare it on `dev-v2` and have both changes merge fine. That said, we should get this merged tomorrow, so if you want to avoid experimenting you should probably just wait :).

Issue #4860, opened 2018-09-24
Pause / stop a single download, not all (Offline)


Mariovc:
### Issue description
When downloading HLS / DASH videos for offline mode, can I stop a single download while having all the rest downloading?
I've found a similar issue "[Pause/Resume download with HlsDownloader or DashDownloader](https://github.com/google/ExoPlayer/issues/3907)" where the solution is stopping all downloads, using `stopDownloads` from `DownloadManager` class. But can I stop only one selected download? 
In case a user wants to download a queued video without losing progress on the current downloading video. It is a nice feature to have, if it doesn't exist.

### Version of ExoPlayer being used
2.8.4

erdemguven:
It isn't possible currently. We'll consider supporting it.

ojw28:
Support will be added in `2.10`.

marcovietovega:
Do you have an estimate on when 2.10 will be released?

PR #5165, opened 2018-11-28
Pause/resume a single download #4860

AleksanderMielczarek:
This PR adds possibility to pause/resume single download (#4860). Please take a look and decide whether this is correct approach, if so I'll add tests.


AleksanderMielczarek:
I signed it!

AleksanderMielczarek:
Implementation seems to be working ok. Added some tests.

erdemguven:
Thanks for the pull request. Actually we're doing some major changes on DownloadManager which breaks this change. The good news is we're planning the add this feature as well. I'll check if we can a least partially use this code.

volkansahin45:
Will it be ready at next release? @erdemguven

meenukrishnamurthy:
Is this feature merged in the current release??

erdemguven:
We're working on a similar change currently. There is no certain date or release it will be available but I can say sometime soon.

AleksanderMielczarek:
@erdemguven Do you have any updates?

ojw28:
The ability to stop and resume an individual download is supported in `2.10.0` with `DownloadManager.setStopReason`. In the normal case of using a `DownloadService`, you should use `DownloadService.sendSetStopReason` or `DownloadService.buildSetStopReasonIntent`. The `2.10.0` release should happen next week (and we've already made a `2.10.0-alpha2` available that contains this functionality.

PR #5187, opened 2018-12-03
feat(MediaSource): client code can get the tag of a MediaSource

### LINKED ISSUE #5177: [Question] Get tag for MediaSource beside current in a playlist
We're trying to retrieve some metadata associated to `MediaSource`s found in a `ConcatenatingMediaSource` playlist.

Searching through past issues (#5131, #4707) and reference doc, we got that it is indeed possible to associate some `Object` as a tag when `MediaSource` is built, then it is possible to get the tag (from player) only for the currently playing item.

What we're looking for is to get that tag for any source item already added to a ConcatenatingMediaSource, so that we can have code similar to this:

```java
    private static void logTracklist(ConcatenatingMediaSource trackList) {
        int nrOfTracks = trackList.getSize();

        for (int i = 0; i < nrOfTracks; i++) {
            MediaSource source = trackList.getMediaSource(i);
            // previously, name has been set with Factory.setTag()

            String name = (String) source.getTag();  // this API does not exist

            Log.d(TAG, "#" + i + ": " + name);
        }
    }
```

[This comment](https://github.com/google/ExoPlayer/issues/4707#issuecomment-443131340) correctly suggested to open a new issue for this request, so here we are.

Is that something already doable? If not, would you consider a PR to add this feature? Thanks

BrainCrumbz:
Closes #5177

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
@@ -222,7 +222,9 @@ public interface MediaSource {
   /**
    * Returns the tag set on the media source, or null when none was set.
    */
-  @Nullable Object getTag();
+  @Nullable default Object getTag() {
+    return null;
+  }
 
   /**
    * Starts source preparation if not yet started, and adds a listener for timeline and/or manifest
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
@@ -276,6 +276,11 @@ public interface MediaSource {
   void releasePeriod(MediaPeriod mediaPeriod);
 
   /**
+   * Returns the tag set on media source, or null when none was set.
+   */
+  Object getTag();
+
+  /**
    * Removes a listener for timeline and/or manifest updates and releases the source if no longer
    * required.
    *
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
@@ -220,6 +220,11 @@ public interface MediaSource {
   void removeEventListener(MediaSourceEventListener eventListener);
 
   /**
+   * Returns the tag set on the media source, or null when none was set.
+   */
+  @Nullable Object getTag();
+
+  /**
    * Starts source preparation if not yet started, and adds a listener for timeline and/or manifest
    * updates.
    *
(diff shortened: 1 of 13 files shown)


GiuseppePiscopo:
I'm ok with my commits being contributed to this project.

BrainCrumbz:
For completeness: when running tests (on a Windows dev machine) we experienced 28 failed tests both before and after the change.

tonihei:
> For completeness: when running tests (on a Windows dev machine) we experienced 28 failed tests both before and after the change.

That's interesting :) Would you mind giving us a list? No one in our team is using Windows, so probably difficult to fix, but still seems worth knowing why they fail.

library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
```diff
@@ -275,6 +275,11 @@ void prepareSource(
    */
   void releasePeriod(MediaPeriod mediaPeriod);
 
+  /**
+   * Returns the tag set on media source, or null when none was set.
```
tonihei: on *the* media source

library/core/src/main/java/com/google/android/exoplayer2/source/MediaSource.java
```diff
@@ -275,6 +275,11 @@ void prepareSource(
...
+   */
+  Object getTag();
```
tonihei: Can you move the method somewhere else to keep the release methods bundled together? For example, above addEventListener or below removeEventListener.

tonihei: Also, can you make this a default method which returns null for now? That prevents breaking other existing implementations.

tonihei: @Nullable is missing here too.

tonihei (commented):
Thanks, looks good in general.

BrainCrumbz:
Re. test failing on Windows 10: on branch dev-v2, commit

> ffbb0da8932b28a0bb21f49eb02cf39275044bdd Prevent Cea608Decoder from generating Subtitles with null Cues list.

here are attached the test results (exported as zipped HTML): 

[Test Results - exoplayer2_in_library-core - ffbb0da.zip](https://github.com/google/ExoPlayer/files/2640476/Test.Results.-.exoplayer2_in_library-core.-.ffbb0da.zip)

BrainCrumbz:
Requested changes have been added. Not sure whether those `return null` should stay in ConcatenatingMediaSource.java. After all, they could be totally replaced by default implementation.

tonihei:
> Requested changes have been added. 

Thanks! Looks good on a code level. I'd also keep the `return null` where it is because it's intentional (compared to just not implemented yet).
Let's use the issue to clarify the need to do this. 

> attached the test results

That's good to see. I suspect they all fail because of some line separator or encoding difference.

BrainCrumbz:
> I'd also keep the return null where it is because it's intentional (compared to just not implemented yet).

True, totally agreeing on making it explicit.

> Let's use the issue to clarify the need to do this.

Not sure what you mean. If there's some action we should do on this side, just tell us please

tonihei:
> Not sure what you mean.

Never mind, I was referring to the ongoing discussion in #5177.

PR #5201, opened 2018-12-05
CEA608: Match alignment of consecutive lines in POP-ON mode

zsmatyas:
[Problem]
As the alignment of each lines are calculated individually, the
artificially introduced CENTER and RIGHT alignments can be mixed
with each other and with the default LEFT alignment. The mixed results
look worse than just using the original DEFAULT, and the readability
is radically decreased.

[Solution]
Pre-calculate the alignment for each line, and use the leftmost of them
for all lines. The intention of caption provider could be different, but
the worst case scenario is that we revert back everything to LEFT
alignment that is the only option in the CEA608 standard so we
show everything at the exact location it is set in the content.
The readability is never worse than the one given by the content
provider.

[Test]
This should influence CEA608 pop-up cations only. (Not roll up).
Test various live channels with this change.
Note: Beginning of commercials tend to have PAINT-ON mode as that
is immediately showing incoming characters, not just at the end
of the line like POP-on mode.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -542,8 +542,28 @@ public final class Cea608Decoder extends CeaDecoder {
 
   private List<Cue> getDisplayCues() {
     List<Cue> displayCues = new ArrayList<>();
-    for (int i = 0; i < cueBuilders.size(); i++) {
-      Cue cue = cueBuilders.get(i).build();
+
+    // CEA608 does not define Center and right alignment, content providers artificially introduced
+    // them by adding whitespaces before the real caption characters to get more visually appealing
+    // results. This decoder tries to optimize by switching to center or right alignment based on
+    // the number of whitespaces on the left and right side of the lines.
+    // This works in many cases, but has a side effect:
+    // consecutive lines on the screen might be differently aligned depending in the leftover
+    // space and the accuracy of the whitespace balancing of the original caption provider.
+    // Fixing this by introducing a new rule: lines shown at the same time, should be aligned the
+    // same way with the left most alignment being the highest priority. This will result in any
+    // exoPlayer optimisation only kick in when all lines can be re-aligned.
+    // LEFT is dominant as that was the only existing option in the original CEA608 standard, so
+    // the new worst case scenario is "showing the content exactly as the content provider
+    // intended"
+    int leftMostAnchor = Cue.ANCHOR_TYPE_END;
+    int cueCount = cueBuilders.size();
+    for (int i = 0; i < cueCount; i++) {
+      leftMostAnchor = Math.min(cueBuilders.get(i).calculatePreferredAlignment(), leftMostAnchor);
+    }
+
+    for (int i = 0; i < cueCount; i++) {
+      Cue cue = cueBuilders.get(i).build(leftMostAnchor);
       if (cue != null) {
         displayCues.add(cue);
       }
@@ -656,6 +676,9 @@ public final class Cea608Decoder extends CeaDecoder {
     private int captionMode;
     private int captionRowCount;
 
+    private SpannableStringBuilder finalCueString;
+    private int startPadding;
+    private int endPadding;
     public CueBuilder(int captionMode, int captionRowCount) {
       cueStyles = new ArrayList<>();
       rolledUpCaptions = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


zsmatyas:
CEA608 is hardly ever used in on-demand content, but it is widely used for over-the-air television. The artificial alignment of the content providers often creates scenarios when consecutive lines aligned differently become very hard to read. The existing optimization of ExoPlayer does not consider multiple lines when decides about alignment of each line individually.

This change fixes lots of issues when the **data contains** something like (note that the left sides of the lines were matched fore easy readability):
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                      What's up
```
----
With the optimization **before this change** (aligning lines individually)
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                                       What's up        // this line gets switched to right alignment as has less space on the right hand side.
```
----
**Possible side effect** of the new change: both lines are accidentally matching the criteria for center alignment. That is not a real issue, as it is still easy to read:
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                    What's up                       // this line is also getting center aligned
```
----
**worst case scenario** (everything reverted back to the left alignment and gets displayed exactly as defined in the content, easy to read (this is the exact placement requested by the specification)
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                       Joe:
                       What's up              // gets left aligned without any optimization
```

zsmatyas:
:shipit:  :D

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -542,8 +542,28 @@ private void handleMiscCode(byte cc2) {
 
   private List<Cue> getDisplayCues() {
     List<Cue> displayCues = new ArrayList<>();
-    for (int i = 0; i < cueBuilders.size(); i++) {
-      Cue cue = cueBuilders.get(i).build();
+
+    // CEA608 does not define Center and right alignment, content providers artificially introduced
```
ojw28, line 546: General comment about comments in code: They should generally not describe historical behaviors in present tense, or describe things as being "new". They should just describe what the current code does and why.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -542,8 +542,28 @@ private void handleMiscCode(byte cc2) {
...
+    // them by adding whitespaces before the real caption characters to get more visually appealing
+    // results. This decoder tries to optimize by switching to center or right alignment based on
+    // the number of whitespaces on the left and right side of the lines.
+    // This works in many cases, but has a side effect:
+    // consecutive lines on the screen might be differently aligned depending in the leftover
+    // space and the accuracy of the whitespace balancing of the original caption provider.
+    // Fixing this by introducing a new rule: lines shown at the same time, should be aligned the
```
ojw28, line 553: I'm not really too convinced this is the correct rule to be applying. In particular:

1. If there's a caption at the top of the screen and another at the bottom, they shouldn't be forced to have the same alignment. Forced alignment probably only makes sense for blocks of cues that are on consecutive rows. Can the code be made to work on such blocks?
1. Is it quite common that subtitles are updated in a way that causes lines to be added (e.g. one line is displayed, then two lines are displayed, then three, then the whole thing is removed)? If so then I think the proposed logic makes it likely that existing lines may change alignment when subsequent lines are added, which is likely quite visually jarring. Would it work better to use the alignment of the first line in each block for the whole block, which would avoid this issue.

If you agree, I think the implementation can probably be a bit simpler also.

zsmatyas: 1.
I wanted to introduce minimal changes as a few code reviewers are very picky with any modifications. ;) 

I tested the suggested modifications thoroughly and they work well as when, for example, the new change reverts to "Left alignment" the position of the text would not change if it was correctly rendered with the other alignments as we have tab and whitespace characters on the left side of the text.

A right-aligned text like:
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                       What's up
```
Is aligned to column 12. If it was correctly aligned, worst case scenario with my change it will be flipped to Left alignment to position 8. That should cause no real change in readability - but of course the pixel position is not exactly the same for both cases - and the position also depends on the font as 608 was specifically designed to use uniform-width fonts. There was no need to complicate the change to let only neighbouring lines influence each other as it is very rare and the worst case scenario is left aligning both an exactly place them to the location where the caption provider set them: very good readability.

----
2.
The use case you describe is common using newer subtitle standards, but in CEA608 it is very rare. I tested live streams in the Sunnyvale area extensively, and had not met the problem. I also assume that 608 is only used on the North American continent, so we do not need to investigate unusual alignment in unexpected languages / countries.

But a very common use case:
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                      What's up
```
These are received as left aligned texts positioned to the 6th column. The exoplayer code middle aligns the first line, right aligns the second one. As the fonts are not uniform width, the second line moves much more to the right than expected. 
I tried to improve readability with minimal change, so the second line will get middle aligned as well with this code in the pull request. 

```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                   What's up
```
Much better case would be to left align both, as the original content provider designed:
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                      What's up
```
If the original content provider intended middle alignment, they would have sent something like
```
1  2  3  4  5  6  7  8  9 10 11 12 13 14 15
                      Joe:
                    What's up
```
This is currently and would still be middle aligned after my suggested changes.

The use case of adding new captions to the screen while keeping previous ones visible is rare.
Pop on captions are sent by 2 byte chunks inserted into video frames. The visible and not visible buffers are swapped by a specific command. If the intention would be to show caption A, then add caption B to the screen without removing caption A, it is only possible by first sending caption A, the memory swap command, repeating caption A, sending caption B and the memory swap command in a short amount of time -> this use case is limited by the standard frame rate (was it 30 in the 90s?) and the 2 byte data in each frame. One exception would be using PAINT ON mode, but I haven't met any working implementation of that.

I have not met the issue of additional lines changing the alignment of already visible lines. Theoretically, my suggested change might cause updates in position of already visible text.

Matching the second line to the first might cause issues if the first line is right aligned. 

All in all I still think it is better to consider all lines and align them together as it greatly improves readability. Aligning everything to the "first line" might also cause problems, the lines might arrive in any order, so preserving location of existing lines is not trivial. 

I am curious how you would simplify the code while introducing the check to match consecutive lines only.

Issue #4468, opened 2018-07-04
setOutputSurface workaround needed for some more devices


qvk:
### Issue description
We're getting multiple records of `ExoPlaybackException` caused by `IllegalArgumentException` (sometimes `MediaCodec#CodecException`) when `MediaCodec#setOutputSurface` is called as seen in collected stacktraces. These happen when exiting the video activity, so I'm assuming `SimpleExoPlayer` sends a `MSG_SET_SURFACE` when `surfaceDestroyed` is invoked by the system, which eventually leads to the playback exception.

We started logging decoder name, error stacktraces and device info inside `onPlayerError` callback only in recent releases, so the actual occurrences are likely much more than we could record.

It sure doesn't look good adding all these devices to the workaround list, but there are only 4 or 5 unique decoder names in the list, so maybe there is some pattern here?

### Reproduction steps
We don't have the affected devices on hand but according to our logs, the issue occurs upon exiting any activity which is playing media using exoplayer.

### Version of ExoPlayer being used
2.7.0

### Device(s) and version(s) of Android being used
These are data collected from last couple of weeks for cases where `onPlayerError` callback is received with the above mentioned exception.

Many devices near the end of the list with very few occurrences might not warrant adding them to the exception list. Although they could be new devices just starting to grow in use.

```
count apiLevel	videoDecoderName		device	model	manufacturer

107	23	OMX.MTK.VIDEO.DECODER.AVC	CPH1609	CPH1609	OPPO
86	23	OMX.MTK.VIDEO.DECODER.AVC	A7020a48	Lenovo A7020a48	LENOVO
76	24	OMX.MTK.VIDEO.DECODER.AVC	1714	vivo 1714	vivo
71	23	OMX.MTK.VIDEO.DECODER.AVC	K50a40	Lenovo K50a40	LENOVO
57	23	OMX.MTK.VIDEO.DECODER.AVC	taido_row	XT1706	motorola
53	23	OMX.MTK.VIDEO.DECODER.AVC	BLACK-1X	BLACK-1XM	XOLO
52	23	OMX.MTK.VIDEO.DECODER.AVC	A10-70F	Lenovo TAB 2 A10-70F	Lenovo
48	23	OMX.MTK.VIDEO.DECODER.AVC	CPY83_I00	Coolpad Y83-I00	Coolpad
48	23	OMX.MTK.VIDEO.DECODER.AVC	GIONEE_GBL7360	F103 Pro	GIONEE
[...]

PR #5210, opened 2018-12-06
output surface workaround for BRAVIA_ATV3_4K

SVPA-ChrisFasulo:


ojw28:
Please can you provide some kind of description (e.g. a stack trace or similar from the affected device)? You will also need to sign the CLA, as described above. Thanks!

SVPA-ChrisFasulo:
Hi @ojw28, I need to track down the model of Android TV we first discovered the issue on and will get back to you with the stack trace, but it was almost identical to what's been reported in https://github.com/google/ExoPlayer/issues/4468. Our fix was simply adding the device name to the surface workaround list and we haven't see any issues arise since. I'll add more info once I track down the device again and get the CLA signed. Thanks!

ojw28:
We added this in https://github.com/google/ExoPlayer/commit/044066f4b12f41bb6db8cee7ffae444fe5530a7d.

Issue #4519, opened 2018-07-17
Samsung devices: Local or networked mp3 file plays twice rather than ending


kimdohyung5:
Before filing an issue:
-----------------------

When reporting a bug:
-----------------------

### Issue description
When I Played the mp3 file, I didn't get  the result "Player.STATE_ENDED" right after the mp3 file was finished.
I think that It happened  in library version over 2.8.0 and not happened below 2.7.3
and I expected that it would have fixed in the version 2.8.2 but, It didn't happed.

### Reproduction steps
Very simple.. when you play the mp3 file in the sample app then, you can find that i am right.

### Link to test content
any contents in the name of mp3s.

### Version of ExoPlayer being used
it has the problem over 2.8.0 but, is ok below 2.7.3

### Device(s) and version(s) of Android being used
i am using the samsung's s-pad device (SM-T 530)

### A full bug report captured from the device
I don't have the error message because it isn't the error when think otherwise.

so I have to attach the sample source which i tested. so that you can find what i mean easily.

ojw28:
I downloaded the test app but cannot reproduce. The mp3 file (duration 2:42) played, the ended event was triggered and the toast you've added was displayed as expected.

ojw28:
Closing for now because everything seems to work fine. Please provide additional information here if you want us to look again, and we'll re-open.

kimdohyung5:
After I received your comment, I realized that I just tested the demo app only on the samsung device
which is samsung pad( Samsung GALAXY Tab 4 10.1 ) and model name is SM-T530.

I tested other devices such as LG Pad , Samsung mobile phone and many others but I didn't make the same 
bug unfortunately.

I recorded the bug scene and attached below and attached the source code again.
I would be happy if you comment some happy news.

below is the source code.
[ExoAudioTestWithActivity_2018.07.18.zip](https://github.com/google/ExoPlayer/files/2204089/ExoAudioTestWithActivity_2018.07.18.zip)

below is the bug scene which I captured.
[20180718_148_rescale.zip](https://github.com/google/ExoPlayer/files/2204106/20180718_148_rescale.zip)

ojw28:
That's really weird. Please could you provide the output of `adb bugreport` captured shortly after the point at which the looping occurs?

kimdohyung5:
I captured the adb bugreport after looping occured.

[...]

PR #5216, opened 2018-12-07
Blacklist OMX.SEC.mp3.dec and OMX.brcm.audio.mp3.decoder on Samsung devices

mseroczynski:
Few confirmed fixes addressing https://github.com/google/ExoPlayer/issues/4519

These devices list below codecs (in this order):
OMX.SEC.mp3.dec
OMX.brcm.audio.mp3.decoder
OMX.google.mp3.decoder

Out of this list only OMX.google.mp3.decoder is working properly.


mseroczynski:
I signed it!

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
```diff
@@ -318,7 +318,14 @@ private static boolean isCodecUsableDecoder(android.media.MediaCodecInfo info, S
     }
 
     // Work around https://github.com/google/ExoPlayer/issues/4519.
-    if ("OMX.SEC.mp3.dec".equals(name) && "SM-T530".equals(Util.MODEL)) {
+    if (("OMX.SEC.mp3.dec".equals(name) || "OMX.brcm.audio.mp3.decoder".equals(name))
```
ojw28: Presumably each device has one of these decoders or the other. Can the code be structured to reflect this?

ojw28: Oh right, I read the change description and you're saying these devices really do have both of these (and the Google one). Is that definitely true for all of the referenced devices? If so then respond here and otherwise ignore the comment. Thanks!

ojw28: Looking at this more, I don't think it is true that all of these devices can have both of these decoders. "OMX.brcm" is probably Broadcom, but I think only two of these devices have Broadcom chipsets (SM-G350 and GT-I9152). I'd like for the code to accurately reflect what decoders are present on each device.

mseroczynski: Thank you for catching this out. I am 100% sure about (SM-G350, GT-I9152 and also GT-S7580) because I own psysical devices and debugged on all three them. As you pointed out, other devices (SM-T231, GT-I9152, GT-P5220) were just an assumption of similarity. I've corrected the PR to meet your suggestions and looking forward for round two of review.

PS: Correct me if I'm wrong but I think these codecs should be just somehow prioritized (not just Collections.singletonList(decoderInfos.get(0))) - I thought of this solution as a temporary fix that will be replaced somewhere in the future, as the list of if's may/will continue growing forever with current approach.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
```diff
@@ -318,7 +318,16 @@ private static boolean isCodecUsableDecoder(android.media.MediaCodecInfo info, S
...
     // Work around https://github.com/google/ExoPlayer/issues/4519.
-    if ("OMX.SEC.mp3.dec".equals(name) && "SM-T530".equals(Util.MODEL)) {
+    if ("OMX.SEC.mp3.dec".equals(name) && ("GT-I9152".equals(Util.MODEL)
+        || "GT-P5220".equals(Util.MODEL)
+        || "SM-T231".equals(Util.MODEL)
+        || "SM-T530".equals(Util.MODEL))) {
+      return false;
+    }
+    if (("OMX.SEC.mp3.dec".equals(name) || "OMX.brcm.audio.mp3.decoder".equals(name))
+        && ("GT-I9515".equals(Util.MODEL)
```
ojw28: I think this one belongs in the block above. It's not a broadcom based device.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
```diff
@@ -318,7 +318,16 @@ private static boolean isCodecUsableDecoder(android.media.MediaCodecInfo info, S
     }
 
     // Work around https://github.com/google/ExoPlayer/issues/4519.
-    if ("OMX.SEC.mp3.dec".equals(name) && "SM-T530".equals(Util.MODEL)) {
+    if ("OMX.SEC.mp3.dec".equals(name) && ("GT-I9152".equals(Util.MODEL)
```
ojw28: I'm surprised the GT-I9152 isn't in the block below, given this one is a broadcom based device.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java
mseroczynski: Overlooked! Swapping GT-I9515 with GT-I9152

PR #5245, opened 2018-12-11
Add protected method to override the AspectRatioFrameLayout aspect ratio

### LINKED ISSUE #5221: [Question] How to crop video that goes beyond some aspect ratio bounds?
I have the requirement that videos should be cropped to not exceed some minimum aspect ratio (say 9:16) and some maximum aspect ratio (say 16:9).

I am not familiar with the source and am wondering what is the efficient solution to this.
As I see it the first thing is to set `videoComponent.videoScalingMode = C.VIDEO_SCALING_MODE_SCALE_TO_FIT_WITH_CROPPING` which I have done.

The second thing would be hook into `AspectRatioFrameLayout` measuring, ideally, override the `setAspectRatio` method and apply my constraints. But the class is final. Tried to override the caller, which is the `ComponentListener` inside `PlayerView`, but there's no way to do that.

If this is the case, could `AspectRatioFrameLayout` be made non-final to support this simple use case?

I guess there are other ways,

- use a custom `SurfaceView` for example, but this requires dropping the `PlayerView`
- add an external video ratio listener, get the ratio, then call `AspectRatioFrameLayout.setAspectRatio()` with the desired value. This is inelegant, it's not sure that this listener would be called after the player  `ComponentListener`, and in any case this means calling setAspectRatio() twice, which means two layouts.
- work with `PlayerView.onMeasure` so it drives AspectRatioFrameLayout measuring? This shares most of the issues with previous idea, because we must wait for the ratio before eventually performing a new layout. This would also affect the player control view which is not what we want. 

Any help?

natario1:
This fixes #5221 : adds a protected method so that `PlayerView` subclasses can change the surface aspect ratio before it is forwarded to the `AspectRatioFrameLayout`.

I personally think that `AspectRatioFrameLayout` could **also** be made non-final, will do if you want.

I have intentionally left aside the drawable artwork case, and named the method with the "video" word, so this does not intercept all `setAspectRatio` calls but has a clearer meaning / usage.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
@@ -1031,6 +1031,19 @@ public class PlayerView extends FrameLayout {
     return subtitleView;
   }
 
+  /**
+   * Called to notify the {@link AspectRatioFrameLayout} of the new video aspect ratio.
+   * The default implementation will simply forward the aspect ratio to
+   * {@link AspectRatioFrameLayout#setAspectRatio(float)}. Subclasses can override this
+   * method, for example, to apply aspect ratio constraints.
+   *
+   * @param aspectRatioFrameLayout the video surface view container
+   * @param widthHeightRatio the computed aspect ratio
+   */
+  protected void onDispatchVideoAspectRatio(@NonNull AspectRatioFrameLayout aspectRatioFrameLayout, float widthHeightRatio) {
+    aspectRatioFrameLayout.setAspectRatio(widthHeightRatio);
+  }
+
   @Override
   public boolean onTouchEvent(MotionEvent ev) {
     if (ev.getActionMasked() != MotionEvent.ACTION_DOWN) {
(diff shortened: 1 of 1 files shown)


natario1:
I signed it!

ojw28:
Thanks! We might rename the method when we merge this, but otherwise all looks fine :). We'll get it merged soon.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java
```diff
@@ -1355,7 +1368,7 @@ public void onVideoSizeChanged(
         videoAspectRatio = 0;
       }
 
-      contentFrame.setAspectRatio(videoAspectRatio);
+      onDispatchVideoAspectRatio(contentFrame, videoAspectRatio);
```
ojw28, line 1371: I think it's a bug that L1344 returns if contentFrame is null, because I think L1351-L1366 still need to execute in that case. We'll fix this as part of merging this change also. No need to do anything on your side.

natario1: Makes sense!

Issue #535, opened 2015-06-10
ExoPlayer SEEK event


agalera-nice:
Hello! I am trying to detect a SEEK event in ExoPlayer, however I revised the demo app and perform several tests in order to check if it is informed to the event logger and I cannot see it.

Is there any way that a SEEK in the ExoPlayer can be detected through any event?

The only thing I could observed is that the player enters the buffering state as soon as the seek begins.

Thank you so much

ojw28:
ExoPlayer will only seek when you tell it to, so you should know when seeks occur without having to listen to the player.

Is your wish to have this just for convenience (i.e. because you want to register a bunch of things as listeners to the player, and see when seek gets called without having to track something else separately)?

agalera-nice:
I would like to listen to the seek event because we we're developing an analytics plugin where the only access to the player that we will have it's the player instance itself. So, we would not have access to any application code (Renders, etc..) and all the information we can get from the player will be obtained thanks to an ExoPlayer.Listener defined inside the plugin.

Is there any chance you could include the seek event in the next releases? Otherwise, I think we can make some workaround to detect this situation

Thank you

Arqu:
If it helps you, you can just extend PlayerControl and override the seekTo function to insert your own callback something like:
```
@Override
  public void seekTo(int timeMillis) {
    callback.seekEvent(timeMillis);
    super.seekTo(timeMillis);
  }
```

jpshelley:
@agalera-nice is this a public analytics project you are creating? I'm currently testing out different android players and need to perform some analysis of my own. Do you have recommendations for such tools?

ojw28:
The best solution as things stand would be for you to provide a thin wrapper around an ExoPlayer instance, which itself could implement ExoPlayer, and require that applications using your analytics project wrap their players and access them through the wrapper. You could then intercept the seek.

We may add a seek callback in the future, but have no immediate plans to do so.

PR #5288, opened 2018-12-21
Made seekTo not final

petersamokhin:
In this commit (https://github.com/google/ExoPlayer/commit/bbd82cf5da8072a0c67f5a422c426ef2d71e315e) you add a `BasePlayer` with `public **final** void seekTo()`, but in #535 (first issue I've found with this question) you advised to override `seekTo` in own instance. Now it's impossible (if you want to intercept `positionMs` too).

Or maybe there are another way to intercept seek event and result position?

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/BasePlayer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/BasePlayer.java
@@ -38,7 +38,7 @@ public abstract class BasePlayer implements Player {
   }
 
   @Override
-  public final void seekTo(long positionMs) {
+  public void seekTo(long positionMs) {
     seekTo(getCurrentWindowIndex(), positionMs);
   }
 


petersamokhin:
I signed it.

tonihei:
The referenced issue is very old and there is generally no guarantee that a piece of advice which is was true 3.5 years ago is still valid today.

If you want to do the same thing as in #535, you should probably override `seekTo(windowIndex, positionMs)` which is still possible.

Not sure what you want to do exactly, but if you just want to know that a seek happened, please listen to `EventListener.onPositionDiscontinuity` with `reason=DISCONTINUITY_REASON_SEEK`.

petersamokhin:
> Not sure what you want to do exactly

> intercept seek event and result position (not only seek event)

tonihei:
You can use `player.getCurrentPosition()` from the callback to the get the result position.

PR #5303, opened 2018-12-24
Add AC-4 format support

ybai001:
- Add AC-4 MIME type definition
- Add AC-4 format support in Mp4Extractor and TsExtractor
- Add AC-4 Extractor
- Add AC-4 playback support in MPEG-4, MPEG-DASH, TS and HLS


library/core/src/main/java/com/google/android/exoplayer2/extractor/DefaultExtractorsFactory.java
```diff
@@ -235,9 +236,10 @@ public synchronized DefaultExtractorsFactory setTsExtractorFlags(
                 | (constantBitrateSeekingEnabled
                     ? AmrExtractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING
                     : 0));
+    extractors[12] = new Ac4Extractor();
```
andrewlewis, line 240: Please could you add this to the list in the class javadoc?

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
```diff
@@ -787,6 +787,11 @@ protected MediaFormat getMediaFormat(
         mediaFormat.setFloat(MediaFormat.KEY_OPERATING_RATE, codecOperatingRate);
       }
     }
+    if (MimeTypes.AUDIO_AC4.equals(format.sampleMimeType)) {
+      // Some devices handle AC-4 raw frame by default.
```
andrewlewis, line 791: Do you have any information about what devices or named OMX components behave in this way? If there's a simple way to identify them, it would be nice to do this conditionally.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1276,10 +1277,17 @@ private boolean readSample(ExtractorInput input) throws IOException, Interrupted
         }
       }
     } else {
+      int sampleHeaderSize = 0;
```
andrewlewis: The code added here needs to go in the `if (parserState == STATE_READING_SAMPLE_START)` block above because the call to `output.sampleData(input, ...)` below may throw if there's a problem reading from the source, at which point we retry reading (and as things stand we'd output the header again). Ditto for `Mp4Extractor`.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac4Reader.java
```diff
@@ -0,0 +1,215 @@
...
+
+  @Override
+  public void consume(ParsableByteArray data) {
+    while (data.bytesLeft() > 0) {
+      switch (state) {
+        case STATE_FINDING_SYNC:
+          if (skipToNextSync(data)) {
+            state = STATE_READING_HEADER;
+            headerScratchBytes.data[0] = (byte)0xAC;
+            headerScratchBytes.data[1] = 0x40;
```
andrewlewis: Nit: Maybe just `headerScratchBytes.data[1] = hasCRC ? 0x41 : 0x40;`?

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * Holds sample format information as presented by a syncframe header.
+   */
+  public static final class SyncFrameInfo {
+
+    /**
+     * The sample mime type of the bitstream is {@link MimeTypes#AUDIO_AC4}.
+     */
+    public final String mimeType;
```
andrewlewis: Looks like this is fixed and only referred to in Ac4Reader. To make it clear that only one MIME type is possible, could you inline the constant where this field is referred to, then remove this field?

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
```diff
@@ -0,0 +1,269 @@
...
+      this.sampleRate = sampleRate;
+      this.frameSize = frameSize;
+      this.sampleCount = sampleCount;
+    }
+  }
+
+  /**
+   * The channel count of AC-4 stream.
+   */
+  // TODO: Parse AC-4 stream channel count.
```
andrewlewis, line 74: Is it feasible to do this as part of this change? It seems like it's important to expose an accurate channel count for surround audio.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
```diff
@@ -0,0 +1,269 @@
...
+  public static final int CHANNEL_COUNT_2 = 2;
+  /**
+   * The header size for AC-4 parser. Only needs to be as big as we need to read, not the full
+   * header size.
+   */
+  public static final int HEADER_SIZE_FOR_PARSER = 16;
+  /**
+   * Number of audio samples in the frame. Defined in IEC61937-14:2017 table 5 and 6. This table
```
andrewlewis, line 82: Is the same information available in a freely-available ETSI specification? If so could you point me to it? I can see ETSI TS 103 190 2014-04 table 83 (subsection 4.3.3.2.6) has some numbers for frame_len_base that are similar but not exactly the same.

andrewlewis (commented):
Do you have a sample .ac4 stream that could be used to add an Ac4ExtractorTest just like Ac3ExtractorTest?

This looks good to submit once that's added and the comments are addressed. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
ybai001: Hi, Andrew,
Could you write an email to me (ybai@dolby.com) so that I can reply that email to provide detailed information about your question? Thanks.

library/core/src/main/java/com/google/android/exoplayer2/extractor/ts/Ac4Reader.java
ybai001: You are right. I'll update it.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
ybai001: Dolby didn't add it into this change because:
1. Dolby has contributed AC-4 extractor to Android, which will be a part of Android Q. In that version, we also hard-coded channel count to 2. To align between Android Q and ExoPlayer, we did same thing in this contribution.
2. There are other reasons. I can provide details through email.

ybai001:
> Do you have a sample .ac4 stream that could be used to add an Ac4ExtractorTest just like Ac3ExtractorTest?

I talked with content creation team. It should be OK to add a sample .ac4 file. After I get approval from legal team, I'll add it into this contribution.

library/core/src/main/java/com/google/android/exoplayer2/extractor/DefaultExtractorsFactory.java
ybai001: Sure.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
ybai001: I see. I'll update the code.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
ybai001: OK, I can do it. But I have a question: we implemented the logic in this way because we wanted to align it with exist Ac3Reader and Ac3Util implementations. Do you want to change Ac3Reader/Ac3Util in the same way in future?

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
ybai001: Your quote is correct. I'd like to explain this table. For example, if frame_rate_index is 0, INTERNAL frame length is 1920. And decoder resampling ratio is 1001/1000*25/24. So EXTERNAL frame length is 1920 * 1001/1000*25/24 = 2002, which is same as the first line in my code.
The background of this table is AC-4 has feature to align the frame length between video and audio. For example, if video frame_rate is 23.976 fps, 2002 audio samples@48kHz will have same time duration as one video frame. (48000/2002 = 23.976)

ybai001: BTW, the latest AC-4 standards in ETSI are
Part 1: https://www.etsi.org/deliver/etsi_ts/103100_103199/10319001/01.03.01_60/ts_10319001v010301p.pdf
and
Part 2: https://www.etsi.org/deliver/etsi_ts/103100_103199/10319002/01.02.01_60/ts_10319002v010201p.pdf

ybai001:
> Do you have a sample .ac4 stream that could be used to add an Ac4ExtractorTest just like Ac3ExtractorTest?
> 
> This looks good to submit once that's added and the comments are addressed. Thanks!

Hi, Andrew,
I have asked content creation team to create sample.ac4 file. And I have one question: How to generate corresponding .dump file in folder library/core/src/test/assets/ts?

andrewlewis:
@ybai001 You can set the field `FakeExtractorOutput.WRITE_DUMP` to true temporarily, then copy the file that gets written to the sdcard into the assets directory.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1277,17 +1282,20 @@ private boolean readSample(ExtractorInput input) throws IOException, Interrupted
         }
       }
     } else {
-      int sampleHeaderSize = 0;
-      if (MimeTypes.AUDIO_AC4.equals(track.format.sampleMimeType)) {
+      if (MimeTypes.AUDIO_AC4.equals(track.format.sampleMimeType) && !isAc4HeaderAdded) {
```
ybai001, line 1285: I didn't put this part of code to "if (parserState == STATE_READING_SAMPLE_START)". Instead, I set "isAc4HeaderAdded" in "if (parserState == STATE_READING_SAMPLE_START)" branch. If there is an exception in input reading, AC-4 header will not be added again.
I didn't put this part of code to if branch because "output" is not available at that time. And I don't want to change your original code a lot.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
andrewlewis: I think `Ac3Util` is different to this one as it handles both `MimeTypes.AUDIO_E_AC3` and `MimeTypes.AUDIO_AC3` and needs to signal which one is used, which this case seems to have only one possible MIME type.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
andrewlewis: Thanks.
- Please could you make the constant private?
- If the input stream is not actually stereo (e.g., it's mono or 5.1) would you expect MediaCodec to throw an exception during configuration?

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
andrewlewis: Got it. Thanks!

andrewlewis (commented):
Thanks.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
ybai001: You are right. I'll update the code.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
ybai001: Please could you make the constant private?
- sure.

If the input stream is not actually stereo (e.g., it's mono or 5.1) would you expect MediaCodec to throw an exception during configuration?
- No. The decoder itself is full functionality. That stream still can be decoded. In current implementation, decoder itself will ignore the input channel configuration information. (Decoder will parse this info by itself.)
Handling channel count is a little complex in AC-4, especially when we consider another concept in AC-4 -- AudioPresentation (https://developer.android.com/reference/android/media/AudioPresentation). This topic is under our internal discussion. I'll definitely submit a new pull request as long as it is done inside Dolby.

ybai001:
Sorry for the untidy commit history. I‘m not familiar with the working process on GitHub. 
BTW, next week is Chinese new year. I'm afraid that I can't reply in time.

library/core/src/main/java/com/google/android/exoplayer2/audio/Ac4Util.java
andrewlewis: We don't currently expose audio presentations so it sounds like we'll need to do this later on. Thanks.

andrewlewis (commented):
Looks good, thanks! This pull request will automatically be marked as merged when we submit this internally and update the development branch.

PR #5328, opened 2018-12-31
Added NotificationDispatcher

### LINKED ISSUE #5301: ExoPlayer manipulate notification based on Playerstate
### Issue description
How can I make the exoplayer notification dismissable when the player is paused and then make it a foreground service when it it resumed? Here are my questions:
1. https://github.com/google/ExoPlayer/issues/4256#issuecomment-388936681: How can I start a foreground service by using `startForeground` inside `onPlayerStateChanged`? Is there a way to retrieve notification object that's already constructed?
2. https://github.com/google/ExoPlayer/issues/4482#issuecomment-407088270: Can we please have the code base on the demo app changed so that we can refer that?

### Reproduction steps
IO18 audio app
https://github.com/google/ExoPlayer/tree/io18

latest

Nexus 6p. API level 28

0xcaff:
`NotificationDispatcher` provides an interface for delivering notifications from a `PlayerNotificationManager`. This can be used to implement a foreground service with a notification which is only pinned (non-dismissable) when the player is playing.

Here's an example of using these changes to implement the desired behavior in the I/O 2018 app.

Running: https://youtu.be/ckKVB4iy5yw
Code: https://gist.github.com/0xcaff/c20e4ea226c33eb68a5cd9cbb80fbb96!

I'd make a PR for the I/O 2018 app too, except there's a bunch of stuff broken in it which I'm not sure how to fix.

Closes #5301

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerNotificationManager.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerNotificationManager.java
@@ -202,6 +202,48 @@ public class PlayerNotificationManager {
     void onNotificationCancelled(int notificationId);
   }
 
+  /** Handles displaying notifications to the user. */
+  public interface NotificationDispatcher {
+    /**
+     * Called when a notification should be displayed to the user.
+     *
+     * @param notificationId The id which the notification should be displayed with.
+     * @param notification The {@link Notification}
+     */
+    void notify(int notificationId, Notification notification);
+
+    /**
+     * Called when a notification should be removed.
+     *
+     * @param notificationId The id of the notification to remove.
+     */
+    void cancel(int notificationId);
+  }
+
+  /** Dispatches notifications directly to a {@link NotificationManagerCompat} */
+  public static class DefaultNotificationDispatcher implements NotificationDispatcher {
+    private final NotificationManagerCompat notificationManager;
+
+    public DefaultNotificationDispatcher(NotificationManagerCompat notificationManagerCompat) {
+      notificationManager = notificationManagerCompat;
+    }
+
+    @Override
+    public void notify(int notificationId, Notification notification) {
+      notificationManager.notify(notificationId, notification);
+    }
+
+    @Override
+    public void cancel(int notificationId) {
+        notificationManager.cancel(notificationId);
+    }
+
+    public static DefaultNotificationDispatcher from(Context context) {
+      NotificationManagerCompat manager = NotificationManagerCompat.from(context);
+      return new DefaultNotificationDispatcher(manager);
+    }
+  }
+
   /** Receives a {@link Bitmap}. */
   public final class BitmapCallback {
     private final int notificationTag;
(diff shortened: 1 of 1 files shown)


gowthamgts:
Just for confirmation, isn't the `setOngoing` supposed to be `true` when playing, and `false` when it's not?

0xcaff:
The ongoing status of the notification is automatically handled by startForeground and stopForeground. Calling setOngoing isn’t needed. 

This is what the UAMP demo app does. https://github.com/googlesamples/android-UniversalMusicPlayer/blob/master/media/src/main/java/com/example/android/uamp/media/MusicService.kt

0xcaff:
I've integrated these changes into my app and almost everything works. In my app, the notification isn't the only consumer of the player. The player is also accessible through a binder exposed in `onBind`.

By default `NotificationDispatcher.cancel` is only called on notification dismissal when `setStopAction(null)` is not called. This means the only way for the cancellation callback to be called is to enable the stop action (and the corresponding button in the notification). Dismissing the notification triggers `STOP_ACTION` stopping the player on dismissal, which is not the desired behavior. At most, the playback should be paused, not stopped.

While it is possible to make this work by extending the notification manager, it's kinda hack-y. You'd have to use this modification, extend notification manager override `getActions` to hide the stop button and use a custom `ControlDispatcher` to disable the stop action. All this, and it only works if you don't need the stop button in the notification.

Are you open to breaking changes? Can we separate the stopping behavior and the notification cancellation behavior?

This is too hard. I want to modify the notification drawables later, so I decided it would be easier to just re-write the notification manager in my application.

Implementation: https://github.com/forte-music/android/blob/18e0eac5487ee6c39526244c531fe1af149eca4e/app/src/main/java/me/a0xcaff/forte/playback/PlayerNotificationManager.kt
Usage: https://github.com/forte-music/android/blob/18e0eac5487ee6c39526244c531fe1af149eca4e/app/src/main/java/me/a0xcaff/forte/playback/PlaybackService.kt

0xcaff:
Closing due to inactivity.

ojw28:
@marcbaechinger - Could you take a look at this?

gowthamgts:
@0xcaff: They might respond slow since it was holiday time.

marcbaechinger:
Yes, sorry for being late.

Thanks for the pull request.

We should certainly provide a way to make the notification dismissible when the player is paused in a service. I was under the impression this works but  I certainly could miss something:

> Dismissing the notification triggers STOP_ACTION stopping the player on dismissal, 
> which is not the desired behavior. At most, the playback should be paused, not stopped.

You can set your custom action with setStopAction("your-action"). When the notification is dismissed, this action is fired which you can handle yourself. Stop is not called in this case. 

------
The current approach which I think should do the trick is the following (and obviouly not well documented). I wonder whether this is working for you.

1. Stop foreground when player is paused
Like you said, the notification can't be dismissed when attached to a service which is in foreground. So you need to call service.stopForeground(/*removeNotification*/ false) when the player is paused (you can listen for playWhenReady within the onPlaybackStateChanged of a listener to stopForegound the service.

2. playerNotificationManager.setOngoing(false)
Needs to be set to false so the notification can be dismissed.

3. playerNotificaitionManager.setStopAction(yourCustomAction)
When the notification is dismmissed and there is a stopAction, that stopAction is fired when the user dismisses the notification. If you set your own, your action is fired.

Hence you can set your custom action to replace the internal stop action. Then you can handle your custom stop action with a CustomActionReceiver and do whatever you want to do in case the user dismisses the notification.

4. Use a CustomActionReceiver to handle you custom stop action which is fired when the user dismisses the notification.

ojw28:
Closing this PR because other changes have been merged that attempt to achieve the same thing. Please see #5301, and let's use that issue for any discussion. Thanks!

Issue #3860, opened 2018-02-16
CEA608 -  Dropped commands


zsmatyas:
Starting at this line the code checks if a CEA608 control code was repeated: 
https://github.com/google/ExoPlayer/blob/2b20780482a9c6b07416bcbf4de829532859d10a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java#L339

The problem is that it does not take into account the timing of the received control codes. There are contents that use the same control code multiple times consecutively, like the Sarnoff test streams required for FCC certification. Their content contains the following commands:

```
02-15 10:32:03.888 MISC: RESUME_CAPTION_LOADING
02-15 10:32:03.938 MISC: ERASE_DISPLAYED_MEMORY
02-15 10:32:03.978 MISC: ERASE_NON_DISPLAYED_MEMORY
02-15 10:32:04.008 PAC: Row:11; Col:1; Color:WHITE; italic:false; underline:false
02-15 10:32:04.568 PAC: Row:12; Col:1; Color:WHITE; italic:false; underline:false
02-15 10:32:05.529 MISC: END_OF_CAPTION
02-15 10:32:05.869 MISC: END_OF_CAPTION // Dropped
02-15 10:32:06.210 MISC: END_OF_CAPTION
02-15 10:32:06.540 MISC: END_OF_CAPTION // Dropped
02-15 10:32:06.870 MISC: END_OF_CAPTION 
...
```
Note: According to the standard END_OF_CAPTION flips the visible and hidden caption buffers, so it should make the 2 lines repeatedly visible then hidden again: effectively blinking in this case.

The current implementation drops every second command of this content.

According to https://www.law.cornell.edu/cfr/text/47/79.101

> If the first transmission of a control code pair passes parity, it is acted upon within one video frame. If **the next frame** contains a perfect repeat of the same pair, the redundant code is ignored.

So the check should only drop control codes if they are in consecutive video frames immediately after each other.

PR #5355, opened 2019-01-09
CEA608: Limiting duplicated command checks to immediate frames

zsmatyas:
Reported in https://github.com/google/ExoPlayer/issues/3860
For failing examples see the github link above.

[Problem]
We drop matching control codes even if they are not received on
consecutive frames.

The specification says
"(4) If the first transmission of a control code pair passes parity,
it is acted upon within one video frame. If the NEXT frame contains
a perfect repeat of the same pair, the redundant code is ignored."

Keyword is the NEXT. The frames must arrive immediately after
each other.

See https://www.law.cornell.edu/cfr/text/47/79.101

[Solution]
Set an additional flag when any data is processed. Control code
duplication checks should be limited only for the first control
byte pairs processed after any control code.

[Test]
Sarnoff tests have equivalent CEA708 and CEA608 Streams.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -292,7 +292,6 @@ public final class Cea608Decoder extends CeaDecoder {
   protected void decode(SubtitleInputBuffer inputBuffer) {
     ccData.reset(inputBuffer.data.array(), inputBuffer.data.limit());
     boolean captionDataProcessed = false;
-    boolean isRepeatableControl = false;
     while (ccData.bytesLeft() >= packetLength) {
       byte ccHeader = packetLength == 2 ? CC_IMPLICIT_DATA_HEADER
           : (byte) ccData.readUnsignedByte();
@@ -337,6 +336,9 @@ public final class Cea608Decoder extends CeaDecoder {
       // If we've reached this point then there is data to process; flag that work has been done.
       captionDataProcessed = true;
 
+      boolean repeatedControlPossible = repeatableControlSet;
+      repeatableControlSet = false;
+
       if (!ODD_PARITY_BYTE_TABLE[ccByte1] || !ODD_PARITY_BYTE_TABLE[ccByte2]) {
         // The data is invalid.
         resetCueBuilders();
@@ -372,7 +374,7 @@ public final class Cea608Decoder extends CeaDecoder {
       // Control character.
       // ccData1 - 0|0|0|X|X|X|X|X
       if ((ccData1 & 0xE0) == 0x00) {
-        isRepeatableControl = handleCtrl(ccData1, ccData2);
+        handleCtrl(ccData1, ccData2, repeatedControlPossible);
         continue;
       }
 
@@ -384,30 +386,26 @@ public final class Cea608Decoder extends CeaDecoder {
     }
 
     if (captionDataProcessed) {
-      if (!isRepeatableControl) {
-        repeatableControlSet = false;
-      }
       if (captionMode == CC_MODE_ROLL_UP || captionMode == CC_MODE_PAINT_ON) {
         cues = getDisplayCues();
       }
     }
   }
 
-  private boolean handleCtrl(byte cc1, byte cc2) {
+  private void handleCtrl(byte cc1, byte cc2, boolean repeatedControlPossible) {
     boolean isRepeatableControl = isRepeatable(cc1);
 
     // Most control commands are sent twice in succession to ensure they are received properly.
     // We don't want to process duplicate commands, so if we see the same repeatable command twice
     // in a row, ignore the second one.
     if (isRepeatableControl) {
-      if (repeatableControlSet
+      if (repeatedControlPossible
           && repeatableControlCc1 == cc1
           && repeatableControlCc2 == cc2) {
-        // This is a duplicate. Clear the repeatable control flag and return.
-        repeatableControlSet = false;
-        return true;
+        // This is a duplicate. Repeatable control flag should be already cleared, let's return.
+        return;
       } else {
-        // This is a repeatable command, but we haven't see it yet, so set the repeatable control
+        // This is a repeatable command, but we haven't seen it yet, so set the repeatable control
         // flag (to ensure we ignore the next one should it be a duplicate) and continue processing
         // the command.
         repeatableControlSet = true;
@@ -425,8 +423,6 @@ public final class Cea608Decoder extends CeaDecoder {
     } else if (isMiscCode(cc1, cc2)) {
       handleMiscCode(cc2);
     }
-
-    return isRepeatableControl;
   }
 
   private void handleMidrowCtrl(byte cc2) {


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -323,19 +325,24 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
...
         if (previousCaptionValid) {
           // The encoder has flipped the validity bit to indicate captions are being turned off.
           resetCueBuilders();
-          captionDataProcessed = true;
         }
         continue;
       }
 
-      // If we've reached this point then there is data to process; flag that work has been done.
-      captionDataProcessed = true;
```
zsmatyas: I moved this line above, as I think it should be set for any invalid caption as well. That will cause the screen update, I do not see a reason why we do not want to update the screen for invalid captions.

ojw28: What does doing this fix? `captionDataProcessed` is already set to `true` on L332 if there's a `valid->invalid` transition. For `invalid->invalid` the screen will already be blank, so updating the screen wont do anything useful?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -230,6 +230,7 @@
   private boolean repeatableControlSet;
   private byte repeatableControlCc1;
   private byte repeatableControlCc2;
+  private boolean dataProcessedSinceLastRepeatableControl = true;
```
ojw28: Why does this default to true? Ditto for flush. We haven't seen any data at all in these cases, so I don't think it makes sense for it to be true. Comment may be obsolete if you can get rid of the field as per comment below.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -323,19 +325,24 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
...
           // The encoder has flipped the validity bit to indicate captions are being turned off.
           resetCueBuilders();
-          captionDataProcessed = true;
         }
         continue;
       }
 
-      // If we've reached this point then there is data to process; flag that work has been done.
-      captionDataProcessed = true;
+      if (dataProcessedSinceLastRepeatableControl) {
```
ojw28: Do you understand what this bit of existing code is doing (which is kept in this change; I'm posting this comment here because GitHub doesn't let me comment on the line I want to comment on):

https://github.com/google/ExoPlayer/blob/71f72c59537711399dc5496136dd6b867acc6d77/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java#L329

I can't figure out what it's supposed to be doing. Perhaps it was an attempt to implement what you're now doing properly, and so becomes unnecessary?

Either way, does it work to get rid of both that bit of code and `dataProcessedSinceLastRepeatableControl`, and just do this here:
```
boolean repeatedControlPossible = repeatableControlSet;
repeatableControlSet = false;
```
and then pass `repeatedControlPossible` as an argument to `handleCtrl`? The line that sets `repeatableControlSet = false` inside `handleCtrl` would also go away.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: I just wanted to reduce redundancy, instead of setting the flag inside and after this if block, we can achieve the same by setting it before the block, can't we?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: The very first "repeatable control" should not be handled as a repetition: default = true.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: * ```repeatableControlSet``` seems to only save the value between 2 iterations of the loop.
* This part is probably only disabling the check for duplications between two calls of ```decode()```:
```
    if (!isRepeatableControl) {
        repeatableControlSet = false;
      }
```
I do not see why that is necessary at the moment. If 608 bytes are the first ones in each packet (as 608 bytes have fixed bandwidth), then 708 bytes are next to each other, then 2 calls to ```decode()``` probably cannot contain originally consecutive bytes in the analogue CEA stream.


Your proposal seems logical, I do not seem any drawbacks of that solution, let me update the code.

(Unfortunately, I cannot run full test cycles including various live content, sarnoff test cases etc. for every code change separately, but I will run them as soon as we have more - maybe all changes merged.)

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: I think the visual result is the same, but in the case that the encoder has turned captions off by flipping the validity bit you're causing a whole load of unnecessary work to happen on every single frame.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -323,19 +322,21 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
...
           resetCueBuilders();
-          captionDataProcessed = true;
         }
         continue;
       }
 
-      // If we've reached this point then there is data to process; flag that work has been done.
-      captionDataProcessed = true;
+      boolean repeatedControlPossible = repeatableControlSet;
+      repeatableControlSet = false;
```
ojw28, line 340: Should these lines go above the block that handles invalid captions? I doubt it's correct to de-duplicate repeated commands that have invalid data in between. Not sure though :).

ojw28:
Two pending comments (the one above, and the one about putting `captionDataProcessed` back). Otherwise good to merge; thanks!

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: I think I had these checks a few line above this current position before and it failed some live streams. My theory was that the original content was made for 30FPS, and when they insert extra frames to be able to play it in 60FPS, they insert zero byte pairs for the new empty frames. Zero check is a little bit earlier, I do not think it would make a difference having this before or after the validity check if it is always after the zero-byte-pair check.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: Updated.

zsmatyas:
Have I addressed all comments? It is hard to follow after I force pushed changes.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: I was thinking it should go after the zero check and before the validity check. In case captions are toggled off by flipping the validity bit after a repeatable command, and then when captions are toggled on again the first piece of data happens to be the same repeatable command. Seems like an edge case, but may as well handle it given it's just moving the lines up a little?

ojw28:
Think so yes; thanks!

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: I will move this up as part of the merge.

PR #5390, opened 2019-01-15
CEA608: handling channel bits properly

zsmatyas:
[Problem]
CC1 is shown even when CC2-3-4 is selected

[Solution]
 * use incoming channel and frame information
 * Misc Codes need to handle use Frame information
 * Add checks everywhere

[Test]
* Live channels and already existing test should not show any difference
in behavior when CC1 is selected.
* Live channels might have CC2-CC3-CC4
* Sarnoff test content has specific test cases for channel support


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -307,8 +329,7 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
         continue;
       }
 
-      if ((selectedField == 1 && (ccHeader & CC_FIELD_FLAG) != NTSC_CC_FIELD_1)
-          || (selectedField == 2 && (ccHeader & CC_FIELD_FLAG) != NTSC_CC_FIELD_2)) {
+      if ((ccHeader & CC_FIELD_FLAG) != selectedField) {
```
ojw28, line 334: Could you explain a bit more about channels (or provide a link/reference). In particular, why is it that the channel check doesn't happen early on in code (e.g. somewhere around here)? It also seems to not apply to the Basic North American character set on L405, so I'm not sure how channels would work for delivering multiple captions?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -414,13 +444,17 @@ private void handleCtrl(byte cc1, byte cc2, boolean repeatedControlPossible) {
       }
     }
 
+    if (!isCodeForSelectedChannel(cc1)) {
```
ojw28: Related question: I probably need to understand channels a bit better, but it feels odd that this isn't at least pulled up to the top of this method.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: The standard has lots of ambiguous parts, but it is probably a good start to check paragraph "4.1 Data types in the Line 21 signal" of the 608 standard.
It describes that Field 1 and Field 2 are independent, they are received by observing different parts of the analogue signal.

Each video frame has 2 bytes of data for each Field.

Inside Field 1 there are bytes belonging to CC1, CC2, Text1 and  Text 2.
Inside Field 2 there are bytes belonging to CC3, CC4, Text3, Text4 and XDS.

As the bandwidth is limited, CC1, CC2, Text 1 and Text 2 are sharing the 2 byte per frame data, usually only one service is used completely: CC1 for the primary language. After the commands and characters used for CC1 the leftover of the 60 bytes per second (2 bytes x 30 FPS) can be used for the other services, but it is rare.

Similarly, CC3 of Frame 2 is conveying the second language, CC4, Text3, Text4 and XDS is rarely used.

But when a single frame transports multiple services, we need to check each CC commands which channel it belongs to. And there are additional commands switching to Text an XDS modes. So after those commands all consecutive characters should be put into the buffers of those modes and not into the Caption buffers (this change will come in the next pull request).

Also note that there is only a single bit marking a channel value inside a single frame. So it is sometimes confusing to call the two channels channel 1 and 2, although channel 1 can mean CC1 and CC3 depending on which frame is decoded at this point. Similarly, Channel 2 can mean CC2 and CC4 depending on the current frame.

I cannot quote the standard as it is copyright protected, but I can link some derived wikipedia content:
https://en.wikipedia.org/wiki/EIA-608

> Special North American character set
...
A command to display a special character has a first byte of 0x11 or 0x19 (depending upon channel)

Similarly,
> Extended Western European character set
...
A command to display an extended Spanish/French or miscellaneous character has a first byte of 0x12 or 0x1A (depending upon channel).
...
Portuguese/German/Danish character has a first byte of 0x13 or 0x1B(depending upon channel).

Additionally:
> Control commands
Bits 15 and 7 are always odd parity bits. Bit 11 is always the channel bit. 

The 608 Standard has list of byte values and their meaning, like:
> F.1.1.4 Miscellaneous Control Codes 

```
Data Channel 1      Data Channel 2        Mnemonic Command       Description 
14  20              1C  20                RCL                Resume caption loading 
```
So the good news is that we always need to check a single bit of the commands and based on that bit we can decide which channel the command belongs to. (The only exceptions are MISC codes).

All consecutive characters not having a channel bit should be directed into the last used channel. So if there is a command to switch to "White Underlined Font" on Channel 1, we need to check the frame value to decide if that is CC1 or CC3. And from that point on all incoming characters will belong to that caption channel.

MISC codes also have one additional bit expected to change based on the Field value. It is described in this very convoluted way in the standard:
>8.4 Closed Caption Mode 
The non-printing character of the miscellaneous control-character pairs that fall in the range of 0x14, 0x20 to 0x14, 0x2F in field 1, shall be replaced with 0x15, 0x20 to 0x15, 0x2F when used in field 2.  
b)   The non-printing character of the miscellaneous control-character pairs that fall in the range of 0x1C, 0x20 to 0x1C, 0x2F in field 1, shall be replaced with 0x1D, 0x20 to 0x1D, 0x2F when used in field 2.  
c)   When any of the control codes from 0x01 to 0x0F is used to begin a control-character pair it indicates the beginning of XDS Data. This XDS data should be 
considered part of a different data channel and can be used to interrupt the captioning data according to the FCC rules for other data channels.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: The output is probably the same, we can optimize the order during these merges.

Note: There are additional comments in the standard, for example, Frame 2 should avoid repeating commands as it has the same bandwidth but more things to transfer.

zsmatyas:
I can merge a few more modifications that come with the next change (handling TEXT and XDS code) into this change set, maybe they belong to this commit more and help with the understanding.

zsmatyas:
I am confused what causes the conflict, I thought my change is based on the latest version Cea608Decoder.java of the dev-2 branch.

ojw28:
Thanks for all the detail. I'll try and block out some time to take a closer look.

As a higher level question, do you think this change is actually fixing something in practice (as opposed to a theoretical issue)? The Wikipedia page ([Channels section](https://en.wikipedia.org/wiki/EIA-608#Channels)) seems to imply that only CC1 and CC3 were used in practice. Aside from the Sarnoff test streams, have you found any real world usage of CC2 and CC4? And if not, do we really need the additional complexity?

ojw28:
What do you think about the question above (i.e. whether this is fixing something that's actually used; or a theoretical issue)? Thanks!

zsmatyas:
Sorry for the slow response, I am off for a few weeks. 

Yes, this change fixes issues with over-the-air broadcasts whenever there is more than one Cea608 channels are present. Next change is also necessary to filter out XDS and TEXT content.

ojw28:
> Yes, this change fixes issues with over-the-air broadcasts whenever there is more than one Cea608 channels are present. Next change is also necessary to filter out XDS and TEXT content.

The question is more whether this ever happens, aside from the Sarnoff test vector. The wikipedia page linked above tends to imply that CC2 and CC4 are basically never used, which is what most of the additional complexity in this change appears to be for. If that's the case, it's unclear to me that it's worthwhile making these changes, other than any parts that are necessary for CC1 and CC3.

zsmatyas:
Yes, by over-the-air broadcasts I meant that if you watch content that is broadcast by local channels in the United States, you will see these problems. I have only tested in the Sunnyvale area, but here when there are more than one channels present or if there is any other type of content (TEXT or XDS), those characters are all mixed into the captions shown by exoplayer.
CC2 and CC4 are rarely used, but when they are, the output is incorrect. I have met XDS content multiple times as well. Supporting these channels and allowing the user to select among the 4 channels is a requirement by FCC in the US for all TV devices.

zsmatyas:
Any news? :)

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -269,6 +292,7 @@ public void flush() {
     repeatableControlSet = false;
     repeatableControlCc1 = 0;
     repeatableControlCc2 = 0;
+    incomingDataTargetChannel = NTSC_CC_CHANNEL_1;
```
ojw28, line 295: For data received prior to the first control character, both initially or after a flush, is assuming `incomingDataTargetChannel = NTSC_CC_CHANNEL_1` a possibly incorrect assumption? And if so would it be more correct to set to an `CHANNEL_UNKNOWN` constant or similar, in both cases, and drop all data until we see a control character to know what channel the data belongs to?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -410,13 +451,17 @@ private void handleCtrl(byte cc1, byte cc2, boolean repeatedControlPossible) {
...
+    }
+
     if (isMidrowCtrlCode(cc1, cc2)) {
       handleMidrowCtrl(cc2);
     } else if (isPreambleAddressCode(cc1, cc2)) {
       handlePreambleAddressCode(cc1, cc2);
     } else if (isTabCtrlCode(cc1, cc2)) {
       currentCueBuilder.tabOffset = cc2 - 0x20;
-    } else if (isMiscCode(cc1, cc2)) {
+    } else if (isMiscCode(cc1, cc2, selectedField)) {
```
ojw28, line 464: Is it really necessary to check the field here? I acknowledge that the field will be set in `cc1` in this case, but wont the header check (`if ((ccHeader & CC_FIELD_FLAG) != selectedField) {`)  near the top of `decode` have already skipped the data if it belonged to a different field? If not then something seems a bit odd about this?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -231,23 +236,41 @@
   private byte repeatableControlCc1;
   private byte repeatableControlCc2;
 
+  private int incomingDataTargetChannel;
```
ojw28, line 239: nit: Maybe just call it `currentChannel`? Not sure the extra words add anything :).

ojw28:
> Any news? :)

Sorry, this fell off my radar a bit. I have two specific questions around correctness (see inline comments above), but once those are addressed/clarified we should be able to get this merged. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: This is weird in the standard, but the Misc code bytes (only) were defined to contain one additional bit based on field value. That bit seems to be redundant information, but needs to be checked, otherwise some rare values would satisfy the condition that should not be handled as Misc Code.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: You may rename at as you like during the pull, or I can upload a new revision with the new name as well.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: You are right, CHANNEL_UNKNOWN might be technically more correct as we should not handle most bytes before identifying the context correctly. The side effect of this change could be that previously working captions for on-demand content not having correctly defined the channel at the very beginning will start missing the first few characters from now on.
 
Defaulting it to CHANNEL_1 might cause problems with live content at the very beginning, but most of the times - even for live content - only one caption channel is used, the decoder will automatically recover from this problem with the first byte pair containing the correct channel information.

As it seems live over-the-air use cases are relatively rare, (as no other people are creating pull requests :) ) it might be safer setting the defaults in a way that would not cause any regression even for not perfect 608 content. I have not tested this situation too much (probability of incorrect on-demand content Vs. probability of live content having more than one channels), so I do not have a definitive suggestion. Maybe CHANNEL_UNKNOWN, as this change is trying to introduce correct handling of channels.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: Ack; no need for you to update.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: Thanks for the clarification. Let's leave it how it is for now. I'll add a comment explaining why as part of the merge. We can change it later if desired.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
ojw28: This still seems odd to me. Either it's completely redundant information, in which case the check is not necessary (even though the field bit is present), or it's not redundant in which case the header field and the misc code field can differ for a legitimate reason.

The way this code is currently written, we'll always ignore misc codes if the fields differ, either because the header specifies the field we want but the misc code doesn't, or because the header field specifies the field we don't want and we never get as far as checking the misc code field. If this is correct behavior, then presumably such data would never be inserted into the stream in the first place, because it doesn't do anything.

The only logical possibilities I can see are:

1. The field is completely redundant and the check is not necessary.
1. We should be handling misc codes if the field bit matches, even if the header specifies another field.

Neither of which is what this code is currently doing. Please clarify.

zsmatyas: What if we consider the other direction:
`boolean isMiscCode(cc1, cc2)`
cannot be a function deciding if the received byte pair CC1, CC2 is a Misc Code or not, as the byte values considered being a Misc Code vary based on the field value, too. So we need the additional third parameter to have a single function deciding if a byte pair is a Misc Code.

Without the third parameter the function might return false positives.

ojw28: I don't think that addresses my comment. I understand that there _exists_ a field bit, but the statement that it's therefore necessary to check it isn't true in possibility (1) above, because you already know what the field bit will be set to. In possibility (2) it is necessary to check the field bit, but in that possibility I think we need to make changes elsewhere to handle misc codes where the header has specified another field. Which of (1) and (2) is true?

zsmatyas: Let me consult the CEA608 Standard

ojw28: Did you figure anything out? If stuck on this I'd propose we proceed on the assumption that (1) is true (since it's easier). If we discover otherwise later on, we can fix in a subsequent change. Would you be OK with that? If so we'll get this merged (no need to change anything on your side; we can drop the check as part of the merge).

zsmatyas: I still do not understand how (1) can work, as it cannot be redundant as you need CC1, CC2, and the field value at the same time to figure out if a Byte pair is a misc code or not. Until we did not handle Field 2 and channel 2, this was not an issue, the original code only worked for Field 1 channel 1 correctly. It dropped anything that was Field 2 so this issue was not present.

I tried to look up if you can get a Field 2 misc code while processing byes from field 1, but there is no mention for that in the standard, so I would expect that (2) is also not valid.

    8.4 Closed Caption Mode When closed captioning is used on line 21, field 2, it shall conform to all of the applicable specifications and recommended practices as defined for field 1 services with the following differences:
    a)   The non-printing character of the miscellaneous control-character pairs that fall in the range of 0x14, 0x20 to 0x14, 0x2F in field 1, shall be replaced with 0x15, 0x20 to 0x15, 0x2F when used in field 2. 
    b)   The non-printing character of the miscellaneous control-character pairs that fall in the range of 0x1C, 0x20 to 0x1C, 0x2F in field 1, shall be replaced with 0x1D, 0x20 to 0x1D, 0x2F when used in field 2.
    c)   When any of the control codes from 0x01 to 0x0F is used to begin a control-character pair it indicates the beginning of XDS Data. This XDS data should be considered part of a different data channel and can be used to interrupt the captioning data according to the FCC rules for other data channels.

So, when we are processing field 2 byte pairs, we need to add this check for every control code processing. If that flag is not set, the value is not a valid Field 2 Control code, so we should drop it (at least from control code processing). We could handle the value of the flag in handleContorlCode() as well, but probably it is easier to handle it in isMiscCode().

ojw28: > it cannot be redundant

If we only had the two CC bytes (`cc_data_1` and `cc_data_2`) then this would be correct. But that's not all we have. We also have `ccHeader`, and we've already done a field check using it before we've got to this point. How is checking the field twice _not_ redundant?

As part of looking at https://github.com/google/ExoPlayer/issues/5656 I coincidentally discovered that this specific check is causing all captions to be discarded for CC3 for that piece of content. It seems the content has the field as expected in `ccHeader`, but not in the misc codes, resulting in them being dropped. This content works correctly in other players, suggesting that they're not checking the misc codes in this way.

For this reason I think we will proceed assuming (1). If you can find some content that's explicitly broken by doing this, please provide a sample and we can revisit. Thanks!

zsmatyas: To comply to section 8.4 of the CEA608 specification I quoted above, byte pairs should be handled as MISC control codes only if they start with 0x15 instead of 0x14 (for frame 2, so it means for channel 3 and 4).

If we do not check it here, do you suggest adding this to the function HandleMiscCode(CC2) then?

Note: Frame 2 (channel 3 and 4) and channel 2 used to be dropped completely, so this was not materializing as an issue before.

I have found these bugs while playing local Over-The-Air content in the Sunnyvale area. Do you have access to live content here?

ojw28: > To comply to section 8.4 of the CEA608 specification I quoted above, byte pairs should be handled as MISC control codes only if they start with 0x15 instead of 0x14 (for frame 2, so it means for channel 3 and 4).

The part you quoted doesn't say anything about how the client should behave, although this is a bit of an aside to the main problem here:

I feel like we're going round in circles. I've been repeatedly asking a very specific question: There's _already_ a check that we're dealing with data for `selectedField` before this point in code, which checks against the field in `ccHeader`. The question I'm asking is why this check would be any different. If the checks are different, that would imply that a misc code for one field was given a `ccHeader` claiming it's actually for a different field. Why would that make sense? The right way to unblock this discussion is to address this specific question.

zsmatyas: (1) is not true, it is not redundant. When we are on Frame 2 (channel 3 or 4), the current code is checking for 0x14 and 0x1F for CC1 to be the first byte of the pair. 

(2) is not possible, I have tested many streams, but never seen a MISC code on Frame "A" addressing frame "B".

I simply added that check on line 464 because MISC codes on Frame 2 are defined to have the first byte 0x15 or 0x1D. Without also checking the frame value I could not decide when the original 0x14 and 0x1C should not be handled as MISC codes. Is  this clear?

I am OK if you go ahead assuming (1), and we get back to this later.

ojw28: You still have not explicitly addressed how the `ccHeader` check is materially different, which is a requirement for checking the field in the MISC code to not be redundant. Simply stating it's not redundant, again, without explaining when it would be expected to give a different result to the `ccHeader` check, does not address the question.

We will proceed assuming (1) for now. Thanks.

PR #5398, opened 2019-01-16
Add ability to start service with string actions

volkansahin45:
* String actions like ACTION_RELOAD_REQUIREMENTS

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
@@ -196,6 +196,27 @@ public abstract class DownloadService extends Service {
   }
 
   /**
+   * Starts the service, adding an action to be executed.
+   *
+   * @param context A {@link Context}.
+   * @param clazz The concrete download service to be started.
+   * @param action The action string to be executed.
+   *  Can be one of {@link #ACTION_INIT} {@link #ACTION_ADD} {@link #ACTION_RELOAD_REQUIREMENTS} {@link #ACTION_RESTART}
+   * @param foreground Whether the service is started in the foreground.
+   */
+  public static void startWithAction(Context context,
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
```diff
@@ -195,6 +195,27 @@ public static void startWithAction(
...
+  /**
+   * Starts the service, adding an action to be executed.
+   *
+   * @param context A {@link Context}.
+   * @param clazz The concrete download service to be started.
+   * @param action The action string to be executed.
+   *  Can be one of {@link #ACTION_INIT} {@link #ACTION_ADD} {@link #ACTION_RELOAD_REQUIREMENTS} {@link #ACTION_RESTART}
+   * @param foreground Whether the service is started in the foreground.
+   */
+  public static void startWithAction(Context context,
```
erdemguven, line 207: Why do you need this method? 
There are already dedicated methods  which use ACTION_INIT and ACTION_ADD. 
ACTION_RESTART is an internal action.
Looks like we forgot to provide a convenient method for ACTION_RELOAD_REQUIREMENTS but it has been already removed in the development branch.

alikemalocalan (approved):

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
volkansahin45: Why ACTION_RELOAD_REQUIREMENTS removed?

erdemguven: There is now setRequirements() method in DownloadManager so requirements can be set directly.

volkansahin45: Okey thank you.

PR #5438, opened 2019-01-25
Dolby Vision

aujohn:


aujohn:
@googlebot rescan

andrewlewis:
@googlebot rescan

andrewlewis:
CLA verified manually. (Committers are ybai and aujohn on corporate CLA.)

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -92,6 +92,10 @@
 
   // Video specific.
 
+  /**
+   * Codec specific profile value or {@link #NO_VALUE} if unknown or not applicable.
+   */
+  public final int profile;
```
andrewlewis: In other cases we have signaled the profile in `Format.codecs` as an RFC 6381 codecs string. [`MediaCodecUtil.getCodecProfileAndLevel`](https://github.com/google/ExoPlayer/blob/34d0e0c/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecUtil.java#L205) parses this. Could you try switching to that way of signaling the profile, as it would avoid having two ways to do this and reduce the number of files that need to be modified?

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -439,6 +440,10 @@ protected abstract int supportsFormat(MediaCodecSelector mediaCodecSelector,
   protected List<MediaCodecInfo> getDecoderInfos(
       MediaCodecSelector mediaCodecSelector, Format format, boolean requiresSecureDecoder)
       throws DecoderQueryException {
+    if (format.sampleMimeType == MimeTypes.VIDEO_DOLBY_VISION) {
```
andrewlewis: Once `format.codecs` contains the profile and `MediaCodec.getCodecProfileAndLevel` handles getting the profile for DV, this addition and the changes to `MediaCodecSelector` shouldn't be needed any more as `MediaCodecInfo.isCodecSupported` will take care of the capabilities check.

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Parses Dolby Vision configuration data.
+   *
+   * @param data A {@link ParsableByteArray}, whose position is set to the start of the DoVi
+   *     configuration data to parse.
+   * @return A parsed representation of the DoVi configuration data.
+   * @throws ParserException If an error occurred parsing the data.
+   */
+  public static DolbyVisionConfig parse(ParsableByteArray data) throws ParserException {
```
andrewlewis, line 39: Perhaps this can return an RFC 6381-style codecs string with the profile and level and be moved to become a private static method in `AtomParsers`?

andrewlewis (commented):
Thanks!

andrewlewis:
One other thing I meant to ask: are there Dolby Vision streams where it's necessary to queue access units to the decoder that consist of the concatenation of samples read from two separate tracks in the MP4 stream (a base layer + an enhancement layer)? I think I read about this somewhere. Thanks.

aujohn:
> One other thing I meant to ask: are there Dolby Vision streams where it's necessary to queue access units to the decoder that consist of the concatenation of samples read from two separate tracks in the MP4 stream (a base layer + an enhancement layer)? I think I read about this somewhere. Thanks.

There will not be any streams that come in dual tracks. Even dual layer profile will only come in single track. So there wont be any need to concatenate tracks.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
aujohn: Ok. I will change the signaling

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
aujohn: Ok. Will upload another patch

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
aujohn: We were thinking of reusing this even in case of TS extractors. Won't that be a problem if we move this as private static in  AtomParsers?

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1055,6 +1055,10 @@ protected MediaFormat getMediaFormat(
     mediaFormat.setInteger(MediaFormat.KEY_HEIGHT, format.height);
     MediaFormatUtil.setCsdBuffers(mediaFormat, format.initializationData);
     // Set format parameters that may be unset.
+    if(MediaCodecUtil.getCodecProfileAndLevel(format.codecs) != null) {
```
aujohn, line 1058: Dolby Vision implementation of platforms may expose multiple codecs each supporting a profile. Ideally I would like to see the MediaCodecSelector doing the selection work and returning the correct codec. But looks that MediaCodecSelector just selects based on mime-type only.

Please suggest a better method if any.

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
andrewlewis: Ah, in that case fine to leave. Is there a need to introduce the new DolbyVisionConfig type or can it just return the codecs string?

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecSelector.java
```diff
@@ -39,7 +39,7 @@
               MediaCodecUtil.getDecoderInfos(mimeType, requiresSecureDecoder);
           return decoderInfos.isEmpty()
               ? Collections.emptyList()
-              : Collections.singletonList(decoderInfos.get(0));
+              : decoderInfos;
```
aujohn: May be, assuming the first codec to the most compatible has to be checked with all attributes in MediaCodecSelector.  Let me check this again to make sure this is correct

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
andrewlewis: We've been planning to propagate the profile information to the MediaCodecSelector for a while but haven't got round to it yet. This is also needed for picking the right VP9 video codec (hardware/software) for HDR vs non-HDR profiles.

Would be okay to support only the 'one DV decoder' case in this change, and I'll make a later commit that handles this in a way that works for arbitrary codecs where we know profile capabilities?

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -768,7 +768,11 @@ private void maybeInitCodecWithFallback(
         return;
       }
       try {
-        initCodec(codecInfo, crypto);
+        if (codecInfo != null && !codecInfo.isCodecSupported(inputFormat.codecs)) {
+            availableCodecInfos.removeFirst();
+        }else {
+            initCodec(codecInfo, crypto);
```
aujohn: This is just to select the codec with the right profile. 
I am almost sure another method exists.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecSelector.java
aujohn: I could have used MediaCodecSelector.DEFAULT_WITH_FALLBACK. But I dont want to change DefaultRendererFactory. I will revert this change

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
aujohn: I think we can think of it after the commit for Dolby Vision TS extractor. It also needed some changes for profile propagation.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
aujohn: I can make the change. But do you think the 'one DV decoder' may fail if the device lists other profiles first than what is currently streamed?

library/core/src/main/java/com/google/android/exoplayer2/util/MimeTypes.java
```diff
@@ -211,6 +212,9 @@ public static boolean isApplication(@Nullable String mimeType) {
     codec = Util.toLowerInvariant(codec.trim());
     if (codec.startsWith("avc1") || codec.startsWith("avc3")) {
```
aujohn, line 213: avc3 will have conflicts with Dolby Vision Profile 9. 
Any recommended ways to list both Dolby Vision codec profile 9 and AVC codec during MediaCodecSelection ?

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
andrewlewis: In `MediaCodecInfo` we iterate through the profileLevels of the codec until we find one which has the required profile so I think the order that the profiles are listed doesn't matter.

So to summarize I think if this change doesn't modify MediaCodecSelector, DV will still work if there exists a codec with the DV MIME type in its supported types, and among its profileLevels there's the required profileLevel.

(I'll take care of doing a later change to pass in the codecs info to `MediaCodecSelector` to pick non-primary decoders based on profile support, for the benefit of those devices with multiple components listed for the type.)

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecSelector.java
andrewlewis: Per the other thread about codec selection, I agree this can be reverted and there should be no changes to this file. (You are absolutely right that we should support selection based on profile though. I'll make a later change for that.)

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
andrewlewis: If needed for the later change then fine to leave. Thanks.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1055,6 +1055,10 @@ protected MediaFormat getMediaFormat(
...
+      MediaFormatUtil.maybeSetInteger(mediaFormat,
+          MediaFormat.KEY_PROFILE, MediaCodecUtil.getCodecProfileAndLevel(format.codecs).first);
```
andrewlewis, line 1060: https://developer.android.com/reference/android/media/MediaFormat#KEY_PROFILE is documented as being needed to for encoders. Is it definitely necessary to set it here or would it work without?

library/core/src/main/java/com/google/android/exoplayer2/util/MimeTypes.java
andrewlewis: Hm, this is a bit inconvenient! Do you know whether "avc3.09" can ever appear for a non-DV H.264 stream? If not I guess we can add a special case that checks for that string. Otherwise I suggest we just say that we won't support profile 9 in this change, and figure out how to handle this later.

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
```diff
@@ -0,0 +1,73 @@
...
+      return new DolbyVisionConfig(Format.NO_VALUE, Format.NO_VALUE);
+    }
+    DolbyVisionConfig config = new DolbyVisionConfig(profile, level);
+    return config;
+  }
+  private DolbyVisionConfig(int profile, int level) {
+
+    this.profile = profile;
+    this.level = level;
+    if (profile < MediaCodecInfo.CodecProfileLevel.DolbyVisionProfileDvheStn) {
```
andrewlewis: Maybe I am looking at the wrong document but appears that profile ID 4, 5, 7 means 'dvhe', 8 means 'hev1' and 9 means 'avc3', whereas the `MediaCodecInfo` constant used here is much larger (`DolbyVisionProfileDvheStn` is 32). Could you point me to the documentation on interpretation of dv_profile from the dvcC/dvvC box?

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
andrewlewis: The expected flow is that `MediaCodecVideoRenderer.supportsFormat` gets called, we pick the first DV decoder, then we check if its profile/level support includes the input media. If so, then we know we can play the input media so the renderer will be used and we'll initialize the first DV decoder. Picking a non-primary DV decoder (further down the list) can wait for a later change.

andrewlewis (commented):
Looks good to merge once these comments are addressed. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/video/DolbyVisionConfig.java
aujohn: You are right. I will correct this.

library/core/src/main/java/com/google/android/exoplayer2/util/MimeTypes.java
aujohn: "avc3.09" is DV stream. We can make this change for "one DV Decoder".

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
aujohn: The implementations in mobile phone looks for the profile information. Any workaround is also fine.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
aujohn: Ok

aujohn:
Can we take a look at the latest patchset?

andrewlewis:
Looks good. Thanks for the contribution. We'll merge this internally and this PR should be marked as merged soon, when the corresponding changed is pushed to the development branch.

library/core/src/main/java/com/google/android/exoplayer2/extractor/mp4/AtomParsers.java
```diff
@@ -882,6 +886,12 @@ private static void parseVideoSampleEntry(ParsableByteArray parent, int atomType
         HevcConfig hevcConfig = HevcConfig.parse(parent);
         initializationData = hevcConfig.initializationData;
         out.nalUnitLengthFieldLength = hevcConfig.nalUnitLengthFieldLength;
+      } else if (childAtomType == Atom.TYPE_dvcC || childAtomType == Atom.TYPE_dvvC) {
```
andrewlewis, line 889: Quick question: is it intentional that this branch doesn't check that mimeType is null? I am wondering if we expect the containing loop to parse any other boxes like avcC or hvcC, and whether dvcC/dvvC is guaranteed to come after the others.

andrewlewis: If this is intentional, could you also explain how fallback works? Is it just for profiles 0 and 2, and do we need to do anything to the access units to make them compatible with non-DV decoders?

aujohn: In all DV streams, avcC and hcCC boxes are guaranteed to come before dvcC/dvvC boxes. non-DV decoders will just ignore the NALUs thats not known to them. and it will only take the base layer and not Dolby Vision metadata.

andrewlewis: Is it the case that none of the supported profiles (4, 5, 8, 9) support backwards compatibility? Just checking that we don't need to implement falling back to an AVC or HEVC decoder in the renderer.

aujohn: Maybe I didn't the question properly. Profile 4,8,9 all are backward compatible profiles. Only Profile 5 is non-backward compatible. If a DV decoder is present, then 4,5,8,9 profiles should be decoded through DV(video/dolby-vision or its variant) decoder. If no DV decoder present profile 4 and 8 can be send through "video/hevc" decoder and profile 9 can be send through "video/avc" decoder. Profile 5 needs DV(video/dolby-vision or its variant) decoder

andrewlewis: Is it the case that after this change, devices that could previously play media in DV profiles 4, 8 and 9 but lack a DV decoder will no longer be able to play the streams? If so, I think we should restrict to profile 5 for now, until we have a working fallback path. If you agree I can take care of this when I merge the change internally (no need to modify anything here).

aujohn: I agree that we should keep only profile 5 in this change. Feel free to make the change. But can you also let me know the fallback that you have in mind? I was thinking of calling the MediaCodecUtil.getDecoderInfos(MimeType, false); to see if we have dv decoders on the platform before assigning the mime-type for proile 4,8,9. Will that work?

andrewlewis: Not sure yet but I think we should probably put this fallback in the renderer or in MediaCodecUtil (to avoid having the extractor's behavior depend on the device decoding capabilities). The codec selector could return a non-DV decoder for those profiles if it detects that no DV decoder is present.

andrewlewis:
This was merged in 9b6caf9edaa873d2b2b77a8b449e36325273124c, so I will close the pull request (it should have been marked as merged by Copybara but unfortunately some tags were missing; we've put in place a check that should ensure this doesn't happen in future). Thanks for the contribution. I'll get in touch once the codec selection fix we discussed is submitted.

Issue #5040, opened 2018-11-01
Update time bar more frequently for short videos


perracodex:
When playing a video the DefaultTimerBar updates the current seek position at 1 second intervals.
Is there any setting to set it at a smaller time precision, so the seekbar updates more smoothly for short videos?

ojw28:
Do you just mean the frequency with which the current position is updated during playback (i.e. nothing to do with the user seeking to a different position)?

If so then the frequency is controlled in `PlayerControlView.updateProgress` rather than `DefaultTimeBar`. It's a bit more complicated than once every second. For example it'll update more frequently if the playback speed is faster than 1x. There is currently no easy way to change the update rate.

Perhaps we should further enhance the calculation take into account the media duration, and have it automatically update more frequently for short duration videos. That's probably better than requiring the application developer to set it themselves. Would that make sense?

perracodex:
Yes this is exactly what I meant.
For long videos it looks fine, but for shorter videos of just a few seconds would be better to see a smoother update instead of a second to second big jump.

ojw28:
Thanks for confirming! We should probably do something like calculate what duration of elapsed time corresponds to 1dp of the screen, and incorporate that into the calculation. Doing this would also update the time bar more frequently on physically larger devices than on small ones, which also seems like a good idea.

I've already marked this as an enhancement, so we can use it for tracking.

viethoa:
@ojw28 might I know the issue is fixed? And that pretty good if you can share your plan on it. Currently, I am playing a 30seconds video and this looks not really good now.

szaboa:
@ojw28 
I had a quick look on this and I have an idea.. maybe it's the same as yours (?), I am not sure.

So we could specify the update length (thickness) in dp, let's say - update the time bar's progress always with 20dp, regardless of the duration.

For example:
`timeBarWidth`= 400 dp (this will vary on devices / tablets).
`duration`= 20 sec
`updateDp`= 20 dp

With these, we could calculate how many times we need to update the time bar with `updateDp` dps to fill it up, `numberOfUpdates`= `timeBarWidth`/ `updateDp`= 20.

To get how frequently we need to update, we just need to calculate `duration`/ `numberOfUpdate ` which is 1 sec in this case.

If the duration is 10 sec, then with this logic the time bar would need to be updated in every 0.5 secs.

[...]

PR #5496, opened 2019-02-10
Tune time bar update based on duration

szaboa:
Hi @tonihei , @ojw28 
I've started to implement the thing what we've discussed on #5040.

Two things are unclear where I would need some guidance:
1. I didn't quite understand some calculations (line 841-845). Why exactly we need those? How can we include them into the current implementation if needed? 
2. How this would work for live streams? How the `duration `works in that cases?

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
@@ -205,6 +205,7 @@ public class DefaultTimeBar extends View implements TimeBar {
   private final CopyOnWriteArraySet<OnScrubListener> listeners;
   private final int[] locationOnScreen;
   private final Point touchPosition;
+  private final float density;
 
   private int keyCountIncrement;
   private long keyTimeIncrement;
@@ -218,8 +219,6 @@ public class DefaultTimeBar extends View implements TimeBar {
   private int adGroupCount;
   private @Nullable long[] adGroupTimesMs;
   private @Nullable boolean[] playedAdGroups;
-
-  private int densityDpi;
 
   /** Creates a new time bar. */
   // Suppress warnings due to usage of View methods in the constructor.
@@ -244,7 +243,7 @@ public class DefaultTimeBar extends View implements TimeBar {
     // Calculate the dimensions and paints for drawn elements.
     Resources res = context.getResources();
     DisplayMetrics displayMetrics = res.getDisplayMetrics();
-    densityDpi = displayMetrics.densityDpi;
+    density = displayMetrics.density;
     fineScrubYThreshold = dpToPx(displayMetrics, FINE_SCRUB_Y_THRESHOLD_DP);
     int defaultBarHeight = dpToPx(displayMetrics, DEFAULT_BAR_HEIGHT_DP);
     int defaultTouchTargetHeight = dpToPx(displayMetrics, DEFAULT_TOUCH_TARGET_HEIGHT_DP);
@@ -451,8 +450,8 @@ public class DefaultTimeBar extends View implements TimeBar {
   }
 
   @Override
-  public int getTimeBarWidth() {
-    return pxToDp(densityDpi, getWidth());
+  public int getTimeBarWidthDp() {
+    return pxToDp(density, getWidth());
   }
 
   // View methods.
@@ -844,7 +843,7 @@ public class DefaultTimeBar extends View implements TimeBar {
     return (int) (dps * displayMetrics.density + 0.5f);
   }
 
-  private static int pxToDp(int densityDpi, int px) {
-    return (int) (px / ((float) densityDpi / DisplayMetrics.DENSITY_DEFAULT));
+  private static int pxToDp(float density, int px) {
+    return (int) (px / density);
   }
 }
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
@@ -219,6 +219,8 @@ public class DefaultTimeBar extends View implements TimeBar {
   private @Nullable long[] adGroupTimesMs;
   private @Nullable boolean[] playedAdGroups;
 
+  private int densityDpi;
+
   /** Creates a new time bar. */
   // Suppress warnings due to usage of View methods in the constructor.
   @SuppressWarnings("nullness:method.invocation.invalid")
@@ -242,6 +244,7 @@ public class DefaultTimeBar extends View implements TimeBar {
     // Calculate the dimensions and paints for drawn elements.
     Resources res = context.getResources();
     DisplayMetrics displayMetrics = res.getDisplayMetrics();
+    densityDpi = displayMetrics.densityDpi;
     fineScrubYThreshold = dpToPx(displayMetrics, FINE_SCRUB_Y_THRESHOLD_DP);
     int defaultBarHeight = dpToPx(displayMetrics, DEFAULT_BAR_HEIGHT_DP);
     int defaultTouchTargetHeight = dpToPx(displayMetrics, DEFAULT_TOUCH_TARGET_HEIGHT_DP);
@@ -445,6 +448,11 @@ public class DefaultTimeBar extends View implements TimeBar {
     this.adGroupTimesMs = adGroupTimesMs;
     this.playedAdGroups = playedAdGroups;
     update();
+  }
+
+  @Override
+  public int getTimeBarWidth() {
+    return pxToDp(densityDpi, getWidth());
   }
 
   // View methods.
@@ -834,5 +842,9 @@ public class DefaultTimeBar extends View implements TimeBar {
 
   private static int dpToPx(DisplayMetrics displayMetrics, int dps) {
     return (int) (dps * displayMetrics.density + 0.5f);
+  }
+
+  private static int pxToDp(int densityDpi, int px) {
+    return (int) (px / ((float) densityDpi / DisplayMetrics.DENSITY_DEFAULT));
   }
 }
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
@@ -196,6 +196,9 @@ public class PlayerControlView extends FrameLayout {
   public static final int MAX_WINDOWS_FOR_MULTI_WINDOW_TIME_BAR = 100;
 
   private static final long MAX_POSITION_FOR_SEEK_TO_PREVIOUS = 3000;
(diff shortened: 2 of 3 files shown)


tonihei:
1. 
mediaTimeUpdatePeriodMs is the number of media time seconds needed until one real time second passed (that's our designated update interval). For example for playback speed 0.5, this is 500ms.
mediaTimeDelayMs is the media time delay needed from the current position until the next update is due. Let's say our position is 102300ms, then we need to wait 500 - (102300 % 500) = 200ms in media time until it makes sense to update. 
delayMs is then calculated as the real time corresponding to the media time we'd like to wait. So in the example, we would delay the update by 200/0.5 = 400ms.

I agree that is is very complicated and took me a while to figure out what's going on here. Feel free to simplify this with your change, but keep in mind different playback speeds when doing this.

2.
The duration of live windows is either C.TIME_UNSET in which case we don't display a progress at all (because we don't anything about it), or it's the duration of the current live window and will change from time to time. For this purpose it's fine to assume that the live duration works the same as a VOD duration and you can also assume the duration itself won't change (even if the live window moves).

szaboa:
@tonihei 
1. I think I understand now, thank you.
 I've updated the calculations to follow this existing logic (and take the current position into consideration). I don't know how this could be simplified, however I've added some comments to be easier to understand. Please have a look on the changes and let me know if something needs to adjusted. 

2. Got it!

library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
```diff
@@ -219,6 +219,8 @@
   private @Nullable long[] adGroupTimesMs;
   private @Nullable boolean[] playedAdGroups;
 
+  private int densityDpi;
```
tonihei: Can you make that final and move up to the other final variables?

library/ui/src/main/java/com/google/android/exoplayer2/ui/TimeBar.java
```diff
@@ -97,6 +97,13 @@
   void setAdGroupTimesMs(@Nullable long[] adGroupTimesMs, @Nullable boolean[] playedAdGroups,
       int adGroupCount);
 
+    /**
+     * Get width of time bar in dps
+     *
+     * @return Width of time bar in dps
+     */
+    int getTimeBarWidth();
```
tonihei: Can you add dp to the name, i.e. getTimeBarWidthDp?

library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
```diff
@@ -835,4 +843,8 @@ public static int getDefaultPlayedAdMarkerColor(int adMarkerColor) {
   private static int dpToPx(DisplayMetrics displayMetrics, int dps) {
     return (int) (dps * displayMetrics.density + 0.5f);
   }
+
+  private static int pxToDp(int densityDpi, int px) {
+    return (int) (px / ((float) densityDpi / DisplayMetrics.DENSITY_DEFAULT));
```
tonihei: Isn't (float) densityDpi / DisplayMetrics.DENSITY_DEFAULT the same as displayMetrics.density? See method dpToPx avove. If so, it's probably better to use the same conversion such that both methods obviously do the same thing, just the other way round.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
```diff
@@ -196,6 +197,9 @@
   public static final int MAX_WINDOWS_FOR_MULTI_WINDOW_TIME_BAR = 100;
 
   private static final long MAX_POSITION_FOR_SEEK_TO_PREVIOUS = 3000;
+  private static final int DEFAULT_UPDATE_DP = 5;
+  private static final long MIN_UPDATE_FREQUENCY_MS = 50;
```
tonihei, line 201: It seems we already use a minimum delay of 200ms for the cases where the playback speed is > 5.0f. Can you change that accordingly, and also use the same constants in updateProgress()?

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
```diff
@@ -196,6 +197,9 @@
   public static final int MAX_WINDOWS_FOR_MULTI_WINDOW_TIME_BAR = 100;
 
   private static final long MAX_POSITION_FOR_SEEK_TO_PREVIOUS = 3000;
+  private static final int DEFAULT_UPDATE_DP = 5;
```
tonihei, line 200: Just curious: How did you choose this value?

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
```diff
@@ -835,13 +839,33 @@ private void updateProgress() {
...
+
+          int timeBarWidth = timeBar.getTimeBarWidth();
+          // Calculate how many updates needs to be done with DEFAULT_UPDATE_DP
+          // to fill up the time bar
+          int numberOfUpdates = timeBarWidth / DEFAULT_UPDATE_DP;
+
+          // Calculate the designated update interval, taking duration into consideration as well
+          long mediaTimeUpdatePeriodMs = duration / numberOfUpdates;
+
+          // Limit the designated update interval, to avoid too frequent / infrequent updates
```
tonihei: I think that should happen at the very end after taking playback speed into account. Otherwise the playback speed may still scale the delay to very large or small values.

tonihei: Be aware though, that the duration can be 0 (e.g. for live streams with unset duration).

tonihei: Same is probably true for the timeBarWidth? Could be 0 if the view is embedded in a strange way where the the timebar isn't really visible anymore.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
```diff
@@ -835,13 +839,33 @@ private void updateProgress() {
         if (playbackSpeed <= 0.1f) {
```
tonihei: These test for speed <0.1 and speed > 5.0 where specifically chosen because all speed values between 0.1 and 5.0 would result in a delay between 200ms and 1000ms anyway. 

As you now limit to the delay in another way and based on other criteria, I think you should be able to just remove these extra tests and let all cases go through the central code.

tonihei (commented):
General approach looks good, thanks! I've added some suggestions to the code for further improvements. Please take a look.

library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
szaboa: Yes, it's the same, I can use that.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
szaboa: Not sure if I understood correctly. Should I change MIN_UPDATE_FREQUENCY_MS to 200ms? The other half is done.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
szaboa: By checking a couple of values, how it looks visually. 20dp was too high and the progress jumped way too much. In my opinion of course.

szaboa:
Done, let me know if any further changes are needed :)

library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java
szaboa: But I can't pass the DisplayMetrics instead of density, because the width of the view can't be obtained yet in the ctr.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
tonihei: Thinking about it again, 50ms might actually be a better minimum value. :) 
After your change. we'll only use such low values for very short video durations and given that the human eye can easily see 20 updates per second, this seems like a good value to get smooth movements.

library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerControlView.java
```diff
@@ -832,21 +836,39 @@ private void updateProgress() {
...
+
+        if (timeBarWidth == 0 || duration == 0) {
+          delayMs = MAX_UPDATE_FREQUENCY_MS;
+        } else {
+          // Calculate how many updates needs to be done with DEFAULT_UPDATE_DP
+          // to fill up the time bar
+          int numberOfUpdates = timeBarWidth / DEFAULT_UPDATE_DP;
+
+          // Calculate the designated update interval, taking duration into consideration as well
+          long mediaTimeUpdatePeriodMs = duration / numberOfUpdates;
```
tonihei: Sorry for being nitty about this, but this can still result in divisions by zero.  For example if timeBarWidth==3. The same is true for the other two divisions on line 852 and 861. 

Maybe the easiest solution is to move this code to a separate private static method getTimeBarUpdateDelayMs. Then you do the steps and return early if an intermediate value is outside the valid range. This avoids a lot of nested if conditions. Something like:
int numberOfUpdate = ...
if (numberOfUpdates <= 0) {
  return MAX_UPDATE_FREQUENCY_MS;
}
etc.

[ We usually put static methods below non-static ones, so please move that new method down to the other static methods. Thanks. ]

szaboa:
Moved the calculations to a separate method, and added some verifications. Let me know if it's OK or something else needs to be adjusted.

tonihei:
Thanks! Looks good now. I'll merge this change next week.

tonihei:
Going to merge this PR soon. Note that we did some updates to your PR:
 - We moved part of the logic into `DefaultTimeBar` so that the return value is independent of how the time bar actually looks like (e.g. doesn't rely on it being horizontal).
 - Removed the DEFAULT_UPDATE_DP because the only reason for this is to limit the number of updates. The same can be achieved by setting the minimum update interval. 
- Also made the minimum update interval configurable in the view and with a method. That is because we had some discussion around the best value. Using something like 33ms (=30fps) gives a nice smooth movement, but increases CPU usages quite a bit while the time bar is visible. So we went for a default of 200ms, which corresponds more to a step-wise update with no serious CPU increase. If you prefer the smoother version and you don't care about the CPU because the time bar is only visible for a short amount of time anyway, you can adjust this value to your preference.

szaboa:
Alright, sounds good!

PR #5508, opened 2019-02-13
Dev v2 Enable rewrite of the Audio KID for separate license requests with different KIDs

bjoernharrer:
This PR adds a configuration option “audio_kid” to media.exolist.json and aims at rewriting the Audio KID of an MPEG-DASH stream within ExoPlayer. This enables separate license requests with different security levels for Audio and Video tracks. These modifications are useful for legacy content, where all tracks use the same KID. Widevine recommends to use different KIDs for different tracks. 

“drm_multi_session” needs to be set to “true”. In HttpMediaDrmCallback.java function executeKeyRequest() the differentiation between audio license request and video license request can be made. 

Note that the current code to parse ‘tenc’ from the init segment and PSSH from the MPD is not generic enough. We first wanted to check if there is enough support to merge this PR.


marcbaechinger:
The diff which should be merged in with this PR appears to me to be like a diff from ExoPlayer 2.9.0 to 2.9.5 or something. I can't see a commit coming from a user bjoernharrer. Please correct me if I'm mistaken. 

May I kindly ask you to check whether something went wrong with creating that pull request and sign the CLA before we look into this some further?

Many thanks.

andrewlewis:
Please could you send this to the `dev-v2` branch instead?

ojw28:
Closing this because the pull request is to the wrong branch and includes a lot of unrelated changes. We'd take a look at pull request sent to `dev-v2` to see exactly what's involved, although the description of the issue sounds pretty strange to me. I think the correct solution is to re-transcode the content, not push hacks/complexity into client code. Thanks!

PR #5525, opened 2019-02-18
Make DefaultRenderersFactory subclass-friendly

noamtamim:
Change the `private` fields to `protected` so that subclasses can use them. 

An alternative might can be adding protected/public getters, but that seems to be an overkill.

Currently our workaround is having our own copies of private allowedVideoJoiningTimeMs and playClearSamplesWithoutKeys.


ojw28:
Could you clarify where you need `allowedVideoJoiningTimeMs` and `playClearSamplesWithoutKeys`. They're already passed to the methods that should need them (e.g. `allowedVideoJoiningTimeMs` is passed to `buildVideoRenderers`). If you're overriding `createRenderers` directly then perhaps at that point you should just be forking the whole class?

noamtamim:
@ojw28 you're right, we missed the fact that `allowedVideoJoiningTimeMs` and `playClearSamplesWithoutKeys` are passed to `buildVideoRenderers`.

PR #5530, opened 2019-02-19
Sets whether an option is available for auto selecting the renderer

BROUDING:
For any renderer, there are "needs" to remove "AUTO" option as "NONE" option just like `setShowDisableOption`


BROUDING:
Wrong pull request.

PR #5531, opened 2019-02-19
Sets whether an option is available for auto selecting the renderer

BROUDING:
There are people who wants this option.

When you're selecting subtitle, "Auto" option is unnecessary.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
@@ -173,6 +173,15 @@ public class TrackSelectionView extends LinearLayout {
   }
 
   /**
+   * Sets whether an option is available for auto selecting the renderer.
+   *
+   * @param showAutoOption Whether the disable option is shown.
+   */
+  public void setShowAutoOption(boolean showAutoOption) {
+    defaultView.setVisibility(showAutoOption ? View.VISIBLE : View.GONE);
+  }
+
+  /**
    * Sets the {@link TrackNameProvider} used to generate the user visible name of each track and
    * updates the view with track names queried from the specified provider.
    *


ojw28:
The problem with doing this is that "auto" is the default selection. If you hide it, it's unclear what should be selected instead (selecting nothing is strange, and selecting "None" isn't guaranteed to do the right thing).

What should probably happen is that instead of "auto" being the default selection, we instead actually select what the default selection is. "Auto" then becomes a "reset" button that'll reset back to the default selection. At this point having an option to hide the "reset" button makes a lot of sense.

BROUDING:
Okay, here I may have misunderstood the usage of the `-flag` for
`createTextFormatSample`.
Some of them are setting it `DEFAULT`, some are `FORCED`, some are
`AUTO_SELECT`...

But most of all, `auto` selecting for videos or audios, I can understand,
but for subtitles.. I really think it's even hard to understand for service users
too. (e.g. watching NETFLIX, and I click on changing subtitle button and it shows current option is on  "AUTO")...?
And that's why I wanted to remove that option from the first place.

Since not using this method doesn't affect anything else anyway, how about
throwing `illegalStateException` in `setShowAutoOption` method when the
user didn't set default value?
And so there will be `setDefaultOption( int defaultOption )` and make
constant value for `None` option = -1

(I really want to remove `Auto` option...)

2019년 2월 19일 (화) 오후 10:53, Oliver Woodman <notifications@github.com>님이 작성:

tonihei:
I'll close this pull request because we won't add an option to remove "Auto" as it is for the reasons outlined above. Having "auto" as the reset button is also slightly difficult, because we wouldn't be able to distinguish between an override and an auto-selection. If you like to display the currently selected tracks, it's probably best to create an UI element which queries `player.getCurrentTrackSelections()` and creates a suitable user-readable String.

PR #5532, opened 2019-02-19
Added Constants for Renderer indexes

BROUDING:
Easy to indicate what those integer values for.

### DIFF
--- a/demos/main/src/main/java/com/google/android/exoplayer2/demo/TrackSelectionDialog.java
+++ b/demos/main/src/main/java/com/google/android/exoplayer2/demo/TrackSelectionDialog.java
@@ -143,7 +143,7 @@ public final class TrackSelectionDialog extends DialogFragment {
    * @param rendererIndex Renderer index.
    * @return Whether the renderer is disabled.
    */
-  public boolean getIsDisabled(int rendererIndex) {
+  public boolean getIsDisabled(@C.RendererIndex int rendererIndex) {
     TrackSelectionViewFragment rendererView = tabFragments.get(rendererIndex);
     return rendererView != null && rendererView.isDisabled;
   }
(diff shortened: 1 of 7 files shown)


ojw28:
Whilst it's the common case, there's no guarantee that renderers are always indexed as you've defined these constants. In particular if you have one or more of the software decoder extensions enabled, there will be additional renderers and the indices will change. If you want to know the type of a renderer at any given index, you should use `Player.getRendererType` or `MappedTrackInfo.getRendererType` instead.

BROUDING:
Well my point was "initializing the renderer".
So when you make the `dialogPair.first.show()` call for example, you'd set
the `rendererIndex` in `getDialog`, and that was really hard to recognize
for me to change that option until I finally understand most of the classes
in the library. So that's why I suggested those constant values.

So I may not have understood this `rendererIndex` fully... Can you give me
example case for extra renderers? I just cannot think of other option
renderers yet..? (And even in that case, you can just add another constant
value in `C.java` and be more indicative..

2019년 2월 19일 (화) 오후 10:56, Oliver Woodman <notifications@github.com>님이 작성:

ojw28:
You can inject whatever renderers you want into an ExoPlayer instance, including custom implementations, if you really want to.

You can see what the default case does [here](https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/DefaultRenderersFactory.java#L323). Note in particular there are various extension renderers (e.g. Libvpx, Opus, Flac software decoder renderers) that'll be included if they're available in the binary.

BROUDING:
But most of the cases for `rendererIndex` would be `video, audio, subtitle` so I think having constant values for them are very helpful. And other heavy users can add constants themselves.

I was just thinking, since it's an open library, it should be more straightforward to "light users" than "heavy users" (people who'd use extension renderers)

ojw28:
`IntDef` isn't appropriate for constant values that are only correct most of the time. Lint checks may give you a compile error or a warning (depending on how you configure lint) if you try and pass any other values, which is something other developers legitimately need to do for non-standard use cases.

The proposal also makes the API a bit strange. In particular `getRendererType(@C.RendererIndex int rendererIndex)` doesn't make sense when all of the constants you can pass to the method contain a type already in their name.

BROUDING:
I think I was on first-time-user-perspective too much.
In this way, non-standard-user cannot use this funtion untill they commit
another file.
I totally understand now.

It still is confusing or was hard to figure out until I dig into most of
the important classes. (Which was actually helpful later on)

I think adding guidance about this part to "Developer guide" would be
really helpful and saving your team's time..?

2019년 2월 20일 (수) 오후 9:26, Oliver Woodman <notifications@github.com>님이 작성:

ojw28:
I think track selection in general needs better describing in the guide, of which that would be a part.

Issue #5529, opened 2019-02-18
AudioPurpose parsing of subtitle track from DASH manifest


szaboa:
### Issue description

**Request for enhancement**
Parse `<Accessibility>` tag of subtitle tracks from mpd (DASH), more specifically the `AudioPurpose` part.
Use case: display HOH (hard of hearing) suffix after the subtitle track's language in the chooser. Useful when you have two subtitle tracks in the same language, but one of them is HOH.

### Link to test content
We have some production streams, but trying to find a sample stream where this is present.

### Version of ExoPlayer being used
I've searched whether this is already implemented or not in version 2.9.5. I didn't find support for this.

**Note.** If this request is accepted, then I would like to work on this and submit a pull request when it's done.

ojw28:
Where is `AudioPurpose` defined? I can't find it in the DASH spec, or in the DASH-IF guidelines. The only relevant result when I try and search is this issue.

szaboa:
`Accessibility` is a metadata (or descriptor?) of a `<AdaptationSet>` which can contain this `AudioPurpose`.

I guess it should be defined here: https://dashif.org/identifiers/role-and-accessibility but this link gives 404.

A small description can be found here: https://www.brendanlong.com/common-informative-metadata-in-mpeg-dash.html

A more official doc is this, where the `AudioPurpose` is described: https://www.etsi.org/deliver/etsi_ts/103200_103299/103285/01.01.01_60/ts_103285v010101p.pdf

ojw28:
Could you please paste a relevant snippet from one of your manifests (it doesn't need to be a fully functioning piece of test content)?

ojw28:
As an additional point, is it not possible for your manifests to use a DASH-IF recommended way of doing this? Which I think is probably to use:

`@schemeIdUri="urn:mpeg:dash:role:2011"`
`@value="enhanced-audio-intelligibility"`

Similarly I think:

`@schemeIdUri="urn:mpeg:dash:role:2011"`
`@value="description"`

is probably equivalent to "visually impaired" in the scheme you're referring to. There's not going to be a good DASH interoperability story if everyone is using different schemes for the same thing.

szaboa:
Relevant snippet:

```
<!--Text-->
<AdaptationSet id="2" lang="nor" segmentAlignment="true">
      <Role schemeIdUri="urn:mpeg:dash:role:2011" value="subtitle"/>
[...]

PR #5578, opened 2019-02-27
Parse role and accessibility descriptor values from adaptation sets

szaboa:
Pull request for Issue #5529 .

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/C.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/C.java
@@ -977,52 +977,36 @@ public final class C {
   public static final int NETWORK_TYPE_OTHER = 8;
 
   /**
-   * Adaptation set's role descriptor value (ISO 23009-1)
+   * Adaptation set's role and accessibility descriptor value.
    */
   @Documented
   @Retention(RetentionPolicy.SOURCE)
-  @IntDef({
-    ROLE_UNSET,
-    ROLE_MAIN,
-    ROLE_ALTERNATE,
-    ROLE_SUPPLEMENTARY,
-    ROLE_COMMENTARY,
-    ROLE_DUB,
-    ROLE_EMERGENCY,
-    ROLE_CAPTION,
-    ROLE_SIGN
+  @IntDef(
+      flag = true,
+      value = {
+        ROLE_FLAGS_MAIN,
+        ROLE_FLAGS_ALTERNATE,
+        ROLE_FLAGS_SUPPLEMENTARY,
+        ROLE_FLAGS_COMMENTARY,
+        ROLE_FLAGS_DUB,
+        ROLE_FLAGS_EMERGENCY,
+        ROLE_FLAGS_CAPTION,
+        ROLE_FLAGS_SIGN,
+        ROLE_FLAGS_ENHANCED_AUDIO_INTELLIGIBILITY,
+        ROLE_FLAGS_DESCRIPTION
   })
-  public @interface Role {}
+  public @interface RoleFlags {}
 
-  public static final int ROLE_UNSET = -1;
-  public static final int ROLE_MAIN = 0;
-  public static final int ROLE_ALTERNATE = 1;
-  public static final int ROLE_SUPPLEMENTARY = 2;
-  public static final int ROLE_COMMENTARY = 3;
-  public static final int ROLE_DUB = 4;
-  public static final int ROLE_EMERGENCY = 5;
-  public static final int ROLE_CAPTION = 6;
-  public static final int ROLE_SIGN = 7;
-
-  /**
-   * Adaptation set's accessibility descriptor value (ISO 23009-1)
-   */
-  @Documented
-  @Retention(RetentionPolicy.SOURCE)
-  @IntDef({
-    ACCESSIBILITY_UNSET,
-    ACCESSIBILITY_ENHANCED_AUDIO_INTELLIGIBILITY,
-    ACCESSIBILITY_DESCRIPTION,
-    ACCESSIBILITY_CAPTION,
-    ACCESSIBILITY_SIGN
-  })
-  public @interface Accessibility {}
-
-  public static final int ACCESSIBILITY_UNSET = -1;
-  public static final int ACCESSIBILITY_ENHANCED_AUDIO_INTELLIGIBILITY = 1;
-  public static final int ACCESSIBILITY_DESCRIPTION = 2;
-  public static final int ACCESSIBILITY_CAPTION = 3;
-  public static final int ACCESSIBILITY_SIGN = 4;
+  public static final int ROLE_FLAGS_MAIN = 1;
+  public static final int ROLE_FLAGS_ALTERNATE = 1 << 1;
+  public static final int ROLE_FLAGS_SUPPLEMENTARY = 1 << 2;
+  public static final int ROLE_FLAGS_COMMENTARY = 1 << 3;
+  public static final int ROLE_FLAGS_DUB = 1 << 4;
+  public static final int ROLE_FLAGS_EMERGENCY = 1 << 5;
+  public static final int ROLE_FLAGS_CAPTION = 1 << 6;
+  public static final int ROLE_FLAGS_SIGN = 1 << 7;
+  public static final int ROLE_FLAGS_ENHANCED_AUDIO_INTELLIGIBILITY = 1 << 8;
+  public static final int ROLE_FLAGS_DESCRIPTION = 1 << 9;
 
   /**
    * Converts a time in microseconds to the corresponding time in milliseconds, preserving
--- a/library/core/src/main/java/com/google/android/exoplayer2/C.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/C.java
@@ -977,6 +977,54 @@ public final class C {
   public static final int NETWORK_TYPE_OTHER = 8;
 
   /**
+   * Adaptation set's role descriptor value (ISO 23009-1)
+   */
+  @Documented
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({
+    ROLE_UNSET,
+    ROLE_MAIN,
+    ROLE_ALTERNATE,
+    ROLE_SUPPLEMENTARY,
+    ROLE_COMMENTARY,
+    ROLE_DUB,
+    ROLE_EMERGENCY,
+    ROLE_CAPTION,
+    ROLE_SIGN
+  })
+  public @interface Role {}
+
+  public static final int ROLE_UNSET = -1;
+  public static final int ROLE_MAIN = 0;
+  public static final int ROLE_ALTERNATE = 1;
+  public static final int ROLE_SUPPLEMENTARY = 2;
+  public static final int ROLE_COMMENTARY = 3;
+  public static final int ROLE_DUB = 4;
+  public static final int ROLE_EMERGENCY = 5;
+  public static final int ROLE_CAPTION = 6;
+  public static final int ROLE_SIGN = 7;
+
+  /**
+   * Adaptation set's accessibility descriptor value (ISO 23009-1)
+   */
+  @Documented
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({
+    ACCESSIBILITY_UNSET,
+    ACCESSIBILITY_ENHANCED_AUDIO_INTELLIGIBILITY,
+    ACCESSIBILITY_DESCRIPTION,
+    ACCESSIBILITY_CAPTION,
+    ACCESSIBILITY_SIGN
+  })
+  public @interface Accessibility {}
+
+  public static final int ACCESSIBILITY_UNSET = -1;
+  public static final int ACCESSIBILITY_ENHANCED_AUDIO_INTELLIGIBILITY = 1;
+  public static final int ACCESSIBILITY_DESCRIPTION = 2;
+  public static final int ACCESSIBILITY_CAPTION = 3;
+  public static final int ACCESSIBILITY_SIGN = 4;
+
+  /**
    * Converts a time in microseconds to the corresponding time in milliseconds, preserving
    * {@link #TIME_UNSET} and {@link #TIME_END_OF_SOURCE} values.
    *
--- a/library/core/src/main/java/com/google/android/exoplayer2/C.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/C.java
@@ -977,7 +977,10 @@ public final class C {
   public static final int NETWORK_TYPE_OTHER = 8;
 
   /**
-   * Adaptation set's role and accessibility descriptor value.
+   * Track role flags. Possible values are {@link #ROLE_FLAGS_MAIN}, {@link #ROLE_FLAGS_ALTERNATE},
+   * {@link #ROLE_FLAGS_SUPPLEMENTARY}, {@link #ROLE_FLAGS_COMMENTARY}, {@link #ROLE_FLAGS_DUB},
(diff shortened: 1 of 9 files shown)


library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1266,6 +1262,32 @@ protected static String parseEac3SupplementalProperties(List<Descriptor> supplem
     return MimeTypes.AUDIO_E_AC3;
   }
 
+  protected static String parseRole(List<Descriptor> roleDescriptors) {
```
ojw28: Making these non-static to allow overriding seems fine to me. And move them up to be grouped with the other protected non-static methods.

Note that after the suggestions I made on the corresponding issue, these methods will return `@Role` and `@Accessibility` integers respectively.

ojw28: I'd be tempted to make a `parseSelectionFlags` method too, that can parse selection flags from the role descriptors. The default implementation will contain the little piece of code that decides whether to set `C.SELECTION_FLAG_DEFAULT` that's currently further up in this class.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -167,6 +167,16 @@
    */
   public final int accessibilityChannel;
 
+  /**
+   * The Role descriptor value.
+   */
+  public final String role;
```
ojw28: Note that video formats can have role/accessibility as well. As can selection flags actually. So these three params shouldn't be under "// Audio and text specific.". They should probably be right up under the `label` param, or somewhere near there.

szaboa: Then I guess the parsing also needs to be moved up, right now it is called only in case of `mimeTypeIsRawText(sampleMimeType) == true` in `DashManifestParser`. Right?

szaboa:
Updated.

library/core/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -976,6 +976,54 @@ private C() {}
    */
   public static final int NETWORK_TYPE_OTHER = 8;
 
+  /**
+   * Adaptation set's role descriptor value (ISO 23009-1)
+   */
+  @Documented
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({
```
ojw28: As per the issue thread, I'd suggest merging the two IntDefs and making it `flag = true`. You can then also remove the `UNSET` constants, since unset is just 0 for flags.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -48,7 +48,18 @@
   public final @Nullable String id;
   /** The human readable label, or null if unknown or not applicable. */
   public final @Nullable String label;
-
+  /** Track selection flags. **/
+  @C.SelectionFlags
+  public final int selectionFlags;
+  /** Track role descriptor value, or {@link C#ROLE_UNSET} if unknown or not applicable. **/
+  @C.Role
+  public final int role;
```
ojw28: And having a single `roleFlags`, rather than separate `role` and `accessibility` fields.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1061,77 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+            return C.ROLE_SIGN;
+          default:
+            return C.ROLE_UNSET;
+        }
+      }
+    }
+    return C.ROLE_UNSET;
+  }
+
+  protected @C.Accessibility int parseAccessibility(List<Descriptor> accessibilityDescriptors) {
```
ojw28: If going with my suggested approach, this method will return C.RoleFlags, which is slightly strange. But then it's somewhat strange DASH-IF separated use of `urn:mpeg:dash:role:2011` across `Role` and `Accessibility` elements to start with, in my opinion :)... They should then be OR'd with the `parseRole` result to get the final flags.

You could probably have a single method parse "urn:mpeg:dash:role:2011" that you call from both `parseRole` and `parseAccessibility`. I think we'd probably want to handle things like "description" appearing as Role elements as well.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1061,77 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+          case "caption":
+            return C.ACCESSIBILITY_CAPTION;
+          case "sign":
+            return C.ACCESSIBILITY_SIGN;
+          default:
+            return C.ACCESSIBILITY_UNSET;
+        }
+      }
+
+      if ("urn:tva:metadata:cs:AudioPurposeCS:2007".equalsIgnoreCase(descriptor.schemeIdUri) &&
```
ojw28: See my comment on the corresponding issue around mapping more of these.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
szaboa: > You could probably have a single method parse "urn:mpeg💨role:2011" that you call from both parseRole and parseAccessibility. I think we'd probably want to handle things like "description" appearing as Role elements as well.

This would mean that the roleFlags would have a common flag for role's "description" and accessibility's "description" values. In this case we couldn't decide if the "description" came from role or from accessibility. 

Even if a couple of values could be the same for role and accessibility, I would store them as separate flags, as at the end we need to know where a value comes from (role or accessibility descriptor). Or am I missing something?

szaboa: Nevermind, probably we can merge these as you said under the correspondent issue.

szaboa:
Merged role and accessibility fields together as you've suggested.

library/core/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -976,6 +976,38 @@ private C() {}
    */
   public static final int NETWORK_TYPE_OTHER = 8;
 
+  /**
+   * Adaptation set's role and accessibility descriptor value.
```
ojw28: nit - Avoid DASH specific terms; there's no reason roles can't be used for other media formats. I would also avoid mentioning accessibility; since outside of the DASH specific parsing where they're merged, they're just "roles" and nothing else. Would suggest documenting like:
```
Track role flags. Possible values are {@link XXX}, ...
```
See how `SelectionFlags` is documented for an example.

library/core/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -976,6 +976,38 @@ private C() {}
...
+        ROLE_FLAGS_DUB,
+        ROLE_FLAGS_EMERGENCY,
+        ROLE_FLAGS_CAPTION,
+        ROLE_FLAGS_SIGN,
+        ROLE_FLAGS_ENHANCED_AUDIO_INTELLIGIBILITY,
+        ROLE_FLAGS_DESCRIPTION
+  })
+  public @interface RoleFlags {}
+
+  public static final int ROLE_FLAGS_MAIN = 1;
```
ojw28, line 1003: Can we please add Javadoc on each of these constants?

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -48,7 +48,12 @@
   public final @Nullable String id;
   /** The human readable label, or null if unknown or not applicable. */
   public final @Nullable String label;
-
+  /** Track selection flags. **/
+  @C.SelectionFlags
+  public final int selectionFlags;
+  /** Track role and accessibility descriptor values. **/
```
ojw28: nit: Just "Track role flags."

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -356,7 +360,8 @@ public static Format createAudioContainerFormat(
       int sampleRate,
       @Nullable List<byte[]> initializationData,
       @C.SelectionFlags int selectionFlags,
-      @Nullable String language) {
+      @Nullable String language,
+      @C.RoleFlags int roleFlags) {
```
ojw28: nit: Seems nice to group selectionFlags and roleFlags together, as in the methods above? Ditto for methods below this one.

library/core/src/main/java/com/google/android/exoplayer2/Format.java
```diff
@@ -906,6 +923,7 @@ public static Format createSampleFormat(
       @C.SelectionFlags int selectionFlags,
       @Nullable String language,
       int accessibilityChannel,
+      @C.RoleFlags int roleFlags,
```
ojw28: nit: Put next to selectionFlags? Ditto x lots below.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -300,7 +301,9 @@ protected AdaptationSet parseAdaptationSet(XmlPullParser xpp, String baseUrl,
         language = checkLanguageConsistency(language, xpp.getAttributeValue(null, "lang"));
         contentType = checkContentTypeConsistency(contentType, parseContentType(xpp));
       } else if (XmlPullParserUtil.isStartTag(xpp, "Role")) {
-        selectionFlags |= parseRole(xpp);
+        Descriptor descriptor = parseDescriptor(xpp, "Role");
+        selectionFlags |= parseSelectionFlags(descriptor);
```
ojw28: It seems cleaner to have a `parseSelection(roleDescriptors)` that you can call in `buildFormat`. This saves you from having to pass `selectionFlags` into that method separately.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -646,7 +640,8 @@ protected Format buildFormat(
             bitrate,
             selectionFlags,
             language,
-            accessibilityChannel);
+            accessibilityChannel,
+            roleFlags);
```
ojw28: I think you probably need to pass `roleFlags` into the `Format.createContainerFormat` on L647 (below) as well as the ones you've already done.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1059,94 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+  }
+
+  protected @C.RoleFlags int parseAccessibility(List<Descriptor> accessibilityDescriptors) {
+    @C.RoleFlags int result = 0;
+    for (int i = 0; i < accessibilityDescriptors.size(); i++) {
+      Descriptor descriptor = accessibilityDescriptors.get(i);
+      if ("urn:mpeg:dash:role:2011".equalsIgnoreCase(descriptor.schemeIdUri) && descriptor.value != null) {
+        result |= parseRoleSchemeValue(descriptor.value);
+      }
+      if ("urn:tva:metadata:cs:AudioPurposeCS:2007".equalsIgnoreCase(descriptor.schemeIdUri) &&
```
ojw28: nit: Use `else if`

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1059,94 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+              descriptor.value != null) {
+        switch (descriptor.value){
```
ojw28: Although it's only used here, it might be nice to split this into a `parseAudioPurposeValue` method, for consistency with `parseRoleSchemeValue`.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1059,94 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+      }
+    }
+    return result;
+  }
+
+  protected @C.RoleFlags int parseRoleSchemeValue(String value){
+    @C.RoleFlags int result = 0;
+    switch (value) {
+      case "main":
+        result |= C.ROLE_FLAGS_MAIN;
```
ojw28: You can just `return C.ROLE_FLAGS_X;` for each case, and add a `default` case that returns 0. Ditto for `parseAudioPurposeValue` if you make that a method too.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1064,6 +1059,94 @@ protected int parseAudioChannelConfiguration(XmlPullParser xpp)
...
+      && "main".equals(roleDescriptor.value) ? C.SELECTION_FLAG_DEFAULT : 0;
+  }
+
+  // Role and Accessibility parsing.
+
+  protected @C.RoleFlags int parseRole(List<Descriptor> roleDescriptors) {
+    @C.RoleFlags int result = 0;
+    for (int i = 0; i < roleDescriptors.size(); i++) {
+      Descriptor descriptor = roleDescriptors.get(i);
+      if ("urn:mpeg:dash:role:2011".equalsIgnoreCase(descriptor.schemeIdUri) && descriptor.value != null) {
```
ojw28: It would IMO be neater for the null check to happen at the start of `parseRoleSchemeValue` (ditto `parseAudioPurposeValue` if you make that).

ojw28 (commented):
Looking good. Just a bunch of nitty little things really, and then we'll get this merged. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/C.java
szaboa: Yeah, I wanted but I didn't find the standard you were referred to (draft of third version of 23009-1) and I wanted to take the definition from there. In the second edition fo 23009-1 a couple of role values are missing (e.g. "sign"). But I will search for it again.

szaboa:
I will do these minor changes later today. You would like to have some test cases for this whole role flags, or do you think isn't so important right now?

library/core/src/main/java/com/google/android/exoplayer2/C.java
ojw28: Feel free to leave them if you don't have the doc. I can add in some documentation when I merge.

ojw28:
Thanks! Feel free to add some test cases if you think they'd be helpful, but I don't think it's hugely important in this case. The non-parser parts are trivial, and parsing code tends to be write-once-never-touch-again.

library/core/src/main/java/com/google/android/exoplayer2/C.java
szaboa: I haven't found it yet, please add these comments.

szaboa:
Updated. I hope I've covered everything, please say so if not. I decided to not to add tests for now :)

ojw28:
Thanks!

Issue #5536, opened 2019-02-19
Player stops when device is sleeping


JonathanImperato:
### Issue description
I created an Audio Streaming app that works with HLS and obviously should work in background. It seems that on some devices (I found this bug only on the reported device ahead) the player stops after some seconds due to the power save manager.
I am running a foreground service and holding the WifiLock (WIFI_MODE_FULL_HIGH_PERF) and a WakeLock (PARTIAL_WAKE_LOCK) but it seems that the system does not respect them.
But, if we use for example bluetooth headphones, then everything works as expected.

### Reproduction steps
Just run exoplayer on that device, press the home button and then the sleep/power button and after less than 10 secs the process will be killed.

### Version of ExoPlayer being used
ExoPlayer 2.9.4

### Device(s) and version(s) of Android being used
Umidigi Z2, Android 8.1

andrewlewis:
If you have a service for audio playback and hold Wi-Fi and wake locks that should be sufficient for background audio playback (as per the [FAQs](https://google.github.io/ExoPlayer/faqs.html#how-do-i-keep-audio-playing-when-my-app-is-backgrounded)), so I think you are doing the right thing already.

I have heard that some devices have OEM customizations that will more aggressively kill processes to give improved battery life, which may be what is going on here. It might be worth playing around with the device settings to see if there are any settings to control this behavior. You could also try following up with the OEM.

A full bug report may also include some clues about what is causing the process to be killed. I'll leave this open for a while in case there is more information to investigate, but unfortunately there is likely nothing we can do at the app level to work around this.

JonathanImperato:
I forgot i wrote a stackoverflow question at the beginning, so more info can be found [here](https://stackoverflow.com/questions/54419823/powersave-manager-stops-background-service-for-exoplayer)

andrewlewis:
Closing due to inactivity.

andrewlewis:
If you can provide a full bugreport captured with `adb bugreport` (attached here or emailed to `dev.exoplayer@gmail.com`) then that may help diagnose what's going on, but without it I don't think there's much we can do to help.

Issue #930, opened 2015-11-05
Explayer and wakelock


diabloo:
hi, 
I hava a question I use exoplayer for online audio streaming and i see by testing and this without aquiring wakelock and wifilock that the player   keeps playing for more than 20 min without touching the phone . does this mean that i dont need wakelock and wifilock when i use exoplayer

ps: tested on LG G2 (4.4.2)

andrewlewis:
ExoPlayer does not acquire a WakeLock or WifiLock directly, but MediaCodecAudioTrackRenderer uses an AudioTrack which should keep the device awake while audio is playing (and for a few seconds after audio playback stops). If audio playback was continuous that might explain why the device stayed awake with the screen off.

However, the device can sleep if audio playback is interrupted and it is not being kept awake by a WakeLock. This could happen if the player runs out of data and needs to rebuffer from the network, for example. So streaming apps that play audio when the screen is off should acquire a WakeLock and WifiLock during playback. It is important to release the locks if/when playback is finished so that the device can sleep.

diabloo:
Thank your for this clear answer

PR #5628, opened 2019-03-13
Added setWakeMode implementation

JonathanImperato:
Added setWakeMode implementation that follows defalt MediaPlayer setWakeMode method implementation.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java
@@ -511,7 +511,7 @@ public class SimpleExoPlayer extends BasePlayer
     }
 
     PowerManager pm = (PowerManager)context.getSystemService(Context.POWER_SERVICE);
-    mWakeLock = pm.newWakeLock(mode|PowerManager.ON_AFTER_RELEASE, MediaPlayer.class.getName());
+    mWakeLock = pm.newWakeLock(mode|PowerManager.ON_AFTER_RELEASE, SimpleExoPlayer.class.getName());
     mWakeLock.setReferenceCounted(false);
     if (washeld) {
       mWakeLock.acquire();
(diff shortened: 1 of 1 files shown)


JonathanImperato:
I signed it

ojw28:
I don't think this code does anything, since the lock is never acquired...

I'm also not convinced such a feature is necessary. What would you use it for, and why would you need the ability to pick the mode (even if a lock is necessary, wouldn't there always be a single "correct" mode to be using)?

ojw28:
Assigning to @andrewlewis because he may know about what wake locks are held anyway during playback (e.g. by the underlying `AudioTrack`), and what the possible use cases for `MediaPlayer.setWakeMode` are.

andrewlewis:
@JonathanImperato Is this related to #5536? Perhaps you could share your findings if you discovered a way to stop the process getting killed in the background.

I think #930 is still accurate regarding what wake locks are held by the system. I'll also follow up with the framework team to get more information about the reasoning behind including this in the framework player's API. It may be that it's just provided via the player API so that it can be acquired/released automatically depending on the state.

JonathanImperato:
> I don't think this code does anything, since the lock is never acquired...
> 
> I'm also not convinced such a feature is necessary. What would you use it for, and why would you need the ability to pick the mode (even if a lock is necessary, wouldn't there always be a single "correct" mode to be using)?

Sorry for the late answers.
I thought about adding this feature because it easier for the user to create wakelock if they are managed by the library, also because some do not even know what WakeLocks are.

> @JonathanImperato Is this related to #5536? Perhaps you could share your findings if you discovered a way to stop the process getting killed in the background

I am sorry but it is not.



> I think #930 is still accurate regarding what wake locks are held by the system. I'll also follow up with the framework team to get more information about the reasoning behind including this in the framework player's API. It may be that it's just provided via the player API so that it can be acquired/released automatically depending on the state.

I think that is the idea.

tonihei:
Based on the discussion above and the linked issues it seems that we are not planning to provide a method to acquire the wake lock because it's either done automatically by the system audio components or needs to be done by the app for very specific purposes. 

I'll close for now. If you see a specific issue that needs to be solved, it's probably better to file a bug report or feature request first to simplify discussing independent of a concrete pull request.

andrewlewis:
I've filed #5846 as I think there's value in handling acquiring/releasing the wake lock based on player state, but it will probably look different to this PR (I'm not sure yet that there is value in specifying a mode for the wake lock, for example).

Issue #5633, opened 2019-03-14
Full screen bitmap smpte-tt subtitle not rendered correctly 


szaboa:
### [REQUIRED] Issue description
Support for image based SMPTE-TT subtitles was implemented in #1583, however it's not working quite well for "full screen" subtitles as it renders these out of the screen and with wrong height. By full screen I mean when the subtitle should take size of the surface. This is the case when **tt** tts:extent = **region** tts:extent in case of pixel based regions.

### [REQUIRED] Reproduction steps
It can be reproduced with the ExoPlayer demo app, just needs a stream with image based SMPTE-TT subtitle track where the Base64 encoded bitmap is "full screen".

### [REQUIRED] Link to test content
Email

### [REQUIRED] A full bug report captured from the device
No logs required. 

### [REQUIRED] Version of ExoPlayer being used
ExoPlayer 2.9.3 - 2.9.6

### [REQUIRED] Device(s) and version(s) of Android being used
It can be reproduced every time and not device dependent.
In my tests, I've used Pixel 1.

szaboa:
Example:
The manifest contains in tt tag let's say `tts:extent="720px 576px"` and there are regions like `<region xml:id="region_0" tts:extent="720px 576px" tts:origin="0px 0px"/>`.

Two problems causing this issue:

**1. Incorrect height calculation**
In the example above the bitmap will take the width of the parent correctly, but the height is scaled up to keep the bitmap's aspect ration. 

Height calculation in `com.google.android.exoplayer2.ui.SubtitlePainter#setupBitmapLayout`:
```
int height = cueBitmapHeight != Cue.DIMEN_UNSET ? Math.round(parentHeight * cueBitmapHeight)
        : Math.round(width * ((float) cueBitmap.getHeight() / cueBitmap.getWidth()));
```
The `cueBitmapHeight` will be `Cue.DIMEN_UNSET` because that's what we pass in `com.google.android.exoplayer2.text.ttml.TtmlNode#getCues` (line 238). 

This logic is OK for non-full screen subtitles, but for a full screen one it doesn't force the bitmap to fit the parent size. This results incorrect positioning.

Inside `com.google.android.exoplayer2.text.ttml.TtmlDecoder#parseRegionAttributes` the width and height calculations are OK and both of them are 1 (which means 100% width and 100% height). But the height is not passed to the `TtmlRegion`.

_Proposed solution:_

[...]

PR #5711, opened 2019-04-01
Fix positioning of full screen image based subtitles

szaboa:
Fix for issue #5633


AquilesCanta:
We'll fix this internally and push soon.

PR #5732, opened 2019-04-04
#5731 Add license information to generated POM files

### LINKED ISSUE #5731: Generate module POM files to include licenses node
### Issue description
Currently, the Maven POM is generated without a licenses node for all modules. Because of this, it's difficult for consumers to properly display licensing when attempting to do so by displaying the modules defined license information.

### Reproduction steps
Review the pom files of various ExoPlayer modules on Bintray such as the main module and note the lack of a licenses node.

https://bintray.com/google/exoplayer/exoplayer/2.9.6

### Link to test content
N/A

N/A

###  Version of ExoPlayer being used
2.9.6

###  Device(s) and version(s) of Android being used
N/A

ToxicBakery:
Resolves #5731 by adding license information to each generated POM file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,7 @@ allprojects {
         }
         buildDir = "${externalBuildDir}/${project.name}"
     }
+    group = 'com.google.android.exoplayer'
 }
 
 apply from: 'javadoc_combined.gradle'
--- a/publish.gradle
+++ b/publish.gradle
@@ -23,10 +23,42 @@ if (project.ext.has("exoplayerPublishEnabled")
         groupId = 'com.google.android.exoplayer'
         website = 'https://github.com/google/ExoPlayer'
     }
+
+    gradle.taskGraph.whenReady { taskGraph ->
+        project.tasks
+                .findAll { task -> task.name.contains("generatePomFileFor") }
+                .forEach { task ->
+                    task.doLast {
+                        task.outputs.files
+                                .filter { File file ->
+                                    file.path.contains("publications") && file.name.matches("^pom-.+\\.xml\$") 
+                                }
+                                .forEach { File file -> addLicense(file) }
+                    }
+                }
+    }
 }
 
 def getBintrayRepo() {
     boolean publicRepo = hasProperty('publicRepo') &&
         property('publicRepo').toBoolean()
     return publicRepo ? 'exoplayer' : 'exoplayer-test'
+}
+
+static void addLicense(File pom) {
+    def licenseNode = new Node(null, "license")
+    licenseNode.append(new Node(null, "name", "The Apache Software License, Version 2.0"))
+    licenseNode.append(new Node(null, "url", "http://www.apache.org/licenses/LICENSE-2.0.txt"))
+    licenseNode.append(new Node(null, "distribution", "repo"))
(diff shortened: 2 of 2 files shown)


andrewlewis:
I think this would need to go into the dev-v2 branch. Setting that aside, it may not be worth merging anyway if we are going to switch to Android Maven soon. Assigning to @ojw28 to decide how to proceed.

ToxicBakery:
Cherry picked the changes onto the dev-v2 branch. Unable to test that they still work as the head currently does not seem to build.

ojw28:
What doesn't build, exactly?

ToxicBakery:
It was failing to compile in Intellij, reimport fixed the issue though and I updated the script changes to work with the changes made in the dev branch.

ojw28:
Not sure, but is it nicer to do it more like [this](https://github.com/novoda/bintray-release/issues/210)? That approach seems to avoid reading the file and writing it out again (I've really no idea though; totally not my area of expertise!).

ToxicBakery:
I believe I tried that exact approach and it resulted in the maven pom file losing all dependency information which is obviously unacceptable. The best option would be dropping the current approach entirely and replace everything with maven-publish which would simplify everything at the cost of touching a lot of files.

ojw28:
There's not much point given we're planning to move to Google Maven soon anyway. I guess we can just merge this as is and have it as the solution until we do that.

ToxicBakery:
what do you mean by Google Maven? Google as a maven host or maven based builds?

ojw28:
I mean [maven.google.com](https://maven.google.com).

ToxicBakery:
Right, then maven-publish is likely still the correct answer

ojw28:
@ToxicBakery - This change seems to strip `<?xml version="1.0" encoding="UTF-8"?>` from the start of the generated POM file? Do you know of a nice way to bring it back? Should we just hardcode it in a `writer.write` call before `XmlNodePrinter` does its thing?

ToxicBakery:
Hmm, I wasn't aware of that. Does that change cause an issue with deployment to the artifact repository?

It could be hacked but the easiest thing to do is to move to the `maven-publish` plugin imo. Doing so would be a bit of work on the front-end but once setup it's far more flexible and correct compared to the current bintray plugin which appears to let you release things like it's the wild west (where we're going we don't need ~roads~ licenses, apparently)

ToxicBakery:
It seems `XmlUtil` may be the solution as it apparently prints the correct XML declaration that `XmlNodePrinter` does not.

```groovy
static void addLicense(File pom) {
    def licenseNode = new Node(null, "license")
    licenseNode.append(new Node(null, "name", "The Apache Software License, Version 2.0"))
    licenseNode.append(new Node(null, "url", "http://www.apache.org/licenses/LICENSE-2.0.txt"))
    licenseNode.append(new Node(null, "distribution", "repo"))
    def licensesNode = new Node(null, "licenses")
    licensesNode.append(licenseNode)

    def xml = new XmlParser().parse(pom)
    xml.append(licensesNode)

    def writer = new PrintWriter(new FileWriter(pom))
    writer.writer(XmlUtil.serialize(xml))
    writer.close()
}
```

PR #5760, opened 2019-04-12
Allow hex format tags when parsing URL templates

matamegger:
Currently the method that parses the URL templates, only recognizes decimal format. However, if it does not find the decimal identifier `d`, it places it in the template.
Streams with other formats are therefore corrupted (they might skip over parts of the video).

So far I only encountered streams with either decimal or hex format. Therefore a simple extension of the `if` adding the "missing" `d` is sufficient.

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
@@ -139,7 +139,7 @@ public final class UrlTemplate {
           String formatTag = DEFAULT_FORMAT_TAG;
           if (formatTagIndex != -1) {
             formatTag = identifier.substring(formatTagIndex);
-            if (!formatTag.endsWith("d")) {
+            if (!formatTag.endsWith("d") && !formatTag.endsWith("x")) {
               formatTag += "d";
             }
             identifier = identifier.substring(0, formatTagIndex);


ojw28:
Please could you provide an example manifest that has hex format tags? Note that hex format tags are not valid according to the DASH spec, which mandates the form `%0[width]d`.

matamegger:
According to the DASH spec (`ISO/IEC 23009-1:2014`), Section `5.3.9.4.4 Template-based Segment URL construction`
> Each identifier may be suffixed, within the enclosing ‘$’ characters, with an additional format tag aligned with the `printf` format tag as defined in IEEE 1003.1-2008 [...]

So, actually there are even more valid formats, than decimal and hex.

_working on an providable example manifest_

ojw28:
It's not valid to quote half a sentence from a spec ;). The full sentence is:

> Each identifier may be suffixed, within the enclosing ‘$’ characters, with an additional format tag aligned with the printf format tag as defined in IEEE 1003.1-2008 [10] **following this prototype: %0[width]d**.

The end of the sentence is important!

matamegger:
~While this might be a problem (and a not scientific correct cite by me), I think the last part is not important, as it refers to an **example** (out of the various options available).~

I am very sorry for misinterpreting the text passage. As already discussed in https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/177 you are totally right, and only  the `d` suffix is specified. 

Also in the linked issue, the author mentions that `dash.js` supports the more general format. So, it might be still worth considering the support for `x` and other suffixes.

ojw28:
We can support `x` if there's evidence of a significant amount of content already in production that uses it. Is there evidence of this? If so then where, and what packager generated it?

In general I'd rather we didn't just support everything. Players accepting non-compliant streams makes it more likely that packagers will produce non-compliant streams, which in turn encourages other players to also widen the range of non-compliant streams they accept. At some point you end up in a mess where the original specifications don't accurately reflect what the industry is doing, which results in all kinds of interop problems and significantly increases the bar for writing new player implementations (because just implementing what's in the specifications you're trying to support is no longer sufficient). I'd actually suggest that `dash.js` should remove support for the more general format, to help the industry remain aligned with what the specification actually says.

ojw28:
Could you address my questions above? Thanks.

matamegger:
I signed it!

PR #5766, opened 2019-04-15
Add AV Sync config 

endiirawan:
Add AV sync config in PlaybackParameter for fixing issue where the video & audio is not sync, due to device limitation.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -561,7 +561,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
     TraceUtil.beginSection("doSomeWork");
 
     updatePlaybackPositions();
-    long rendererPositionElapsedRealtimeUs = SystemClock.elapsedRealtime() * 1000;
+    long rendererPositionElapsedRealtimeUs = (SystemClock.elapsedRealtime() * 1000) + mediaClock.getPlaybackParameters().videoOffsetUs;
 
     playingPeriodHolder.mediaPeriod.discardBuffer(playbackInfo.positionUs - backBufferDurationUs,
         retainBackBufferFromKeyframe);
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
@@ -66,6 +66,7 @@ public final class SonicAudioProcessor implements AudioProcessor {
   private int sampleRateHz;
   private float speed;
   private float pitch;
+  private long videoOffsetUs;
   private int outputSampleRateHz;
   private int pendingOutputSampleRateHz;
 
@@ -125,6 +126,21 @@ public final class SonicAudioProcessor implements AudioProcessor {
     }
     flush();
     return pitch;
+  }
+
+  /**
+   * Sets the playback video offset for AV sync. Calling this method will discard any data buffered within the
+   * processor, and may update the value returned by {@link #isActive()}.
+   * @param videoOffsetUs
+   * @return the actual new video offset in nanosecond
+   */
+  public long setVideoOffsetUs(long videoOffsetUs) {
+    if (this.videoOffsetUs != videoOffsetUs) {
+      this.videoOffsetUs = videoOffsetUs;
+      pendingSonicRecreation = true;
+    }
+    flush();
+    return videoOffsetUs;
   }
 
   /**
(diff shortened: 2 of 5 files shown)


library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
```diff
@@ -66,6 +66,7 @@
   private int sampleRateHz;
   private float speed;
   private float pitch;
+  private long videoOffsetUs;
```
ojw28, line 69: The changes to this class don't seem to do anything (in particular `videoOffsetUs` doesn't seem to be read by anything? Ditto to the change in `DefaultAudioSink`.

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -561,7 +561,7 @@ private void doSomeWork() throws ExoPlaybackException, IOException {
     TraceUtil.beginSection("doSomeWork");
 
     updatePlaybackPositions();
-    long rendererPositionElapsedRealtimeUs = SystemClock.elapsedRealtime() * 1000;
+    long rendererPositionElapsedRealtimeUs = (SystemClock.elapsedRealtime() * 1000) + mediaClock.getPlaybackParameters().videoOffsetUs;
```
ojw28, line 564: This isn't really the right place to apply this offset. It should probably be applied somewhere under `DefaultAudioSink.getCurrentPositionUs`, if anywhere.

ojw28 (commented):
I'm not sure this is the best way of doing this. Also, it's unclear what the problem is that you're actually trying to solve. Please could you file a bug report [here](https://github.com/google/ExoPlayer/issues/new/choose) that fills in all the sections of the template, so we have a clear idea of what this pull request is actually addressing, what device(s) have this apparently device limitation, and what that limitation is? Thanks.

endiirawan:
@ojw28 this issue is only can be reproduced on the specific device only (which is has closed market in Indonesia). here is the detail issue:

Device          : JellyBean API 17
Device Type : STB
Issue             : 
The audio and video are out of sync (audio is behind the video), the delay between audio and video is constant about 500 ms.   
Not only on Exoplayer, this issue also happen on another player (e.g: NexStreaming, etc)
This issue is happen on this kind of device type only.
This issue was tested happen on all content, both clear & protected content.

By using this workaround, this issue is gone (tested). we can adding the offset minus/plus number to adjust it.  and if we not set the offset value it wont break another part I think.

library/core/src/main/java/com/google/android/exoplayer2/audio/SonicAudioProcessor.java
endiirawan: @ojw28  this is used on this class:
https://github.com/google/ExoPlayer/pull/5766/files#diff-d7c4edd055ad44ffd9ee2b757ac41786

when we set the playbackParameters e.g:
```suggestion
  final long videoOffsetUs = 500 * 1000; // video offset in micro seconds
      PlaybackParameters playbackParameters = new PlaybackParameters(1f, 1f, videoOffsetUs, false);
      player.setPlaybackParameters(playbackParameters);
```

somehow exoplayer will copying that parameters into SonicAudioProcessor:
library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java **Line:140**
```suggestion
  return new PlaybackParameters(
          sonicAudioProcessor.setSpeed(playbackParameters.speed),	          sonicAudioProcessor.setSpeed(playbackParameters.speed),
          sonicAudioProcessor.setPitch(playbackParameters.pitch),	          sonicAudioProcessor.setPitch(playbackParameters.pitch),
          sonicAudioProcessor.setVideoOffsetUs(playbackParameters.videoOffsetUs),
          playbackParameters.skipSilence);	         
   }
```
so if we dont copy the video offset into SonicAudioProcessor, we will get null when call this:
`mediaClock.getPlaybackParameters().videoOffsetUs`
hence I am adding the video offset on this class too.

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
endiirawan: @ojw28  oh i see, it is works too. you are right, this logic is more relevant to be applied into class DefaultAudioSink.
OK I will move this part into  `DefaultAudioSink.getCurrentPositionUs`.

ojw28:
I don't think we really want to take this kind of pull request, for what sounds like a fairly severe issue in the underlying platform, on one very specific device, which will be affecting multiple video players running on that device. It's up to the device manufacturer to fix the underlying problem.

Issue #5756, opened 2019-04-11
HLS: Consider providing hooks for non-standard (full segment encryption) DRM schemes


bjitivo:
### [REQUIRED] Searched documentation and issues
Tell us where you’ve already looked for an answer to your question. It’s
important for us to know this so that we can improve our documentation.

I searched on Google for information on integrating the DRM system that we are using with ExoPlayer.  I found some discussions but they were always closed by the ExoPlayer team as 'not something we will support'.

### [REQUIRED] Question
Describe your question in detail.

I would like to know if the ExoPlayer team would consider a change that would enable what I call "HLS whole segment decryption".  The motivation for this change is that some DRM vendors provide solutions that require that the mechanism for decrypting HLS segments be more controlled by them than the ExoPlayer code base currently allows.

The gist of these DRM vendors' design is that they implement their own mechanism for downloading the segment encryption key and decrypting the segment, given these inputs:

1. The segment decryption key URI
2. The initialization vector
3. The entire segment as bytes

I do not know the reasons for these design choices, but I can guess:

- HLS as specified allows the client to request the decryption key from the key URI and perform the decryption of the segment
- However, naive implementations of this mechanism may make the key 'sniffable' by motivated individuals who may have means for intercepting data on the host device, such as somehow acquiring memory snapshots, or side channel attacks on the processor to detect operations on the key that reveal details that could be used to help break the key
- So these vendors provide obfuscation and risk mitigation solutions that do things like scrambling the keys in memory, implementing alternate algorithms for decryption that are not as vulnerable to side channel attacks, never piecing the key together except in CPU registers that are very difficult to sniff from the outside, etc
- Finally, the mechanism for fetching the key itself may be more than a standard HTTP GET operation - the vendor may implement some means for having the client provide authentication credentials at the time of requesting the key from the key URI and may do so using nonstandard (outside of normal HTTP authentication mechanisms) means

This last part seems to me like it would violate the HLS spec; but be that as it may, it is a reality that these systems exist and I don't think it would be very hard to make ExoPlayer more amenable to supporting them.

Currently ExoPlayer assumes that if segments are encrypted in a way that is described by standard HLS tags (EXT-X-KEY) that it should:

a. Fetch the decryption key as a precondition to creating a new HlsMediaChunk for the segment
b. Create an Aes128DataSource to perform a decryption step before providing the raw data of the chunk

[...]

PR #5780, opened 2019-04-19
Added HlsDecryptingDataSource interface and use it within HlsStreamChunk

bjitivo:
to allow an application using ExoPlayer to perform decryption instead of
relying on the HlsChunkSource/HlsMediaChunk to do streaming decryption.  This
allows custom decryption methodologies should they be needed.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -309,16 +309,26 @@ import java.util.List;
     HlsMediaPlaylist.Segment segment = mediaPlaylist.segments.get(chunkIndex);
 
     // Check if the segment is completely encrypted using the identity key format.
+    DataSource dataSource = mediaDataSource;
     if (segment.fullSegmentEncryptionKeyUri != null) {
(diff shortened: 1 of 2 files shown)


bjitivo:
Will submit once I have actual authorization ... didn't realize that law paranoia was going to be a factor here ...

AquilesCanta:
Hi @bjitivo, CLAs are a requirement before taking in any pull request. However, it's fine to ignore until we decide to accept the PR.

I would be happy to have a bit more context for this patch. Could you provide some form of specification? 

A few things that would be useful:

* A sample stream that this patch aims to provide support for.
* A description of the encryption mechanism to use. Also great to know why existing mechanisms are not sufficient.
* Alternative approaches you've looked into (a very brief overview is fine), and why they were discarded.

ojw28:
@AquilesCanta - The context is https://github.com/google/ExoPlayer/issues/5756 (@bjitivo - Please link to the issue from the change description).

bjitivo:
This issue gives some explanation:

https://github.com/google/ExoPlayer/issues/5756

Is that sufficient or do you need more?  Thanks!

bjitivo:
> @AquilesCanta - The context is #5756 (@bjitivo - Please link to the issue from the change description).

Thank you I will do that!

bjitivo:
Also, the legal department at my company is reviewing the CLA and will hopefully approve it soon.  As an individual I fight against the encumbrance of legal agreements on code sharing but as an employee and a pull requestor I guess I have no choice :)

PR #5807, opened 2019-04-26
CEA608: Handling XDS and TEXT modes

zsmatyas:
[Problem]
There are 3 services / modes transported on line 21:
- Captioning
- TEXT (generally not program related)
- XDS (eXtended Data Services)

Bytes belonging to the unsupported modes are interleaved with the
bytes of the captioning mode.
See Chapter 7, Chapter 8.5 and Chapter 8.6 of the CEA608 Standard for
more details.

[Solution]
Drop all bytes belonging to unsupported modes.

[Test]
- All streams containing only captioning services should not be influenced
- Test all 4 CEA 608 channels with live over-the-air content and
using all available TEXT and XDS streams.

[Links]
https://www.law.cornell.edu/cfr/text/47/79.101
https://en.wikipedia.org/wiki/EIA-608#Channels

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -83,12 +83,6 @@ public final class Cea608Decoder extends CeaDecoder {
 
   private static final byte CTRL_BACKSPACE = 0x21;
 
-  @SuppressWarnings("unused")
-  private static final byte CTRL_ALARM_OFF= 0x22; // not supported any more
-
-  @SuppressWarnings("unused")
-  private static final byte CTRL_ALARM_ON= 0x23; // not supported any more
-
   private static final byte CTRL_DELETE_TO_END_OF_ROW = 0x24;
 
   /**
@@ -258,7 +252,7 @@ public final class Cea608Decoder extends CeaDecoder {
   // The incoming characters may belong to 3 different services based on the last received control
   // codes. The 3 services are Captioning, Text and XDS. In this decoder we only intend to process
   // bytes belonging to the Captioning service.
-  private boolean isInCaptionMode = true;
+  private boolean isInCaptionMode;
 
   public Cea608Decoder(String mimeType, int accessibilityChannel) {
     ccData = new ParsableByteArray();
@@ -291,6 +285,7 @@ public final class Cea608Decoder extends CeaDecoder {
 
     setCaptionMode(CC_MODE_UNKNOWN);
     resetCueBuilders();
+    isInCaptionMode = true;
   }
 
   @Override
@@ -330,14 +325,14 @@ public final class Cea608Decoder extends CeaDecoder {
     return new CeaSubtitle(cues);
   }
 
-  private boolean isCodeForUnsupportedMode(byte cc1, byte cc2) {
+  private static boolean isCodeForUnsupportedMode(byte cc1, byte cc2) {
     // Control codes from 0x01 to 0x0F indicate the beginning of XDS Data
     if (0x01 <= cc1 && cc1 <= 0x0F) {
       return true;
     }
 
     // 2 commands switch to TEXT mode.
-    if (((cc1 & 0xF7) == 0x14) // first byte must be 0x14 or 0x1C based on channel
+    if ((isModeSwitchCommand(cc1))
         && (cc2 == CTRL_TEXT_RESTART || cc2 == CTRL_RESUME_TEXT_DISPLAY)) {
       return true;
     }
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -306,6 +330,38 @@ protected Subtitle createSubtitle() {
     return new CeaSubtitle(cues);
   }
 
+  private boolean isCodeForUnsupportedMode(byte cc1, byte cc2) {
+    // Control codes from 0x01 to 0x0F indicate the beginning of XDS Data
+    if (0x01 <= cc1 && cc1 <= 0x0F) {
```
zsmatyas, line 330: Section 8.5 C of the CEA608 Standard:

>    When any of the control codes from 0x01 to 0x0F is used to begin a control-character pair it indicates the beginning of XDS Data. This XDS data should be considered part of a different data channel and can be used to interrupt the Text data according to the FCC rules for other data channels.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -306,6 +330,38 @@ protected Subtitle createSubtitle() {
...
+    return false;
+  }
+
+  private static boolean isControlCodeSwitchingToCaptionMode(byte cc1, byte cc2) {
+    if ((cc1 & 0xF7) != 0x14) { // Matching commands must have the CC1 value: 0|0|0|1|CH|1|0|0 where CH is the channel bit
+      return false;
+    }
+
+    switch (cc2) {
+      case CTRL_END_OF_CAPTION:
```
zsmatyas, line 354: See C.16 of the CEA608 standard.
>  EDM or ENM command does not instruct the decoder to treat subsequent data as caption data
...
There are six caption-specific commands which do terminate a Text Mode data stream: EOC, RCL, RDC, RU2, RU3, and RU4.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -80,6 +80,17 @@
    * at which point the non-displayed memory becomes the displayed memory (and vice versa).
    */
   private static final byte CTRL_RESUME_CAPTION_LOADING = 0x20;
+
+  private static final byte CTRL_BACKSPACE = 0x21;
+
+  @SuppressWarnings("unused")
+  private static final byte CTRL_ALARM_OFF= 0x22; // not supported any more
```
tonihei: We don't usually define constants if we don't use them (even if they are mentioned in the associated standard). Can you remove them again?

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -237,6 +255,11 @@
   private byte repeatableControlCc2;
   private int currentChannel;
 
+  // The incoming characters may belong to 3 different services based on the last received control
+  // codes. The 3 services are Captioning, Text and XDS. In this decoder we only intend to process
+  // bytes belonging to the Captioning service.
+  private boolean isInCaptionMode = true;
```
tonihei: Please move initialization of non-static variables to constructor.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -306,6 +330,38 @@ protected Subtitle createSubtitle() {
...
   }
 
+  private boolean isCodeForUnsupportedMode(byte cc1, byte cc2) {
+    // Control codes from 0x01 to 0x0F indicate the beginning of XDS Data
+    if (0x01 <= cc1 && cc1 <= 0x0F) {
+      return true;
+    }
+
+    // 2 commands switch to TEXT mode.
+    if (((cc1 & 0xF7) == 0x14) // first byte must be 0x14 or 0x1C based on channel
```
tonihei: Can you move (cc1 & 0xF7) == 0x14 to a private static method isModeSwitchCommand(cc1) for readability? And use the same method on L349 below. Then you only need to add the explaining comment once.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -363,6 +419,16 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
         continue;
       }
 
+      if (isCodeForUnsupportedMode(ccData1, ccData2)) {
```
tonihei, line 422: If I'm not mistaken, it seems that this code can move to handleCrtl? That is probably a better place to handle these transitions as it's part of "control code handling" in general.

tonihei (commented):
Looks good in general. Just some code structure comments.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: I do not think so. We need stop processing all data (characters included) in case we are not in CAPTION mode. Moving this down would add the characters of the TEXT and XDS modes to the buffers of our processing of CAPTION mode.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
zsmatyas: I could not come up with a better name, but the suggested function signature could be misleading. Usually both bytes are needed to figure out what the command is doing, and there could be a long range of values this function returns true, although there is no mode switch expected.

zsmatyas:
Added the suggested improvements as a new commit, now the comments given to the first commit might point to locations that might be "Outdated"

tonihei (commented):
Should be able to merge now. I'll do some further formatting changes on our side, but otherwise looks good.

PR #5894, opened 2019-05-15
Fix package name in PlatformScheduler java doc

bidrohi:


bidrohi:
I signed it!

andrewlewis:
Thanks for the contribution. For future reference, pull requests need to be sent to the dev-v2 branch. As this is a small change and the branch is wrong, I'll just submit an equivalent change internally and close this.

PR #5908, opened 2019-05-19
Support signalling of last segment number via supplemental descriptor in mpd.

sr1990:
Last segment indicated Supplemental Property by mpd is parsed but not taken into consideration by the player.

This patch supports Supplemental Descriptor with @schemeIdUri set to "http://dashif.org/guide-
lines/last-segment-number" with the @value set to the last segment number.

Example stream: http://dash.akamaized.net/dash264/TestCasesIOP41/LastSegmentNumber/1/manifest_last_segment_num.mpd

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -242,7 +242,7 @@ public class DashManifestParser extends DefaultHandler
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentList")) {
         segmentBase = parseSegmentList(xpp, null);
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentTemplate")) {
-        segmentBase = parseSegmentTemplate(xpp, null,null);
+        segmentBase = parseSegmentTemplate(xpp, null,null,null);
       } else {
         maybeSkipTag(xpp);
       }
@@ -333,7 +333,8 @@ public class DashManifestParser extends DefaultHandler
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentList")) {
         segmentBase = parseSegmentList(xpp, (SegmentList) segmentBase);
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentTemplate")) {
-        segmentBase = parseSegmentTemplate(xpp, (SegmentTemplate) segmentBase,supplementalProperties);
+        segmentBase = parseSegmentTemplate(xpp, (SegmentTemplate) segmentBase,
+            supplementalProperties,null);
       } else if (XmlPullParserUtil.isStartTag(xpp, "InbandEventStream")) {
         inbandEventStreams.add(parseDescriptor(xpp, "InbandEventStream"));
       } else if (XmlPullParserUtil.isStartTag(xpp)) {
@@ -487,7 +488,7 @@ public class DashManifestParser extends DefaultHandler
       List<Descriptor> adaptationSetRoleDescriptors,
       List<Descriptor> adaptationSetAccessibilityDescriptors,
       SegmentBase segmentBase,
-      ArrayList<Descriptor> parentSupplementalProperties)
+      ArrayList<Descriptor> adaptationSetSupplementalProperties)
       throws XmlPullParserException, IOException {
     String id = xpp.getAttributeValue(null, "id");
     int bandwidth = parseInt(xpp, "bandwidth", Format.NO_VALUE);
@@ -520,7 +521,7 @@ public class DashManifestParser extends DefaultHandler
         segmentBase = parseSegmentList(xpp, (SegmentList) segmentBase);
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentTemplate")) {
         segmentBase = parseSegmentTemplate(xpp, (SegmentTemplate) segmentBase,
-            parentSupplementalProperties);
+            adaptationSetSupplementalProperties,supplementalProperties);
       } else if (XmlPullParserUtil.isStartTag(xpp, "ContentProtection")) {
         Pair<String, SchemeData> contentProtection = parseContentProtection(xpp);
         if (contentProtection.first != null) {
@@ -760,7 +761,8 @@ public class DashManifestParser extends DefaultHandler
   }
 
   protected SegmentTemplate parseSegmentTemplate(XmlPullParser xpp, SegmentTemplate parent,
-      ArrayList<Descriptor> parentSupplementalProperties)
+      ArrayList<Descriptor> adaptationSetSupplementalProperties,
+      ArrayList<Descriptor> representationSupplementalProperties)
       throws XmlPullParserException, IOException {
     long timescale = parseLong(xpp, "timescale", parent != null ? parent.timescale : 1);
     long presentationTimeOffset = parseLong(xpp, "presentationTimeOffset",
@@ -793,7 +795,18 @@ public class DashManifestParser extends DefaultHandler
 
     return buildSegmentTemplate(initialization, timescale, presentationTimeOffset,
         startNumber, duration, timeline, initializationTemplate, mediaTemplate,
-        parentSupplementalProperties);
+        adaptationSetSupplementalProperties,representationSupplementalProperties);
+  }
+
+  protected String parseLastSegmentNumberSupplementalProperty
+      (List<Descriptor> supplementalProperties){
+    for (Descriptor descriptor : supplementalProperties) {
+      if (descriptor.schemeIdUri.equalsIgnoreCase
+          ("http://dashif.org/guidelines/last-segment-number")) {
+        return descriptor.value;
+      }
+    }
+    return null;
   }
 
   protected SegmentTemplate buildSegmentTemplate(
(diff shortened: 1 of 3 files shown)


sr1990:
I have signed the CLA

ojw28:
Thanks for the contribution!

I see what this is doing, but I think I'd expect the information to be incorporated into `SegmentTemplate` and returned via `SegmentTemplate.getSegmentCount`, rather than applied as a special case in chunk source. I think that would involve passing an extra argument into `DashManifestParser.parseSegmentTemplate`. Do you think that would work (and make sense)?

sr1990:
@ojw28 Thanks for the feedback. It does make sense to incorporate the information in SegmentTemplate. I have re-submitted the patch.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -517,7 +519,8 @@ protected RepresentationInfo parseRepresentation(
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentList")) {
         segmentBase = parseSegmentList(xpp, (SegmentList) segmentBase);
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentTemplate")) {
-        segmentBase = parseSegmentTemplate(xpp, (SegmentTemplate) segmentBase);
+        segmentBase = parseSegmentTemplate(xpp, (SegmentTemplate) segmentBase,
+            parentSupplementalProperties);
```
ojw28: Is it defined somewhere at what level the supplemental property will be defined? This seems to assume it'll be defined at the `AdaptationSet` level, but presumably `Period` and `Representation` levels could both be possible too?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java
```diff
@@ -338,7 +408,9 @@ public RangedUri getSegmentUrl(Representation representation, long sequenceNumbe
 
     @Override
     public int getSegmentCount(long periodDurationUs) {
-      if (segmentTimeline != null) {
+      if( endNumber != C.INDEX_UNSET) {
+        return endNumber;
+      } else if (segmentTimeline != null) {
```
ojw28: I think the timeline should still take priority, if it's set.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java
```diff
@@ -338,7 +408,9 @@ public RangedUri getSegmentUrl(Representation representation, long sequenceNumbe
 
     @Override
     public int getSegmentCount(long periodDurationUs) {
-      if (segmentTimeline != null) {
+      if( endNumber != C.INDEX_UNSET) {
+        return endNumber;
```
ojw28: Shouldn't this be `endNumber - startNumber + 1`?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
sr1990: Level of the supplemental property is not defined. Have made changes to pass AdaptationSet and Representation supplemental properties to parse parseSegmentTemplate method.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java
```diff
@@ -102,6 +102,7 @@ public RangedUri getIndex() {
     /* package */ final long startNumber;
     /* package */ final long duration;
     /* package */ final List<SegmentTimelineElement> segmentTimeline;
+    /* package */ final int endNumber;
```
ojw28: * Is there a reason this is defined here rather than on `SegmentTemplate`? Given it's only used there, it seems simpler just to define it on that class directly?
* I think this needs to be a long rather than an int (given `startNumber` is a long and it must be true that `endNumber >= startNumber`. Ditto all the plumbing of `endNumber`

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java
```diff
@@ -340,14 +410,15 @@ public RangedUri getSegmentUrl(Representation representation, long sequenceNumbe
     public int getSegmentCount(long periodDurationUs) {
       if (segmentTimeline != null) {
         return segmentTimeline.size();
+      } else if (endNumber != C.INDEX_UNSET) {
+        return endNumber - (int) startNumber + 1;
```
ojw28: The cast should happen as the last thing in the calculation.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -799,9 +817,33 @@ protected SegmentTemplate buildSegmentTemplate(
...
+      UrlTemplate mediaTemplate,ArrayList<Descriptor> adaptationSetSupplementalProperties,
+      ArrayList<Descriptor> representationSupplementalProperties ) {
+
+    if (representationSupplementalProperties != null) {
+      if (parseLastSegmentNumberSupplementalProperty
+          (representationSupplementalProperties) != null) {
+        String lastSegment = parseLastSegmentNumberSupplementalProperty
+            (representationSupplementalProperties);
+        return new SegmentTemplate(initialization, timescale, presentationTimeOffset,
+            startNumber, Integer.valueOf(lastSegment),duration, timeline,
```
ojw28: I think it probably needs parsing as a `long` rather than an `int`?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -799,9 +817,33 @@ protected SegmentTemplate buildSegmentTemplate(
       long duration,
       List<SegmentTimelineElement> timeline,
       UrlTemplate initializationTemplate,
-      UrlTemplate mediaTemplate) {
+      UrlTemplate mediaTemplate,ArrayList<Descriptor> adaptationSetSupplementalProperties,
+      ArrayList<Descriptor> representationSupplementalProperties ) {
+
+    if (representationSupplementalProperties != null) {
```
ojw28: All other non-trivial code is in `parseSegmentTemplate`, with `buildSegmentTemplate` just for the trivial building part. So this probably belongs in `parseSegmentTemplate`, and `buildSegmentTemplate` should probably be changed only to have an additional `long endNumber` argument.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
ojw28: It seems non-ideal to push complexity into the client because something hasn't been defined in the DASH-IF doc. Can we assume it'll be defined at the same level as the `SegmentTemplate` for now, since this avoids having to pass around the supplemental properties from multiple layers.

I can open a request with the DASH-IF folks to get them to try and clarify how it's supposed to work in the spec.

sr1990: Makes sense. 
However, the asset provided by DASH-IF (https://testassets.dashif.org/#testvector/details/5a1c4b447cccbb15567e567b) has supplemental property at the AdaptationSet level and not SegmentTemplate level. 
Do you think it would make sense to assume that the supplemental property is specified at AdaptationSet level based on the example provided and pass around only the AdaptationSet supplemental properties? or request DASH-IF for clarification?

ojw28: Passing just the `AdaptationSet` ones seems OK for now (which I think is what you were doing before; sorry!). Thinking about this a bit more:

1. It wouldn't make sense for the property to be defined at the `Representation` level. Segments are supposed to be aligned across representations within an adaptation set, so the value would be the same in each one (i.e. may as well move the property up to the `AdaptationSet`)
1. I guess it's possible it could be defined at the `Period` level, but we've not seen any examples of that.

I will open a DASH-IF issue to request that this is clarified explicitly, also.

ojw28: Filed https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/340

ojw28:
See inline comment (from yesterday) in response to your question about what levels we should assume the property can be set on. Once that's updated I'll get this merged; thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java
sr1990: Have moved endNumber to SegmentTemplate.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
sr1990: Sounds good. Thanks

PR #5951, opened 2019-05-27
Do not update with empty metadata when metadata provider is set to null

thibseisel:
When `MediaSessionConnector.setMediaMetadataProvider` is called with `null`, I'd expect `MediaSessionConnector` to no longer update the media session metadata, even with an empty placeholder (`METADATA_EMPTY`).
This brings more flexibility to how metadata can be updated: the current `MediaMetadataProvider` interface does not support setting metadata retrieved from a callback/asynchronous API (RxJava, Kotlin Coroutines, etc.)

### DIFF
--- a/extensions/mediasession/src/main/java/com/google/android/exoplayer2/ext/mediasession/MediaSessionConnector.java
+++ b/extensions/mediasession/src/main/java/com/google/android/exoplayer2/ext/mediasession/MediaSessionConnector.java
@@ -662,11 +662,12 @@ public final class MediaSessionConnector {
    * {@link MediaMetadataProvider#getMetadata(Player)}.
    */
   public final void invalidateMediaSessionMetadata() {
-    MediaMetadataCompat metadata =
-        mediaMetadataProvider != null && player != null
-            ? mediaMetadataProvider.getMetadata(player)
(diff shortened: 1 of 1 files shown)


ojw28:
> The current MediaMetadataProvider interface does not support setting metadata retrieved from a callback/asynchronous API

It is possible to retrieve metadata asynchronously. You have to return something synchronously to be displayed, which makes sense, but you can also start an asynchronous operation in `getMetadata`, and when that completes call `MediaSessionConnector.invalidateMediaSessionMetadata` to trigger another `getMetadata`, from which you can return the result if it's still applicable. Doing this requires a small amount of extra work, but it's not particularly hard to do.

I think it makes sense for the `MediaSessionConnector` to be the only thing setting the metadata directly on the session, else it will be confusing who has this responsibility.

Given this, I'm not sure this change is necessary. Perhaps we just need to better document how we think this should be done. What do you think?

thibseisel:
Oh, I didn't thought about that, that should do the trick.
Yes, we could improve the documentation a bit to explain how to handle asynchronous retrieval of metadata.

ojw28:
Filed https://github.com/google/ExoPlayer/issues/6047 to better document this. Thanks!

PR #5970, opened 2019-05-29
Fix for CloudFront malformed manifest

tjviking:
CloudFront uses malformed manifests in their mpd files, eg:
<SegmentTemplate duration="4000000" initialization="../video/360/stream_2-init.mp4" media="../video/360/stream_2-seg_$Number%00d$.mp4" startNumber="1" timescale="1000000"></SegmentTemplate>

%00d does not format well when passed to formatter and throws exception

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
@@ -115,6 +115,7 @@ public final class UrlTemplate {
    */
   private static int parseTemplate(String template, String[] urlPieces, int[] identifiers,
       String[] identifierFormatTags) {
+    template = template.replace("%00","%0");
     urlPieces[0] = "";
     int templateIndex = 0;
     int identifierCount = 0;
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
@@ -115,7 +115,7 @@ public final class UrlTemplate {
    */
   private static int parseTemplate(String template, String[] urlPieces, int[] identifiers,
       String[] identifierFormatTags) {
-    template = template.replace("%00","%0");
+    template = template.replace("%00d","");
     urlPieces[0] = "";
     int templateIndex = 0;
     int identifierCount = 0;


tjviking:
I signed it!

ojw28:
Are you sure this is the right fix? `%0d` also crashes the formatter for me. Could you also please point me to where you've reported this issue to CloudFront? Thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/UrlTemplate.java
```diff
@@ -115,6 +115,7 @@ public String buildUri(String representationId, long segmentNumber, int bandwidt
    */
   private static int parseTemplate(String template, String[] urlPieces, int[] identifiers,
       String[] identifierFormatTags) {
+    template = template.replace("%00","%0");
```
ojw28: If you're going to do this anywhere, shouldn't it be directly on each `formatTag` down at L148.

tjviking: yeah, maybe I will move  it and then will submit new patch, thx

tjviking:
> Are you sure this is the right fix? `%0d` also crashes the formatter for me. Could you also please point me to where you've reported this issue to CloudFront? Thanks!

boy, sorry for my mistake.... meant CloudFlare Stream, not yours CloudFront service,
sorry!

And there is no chance to submit any fixes to them, no code base no support nor anything,

I will test it on my fork and will resubmit it proper way in better place, sorry for disturbing,
I was too fast and did not do enough tests.

ojw28:
> And there is no chance to submit any fixes to them, no code base no support nor anything,

Have you tried? I imagine they provide some level of support if you're paying them for a service. They also have https://community.cloudflare.com. Pushing fixes into every player that you want to use, rather than fixing the actual underlying problem, is not good for anyone.

tjviking:
> > And there is no chance to submit any fixes to them, no code base no support nor anything,
> 
> Have you tried? I imagine they provide some level of support if you're paying them for a service. They also have https://community.cloudflare.com. Pushing fixes into every player that you want to use, rather than fixing the actual underlying problem, is not good for anyone.

just answer yourself listing level of merit of those topics:
https://community.cloudflare.com/search?q=stream

Its level of the ground, nor any client support :D 

So to answer your question, I did not, because it seems dead community and
unsupported product.

We solved it by forking your repo and fixing it on our own.

Sorry for wasting your time. I just hope there will be some checks 
before blindly calling formatting rules to not throw exceptions :/

ojw28:
To clarify, I've no particular objection to taking a patch for this, provided it's done right (as per the other comments) :). Feel free to update the pull request if you'd like to get a fix merged in.

I'd like for the issue to be at least reported as well though. Even if it's going to be ignored. It doesn't take long to report it.

tjviking:
> To clarify, I've no particular objection to taking a patch for this, provided it's done right (as per the other comments) :). Feel free to update the pull request if you'd like to get a fix merged in.
> 
> I'd like for the issue to be at least reported as well though. Even if it's going to be ignored. It doesn't take long to report it.

ok thx, we abandoned this idea and switched to m3u8 links instead of dash/mpeg manifest files

PR #6032, opened 2019-06-14
Support for UDP data source

angelrc96:
In the current release, there is no support to reproduce videos throughout UDP. As a result, I wanted to modify just this single class to add the functionality. I have tested it with an UDP video and it worked like a charm. Just a few changes were needed. 

In order to test and make sure that these changes can be applied to next release, it's only necessary to put an UDP video in the demo and ensure that it works correctly.

I saw some people asking for this feature in ExoPlayer's issues. So I think it would be very useful to add this functionality in future releases. It would keep us aside of having to instantiate your whole project locally in our repositories.


angelrc96:
I signed it!

angelrc96:
I signed it!

PR #6042, opened 2019-06-15
Add convenience constructors for notification channel descriptions

Timbals:
Currently only the channel name can be supplied when using the convenience constructors for `PlayerNotificationManager` and `DownloadService` to create notification channels.

With this PR it is possible to also conveniently change the description of the notification channel.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/util/NotificationUtil.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/util/NotificationUtil.java
@@ -80,11 +80,48 @@ public final class NotificationUtil {
    */
   public static void createNotificationChannel(
       Context context, String id, @StringRes int nameResourceId, @Importance int importance) {
+    createNotificationChannel(
+        context,
+        id,
+        nameResourceId,
+        importance,
+        /* descriptionResourceId= */ 0);
+  }
+
+  /**
+   * Creates a notification channel that notifications can be posted to. See {@link
+   * NotificationChannel} and {@link
+   * NotificationManager#createNotificationChannel(NotificationChannel)} for details.
+   *
+   * @param context A {@link Context}.
+   * @param id The id of the channel. Must be unique per package. The value may be truncated if it's
+   *     too long.
+   * @param nameResourceId A string resource identifier for the user visible name of the channel.
+   *     You can rename this channel when the system locale changes by listening for the {@link
+   *     Intent#ACTION_LOCALE_CHANGED} broadcast. The recommended maximum length is 40 characters.
+   *     The value may be truncated if it is too long.
+   * @param importance The importance of the channel. This controls how interruptive notifications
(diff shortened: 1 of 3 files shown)


ojw28:
I'm inclined not to merge this. There are lots of things that you can set on `NotificationChannel`. The intention of the convenience methods is really just to handle the simplest cases. For other cases app code can do its own `NotificationChannel` setup. Is there any reason why this is particularly difficult to do, that would justify the addition being proposed here?

Timbals:
Both `PlayerNotificationManager` and `DownloadService` use notifications with well-defined use cases. I feel like all other attributes of `NotificationChannel` use sensible defaults for these types of notifications. The name and description are most likely to differ for different use cases of the library. Changing the description is not necessarily more difficult, but probably a lot more common than changing other attributes of the `NotificationChannel`.

ojw28:
Do you think that applications always be providing a description? Is it best practice to do so? I'm mainly just wondering to what extent the existing constructors we provide are useful. If applications should always be providing a description, presumably the answer is "not very"...

ojw28:
@Timbals - It would be good to get your take on my question above. Thanks!

Timbals:
I always add a description to give the user more information for if they want to customize the notification behavior. I agree that not all applications will always want to provide a description but I think it is best practice to provide the user with more information. You can reject the PR if you think it is not useful enough to be included.

ojw28:
Thanks for the context. We'll get this merged (we'll probably deprecate the old constructors to encourage descriptions to be provided, given it's possible to pass 0 to the new ones if a developer explicitly wants to opt out of providing one). No need to make any changes; we can do this as part of the way we merge changes.

library/core/src/main/java/com/google/android/exoplayer2/util/NotificationUtil.java
```diff
@@ -80,11 +80,48 @@
...
+   *     #IMPORTANCE_DEFAULT} and {@link #IMPORTANCE_HIGH}.
+   * @param descriptionResourceId A String resource identifier for the user visible description of
+   *     the channel. You can change the description of this channel when the system locale changes
+   *     by listening for the {@link Intent#ACTION_LOCALE_CHANGED} broadcast. Ignored if set to 0.
+   */
+  public static void createNotificationChannel(
+      Context context,
+      String id,
+      @StringRes int nameResourceId,
+      @Importance int importance,
```
ojw28, line 115: No need to change this either (we'll swap it around), but just sanity checking: This arg should come last, right? The places you call the new method from (other than the one right above in this class) are using the ordering: `nameResourceId, descriptionResourceId, importance`.

Timbals: Yes, these should be swapped for consistency.

ojw28:
Merged as https://github.com/google/ExoPlayer/commit/0e7f015020b515d0b33e3d80c41448ae35488293

PR #6055, opened 2019-06-18
Parse SmoothStreaming text track subtype into Format.roleflags.

xirac:
According to the SmoothStreaming specification available here https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-sstr/8383f27f-7efe-4c60-832a-387274457251 "Subtype" can describe the role of a text track like _"Closed captions for people who are deaf."_


xirac:
Updating cla

PR #6056, opened 2019-06-18
issue 5550

johnshazhu:
stream reached end but still read data


ojw28:
I don't think this change is doing the right thing. If result is `RESULT_NOTHING_READ` or `RESULT_NOTHING_READ` then the contents of `inputBuffer` aren't valid.

johnshazhu:
I signed it!

johnshazhu:
> I don't think this change is doing the right thing. If result is `RESULT_NOTHING_READ` or `RESULT_NOTHING_READ` then the contents of `inputBuffer` aren't valid.

The result is RESULT_NOTHING_READ and inputBuffer.isEndOfStream is ture, in this condition it will keep read data and can't finish play correctly.

ojw28:
The value of `isEndOfStream` isn't meaningful if the result is `RESULT_NOTHING_READ`.

ojw28:
Closing because this doesn't appear to be doing the correct thing.

Issue #5527, opened 2019-02-18
Expose FLAC VORBIS_COMMENT block in metadata


nlgtuankiet:
Is there a way to pull out .flac file metadata (title, album name...)? I tried the EventLogger in the sample app, It work for .mp3 .m4a but not for .flac. I believe the default FlacExtractor already enable peekId3Data.
Not sure this feature is implemented or not.

andrewlewis:
If it's stored as ID3 data at the start of the file and isn't disabled via `FLAG_DISABLE_ID3_METADATA` this should work. If you provide the media you're trying to play we'll take a look (see the [issue template](https://github.com/google/ExoPlayer/issues/new)).

nlgtuankiet:
Sure, I have tried to retrieve the title with MediaMetadataRetriever, it works, so it definitely in there, I just don't know the proper way to get it 
[Muon Ruou To Tinh.zip](https://github.com/google/ExoPlayer/files/2884270/Muon.Ruou.To.Tinh.-.BigDaddy_.Emily.zip)

andrewlewis:
This stream does not have an ID3 prefix, but after the STREAMINFO block at the start there is a PICTURE block and a VORBIS_COMMENT block. The latter has a title and artist, but I don't think we're parsing it at the moment. (It looks like we only handle STREAMINFO and SEEKTABLE in our libflac wrapper.)

I will mark this as a enhancement to store the vorbis comment payload in the metadata but it's low priority for now. I think we'd also be happy to accept a pull request for this.

szaboa:
@ojw28 
I'd like to solve this, if there is no ongoing work on this currently.

ojw28:
This is being worked on already, I think by @vavadhani or one of his colleagues. We expect support to be added soon.

szaboa:
Alright, then I will search for something else :)

PR #6151, opened 2019-07-09
Add vorbis comments support to flac extractor

vavadhani:
Decode and add vorbis comments from the flac file to metadata.

@ojw28 @umangsaini @harishdm

 #5527


vavadhani:
I signed it

ojw28:
@vavadhani - Please ensure you're registered under your company's Google CLA (this is distinct from the Android CLA). Thanks!

extensions/flac/src/main/jni/flac_jni.cc
```diff
@@ -110,6 +110,32 @@ DECODER_FUNC(jobject, flacDecodeMetadata, jlong jContext) {
                         streamInfo.total_samples);
 }
 
+DECODER_FUNC(jobject, flacDecodeVorbisComment, jlong jContext) {
```
ojw28: For this to return something useful, does it rely on `flacDecodeMetadata` having been called previously (and specifically, on the call to `context->parser->decodeMetadata()` made on L94 above)?

If so, I think it would be much cleaner to rename `FlacDecoderJni.decodeStreamInfo` -> `FlacDecoderJni.decodeMetadata` and `FlacStreamInfo` -> `FlacMetadata`, and put the `Arraylist<String> vorbisComment` inside `FlacMetadata`

Here you'd pass the `vorbisComment` back as part of `flacDecodeMetadata` above, and the need for the separate `flacDecodeVorbisComment` would go away.

What do you think?

extensions/flac/src/main/jni/include/flac_parser.h
```diff
@@ -116,6 +126,8 @@ class FLACParser {
   const FLAC__StreamMetadata_SeekTable *mSeekTable;
   uint64_t firstFrameOffset;
 
+  bool mVorbisCommentValid;
```
ojw28, line 134: nit: Can we keep `mVorbisCommentValid` and `mVorbisComment` grouped together, maybe just under the `mStreamInfo...` params, with a similar comment to those.

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -172,6 +172,30 @@ void FLACParser::metadataCallback(const FLAC__StreamMetadata *metadata) {
     case FLAC__METADATA_TYPE_SEEKTABLE:
       mSeekTable = &metadata->data.seek_table;
       break;
+    case FLAC__METADATA_TYPE_VORBIS_COMMENT:
+      if (!mVorbisCommentValid) {
+        FLAC__uint32 count = 0;
+        const FLAC__StreamMetadata_VorbisComment *vc =
+            &metadata->data.vorbis_comment;
+        mVorbisCommentValid = true;
+        mVorbisComment.metadataArray =
```
harishdm: These aren't freed anywhere.
Instead of malloc and copy, can we hold reference to &vc->comments[i]; and create required strings in flac_jni.cc flacDecodeVorbisComment()?
i.e. getVorbisComment() can return FLAC__StreamMetadata_VorbisComment* 
That would avoid the need for malloc and copy here.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package com.google.android.exoplayer2.metadata.vorbis;
+
+import androidx.annotation.Nullable;
+import com.google.android.exoplayer2.metadata.Metadata;
+import java.util.ArrayList;
+
+/** Decodes vorbis comments */
+public class VorbisCommentDecoder {
```
ojw28: nit: Please make classes final where possible, as it helps reduce our API surface. Ditto elsewhere.

Also, all other `XXXDecoder` classes in the metadata package implement `MetadataDecoder`, so this is slightly confusing. Given the data is already half decoded when it comes out of the JNI layer, it might be simpler just to merge this code into `FlacDecoderJni` or `FlacExtractor`.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+  private static final String SEPARATOR = "=";
+
+  /**
+   * Decodes an {@link ArrayList} of vorbis comments.
+   *
+   * @param metadataStringList An {@link ArrayList} containing vorbis comments as {@link String}
+   * @return A {@link Metadata} structure with the vorbis comments as its entries.
+   */
+  public Metadata decodeVorbisComments(@Nullable ArrayList<String> metadataStringList) {
+    if (metadataStringList == null || metadataStringList.size() == 0) {
```
ojw28: nit: Use `.isEmpty()` rather than `size() == 0`. Ditto below.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+      return null;
+    }
+
+    ArrayList<VorbisCommentFrame> vorbisCommentFrames = new ArrayList<>();
+    VorbisCommentFrame vorbisCommentFrame;
+
+    for (String commentEntry : metadataStringList) {
+      String[] keyValue;
+
+      keyValue = commentEntry.split(SEPARATOR);
```
ojw28: I suspect that it's allowed to have the `=` character in the value. In other words `KEY=VAL=UE` is probably valid. In which case you should be looking the first instance of the separator only. You may find `Util.splitAtFirst` helpful for this.

If you agree, it would be good to have a test case for this as well.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentFrame.java
```diff
@@ -0,0 +1,102 @@
...
+
+import android.os.Parcel;
+import android.os.Parcelable;
+import androidx.annotation.Nullable;
+import com.google.android.exoplayer2.metadata.Metadata;
+
+/** Base class for Vorbis Comment Frames. */
+public class VorbisCommentFrame implements Metadata.Entry {
+
+  /** The frame key and value */
```
ojw28: When generating HTML Javadoc, this is going to attach "The frame key and value" to `key`, and nothing to `value`. Could you document each of them separately?

extensions/flac/src/main/jni/flac_jni.cc
vavadhani: Yes, it does depend on the call to `context->parser->decodeMetadata()`. I will make this change and remove the separate call to `flacDecodeVorbisComment`.

Shall I push the refactor as a separate commit? Would it make the review easier?

extensions/flac/src/main/jni/flac_parser.cc
vavadhani: I have reorganised the code according to other comments, and added the free to the character arrays that will be used.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoder.java
vavadhani: I will remove this class. I will also see how I can still run the tests without this.

extensions/flac/src/main/jni/flac_jni.cc
ojw28: Just adding separate commits as the code evolves is fine (and better than squashing + force pushing I think).

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoder.java
vavadhani: Would making `decodeVorbisComments` a `public static` method in the `StreamMetadata` (previously `StreamInfo`) be acceptable?

ojw28: You could pass the `ArrayList<String>` to the `StreamMetadata` constructor, have the constructor do the decoding, and have `StreamMetadata` hold the decoded `Metadata` directory rather than the `ArrayList<String>`.

 do the decoding as part of the `StreamMetadata` constructor, and then have `StreamMetadata` hold the decoded `Metadata` directly rather than the `ArrayList<String>`.

The test would then be building a `StreamMetadata` and checking that the metadata field after construction is as expected.

extensions/flac/src/main/jni/include/flac_parser.h
vavadhani: Done

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -174,24 +174,16 @@ void FLACParser::metadataCallback(const FLAC__StreamMetadata *metadata) {
...
-                (char *) malloc((vce->length + 1) * sizeof(char));
-            memcpy(mVorbisComment.metadataArray[count], vce->entry,
-                vce->length);
-            mVorbisComment.metadataArray[count][vce->length] = '\0';
-            count++;
+        FLAC__StreamMetadata_VorbisComment vc = metadata->data.vorbis_comment;
+        for (FLAC__uint32 i = 0; i < vc.num_comments; ++i) {
+          FLAC__StreamMetadata_VorbisComment_Entry vce = vc.comments[i];
+          if (vce.entry != NULL) {
+            std::string comment(reinterpret_cast<char *>(vce.entry),
```
vavadhani: Removed the mallocs here and the free. Using `std::vector` and `std::string` instead.

library/core/src/main/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentFrame.java
vavadhani: Done

extensions/flac/src/main/jni/include/flac_parser.h
vavadhani: Done

extensions/flac/src/main/jni/flac_parser.cc
vavadhani, line 181: Removed the mallocs here and the free. Using `std::vector` and `std::string` instead.

vavadhani:
@ojw28 @harishdm @umangsaini I've made some changes, please take a look.

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -172,6 +172,22 @@ void FLACParser::metadataCallback(const FLAC__StreamMetadata *metadata) {
     case FLAC__METADATA_TYPE_SEEKTABLE:
       mSeekTable = &metadata->data.seek_table;
       break;
+    case FLAC__METADATA_TYPE_VORBIS_COMMENT:
+      if (!mVorbisCommentValid) {
+        FLAC__StreamMetadata_VorbisComment vc = metadata->data.vorbis_comment;
```
ojw28, line 177: nit (don't update): Avoid shortened variables names (e.g. use `vorbisComment` rather than `vc`).

library/core/src/main/java/com/google/android/exoplayer2/util/FlacStreamMetadata.java
```diff
@@ -85,6 +94,78 @@ public FlacStreamInfo(
...
+   *
+   * @param minBlockSize Minimum block size of the FLAC stream.
+   * @param maxBlockSize Maximum block size of the FLAC stream.
+   * @param minFrameSize Minimum frame size of the FLAC stream.
+   * @param maxFrameSize Maximum frame size of the FLAC stream.
+   * @param sampleRate Sample rate of the FLAC stream.
+   * @param channels Number of channels of the FLAC stream.
+   * @param bitsPerSample Number of bits per sample of the FLAC stream.
+   * @param totalSamples Total samples of the FLAC stream.
+   * @param vorbisCommentList An {@link ArrayList<String>} that contains vorbis comments, which will
```
ojw28, line 111: nit (don't update): No need to document the type of the argument in the comment, since it's already documented by the method signature.

library/core/src/test/java/com/google/android/exoplayer2/metadata/vorbis/VorbisCommentDecoderTest.java
```diff
@@ -0,0 +1,90 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import com.google.android.exoplayer2.metadata.Metadata;
+import com.google.android.exoplayer2.util.FlacStreamMetadata;
+import java.util.ArrayList;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/** Test for {@link FlacStreamMetadata}'s conversion of {@link ArrayList} to {@link Metadata}. */
+@RunWith(AndroidJUnit4.class)
+public final class VorbisCommentDecoderTest {
```
ojw28, line 29: nit (don't update): Should be `FlacStreamMetadataTest` now (and in the `util` package).

ojw28 (commented):
Thanks for the changes! This looks good to merge. I put some comments inline for future reference, but there's no need to push any changes. We'll take care of them as part of the process we use to get the change merged.

Issue #5749, opened 2019-04-11
Automatically show closed captioning/hard of hearing track based on accessibility settings


sobbe81:
Hi,

I'm using ExoPlayer for playback of DASH-content. I'm looking for a way to 
 * Pick a certain language as default for subtitles, now done by using setPreferredTextLanguage
 * Pick cc/hard of hearing track if this setting in accessibility settings is enabled.   

The DefaultTrackSelector has an API to setPreferredTextLanguage which works fine for displaying subtitle tracks.

However, my app needs to support both subtitles and closed captions and my problem is that if I use setPreferredTextLanguage it will always display the closed captions since these are using the same language code.

The CaptioningManager, https://developer.android.com/reference/android/view/accessibility/CaptioningManager, offers an API to read accessibility settings. From what I understand, ExoPlayer uses this setting to determine the styling of the closed captions but not to determine whether text should be shown or not. 

Is there any way to get the ExoPlayer automatically enable closed captioning based on accessibility settings? And that setPreferredTextLanguage only applies to subtitles and not closed captions?

Thanks,
Martin

tonihei:
Thanks for the detailed description of the problem. I think there are two enhancements in here that need to be addressed:

 1. Automatically select closed captions if `CaptioningManager.isEnabled()` is true. This should only be done if a closed caption track for `CaptioningManager.getLocale()` is available. Or the closed caption track has no language and the `selectUndeterminedTextLanguage` option is enabled.
 2. If the `preferredTextLanguage` option is set and multiple matching text tracks are found, choose the closed caption track if `CaptioningManager.isEnabled()` is true, and the subtitle track otherwise.

Does that summarize your proposal correctly?

sobbe81:
Yes, that's sounds like it would solve my problem.

Just a clarification for one scenario:
if `CaptioningManager.isEnabled()` is false and `preferredTextLanguage` is set to "swe" and there is a closed captioning track, i.e. AdaptationSet with role=caption, available with language code "swe" this track would not be displayed. Correct?

tonihei:
Yes, that's my understanding of the `isEnabled` flag.

sobbe81:
Great!

ojw28:
[...]

PR #6158, opened 2019-07-10
Automatically show closed captioning/hearing impaired text track

xirac:
Following #6055 and after reading #5749 , here are changes on `setPrerredTextLanguage` and `getFormatLanguageScore` allowing to automatically show closed captioning/hearing impaired text track by using the following trackSelectorParameter for instance:
`trackSelectorParameters=newDefaultTrackSelector.ParametersBuilder().setPreferredTextLanguage("fra",C.ROLE_FLAG_CAPTION).build();`


tonihei:
Thanks for taking a look at this!

I'll add some comments to the code for discussion.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -420,6 +420,12 @@ public ParametersBuilder setPreferredTextLanguage(@Nullable String preferredText
       return this;
     }
 
+    @Override
+    public ParametersBuilder setPreferredTextLanguage(@Nullable String preferredTextLanguage, int preferredRoleFlags) {
```
tonihei: I think the API we had in mind when discussing this issue a while ago was having a separate method to set the preferred role flags.

This would allow the following combinations:
1. No language specified, no role flags => select no text track unless marked as "default" or "forced".
3. No language specified, but role flags specified => select text track with matching or partly matching role flags and the selected audio language, select none (or the default/forced track) if no text track with these flags and language is available
2. Language specified, no role flags => select text track of the best matching language, prefer default, non-forced and  tracks without caption flags. If no track with this language is available, select "default" or "forced" in other language if possible.
4. Language and role flags specified => select text track of the best matching language, prefer track with matching or partly matching role flags. If not available, fall back to default, non-forced tracks of the preferred language if possible. If no track of the language is available, fall back to matching role flags in other languages, or default or forced tracks in other languages.

So if someone wants to integrate with CaptioningManager, they would use it like that:
```
if (CaptioningManager.isEnabled()) {
  paramsBuilder.setPreferredRoleFlags(C.ROLE_FLAG_CAPTION | C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND);
  Locale preferredLocale = CaptioningManager.getLocale();
  if (preferredLocale != null) {
    paramsBuilder.setPreferredTextLanguage(Util.getLocaleLanguageTag(preferredLocale));
  }
}
```
This method is complicated enough that we should probably integrate it into SimpleExoPlayer for convenience (or document it somewhere). Not in this PR, that's a just a general reminder for us to do this :)

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2079,7 +2089,7 @@ private static boolean isSupportedAdaptiveAudioTrack(
           boolean isDefault = (maskedSelectionFlags & C.SELECTION_FLAG_DEFAULT) != 0;
           boolean isForced = (maskedSelectionFlags & C.SELECTION_FLAG_FORCED) != 0;
           int trackScore;
-          int languageScore = getFormatLanguageScore(format, params.preferredTextLanguage);
+          int languageScore = getFormatLanguageScore(format, params.preferredTextLanguage, params.preferredRoleFlags);
```
tonihei: The track score logic in selectTextTrack gets increasingly complicated, especially if we follow the proposal I wrote down above. I think before continuing with this change, it would be best to switch from the score based logic to a comparison based logic similar to the one we use for audio track selection (see AudioTrackScore). This allows to get rid of these seemingly arbitrary score numbers (they are not arbitrary, but difficult to understand) and also make the logic more straight-forward. This first change should be a separate PR without changing the actual result and the new functionality here can be then added on top. Are you planning to take a look at this? Otherwise you can wait until we get around to do that ourselves and we'll revisit this PR once that is done.

xirac: Thanks for your comment.
Here is the separate PR moving logic to a TextTrackScore #6178 .

ojw28:
Is the plan to update this now that #6178 has been merged? Thanks!

xirac:
Thanks for the merge of #6178 . I plan to update this one within next 2 weeks.

xirac:
Here is the PR update using TextTrackScore.

xirac:
Do you need some code changes on this pull request?

ojw28:
@tonihei - PTAL

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2702,6 +2713,7 @@ public TextTrackScore(
       preferredLanguageScore =
           getFormatLanguageScore(
               format, parameters.preferredTextLanguage, parameters.selectUndeterminedTextLanguage);
+      preferredRoleFlagsScore = format.roleFlags & parameters.preferredRoleFlags;
```
tonihei: I think you actually need to count the number of matches: Integer.bitCount(format.roleFlags & parameters.preferredRoleFlags)

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2712,7 +2724,7 @@ public TextTrackScore(
       selectedAudioLanguageScore =
           getFormatLanguageScore(format, selectedAudioLanguage, selectedAudioLanguageUndetermined);
       isWithinConstraints =
-          preferredLanguageScore > 0 || isDefault || (isForced && selectedAudioLanguageScore > 0);
+          (preferredLanguageScore > 0 || (parameters.preferredTextLanguage == null && selectedAudioLanguageScore > 0 && preferredRoleFlagsScore > 0)) || isDefault || (isForced && selectedAudioLanguageScore > 0);
```
tonihei: I'm not sure the `selectedAudioLanguageScore > 0` part makes sense. If someone specifies preferred role flags without providing a preferred language, we should probably still select a track with the matching role flags if available.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2730,6 +2742,9 @@ public int compareTo(TextTrackScore other) {
       if (this.preferredLanguageScore != other.preferredLanguageScore) {
         return compareInts(this.preferredLanguageScore, other.preferredLanguageScore);
       }
+      if (this.preferredRoleFlagsScore != other.preferredRoleFlagsScore) {
```
tonihei: If everything else is exactly the same, and no role flag preference is provided in the parameters, we should probably still compare by the existing role flags and prefer the tracks *without* CAPTION or DESCRIBES_MUSIC_AND_SOUND role flags. This under the assumption that no specification means that non-captioning text tracks are preferred.  This comparison should be the very last one after the selectedAudioLanguageScore.

tonihei:
Sorry for the delay. The PR looks good in general, I just left some suggestions for improvement.

xirac:
Thanks. I pushed the 3 changes corresponding to your suggestions.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2736,7 +2753,17 @@ public int compareTo(TextTrackScore other) {
...
+        return compareInts(this.selectedAudioLanguageScore, other.selectedAudioLanguageScore);
+      }
+
+      // If no role flag preference is provided in the parameters, prefer tracks
+      // without C.ROLE_FLAG_CAPTION or C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND
+      // No specification means that non-captioning text tracks are preferred.
+      if (!this.hasPreferredRoleFlags) {
+        return compareInts(other.hasCaptionRoleFlags, this.hasCaptionRoleFlags);
+      }
+      return compareInts(this.preferredRoleFlagsScore, other.preferredRoleFlagsScore);
```
tonihei, line 2766: I actually meant that this comparison stays where it was (after the language score), but never mind, I just move it up again when merging.

tonihei:
Change is merged now. Thanks again for all the work!

xirac:
Thanks for the merge, you're welcome.

PR #6178, opened 2019-07-12
TextTrackScore in a same manner than AudioTrackScore

xirac:
Following this [comment](https://github.com/google/ExoPlayer/pull/6158/files/f4c543445c3441d0a6ad175beeccc552017473d0#r301961272) from @tonihei , here is a change switching text track score from the score based logic to a comparison based logic similar to the one we use for audio track selection (see AudioTrackScore).

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
@@ -1555,29 +1555,39 @@ public class DefaultTrackSelector extends MappingTrackSelector {
     TextTrackScore selectedTextTrackScore = null;
     int selectedTextRendererIndex = C.INDEX_UNSET;
     for (int i = 0; i < rendererCount; i++) {
-      // The below behaviour is different from video and audio track selection
-      // i.e. do not perform a text track pre selection if there are no preferredTextLanguage requested.
-      if (C.TRACK_TYPE_TEXT == mappedTrackInfo.getRendererType(i) && params.preferredTextLanguage != null) {
-        Pair<TrackSelection.Definition, TextTrackScore> textSelection =
-            selectTextTrack(
-                mappedTrackInfo.getTrackGroups(i),
-                rendererFormatSupports[i],
-                params);
-        if (textSelection != null
-            && (selectedTextTrackScore == null
-            || textSelection.second.compareTo(selectedTextTrackScore) > 0)) {
-          if (selectedTextRendererIndex != C.INDEX_UNSET) {
-            // We've already made a selection for another text renderer, but it had a lower
-            // score. Clear the selection for that renderer.
-            definitions[selectedTextRendererIndex] = null;
+      int trackType = mappedTrackInfo.getRendererType(i);
+      switch (trackType) {
+        case C.TRACK_TYPE_VIDEO:
+        case C.TRACK_TYPE_AUDIO:
+          // Already done. Do nothing.
+          break;
+        case C.TRACK_TYPE_TEXT:
+          Pair<TrackSelection.Definition, TextTrackScore> textSelection =
+              selectTextTrack(
+                  mappedTrackInfo.getTrackGroups(i),
+                  rendererFormatSupports[i],
+                  params,
+                  selectedAudioLanguage);
+          if (textSelection != null
+              && (selectedTextTrackScore == null
+              || textSelection.second.compareTo(selectedTextTrackScore) > 0)) {
+            if (selectedTextRendererIndex != C.INDEX_UNSET) {
+              // We've already made a selection for another text renderer, but it had a lower
+              // score. Clear the selection for that renderer.
+              definitions[selectedTextRendererIndex] = null;
+            }
+            definitions[i] = textSelection.first;
+            selectedTextTrackScore = textSelection.second;
+            selectedTextRendererIndex = i;
           }
-          TrackSelection.Definition definition = textSelection.first;
-          definitions[i] = definition;
-          selectedTextTrackScore = textSelection.second;
-          selectedTextRendererIndex = i;
+          break;
+        default:
+            definitions[i] =
+                  selectOtherTrack(
+                      trackType, mappedTrackInfo.getTrackGroups(i), rendererFormatSupports[i], params);
+            break;
         }
       }
-    }
 
     return definitions;
   }
@@ -2052,7 +2062,8 @@ public class DefaultTrackSelector extends MappingTrackSelector {
   protected Pair<TrackSelection.Definition, TextTrackScore> selectTextTrack(
       TrackGroupArray groups,
       int[][] formatSupport,
-      Parameters params)
+      Parameters params,
+      @Nullable String selectedAudioLanguage)
       throws ExoPlaybackException {
     TrackGroup selectedGroup = null;
     int selectedTrackIndex = C.INDEX_UNSET;
@@ -2064,7 +2075,7 @@ public class DefaultTrackSelector extends MappingTrackSelector {
         if (isSupported(trackFormatSupport[trackIndex],
             params.exceedRendererCapabilitiesIfNecessary)) {
           Format format = trackGroup.getFormat(trackIndex);
-          TextTrackScore trackScore = new TextTrackScore(format, params, trackFormatSupport[trackIndex]);
+          TextTrackScore trackScore = new TextTrackScore(format, params, trackFormatSupport[trackIndex], selectedAudioLanguage);
           if ((selectedTrackScore == null) || trackScore.compareTo(selectedTrackScore) > 0) {
             selectedGroup = trackGroup;
             selectedTrackIndex = trackIndex;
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -1552,38 +1552,30 @@ public void experimental_allowMultipleAdaptiveSelections() {
...
+            || textSelection.second.compareTo(selectedTextTrackScore) > 0)) {
+          if (selectedTextRendererIndex != C.INDEX_UNSET) {
+            // We've already made a selection for another text renderer, but it had a lower
+            // score. Clear the selection for that renderer.
+            definitions[selectedTextRendererIndex] = null;
           }
-          break;
-        default:
-          definitions[i] =
-              selectOtherTrack(
```
tonihei, line 1584: You can't remove that. We still need to select "other" tracks.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -1552,38 +1552,30 @@ public void experimental_allowMultipleAdaptiveSelections() {
...
-            if (selectedTextRendererIndex != C.INDEX_UNSET) {
-              // We've already made a selection for another text renderer, but it had a lower score.
-              // Clear the selection for that renderer.
-              definitions[selectedTextRendererIndex] = null;
-            }
-            definitions[i] = textSelection.first;
-            selectedTextTrackScore = textSelection.second;
-            selectedTextRendererIndex = i;
+      // The below behaviour is different from video and audio track selection
+      // i.e. do not perform a text track pre selection if there are no preferredTextLanguage requested.
```
tonihei: I don't fully understand this comment. We sometimes want to select text tracks even without a language (e.g. if the track is the default or a forced track), so you need to call selectTextTrack in all cases.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2494,59 @@ public int hashCode() {
...
+    private final int preferredLanguageScore;
+    private final int localeLanguageMatchIndex;
+    private final int localeLanguageScore;
+    private final boolean isDefaultSelectionFlag;
+
+    public TextTrackScore(Format format, Parameters parameters, int formatSupport) {
+      isWithinRendererCapabilities = isSupported(formatSupport, false);
+      preferredLanguageScore = getFormatLanguageScore(format, parameters.preferredTextLanguage);
+      isDefaultSelectionFlag = (format.selectionFlags & C.SELECTION_FLAG_DEFAULT) != 0;
+      String[] localeLanguages = Util.getSystemLanguageCodes();
```
tonihei: Using System locale languages isn't part of the original logic. Can you leave that out and ensure that the logic stays the same?  The same is true for other parts of the logic that just disappeared with this change. For example selectUndeterminedTextLanguage isn't handled anymore and neither is the forced flag.

tonihei (commented):
The general approach looks right, but please ensure that the existing logic stays the same even if the code is restructured.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
xirac: Thanks for your comment, I re-added it in the next commit.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
xirac: Thanks for your comment, I re-added it in the next commit.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
xirac: My bad, I misunderstood what you were expecting. So I first re-introduced the existing score logic into TextTrackScore. However, I still need to move to a comparison logic, am I right?

tonihei: Yes, that was the idea. The code right now is correct and a step in the right direction, but the main idea was to change all the conditions and scores that are currently in the constructor of TextTrackScore to comparisons in TextTrackScore.compareTo. Because the comparisons are easier to maintain than the score numbers. And it becomes easier to add new features like the role flag parsing.

xirac: Thanks for pointing the right direction.
I've almost done the move to comparisons in `TextTrackScore.compareTo` but I have one issue.
In the existing logic of `selectTextTrack`, it allows to avoid selecting a text track when none conditions matched above:
`          } else {
            // Track should not be selected.
            continue;
          }`

In my last commit, it always selects and enables the first TextTrack, also when not calling `.setPreferredTextLanguage()` on DefaultTrackSelector:

    Renderer:2 [
     Group:0, adaptive_supported=N/A [
       [X] Track:0, id=DVBSUB_TTML_fra_1947161=10000.track_id=30003, mimeType=application/ttml+xml, bitrate=10000, codecs=stpp, language=fra, supported=YES
     ]
     Group:1, adaptive_supported=N/A [
       [ ] Track:0, id=DVBSUB_TTML_fra_1947361=10000.track_id=30001, mimeType=application/ttml+xml, bitrate=10000, codecs=stpp, language=fra, supported=YES
     ]


So my implementation behaviour differs from the existing one right now.

Should I add a `TextTrackScore.shouldNotBeSelected` and test it in `selectTextTrack` to match the existing logic ?


          if (!trackScore.shouldNotBeSelected && ((selectedTrackScore == null) || trackScore.compareTo(selectedTrackScore) > 0)) {
            selectedGroup = trackGroup;
            selectedTrackIndex = trackIndex;
            selectedTrackScore = trackScore;
          }

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+
+    public TextTrackScore(
+        Format format,
+        Parameters parameters,
+        int trackFormatSupport,
+        @Nullable String selectedAudioLanguage) {
+      isWithinRendererCapabilities = isSupported(trackFormatSupport, false);
+      int maskedSelectionFlags =
+          format.selectionFlags & ~parameters.disabledTextTrackSelectionFlags;
+      isDefault = (format.selectionFlags & C.SELECTION_FLAG_DEFAULT) != 0;
```
tonihei: That needs to use maskedSelectionFlags too.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+      isForced = (maskedSelectionFlags & C.SELECTION_FLAG_FORCED) != 0;
+      preferredLanguageScore = getFormatLanguageScore(format, parameters.preferredTextLanguage);
+      selectedAudioLanguageScore = getFormatLanguageScore(format, selectedAudioLanguage);
+      trackHasNoLanguage = formatHasNoLanguage(format);
+      selectUndeterminedTextLanguage = parameters.selectUndeterminedTextLanguage;
+      stringDefinesNoLang = stringDefinesNoLanguage(selectedAudioLanguage);
```
tonihei: The comparison below which use a formatLanguageScore and the undetermined language is quite long. I think you could simplify this a lot by including it all into one "boolean hasLanguageMatch = preferredLanguageScore > 0 || (parameters.selectUndeterminedTextLanguage && trackHasNoLanguage)" and then just compare the boolean.

The same is true for the selectedAudioLanguage. You just need a "boolean hasSelectedAudioLanguageMatch = selectedAudioLanguageScore || (trackHasNoLanguage && stringDefinesNoLanguage(selectedAudioLanguage))"

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+          if ((this.isForced && (this.selectedAudioLanguageScore > 0 || (this.trackHasNoLanguage && this.stringDefinesNoLang))) !=
+              (other.isForced && (other.selectedAudioLanguageScore > 0 || (other.trackHasNoLanguage && other.stringDefinesNoLang)))) {
+            return (this.isForced && (this.selectedAudioLanguageScore > 0
+                || (this.trackHasNoLanguage && this.stringDefinesNoLang))) ? 1 : -1;
+          }
+          // Track should not be selected.
+          return -1;
+        }
+      } else {
+        return (this.preferredLanguageScore > 0 || (this.selectUndeterminedTextLanguage && this.trackHasNoLanguage)) ? 1 : -1;
```
tonihei: Can you move this "else" path up to avoid the high level of indentation?  Just invert the condition to "if (hasLanguageMatch != other.hasLanguageMatch" and return immediately.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+    @Override
+    public int compareTo(@NonNull TextTrackScore other) {
+      if (this.isWithinRendererCapabilities != other.isWithinRendererCapabilities) {
+        return this.isWithinRendererCapabilities ? 1 : -1;
+      }
+      if ((this.preferredLanguageScore > 0 || (this.selectUndeterminedTextLanguage && this.trackHasNoLanguage)) ==
+          (other.preferredLanguageScore > 0 || (other.selectUndeterminedTextLanguage && other.trackHasNoLanguage))) {
+        if (this.preferredLanguageScore > 0 || (this.selectUndeterminedTextLanguage
+            && this.trackHasNoLanguage)) {
+          if (this.isDefault != other.isDefault) {
```
tonihei: This condition is the first one here and in the corresponding else path below. That means you can just move it out of the if condition to avoid the duplication.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+            return this.isDefault ? 1 : -1;
+          }
+          if (this.isForced != other.isForced) {
+            // Prefer non-forced to forced if a preferred text language has been specified. Where
+            // both are provided the non-forced track will usually contain the forced subtitles as
+            // a subset.
+            return !this.isForced ? 1 : -1;
+          }
+          return (this.preferredLanguageScore > other.preferredLanguageScore) ? 1 : -1;
```
tonihei: That should be "this.preferredLanguageScore - other.preferredLanguageScore" because they may also be equal.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2535,4 +2505,76 @@ public int hashCode() {
...
+          if (this.isDefault != other.isDefault) {
+            return this.isDefault ? 1 : -1;
+          }
+          if ((this.isForced && (this.selectedAudioLanguageScore > 0 || (this.trackHasNoLanguage && this.stringDefinesNoLang))) !=
+              (other.isForced && (other.selectedAudioLanguageScore > 0 || (other.trackHasNoLanguage && other.stringDefinesNoLang)))) {
+            return (this.isForced && (this.selectedAudioLanguageScore > 0
+                || (this.trackHasNoLanguage && this.stringDefinesNoLang))) ? 1 : -1;
+          }
+          // Track should not be selected.
+          return -1;
```
tonihei: That's interesting. We need to return 0 here to fulfill the Comparable interface (e.g. comparing the same unsupported text format with itself should always return 0).

To ensure we don't select a track for this case, we should add a "isWithinConstraints" field similar to the AudioTrackScore. This should be something like "=hasLanguageMatch || isDefault || (isForced && hasSelectedAudioLanguageMatch)". And then we can filter out TextTrackScores that are not within the constraints in selectTextTrack().

tonihei (commented):
Thanks - that looks good! I left a couple of comments to improve the code readability. And there is also an issue with the tracks that should not be selected.

xirac:
Thanks for your review. I pushed the changes corresponding to your 6 comments.

tonihei:
Thanks a lot for making this tedious change! Looks correct now and I'm going to merge this. 

Can you wait with updating #6158 until I merged this and pushed it back to the dev branch? Just to ensure it's based on the same code.

xirac:
Yes, I'll wait until merged to resume #6158.
Thanks!

tonihei:
Thanks for waiting. Your change is now merged and we also did some follow-up simplifications to the logic in https://github.com/google/ExoPlayer/commit/e84d88e90f62d1755317cd692496bac58c5b07d3.

Issue #474, opened 2015-05-21
HLS support for trick play with EXT-X-I-FRAMES-ONLY and EXT-X-I-FRAME-STREAM-INF


Arqu:
I can't seem to find any reference in the code to these tags and would like to know if this could be implemented as it would allow more advanced control of the video stream. This would basically lift the ExoPlayer to EXT-X-VERSION 4 of HLS.

ojw28:
Using I-Frame only variants looks pretty optional to me. Marking as enhancement, but probably wont happen any time soon.

Arqu:
Yes, the I-Frame only is not really required for stream playback and core functionality, however it allows us to play in fast forward and reverse which can enhance a users experience.

Arqu:
I've started to work on this, once there is a substantial amount I'll push to the dev branch. Currently I've just added parsing of the media playlist tags, but made sure none of them are processed and executed until this is implemented. 

However, I wanted to ask how should I approach implementing playback in this case, i.e. only play those I-FRAMES and play them in some term of I-FRAMES/second instead of the usual FPS? Maybe even expose this playback speed manipulation so that you can pretty much do 2x 4x 8x... fast forward.

Arqu:
I've found out, that if you simply open up an I-FRAME list, there are two things that happen:
1) It starts playing - thus playback is more or less possible, I just had to play around a little bit with timing and render it immediately to the screen.
2) It gets stuck on the first chunk and keeps reloading it infinitely. Seems like it reads the bytes ok but somewhere along the way the extractor gets stuck and never gets prepared thus throwing the player into an infinite prepare loop.

So any thoughts about case 2? I've spent all day tracking down where it breaks...

Arqu:
I've managed to get a lot of things going regarding this and most of the code is already up on: https://github.com/Arqu/ExoPlayer/tree/dev however I wouldn't want to merge it yet as it is only basic support for directly playing back IFRAME media playlists (It is not really specified that way in the HLS spec but it helps to get things going). However to get this thing done I need to manage a few more things, one is https://github.com/google/ExoPlayer/issues/531 and the other is how to allow switching to an IFRAME track. I've also added some basic code to indicate if a variant has an IFRAME playlist associated with it, but I still have to get down to the part of managing if we are playing the full stream or the IFRAME playlist. So if somebody could help out a bit, especially with the buffering part, that would be great.

deepvernekar:
Hi,
[...]

PR #6270, opened 2019-08-07
Support for HLS EXT-X-IFRAME-ONLY playlists

stevemayhew:
This branch adds support for parsing HLS I-Frame only playlists and adding them as a selectable `Format` in an `TrackGroup`.

All existing unit test cases pass, the branch includes one test case for iFrame specifically.   I modified `AdaptiveTrackSelection` to not include these IDR only `Formats`, so they will not affect current ExoPlayer playback for existing master playlists that already contain iFrame only variants. The expectation is a subclass of same would do this intelligently based on desired playback speed.

I can provide (privately) sample broadcast and VOD streams that have IDR streams you can manually select to see the behavior.  The existing apple streams all parse and present their iFrame only tracks (bipbop), however playback of these is not very interesting ;-).

See my notes in the code in the individual commits for more thoughts on direction.

PS, our CLA is signed and in Google's hands for approval.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -541,16 +541,10 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
 
-    boolean isIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) != 0;
+    boolean isNonIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) == 0;
     boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
 
-    if (Math.abs(playbackSpeed) > 6.0f) {
-      canSelect = isIframeOnly;   // TODO factor in playback speed...
-    } else {
-      canSelect = ! isIframeOnly && canSelect;
-    }
-    return canSelect;
-
+    return canSelect && isNonIframeOnly;    // Default is not to use the IDR only tracks in selection
   }
 
   /**
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -541,10 +541,10 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
 
-    boolean isNonIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) == 0;
+    boolean isIframeOnlyFormat = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) != 0;
     boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
 
-    return canSelect && isNonIframeOnly;    // Default is not to use the IDR only tracks in selection
+    return canSelect && !isIframeOnlyFormat;    // Default is not to use the IDR only tracks in selection
   }
 
   /**
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -540,7 +540,17 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   @SuppressWarnings("unused")
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
-    return Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    boolean isIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) != 0;
+    boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    if (Math.abs(playbackSpeed) > 6.0f) {
+      canSelect = isIframeOnly;   // TODO factor in playback speed...
+    } else {
+      canSelect = ! isIframeOnly && canSelect;
+    }
+    return canSelect;
+
   }
 
   /**
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
@@ -470,7 +470,6 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
             : Collections.emptyMap();
 
     boolean hasVariants = !masterPlaylist.variants.isEmpty();
-    boolean hasIFrameVariants = !masterPlaylist.iFrameVariants.isEmpty();
     List<Rendition> audioRenditions = masterPlaylist.audios;
     List<Rendition> subtitleRenditions = masterPlaylist.subtitles;
 
@@ -564,7 +563,8 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
       long positionUs,
       List<HlsSampleStreamWrapper> sampleStreamWrappers,
       List<int[]> manifestUrlIndicesPerWrapper,
-      Map<String, DrmInitData> overridingDrmInitData, List<Variant> variants) {
+      Map<String, DrmInitData> overridingDrmInitData,
+      List<Variant> variants) {
     int[] variantTypes = new int[variants.size()];
     int videoVariantCount = 0;
     int audioVariantCount = 0;
@@ -627,8 +627,8 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
       List<TrackGroup> muxedTrackGroups = new ArrayList<>();
       if (variantsContainVideoCodecs) {
         Format[] videoFormats = new Format[selectedVariantsCount];
-        for (int i1 = 0; i1 < videoFormats.length; i1++) {
-          videoFormats[i1] = deriveVideoFormat(selectedPlaylistFormats[i1]);
+        for (int i = 0; i < videoFormats.length; i++) {
+          videoFormats[i] = deriveVideoFormat(selectedPlaylistFormats[i]);
         }
         muxedTrackGroups.add(new TrackGroup(videoFormats));
 
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
@@ -35,7 +35,6 @@ import com.google.android.exoplayer2.source.SequenceableLoader;
 import com.google.android.exoplayer2.source.TrackGroup;
 import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.source.hls.playlist.HlsMasterPlaylist;
-import com.google.android.exoplayer2.source.hls.playlist.HlsMasterPlaylist.IFrameVariant;
 import com.google.android.exoplayer2.source.hls.playlist.HlsMasterPlaylist.Rendition;
 import com.google.android.exoplayer2.source.hls.playlist.HlsMasterPlaylist.Variant;
 import com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistTracker;
@@ -479,22 +478,17 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
     ArrayList<HlsSampleStreamWrapper> sampleStreamWrappers = new ArrayList<>();
     ArrayList<int[]> manifestUrlIndicesPerWrapper = new ArrayList<>();
 
-    if (hasVariants) {
+    ArrayList<Variant> allVariants = new ArrayList<>();
+    allVariants.addAll(masterPlaylist.variants);
+    allVariants.addAll(masterPlaylist.iFrameVariants);
+    if (hasVariants){
       buildAndPrepareMainSampleStreamWrapper(
           masterPlaylist,
           positionUs,
           sampleStreamWrappers,
           manifestUrlIndicesPerWrapper,
-          overridingDrmInitData);
-    }
-
-    if (hasIFrameVariants) {
-      buildAndPrepareIFrameSampleStreamWrappers(
-          masterPlaylist,
-          positionUs,
-          sampleStreamWrappers,
-          overridingDrmInitData
-      );
+          overridingDrmInitData,
+          allVariants);
     }
 
     // TODO: Build video stream wrappers here.
@@ -570,12 +564,12 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
       long positionUs,
       List<HlsSampleStreamWrapper> sampleStreamWrappers,
       List<int[]> manifestUrlIndicesPerWrapper,
-      Map<String, DrmInitData> overridingDrmInitData) {
-    int[] variantTypes = new int[masterPlaylist.variants.size()];
+      Map<String, DrmInitData> overridingDrmInitData, List<Variant> variants) {
+    int[] variantTypes = new int[variants.size()];
     int videoVariantCount = 0;
     int audioVariantCount = 0;
-    for (int i = 0; i < masterPlaylist.variants.size(); i++) {
-      Variant variant = masterPlaylist.variants.get(i);
+    for (int i = 0; i < variants.size(); i++) {
+      Variant variant = variants.get(i);
       Format format = variant.format;
       if (format.height > 0 || Util.getCodecsOfType(format.codecs, C.TRACK_TYPE_VIDEO) != null) {
         variantTypes[i] = C.TRACK_TYPE_VIDEO;
@@ -606,10 +600,10 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
     Format[] selectedPlaylistFormats = new Format[selectedVariantsCount];
     int[] selectedVariantIndices = new int[selectedVariantsCount];
     int outIndex = 0;
-    for (int i = 0; i < masterPlaylist.variants.size(); i++) {
+    for (int i = 0; i < variants.size(); i++) {
       if ((!useVideoVariantsOnly || variantTypes[i] == C.TRACK_TYPE_VIDEO)
           && (!useNonAudioVariantsOnly || variantTypes[i] != C.TRACK_TYPE_AUDIO)) {
-        Variant variant = masterPlaylist.variants.get(i);
+        Variant variant =variants.get(i);
         selectedPlaylistUrls[outIndex] = variant.url;
         selectedPlaylistFormats[outIndex] = variant.format;
         selectedVariantIndices[outIndex++] = i;
@@ -633,8 +627,8 @@ public final class HlsMediaPeriod implements MediaPeriod, HlsSampleStreamWrapper
       List<TrackGroup> muxedTrackGroups = new ArrayList<>();
       if (variantsContainVideoCodecs) {
         Format[] videoFormats = new Format[selectedVariantsCount];
-        for (int i = 0; i < videoFormats.length; i++) {
-          videoFormats[i] = deriveVideoFormat(selectedPlaylistFormats[i]);
+        for (int i1 = 0; i1 < videoFormats.length; i1++) {
+          videoFormats[i1] = deriveVideoFormat(selectedPlaylistFormats[i1]);
         }
         muxedTrackGroups.add(new TrackGroup(videoFormats));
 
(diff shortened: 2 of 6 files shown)


library/core/src/main/java/com/google/android/exoplayer2/C.java
```diff
@@ -1073,6 +1073,10 @@ private C() {}
   /** Indicates the track contains a text that has been edited for ease of reading. */
   public static final int ROLE_FLAG_EASY_TO_READ = 1 << 13;
 
+  // TODO - not a 'role' in the sense it is parsed from the CHARACTERISTICS attribute... forced if iFrame only
+  /** Indicates the track is an IDR (IFrame) only track for trick play */
+  public static final int ROLE_FLAG_TRICK_PLAY = 1 << 14;
+
```
stevemayhew: I'm certainly not married to this idea...  But, there needs to be some hook in the `Format` to identify the Iframe only tracks.  It's unfortunate the HLS spec does not mandate (or even allow) specifying the frame rate on the Master Playlist EXT-X-IFRAME-ONLY

library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
```diff
@@ -583,7 +583,11 @@ public int evaluateQueueSize(long playbackPositionUs, List<? extends MediaChunk>
   @SuppressWarnings("unused")
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
-    return Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    boolean isNonIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) == 0;
+    boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    return canSelect && isNonIframeOnly;    // Default is not to use the IDR only tracks in selection
   }
```
stevemayhew, line 548: This will prevent any of the Iframe `Format` from being part of default ABR.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
```diff
@@ -454,20 +454,25 @@ private void buildAndPrepareSampleStreamWrappers(long positionUs) {
...
+    if (hasVariants){
       buildAndPrepareMainSampleStreamWrapper(
           masterPlaylist,
           positionUs,
           sampleStreamWrappers,
           manifestUrlIndicesPerWrapper,
-          overridingDrmInitData);
+          overridingDrmInitData,
+          allVariants);
     }
```
stevemayhew, line 491: Sorry the diff might be a bit hard to read.  Basically added an argument to pass Variants and replaced all references to the `variants` field in the `HlsMasterPlaylist` with same.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylist.java
```diff
@@ -94,6 +95,28 @@ public Variant(
...
+      this(
+          url,
+          format,
+          videoGroupId,
+          /* audioGroupId */null,
+          /* subtitleGroupId */null,
+          /* captionGroupId */null);
+    }
+
     /**
```
stevemayhew, line 120: Alternate 'VIDEO' renditions should be a TODO, it is in the spec but I've never seen it in the wild.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylist.java
```diff
@@ -201,23 +227,25 @@ public Rendition(@Nullable Uri url, Format format, String groupId, String name)
...
+          List<Rendition> videos,
+          List<Rendition> audios,
+          List<Rendition> subtitles,
+          List<Rendition> closedCaptions,
+          Format muxedAudioFormat,
+          List<Format> muxedCaptionFormats,
+          boolean hasIndependentSegments,
+          Map<String, String> variableDefinitions,
+          List<DrmInitData> sessionKeyDrmInitData) {
     super(baseUri, tags, hasIndependentSegments);
```
stevemayhew, line 234: Sorry, looks like I've missed the spacing default.  It should be 2 spaces for indent I assume?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -300,38 +302,11 @@ private static HlsMasterPlaylist parseMasterPlaylist(LineIterator iterator, Stri
...
-        }
-        float frameRate = Format.NO_VALUE;
-        String frameRateString =
-            parseOptionalStringAttr(line, REGEX_FRAME_RATE, variableDefinitions);
-        if (frameRateString != null) {
-          frameRate = Float.parseFloat(frameRateString);
-        }
+        int bitrate = parseStreamBitrate(variableDefinitions, line);
+        Format format = parseVideoContainerFormat(variableDefinitions, line, formatId, bitrate, 0);
         String videoGroupId = parseOptionalStringAttr(line, REGEX_VIDEO, variableDefinitions);
```
stevemayhew, line 314: I refactored to lift methods for the two code blocks to parse out the bitrate and the codecs/frame rate.   As regular Variants (`EXT-X-STREAM-INF`) are nearly identical to I-Frame only variants (`EXT-X-I-FRAME-STREAM-INF`)

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -368,6 +330,16 @@ private static HlsMasterPlaylist parseMasterPlaylist(LineIterator iterator, Stri
         variantInfosForUrl.add(
             new VariantInfo(
                 bitrate, videoGroupId, audioGroupId, subtitlesGroupId, closedCaptionsGroupId));
+      } else if (line.startsWith(TAG_STREAM_IFRAME)) {
+        String formatId = "iFrame-" + iFrameVariants.size();
+        int bitrate = parseStreamBitrate(variableDefinitions, line);
+        Format format = parseVideoContainerFormat(variableDefinitions, line, formatId, bitrate,
+            C.ROLE_FLAG_ALTERNATE | C.ROLE_FLAG_TRICK_PLAY);
+
```
stevemayhew, line 351: ROLE_FLAG_ALTERNATE should probably go, I added this to surface it to the track selector UI.

liambresnahan:
@googlebot I signed it!

stevemayhew:
@googlebot I signed it!

tonihei:
@AquilesCanta  Would you mind reviewing this change? See #474 for context.

biespb:
library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java

AfanasievAnton:
hey, was wondering about your future plans with regard to this pr. 
Do you consider to review it, or decline. And if so, are you planning to provide some kind of support for I-FRAMES extraction?

stevemayhew:
@AfanasievAnton I've got another pull reques that also cover this same set of files, but I need to rebase this one.  Will do that this week.

We are using this with our release version of ExoPlayer (2.10.6 based) for visual trick play.

ojw28:
@AquilesCanta - Did you have a chance to look at this?

stevemayhew:
Added a commit to enable support for "implicit" Media Initialization Section (PAT/PMT for MPEG-TS).  This is generated by Apple's tools (mediafilesegmenter for example).

stevemayhew:
Rebased to latest `dev-v2` (removing the old merge commits).  @AquilesCanta let me know if you need me to rebase and collapse some commits to make it clearer.  I did merge in the updates to use the new `Format.builder()` in the method I refactor extracted.

stevemayhew:
PS I can send you some sample content to try as well.  You can either modify demo player to change speed or simply use track selector to override select the i-Frame only track and play it.

Locally I subclassed `AdaptiveTrackSelection` and change this code to select i-Frame only track if over 8x speed (you'll want to disable audio if you do this):

````Java
  protected boolean canSelectFormat(
      Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {

    boolean isNonIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) == 0;
    boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;

    return canSelect && isNonIframeOnly;    // Default is not to use the IDR only tracks in selection
  }
````

**Comment by AquilesCanta** [comment]
If this is now updated to dev-v2 I'll do my best to look into it this week. Sorry for the delay, there are some high priority deadlines.

**Comment by stevemayhew** [comment]
No worries @AquilesCanta I'm becoming good at merging it ;-)  I'm keeping a back-ported to 2.10.6 version as well for us internally where I've added the support for the implicit init segment (Apple's hack), so that triggered updating this branch too.     I'll keep using rebase to avoid pollution of the history with merges.

File: library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java [hunk 8]
```diff
@@ -540,7 +540,11 @@ public int evaluateQueueSize(long playbackPositionUs, List<? extends MediaChunk>
   @SuppressWarnings("unused")
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
-    return Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    boolean isNonIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) == 0;
```
AquilesCanta: Nit: Can we replace this variable is isTrickPlayTrack or isIFrameOnlyTrack (would require negating the comparison to != 0). Otherwise, it could be read as a track that has no I-frames, which would be inaccurate. Right?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
```diff
@@ -470,20 +470,25 @@ private void buildAndPrepareSampleStreamWrappers(long positionUs) {
...
     List<Rendition> subtitleRenditions = masterPlaylist.subtitles;
 
     pendingPrepareCount = 0;
     ArrayList<HlsSampleStreamWrapper> sampleStreamWrappers = new ArrayList<>();
     ArrayList<int[]> manifestUrlIndicesPerWrapper = new ArrayList<>();
 
-    if (hasVariants) {
+    ArrayList<Variant> allVariants = new ArrayList<>();
+    allVariants.addAll(masterPlaylist.variants);
+    allVariants.addAll(masterPlaylist.iFrameVariants);
```
AquilesCanta, line 482: So my view on the I-FRAME tracks is that they should belong in their own track group, I think.

One can summarise (and simplify) a TrackGroup as a group of tracks that can be  adapted between. 
- However, it's not intended that a player adapts between a regular track and a I-FRAME only tag.
- Additionally, I can conceive a scenario where I-frame tags are track-selected to a different renderer than the variants. For these two reasons I think we should put I-FRAME only tracks in their own track group. To do this, they need to be in their own HlsSampleStreamWrapper.

Please do feel free to provide evidence pointing in a different direction.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
```diff
@@ -559,12 +564,12 @@ private void buildAndPrepareMainSampleStreamWrapper(
       long positionUs,
       List<HlsSampleStreamWrapper> sampleStreamWrappers,
       List<int[]> manifestUrlIndicesPerWrapper,
-      Map<String, DrmInitData> overridingDrmInitData) {
-    int[] variantTypes = new int[masterPlaylist.variants.size()];
+      Map<String, DrmInitData> overridingDrmInitData, List<Variant> variants) {
```
AquilesCanta: Super-nit, since it's easy to fix. Mind moving this to the next line?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
```diff
@@ -622,8 +627,8 @@ private void buildAndPrepareMainSampleStreamWrapper(
       List<TrackGroup> muxedTrackGroups = new ArrayList<>();
       if (variantsContainVideoCodecs) {
         Format[] videoFormats = new Format[selectedVariantsCount];
-        for (int i = 0; i < videoFormats.length; i++) {
-          videoFormats[i] = deriveVideoFormat(selectedPlaylistFormats[i]);
+        for (int i1 = 0; i1 < videoFormats.length; i1++) {
```
AquilesCanta: Unless this change is necessary, i'd avoid it. Just to keep the change easier to review.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -550,6 +575,9 @@ private static HlsMediaPlaylist parseMediaPlaylist(
     long segmentStartTimeUs = 0;
     long segmentByteRangeOffset = 0;
     long segmentByteRangeLength = C.LENGTH_UNSET;
+    boolean isExplicitInitSegment = false;   // Was declared (TAG_INIT_SEGMENT) vs implied.
```
AquilesCanta, line 578: Can you describe the purpose of this variable? The whole code around initialisation segment is not very clear to me.

AquilesCanta (commented):
Thanks for the PR! Two main points to visit, rest of comments just nits.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
stevemayhew: Sure, that sounds fine.  Really this logic is just to prevent selecting these tracks ever, for now.  The right thing to do is add the logic to TrackSelection itself.  In our back-port to release we have subclassed `AdaptiveTrackSelection`.   The idea is when you move to high-speed mode (or scrub based trick-play) a TrackSelection occurs that results in a `TrackSelection.Definition` that includes only formats that are relevant (either all i-Frame or all non-iFrame tracks).   Oliver had other thoughts about adapting based on frame rate.   I think all of these are worth visiting, and the changes would be restricted to track selection so this code could be merged in first.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
stevemayhew: sure, is the preferred pattern once there needs to be an argument list wrap everything goes to it's own line?

Consistency is important...  Wish Intellij had rules for how to do this.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
stevemayhew: Sure..  The Apple sample iFrame playlists do not have EXT-X-MAP for their byte-range segments, so the Media Initialization Segment is "implicit",  one sentence in the Pantos draft explains this (I missed it first pass too).  This commit's 1f11233ba022e00367ed32904092070d8c42f5ee comment explains it.  Their sample streams are authored this way, `mediafilesegmenter` does this.  IMHO, it is a hack, but it is their 'standard'.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
stevemayhew: I originally did it this way, https://github.com/google/ExoPlayer/issues/474#issuecomment-420594414  

@ojw28 Made good points for including all the tracks in the same adaption set https://github.com/google/ExoPlayer/issues/474#issuecomment-514307811  So basically we are adapting based on frame rate as well as bandwidth.  

One advantage to the current approach, that Oliver advocated, is it may make more sense to trick play a very low bandwidths 30FPS track rather than a higher resolution low frame rate i-Frame only track.  Having the tracks all in an adaptive set allows this to occur when adapting to speed changes without the overhead of a full on track selection.

I can go either way, could we get the parsing part in first at least, and do this as a later enhancement?   Rebasing my collection of pull requests that alter this code is becoming laborious ;-), there is a lot of change to these files.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
stevemayhew: Yeah, sorry about that.  Looks like an overzealous refactor operation must have thought duplicate `i` variables were in scope or something like that.  I'll change it back when I fix the nits.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
AquilesCanta: > could we get the parsing part in first at least

Hah, I was going to suggest splitting this into smaller independent pieces, but I didn't want to get too annoying. So I'll assume the scope of this change is only parsing as of now?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylist.java
AquilesCanta: Correct. 2 spaces.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
AquilesCanta: Sounds good.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
AquilesCanta: I've sent an email to Bill May to see if he can explain the reason behind that bit in the spec. If he replies soonish we might be able to get rid of this.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
stevemayhew: latest checkin fixes this.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
stevemayhew: Replaced with a positive logic variable, hopefully this is more clear.   The function should return fall if the track is an iFrame only track.

Logic to return true will depend on the trick-play implementation.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
stevemayhew: removed.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
```diff
@@ -470,7 +470,6 @@ private void buildAndPrepareSampleStreamWrappers(long positionUs) {
             : Collections.emptyMap();
 
     boolean hasVariants = !masterPlaylist.variants.isEmpty();
-    boolean hasIFrameVariants = !masterPlaylist.iFrameVariants.isEmpty();
```
stevemayhew: Variable removed, unused.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylist.java
```diff
@@ -181,6 +206,7 @@ public Rendition(@Nullable Uri url, Format format, String groupId, String name)
    * @param baseUri See {@link #baseUri}.
    * @param tags See {@link #tags}.
    * @param variants See {@link #variants}.
+   * @param iFrameVariants
```
AquilesCanta, line 209: Nit: Can we add See {@link iFrameVariants}?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaPeriod.java
AquilesCanta: Seems to be pretty well contained now. Thanks!

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
AquilesCanta: Sorry for the (new) delay on this. The response from Bill was that it's basically due to historical reasons.

AquilesCanta (commented):
Sorry I forgot to send the previous round of comments. I'll try to merge this next week and address any minor changes manually. Thanks for the pull request!

AquilesCanta:
Hi @stevemayhew, have you been able to play an i-frame-only variant?

I was playing around with this, and it would seem that the fmp4 extractor throws whenever you try to play an i-frame only variant. Could you please confirm for me whether you have used the fmp4 extractor to play i-frame only variants? If not, I think I will disable them until we manage to make it work.

stevemayhew:
We are definitely using fMP4 with HLS (looks like @ojw28 is working on adding DASH support, thanks so much, we’ll need that next).  I can provide you with a test stream if you’d like from our origin servers.

Perhaps you are testing against the Apple bib-bop streams,  I just tried it (with an r2.11.4 build) and it hit an assert in the extractor, can I assign a bug to us and fix it?

AquilesCanta:
> Perhaps you are testing against the Apple bib-bop streams

Correct. I was assuming the issue was that the I-frame only variants were just trimming samples from the original content, so the extractor was expecting the missing samples.

So if you have working I-FRAME only streams, do you use a i-frame-only-specific EXT-X-MAP for them (as opposed to just linking to ranges in the original content)?

I'll file an issue for fixing I-FRAME only support in the HLS bip-bop streams. You are welcome to send a pull request for fixing that.

AquilesCanta:
Filed #7512.

stevemayhew:
Thanks,
Yes, we’ll have a look at the bug and work on a fix.
 
All the origins we have support live, the produce the iFrame tracks from the original segments but it is not the same mp4.   Apple’s mediafilesegmenter (VOD) simply identifies  IDR’s in the original content as byte-ranges from the main mp4.

PR #6286, opened 2019-08-13
Support negative value of the @r attrbute of S in SegmentTimeline element

sr1990:
Hi,
The following patch supports parsing negative value of the @r attribute of S in SegmentTimeline element.

Example streams:
1. http://dash.edgesuite.net/dash264/TestCases/1c/qualcomm/1/MultiRate.mpd
2. http://dash.edgesuite.net/dash264/TestCases/1c/qualcomm/2/MultiRate.mpd


library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -996,7 +1001,12 @@ protected EventMessage buildEvent(
       if (XmlPullParserUtil.isStartTag(xpp, "S")) {
         elapsedTime = parseLong(xpp, "t", elapsedTime);
         long duration = parseLong(xpp, "d", C.TIME_UNSET);
-        int count = 1 + parseInt(xpp, "r", 0);
+
+        //if repeat is -1 : length of each segment = duration / timescale and
+        //                   number of segments  = periodDuration / length of each segment
+        int repeat = parseInt(xpp,"r",0);
+        int count = repeat != -1? 1 + repeat : (int) (((periodDurationMs / 1000) * timescale) / duration);
```
tonihei: There is no mention in the DASH IOP that this value needs to be -1 specifically. It just says  "if the @r value is greater than or equal to zero" for one case and "else" for the other (=this) case. So can you check for < 0 instead?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -733,7 +736,7 @@ protected SegmentList parseSegmentList(XmlPullParser xpp, @Nullable SegmentList
       if (XmlPullParserUtil.isStartTag(xpp, "Initialization")) {
         initialization = parseInitialization(xpp);
       } else if (XmlPullParserUtil.isStartTag(xpp, "SegmentTimeline")) {
-        timeline = parseSegmentTimeline(xpp);
+        timeline = parseSegmentTimeline(xpp,timescale,duration);
```
tonihei: That also needs to use the period duration and not the segment list duration. Can you forward the period duration is the same way as for the other methods?

tonihei (commented):
Thanks for the PR! The approach looks good in general. Could you address the minor issues I commented on? Thanks.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
sr1990: Re-submitted patch with the above change.

tonihei:
Thanks, looks good now. I'll merge that soon.

tonihei:
Merged. We also added additional support for multiple `S` elements with time offsets and negative repeat count.

PR #6289, opened 2019-08-13
Change mime type of AV1 from video/av01 to video/av1

zhouziyan:
The AV1 test vectors provided by Netflix can not be played
due to incorrect mime type. Fix this issue by changing mime type
from video/av01 to video/av1.
The test vectors link:
http://download.opencontent.netflix.com/?prefix=AV1/Chimera/


zhouziyan:
Use 6920 instead. Thanks.

PR #6298, opened 2019-08-14
Checking directory exist before create

bwindlo:
It should check if the parent directory exists before creating it. Otherwise, the directory creation fails and a wrong exception message is thrown.


tonihei:
Please sign the CLA first. Otherwise we are not able to take a look at the pull request. If you want to report a bug instead, please open a new issue (i.e. not a pull request).

ojw28:
This change isn't quite right because it retries creating the `AtomicFileOutputStream` even if nothing has changed. Given there's no CLA here, I'll close this and fix the exception message with a different change.

PR #6321, opened 2019-08-20
Feature discard upstream HLS

stevemayhew:
Pull request is not ready to merge.  The HLS libraries `HlsSampleStreamWrapper` needs a way to manage the queued samples matching up with the `HlsMediaChunk` that will be discarded.  See notes in the diffs below.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -451,6 +451,13 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
     return chunkIterators;
   }
 
+  public int getPreferredQueueSize(long playbackPositionUs, List<? extends MediaChunk> queue) {
+    if (fatalError != null || trackSelection.length() < 2) {
+      return queue.size();
+    }
+    return trackSelection.evaluateQueueSize(playbackPositionUs, queue);
+  }
+
   // Private methods.
 
   /**
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
@@ -27,6 +27,7 @@ import com.google.android.exoplayer2.extractor.PositionHolder;
 import com.google.android.exoplayer2.metadata.Metadata;
 import com.google.android.exoplayer2.metadata.id3.Id3Decoder;
 import com.google.android.exoplayer2.metadata.id3.PrivFrame;
+import com.google.android.exoplayer2.source.SampleQueue;
 import com.google.android.exoplayer2.source.chunk.MediaChunk;
 import com.google.android.exoplayer2.source.hls.playlist.HlsMediaPlaylist;
 import com.google.android.exoplayer2.upstream.DataSource;
@@ -222,6 +223,12 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   private volatile boolean loadCanceled;
   private boolean loadCompleted;
 
+  /**
+   * Index of first sample written to the SampleQueue for the primary track from
+   * this segment.
+   */
+  private int firstSampleIndex = C.INDEX_UNSET;
+
   private HlsMediaChunk(
       HlsExtractorFactory extractorFactory,
       DataSource mediaDataSource,
@@ -291,6 +298,15 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     return loadCompleted;
   }
 
+  /**
+   * Return the index of the first sample from the primary sample stream for this media chunk
+   *
+   * @return sample index {@link SampleQueue#getWriteIndex()}
+   */
+  public int getFirstPrimarySampleIndex() {
+    return firstSampleIndex;
+  }
+
   // Loadable implementation
 
   @Override
@@ -308,6 +324,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
       initDataLoadRequired = false;
       output.init(uid, shouldSpliceIn, /* reusingExtractor= */ true);
     }
+    firstSampleIndex = output.getPrimaryTrackWritePosition();
     maybeLoadInitData();
     if (!loadCanceled) {
       if (!hasGapTag) {
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
@@ -300,7 +300,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   }
 
   /**
-   * Return the index of the first sample from the primary sample stream for this media chunk
+   * Return the indexes of the first samples from each sample queue for this media chunk
    *
    * @return sample index {@link SampleQueue#getWriteIndex()}
    */
@@ -308,6 +308,13 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     return firstSampleIndexes;
   }
 
+  /**
+   * Set the index of the first sample written to a sample queue from this media chunk,
+   * indexed by the caller's stream index for the sample queue
+   *
+   * @param streamIndex - caller's index for the {@link SampleQueue}
+   * @param firstSampleIndex - index value to store
+   */
   void setFirstSampleIndex(int streamIndex, int firstSampleIndex) {
     firstSampleIndexes = Arrays.copyOf(firstSampleIndexes, streamIndex + 1);
     firstSampleIndexes[streamIndex] = firstSampleIndex;
@@ -330,7 +337,6 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
       initDataLoadRequired = false;
       output.init(uid, shouldSpliceIn, /* reusingExtractor= */ true);
     }
-    firstSampleIndexes = output.getTrackWritePositions();
     maybeLoadInitData();
     if (!loadCanceled) {
       if (!hasGapTag) {
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
@@ -40,6 +40,7 @@ import com.google.android.exoplayer2.util.Util;
 import java.io.EOFException;
 import java.io.IOException;
 import java.math.BigInteger;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
@@ -224,10 +225,10 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
   private boolean loadCompleted;
 
   /**
-   * Index of first sample written to the SampleQueue for the primary track from
-   * this segment.
+   * Index of first sample written for each SampleQueue created from the extraction of this
+   * media chunk
    */
-  private int firstSampleIndex = C.INDEX_UNSET;
+  private int [] firstSampleIndexes = new int[0];
 
   private HlsMediaChunk(
       HlsExtractorFactory extractorFactory,
@@ -303,8 +304,13 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
    *
    * @return sample index {@link SampleQueue#getWriteIndex()}
    */
-  public int getFirstPrimarySampleIndex() {
-    return firstSampleIndex;
+  int [] getFirstSampleIndexes() {
+    return firstSampleIndexes;
+  }
+
+  void setFirstSampleIndex(int streamIndex, int firstSampleIndex) {
+    firstSampleIndexes = Arrays.copyOf(firstSampleIndexes, streamIndex + 1);
+    firstSampleIndexes[streamIndex] = firstSampleIndex;
   }
 
   // Loadable implementation
@@ -324,7 +330,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
       initDataLoadRequired = false;
       output.init(uid, shouldSpliceIn, /* reusingExtractor= */ true);
     }
-    firstSampleIndex = output.getPrimaryTrackWritePosition();
+    firstSampleIndexes = output.getTrackWritePositions();
     maybeLoadInitData();
     if (!loadCanceled) {
       if (!hasGapTag) {
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
@@ -1049,23 +1049,6 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     }
   }
 
-  /**
-   * Get the SampleQueue write position indexes.  Used to associate group of
-   * samples with a MediaChunk.
-   *
-   * @return array of write positions {@link SampleQueue#getWriteIndex()}, by sample queue index
-   */
-  int [] getTrackWritePositions() {
-    int [] writePositions;
-
-    writePositions = new int[sampleQueues.length];
-    for (int i=0; i < sampleQueues.length; i++) {
-      writePositions[i] = sampleQueues[i].getWriteIndex();
-    }
-
-    return writePositions;
-  }
-
   // Internal methods.
 
   private HlsMediaChunk findChunkMatching(int chunkUid) {
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
@@ -696,9 +696,65 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 
   @Override
   public void reevaluateBuffer(long positionUs) {
-    // Do nothing.
+    if (loader.isLoading() || isPendingReset()) {
+      return;
+    }
+
+    int currentQueueSize = mediaChunks.size();
+    int preferredQueueSize = chunkSource.getPreferredQueueSize(positionUs, readOnlyMediaChunks);
+    if (currentQueueSize > preferredQueueSize) {
+      Log.d(TAG, "reevaluateBuffer() -  position: " + positionUs + " preferredQueueSize: "
+          + preferredQueueSize + " current size: "+ currentQueueSize);
+
+      long firstRemovedStartTimeUs = discardMediaChunks(preferredQueueSize);
+      long endTimeUs = getLastMediaChunk().endTimeUs;
+      eventDispatcher.upstreamDiscarded(primarySampleQueueType, firstRemovedStartTimeUs, endTimeUs);
+    }
   }
 
+
+  /**
+   * Discards HlsMediaChunks, after currently playing chunk {@see #haveReadFromMediaChunk}, that have
+   * not yet started to play to allow (hopefully) higher quality chunks to replace them
+   *
+   * @param preferredQueueSize - desired media chunk queue size (always < mediaChunks.size())
+   * @return endTimeUs of first chunk removed
+   */
+  private long discardMediaChunks(int preferredQueueSize) {
+    Log.d(TAG, "discardChunksToIndex() -  preferredQueueSize " + preferredQueueSize
+        + " currentSize " + mediaChunks.size()
+        + " write: "+sampleQueues[primarySampleQueueIndex].getWriteIndex()
+        + " read: "+sampleQueues[primarySampleQueueIndex].getReadIndex()
+    );
+    for (int i=0; i<mediaChunks.size(); i++) {
+      HlsMediaChunk chunk = mediaChunks.get(i);
+      Log.d(TAG, "chunk " + chunk.uid + " reading: " + haveReadFromMediaChunk(i)
+          + " start/end: " + chunk.startTimeUs + "/" + chunk.endTimeUs + " sample index: "
+          + chunk.getFirstPrimarySampleIndex() + " format: "+chunk.trackFormat);
+    }
+
+    // Preserve current playing chunk and/or enough following it to reach the desired queue size
+    int firstRemovedChunkIndex = 0;
+    while (haveReadFromMediaChunk(firstRemovedChunkIndex) || preferredQueueSize > 0) {
+      firstRemovedChunkIndex++;
+      preferredQueueSize--;
+    }
+
+    HlsMediaChunk firstRemovedChunk = mediaChunks.get(firstRemovedChunkIndex);
+    Util.removeRange(mediaChunks, firstRemovedChunkIndex, mediaChunks.size() - 1);
+    Log.d(TAG, "discardChunksToIndex() -  discard from: " + firstRemovedChunk.getFirstPrimarySampleIndex());
+
+    sampleQueues[primarySampleQueueIndex].discardUpstreamSamples(firstRemovedChunk.getFirstPrimarySampleIndex());
+
+    return firstRemovedChunk.endTimeUs;
+  }
+
+
+  /** Returns whether samples have been read from primary sample queue of the indicated chunk */
+  private boolean haveReadFromMediaChunk(int mediaChunkIndex) {
+    HlsMediaChunk mediaChunk = mediaChunks.get(mediaChunkIndex);
+    return sampleQueues[primarySampleQueueIndex].getReadIndex() > mediaChunk.getFirstPrimarySampleIndex();
+  }
   // Loader.Callback implementation.
 
   @Override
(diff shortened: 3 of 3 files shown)


library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -433,6 +433,13 @@ public boolean onPlaylistError(Uri playlistUrl, long blacklistDurationMs) {
     return chunkIterators;
   }
 
+  public int getPreferredQueueSize(long playbackPositionUs, List<? extends MediaChunk> queue) {
+    if (fatalError != null || trackSelection.length() < 2) {
+      return queue.size();
+    }
+    return trackSelection.evaluateQueueSize(playbackPositionUs, queue);
+  }
+
```
stevemayhew, line 460: I'm looking at actually implementing `ChunkSource` interface.  This looks mostly possible for HLS library except that the `ChunkHolder` is a final class....  Making it an interface and allowing `HlsChunkSource` to implement it looks like the best path.

One of the reasons I'm putting this pull up early is to get feed back on the history and desired architecture direction

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -652,7 +652,15 @@ public boolean continueLoading(long positionUs) {
...
-    // Do nothing.
+    if (loader.isLoading() || isPendingReset()) {
+      return;
+    }
+
+    int currentQueueSize = mediaChunks.size();
+    int preferredQueueSize = chunkSource.getPreferredQueueSize(positionUs, readOnlyMediaChunks);
+    if (currentQueueSize <= preferredQueueSize) {
+      return;
+    }
```
stevemayhew: Just the first checkin, see next changes to the class for the balance of the code.  It seems extending `ChunkSampleStream` is a better path for `HlsSampleStreamWrapper`.  Again, if that is the desired architecture I'm fine with taking that path.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -652,7 +652,42 @@ public boolean continueLoading(long positionUs) {
...
+
+    Log.d(TAG, "discardChunksToIndex() -  chunksToDiscard " + chunksToDiscard + " count " + mediaChunks.size());
+    for (HlsMediaChunk chunk : mediaChunks) {
+      Log.d(TAG, "chunk " + chunk.uid + " start/end: " + chunk.startTimeUs + "/" + chunk.endTimeUs + " format: "+chunk.trackFormat);
+    }
+
+    // TODO - need a way to discard just the samples owned by the discarded chunks
+    sampleQueues[primarySampleQueueIndex].discardToRead();
+    Util.removeRange(mediaChunks, 0, chunksToDiscard  - 1);
+    return lastDiscardEndTimeUs;
```
stevemayhew: This is just the first part of the implementation, because it is not yet possible to extend `ChunkSampleStream` this results in code copy from it.

It seems extending `ChunkSampleStream` is a better path for `HlsSampleStreamWrapper`.  Again, if that is the desired architecture I'm fine with taking that path.

Note the HlsSampleSteamWrapper's `mediaChunks` need to be discarded from index 0 to the index of the first 'good' format sample.  This seems not to be the way of `ChunkSampleStream#discardUpstreamMediaChunksFromIndex()`

Here's a snapshot of the logging output from above from an actual live stream that has up leveled.

````
08-19 16:27:06.749 21933 21967 D HlsSampleStreamWrapper: discardChunksToIndex() -  chunksToDiscard 1 count 4
08-19 16:27:06.750 21933 21967 D HlsSampleStreamWrapper: chunk 5 start/end: 1792000000/1796000000 format: Format(2, null, application/x-mpegURL, null, mp4a.40.2,avc1.64001F, 2100000, null, [960, 540, -1.0], [-1, -1])
08-19 16:27:06.751 21933 21967 D HlsSampleStreamWrapper: chunk 7 start/end: 1796000000/1800000000 format: Format(2, null, application/x-mpegURL, null, mp4a.40.2,avc1.64001F, 2100000, null, [960, 540, -1.0], [-1, -1])
08-19 16:27:06.751 21933 21967 D HlsSampleStreamWrapper: chunk 8 start/end: 1800000000/1804000000 format: Format(2, null, application/x-mpegURL, null, mp4a.40.2,avc1.64001F, 2100000, null, [960, 540, -1.0], [-1, -1])
08-19 16:27:06.752 21933 21967 D HlsSampleStreamWrapper: chunk 10 start/end: 1800000000/1804000000 format: Format(4, null, application/x-mpegURL, null, mp4a.40.2,avc1.640020, 8250000, null, [1280, 720, -1.0], [-1, -1])
````

So the samples and chunks with UID's 5,7,8 need to be discarded.

File: library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java (hunk 1 above)
Comment (tonihei) [reply library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java:460]: HLS does not implement some of the common interfaces like ChunkSource and ChunkSampleStream, for historical reasons and also because HLS has a slightly more complicated media structure that is different from DASH and Smooth Streaming. It would be easier to keep it that way for now even if it means to copy paste code. We can consolidate that later.

File: library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java (hunk 3 above)
Comment (tonihei) [reply library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java]: To make that work, HlsMediaChunk needs to store a list of the first sample index for each track (similar to BaseMediaChunk). But I haven't checked in detail if and how we can do this.

**Comment by stevemayhew** [comment]
This is the feature request for this work: https://github.com/google/ExoPlayer/issues/6322

File: library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java (hunk 3 above)
Comment (stevemayhew) [reply library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java]: This has been done... See the latest checkin

**Comment by stevemayhew** [comment]
I am testing this now.  It is mostly working, but I see some glitches.   The test case I have is a demux ed stream (so there are two `HlsSampleStreamWrapper`, one for the demux single audio track and the other for the video.  The stream is live and as you can see in `PlayerActivity` I've set the retain time very aggressively (2seconds).

The code is ready to review, but I'm still seeing issues in my testing so don't merge it in yet.

File: demos/main/src/main/java/com/google/android/exoplayer2/demo/PlayerActivity.java [hunk 4]
```diff
@@ -352,7 +357,12 @@ private void initializePlayer() {
       TrackSelection.Factory trackSelectionFactory;
       String abrAlgorithm = intent.getStringExtra(ABR_ALGORITHM_EXTRA);
       if (abrAlgorithm == null || ABR_ALGORITHM_DEFAULT.equals(abrAlgorithm)) {
-        trackSelectionFactory = new AdaptiveTrackSelection.Factory();
+        trackSelectionFactory = new AdaptiveTrackSelection.Factory(
+            DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS,
+            DEFAULT_MAX_DURATION_FOR_QUALITY_DECREASE_MS,
+            2000,
+            DEFAULT_BANDWIDTH_FRACTION
```
stevemayhew: 2 seconds pretty much forces this to happen on initial startup of live stream.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Got it.  Yes the code is similar.  Main difference is how HLS manages lists of `SampleStreams`.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -211,6 +213,12 @@ public static HlsMediaChunk createInstance(
   private volatile boolean loadCanceled;
   private boolean loadCompleted;
 
+  /**
+   * Index of first sample written for each SampleQueue created from the extraction of this
+   * media chunk
+   */
+  private int [] firstSampleIndexes = new int[0];
+
```
stevemayhew, line 232: Keep all the SampleStreams as discovered in the MPEG-TS parse.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -280,6 +288,20 @@ public boolean isLoadCompleted() {
     return loadCompleted;
   }
 
+  /**
+   * Return the index of the first sample from the primary sample stream for this media chunk
```
stevemayhew: samples, I'll fix this comment

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -652,9 +653,93 @@ public boolean continueLoading(long positionUs) {
...
+    int [] firstSamples = firstRemovedChunk.getFirstSampleIndexes();
+    for (int i=0; i < sampleQueues.length; i++) {
+      Log.d(TAG, "discardUpstreamSamples() -  stream: " + " from index: " + firstSamples[i]);
+
+      sampleQueues[i].discardUpstreamSamples(firstSamples[i]);
+    }
+
+    dumpCurrentChunkList();
+
+    return firstRemovedChunk;
```
stevemayhew, line 759: This code is near a literal copy now of `ChunkSampleStream`.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -652,9 +653,93 @@ public boolean continueLoading(long positionUs) {
...
+        + " primary sample write: "+sampleQueues[primarySampleQueueIndex].getWriteIndex()
+        + " primary sample read: "+sampleQueues[primarySampleQueueIndex].getReadIndex()
+    );
+    for (int i=0; i<mediaChunks.size(); i++) {
+      HlsMediaChunk chunk = mediaChunks.get(i);
+      Log.d(TAG, "chunk " + chunk.uid + " reading: " + haveReadFromMediaChunk(i)
+          + " start/end: " + chunk.startTimeUs + "/" + chunk.endTimeUs + " primary sample index: "
+          + chunk.getFirstSampleIndexes()[primarySampleQueueIndex] + " format: "+chunk.trackFormat);
+    }
+  }
```
stevemayhew, line 774: Just for my debugging.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -652,9 +653,93 @@ public boolean continueLoading(long positionUs) {
...
+  private boolean haveReadFromMediaChunk(int mediaChunkIndex) {
+    HlsMediaChunk mediaChunk = mediaChunks.get(mediaChunkIndex);
+    boolean haveRead = false;
+    int [] firstSamples = mediaChunk.getFirstSampleIndexes();
+    for (int i=0; i < firstSamples.length; i++) {
+      haveRead = haveRead || sampleQueues[i].getReadIndex() > firstSamples[i];
+    }
+    return haveRead;
+  }
   // Loader.Callback implementation.
```
stevemayhew, line 787: Again nearly identical to `ChunkSampleStream`

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -849,6 +937,8 @@ public TrackOutput track(int id, int type) {
     sampleQueueTrackIds[trackCount] = id;
     sampleQueues = Arrays.copyOf(sampleQueues, trackCount + 1);
     sampleQueues[trackCount] = trackOutput;
+    HlsMediaChunk mediaChunk = findChunkMatching(chunkUid);
+    mediaChunk.setFirstSampleIndex(trackCount, 0);
     sampleQueueReaders = Arrays.copyOf(sampleQueueReaders, trackCount + 1);
```
stevemayhew: When each elementary stream is discovered and the SampleQueue is created, this is needed for the first chunks loaded right after the initial Extractor is created.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -898,8 +988,36 @@ public void setSampleOffsetUs(long sampleOffsetUs) {
...
+    int [] writePositions;
+
+    writePositions = new int[sampleQueues.length];
+    for (int i=0; i < sampleQueues.length; i++) {
+      writePositions[i] = sampleQueues[i].getWriteIndex();
+    }
+
+    return writePositions;
+  }
+
```
stevemayhew: This method and it's call are no longer needed.  Will remove

library/ui/src/main/java/com/google/android/exoplayer2/ui/DebugTextViewHelper.java
```diff
@@ -18,25 +18,46 @@
 import android.annotation.SuppressLint;
```
stevemayhew: I can leave this file out of the review.  Changes are to expose the upstream as well as the downstream level and the current playback position in US.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: removed

stevemayhew:
I'm currently testing by forcing poor choices on the initial bitrate for a live stream:

```java

      DefaultBandwidthMeter bandwidthMeter = new DefaultBandwidthMeter.Builder(getApplicationContext())
          .setSlidingWindowMaxWeight(DefaultBandwidthMeter.DEFAULT_SLIDING_WINDOW_MAX_WEIGHT).build();
      bandwidthMeter.setNetworkTypeOverride(C.NETWORK_TYPE_3G);
```

I'm also adjusting the sliding weight to cause it to adapt more slowly to the new bandwidth discovered.

I can provide URLs of live broadcasts here that I'm testing with if you give me an email I can send them to you.

stevemayhew:
Sorry, I can get back to you tomorrow.  If the tracks have same GOP rate
(required for HLS) this should work fine. Perhaps we need to limit it based
on the independent playlist supported flag?

tonihei:
Sorry for the delay, but just checking - is this ready for review now or still being worked on?

stevemayhew:
We are testing this code in our own branches.  I’ll rebase it to keep up to
date with dev-v2.  This does make channel changes faster for us

stevemayhew:
Rebased this to the latest `dev-v2` and re-tested, drop me a note if you want a private live URL you can test this against.

Simply forcing a low initial bandwidth estimate and discarding aggressively on level shift takes the time to full bit rate on initial channel change down by 2 seconds (the size of the discard).  I can send a patch, but basically change `PlayerActivity` track selection and bandwidth meter as:

```java
      if (abrAlgorithm == null || ABR_ALGORITHM_DEFAULT.equals(abrAlgorithm)) {
        trackSelectionFactory = new AdaptiveTrackSelection.Factory(
            DEFAULT_MIN_DURATION_FOR_QUALITY_INCREASE_MS,
            DEFAULT_MAX_DURATION_FOR_QUALITY_DECREASE_MS,
            2000,
            DEFAULT_BANDWIDTH_FRACTION
        );
      }
```
...
```java
      player =
          new SimpleExoPlayer.Builder(/* context= */ this, renderersFactory)
              .setTrackSelector(trackSelector)
              .setBandwidthMeter(new DefaultBandwidthMeter.Builder(this)
                  .setInitialBitrateEstimate(500)
                  .build())
              .build();
      player.addListener(new PlayerEventListener());
```
.

biespb (approved):

biespb:
[https://en.av01.tv/](url)

biespb (approved):

biespb (commented):
[snip]

stevemayhew:
I'm not rushing to get this merged in...  We are finding architectural issues (`ConcurrentModificationException`) with this change that I am still looking at the best way to fix.  I would love some  guidance from the ExoPlayer architecture team (@tonihei @ojw28 )

The issue is both the `HlsMediaChunk` list and the `SampleQueue` array are modified by different threads.  This is true before this change as well.

For example, the player thread modifies the list here when reading samples:
https://github.com/google/ExoPlayer/blob/826083db923f26f2000ca42f7b54b9531726d713/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java#L543

And the loading thread of course modifies it here (one of many examples):  https://github.com/google/ExoPlayer/blob/826083db923f26f2000ca42f7b54b9531726d713/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java#L774

This comment:  https://github.com/google/ExoPlayer/blob/826083db923f26f2000ca42f7b54b9531726d713/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java#L818

Is no longer true based on the code (the `Loadable` calls the `init()` method from the `Loader`'s run loop here:  https://github.com/google/ExoPlayer/blob/826083db923f26f2000ca42f7b54b9531726d713/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java#L309 and it is iterating an array that seems more 'owned' by the `HlsSampleStreamWrapper` itself.

The synchronizing by avoidance (make sure the loading threads are not running before calling any method that accesses structures in the wrapper) seems to be breaking down, and I'm not sure which method to fix it best matches the direction the overall architecture for Chunk/Sample loading interfacing with the player thread is headed.

The options I see are:
1. Wrap the lists and arrays that are modified/accessed by both threads with `Collections.synchronizedList()` - big hammer, not obvious what is going on
2. Extract methods for the sections that modify and synchronize them with the player thread (using the `HlsSampleStreamWrapper`
3. Lift an interface for methods in the `HlsSampleStreamWrapper` called by the loader thread (https://github.com/google/ExoPlayer/blob/826083db923f26f2000ca42f7b54b9531726d713/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java#L217) and synchronize them and any other methods that access fields shared by these methods with `HlsSampleStreamWrapper` object (the player thread effectively)

AquilesCanta:
> The issue is both the HlsMediaChunk list and the SampleQueue array are modified by different threads. This is true before this change as well.

I think both modifications quoted above happen on the player thread.

If you think there are threading concerns we should address in the existing code, please file a new issue including all the information required in the issue template. We should ensure the existing code is correct in order to make the review of this pull request constrained to the new feature, and not also fixing a problem in the original code. 

Once we agree on the existing code being thread-safe, we can edit the comment above to limit it to the changes we'd like to introduce.

EDIT: Note that there is indeed an issue (already sent a fix for internal review) with the thread access comments around `HlsSampleStreamWrapper.init()`. I don't think they are related to the thread issues you mentioned above, but still worth mentioning.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -222,6 +224,12 @@ public static HlsMediaChunk createInstance(
   private volatile boolean loadCanceled;
   private boolean loadCompleted;
 
+  /**
+   * Index of first sample written for each SampleQueue created from the extraction of this
+   * media chunk
+   */
+  private int [] firstSampleIndexes = new int[0];
```
AquilesCanta, line 231: Super-nit: Here and elsewhere, use don't introduce a space between int and [], so this should be:

`private int[] firstSampleIndexes = new int[0];`

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -291,6 +299,27 @@ public boolean isLoadCompleted() {
...
+
+  /**
+   * Set the index of the first sample written to a sample queue from this media chunk,
+   * indexed by the caller's stream index for the sample queue
+   *
+   * @param streamIndex - caller's index for the {@link SampleQueue}
+   * @param firstSampleIndex - index value to store
+   */
+  void setFirstSampleIndex(int streamIndex, int firstSampleIndex) {
+    firstSampleIndexes = Arrays.copyOf(firstSampleIndexes, streamIndex + 1);
```
AquilesCanta, line 319: What do you think should happen if a segment does not include samples for a given track? A valid option would be using C.INDEX_UNSET, or zero, in which case we should adjust the javadocs.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +697,93 @@ public boolean isLoading() {
...
   }
 
+  /**
+   * Discard upstream media chunks from {@code chunkIndex} and corresponding samples from sample
+   * queues.
+   *
+   * @param chunkIndex The index of the first chunk to discard.
+   * @return The chunk at given index.
+   */
+  private HlsMediaChunk discardUpstreamMediaChunksFromIndex(int chunkIndex) {
```
AquilesCanta, line 745: We should add a comment in the javadoc saying "Must only be called when there's no active loading thread, or similar, which I think is already fulfilled by reevaluate buffer.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +697,93 @@ public boolean isLoading() {
...
+    HlsMediaChunk firstRemovedChunk = mediaChunks.get(chunkIndex);
+    Util.removeRange(mediaChunks, /* fromIndex= */ chunkIndex, /* toIndex= */ mediaChunks.size());
+
+
+    int [] firstSamples = firstRemovedChunk.getFirstSampleIndexes();
+    for (int i=0; i < sampleQueues.length; i++) {
```
AquilesCanta, line 751: Super-nit: This should be 

`int i = 0; ...`

Ditto elsewhere.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +697,93 @@ public boolean isLoading() {
...
+    for (int i=0; i<mediaChunks.size(); i++) {
+      HlsMediaChunk chunk = mediaChunks.get(i);
+      Log.d(TAG, "chunk " + chunk.uid + " reading: " + haveReadFromMediaChunk(i)
+          + " start/end: " + chunk.startTimeUs + "/" + chunk.endTimeUs + " primary sample index: "
+          + chunk.getFirstSampleIndexes()[primarySampleQueueIndex] + " format: "+chunk.trackFormat);
+    }
+  }
+
+
+  /** Returns whether samples have been read from primary sample queue of the indicated chunk */
```
AquilesCanta, line 777: Could you clarify why the primary sample queue is enough, as opposed to any sample queue?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
stevemayhew: thanks.. Will do.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
stevemayhew: > EDIT: Note that there is indeed an issue (already sent a fix for internal review) with the thread access comments around `HlsSampleStreamWrapper.init()`. I don't think they are related to the thread issues you mentioned above, but still worth mentioning.

What's the issue number?  Because the method is definitely called from the loader thread, this happens concurrently with sample reading on the player thread so concurrent modifications can happen.

Note, of course you know even iterating an collection in one thread that is being modified in another thread will result in an ConcurrentModificationException.   I'll help address this in the issue discussion if I can.  Just let me know how best I can help.

AquilesCanta: The exception you mentioned shouldn't be related to the init issue. Please file a fresh issue for that, it will be easier for us to debug it.

ojw28:
@stevemayhew @AquilesCanta - Just wondering what the current status of this PR is? We have another external contributor who has some changes that build on top of this, so it would be good to figure out what we need to do here and/or what the open issues are. Thanks!

stevemayhew:
Oliver (@ojw28) I need to rebase and test again.  With short segments
(I-Frame for example) I saw `ConcurrentModificationException` on the list
of `MediaChunk`.

You saw my earlier notes as to why this is possible.  My feeling is the
loader threads should never touch this list, the changes to `init()`
recently might have fixed it so they don’t, I’ll have to look at it again

stevemayhew:
Closing this pull request, it is a duplicate of  pull #7244 which is further along.

PR #6303, opened 2019-08-16
Upgrade librtmp-client to 3.1.0

vavadhani:
The upgraded version should fix #4200, #4249, #4319, #4337

PR #6329, opened 2019-08-21
Update com.squareup.okhttp3 to 4.1.0 and update deprecated

fgl27:
There is others out of date dependencies..

**Dependencies that need update:**

 - androidx.test:core [1.1.0 -> 1.2.0]
 - androidx.test:runner [1.1.0 -> 1.2.0]
 - androidx.test.ext:junit [1.1.0 -> 1.1.1]
 - androidx.work:work-runtime [2.1.0 -> 2.2.0]
 - com.android.tools.lint:lint-gradle [26.4.2 -> 26.5.0]
 - com.firebase:firebase-jobdispatcher [0.8.5 -> 0.8.6]
 - com.google.android.gms:play-services-ads-identifier [16.0.0 -> 17.0.0]
 - com.google.auto.value:auto-value [1.6 -> 1.6.6]
 - com.google.auto.value:auto-value-annotations [1.6 -> 1.6.6]
 - com.google.truth:truth [0.44 -> 1.0]
 - com.linkedin.dexmaker:dexmaker [2.21.0 -> 2.25.0]
 - com.linkedin.dexmaker:dexmaker-mockito [2.21.0 -> 2.25.0]
 - net.butterflytv.utils:rtmp-client [3.0.1 -> 3.1.0]
 - org.checkerframework:checker-compat-qual [2.5.0 -> 2.5.5]
 - org.checkerframework:checker-qual [2.5.0 -> 2.10.0]
 - org.chromium.net:cronet-embedded [75.3770.101 -> 76.3809.111]
 - org.mockito:mockito-core [2.25.0 -> 3.0.0]

**net.butterflytv.utils:rtmp-client** has a open pr https://github.com/google/ExoPlayer/pull/6303

### DIFF
--- a/extensions/okhttp/build.gradle
+++ b/extensions/okhttp/build.gradle
@@ -35,7 +35,7 @@ dependencies {
     implementation project(modulePrefix + 'library-core')
     implementation 'androidx.annotation:annotation:1.1.0'
     compileOnly 'org.checkerframework:checker-qual:' + checkerframeworkVersion
-    api 'com.squareup.okhttp3:okhttp:3.12.1'
+    api 'com.squareup.okhttp3:okhttp:4.1.0'
 }
 
 ext {
--- a/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
+++ b/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
@@ -351,10 +351,10 @@ public class OkHttpDataSource extends BaseDataSource implements HttpDataSource {
     }
     RequestBody requestBody = null;
     if (dataSpec.httpBody != null) {
-      requestBody = RequestBody.create(null, dataSpec.httpBody);
+      requestBody = RequestBody.create(dataSpec.httpBody, null);
     } else if (dataSpec.httpMethod == DataSpec.HTTP_METHOD_POST) {
       // OkHttp requires a non-null body for POST requests.
-      requestBody = RequestBody.create(null, Util.EMPTY_BYTE_ARRAY);
+      requestBody = RequestBody.create(Util.EMPTY_BYTE_ARRAY, null);
     }
     builder.method(dataSpec.getHttpMethodString(), requestBody);
     return builder.build();


fgl27:
@googlebot I signed it!

tonihei:
Thanks for the update.

Note that we usually make these dependency updates anyway whenever we do a major release.

noamtamim:
Specifically for OkHttp, version 3.13 and up [require Android 5+](https://github.com/square/okhttp/issues/4481). So assuming ExoPlayer still supports Android 4.1, **this update cannot happen**.

fgl27:
OK I do understand all informed here.

But I think a discussion about this is needed to be opened.

The solution of simple not update the dependencies because will cause issue on old devices that account for less than 10% of all devices is not ideal.

I know you guys know what you are doing but a discussion on a better solution must be done we can just give up on updates.

So from the top of my mind I have a few ideas...

Have a main branch and a old supported but limited branch:

- Main branch: update all the dependencies drop unsupported SDK clean up old code that checks for those give a minor but a general code improve (by removing SDK checks and SDK specific resources).
- Old devices branch: keep as it is and only make **critical security fixes and related things** to keep it working and add workaround for old devices.

One can also use the "Multiple APK" and targeted old device only on a old version of Exoplayer, this way is possible to drop old SDK and let people know they must use that "Multiple APK" option and targeted old ExoPlayer version for those.

This are just my ideais for something that can be used to start a bigger discussion, I assume better ideas may come and a better solution will be found.

ojw28:
> The solution of simple not update the dependencies because will cause issue on old devices that account for less than 10% of all devices is not ideal.

Generally, the lower in the stack a library is the more conservative the library maintainer will be about updating its minimum version requirement, and as a result this kind of problem is normally avoided. OkHttp has been quite aggressive about updating its minimum version requirement in this case.

For this specific case you can just fork the ExoPlayer OkHttp extension into your own app for a while. It's only two classes, one of which is trivial. I don't think there's sufficient need for us to be maintaining multiple branches.

Alternatively, if the `RequestBody` change was non-essential, you can force the OkHttp version in your app's gradle file to use the new version with the existing extension. This approach is also a general solution if the dependency has maintained API compatibility between versions.

tonihei:
I'll close the pull request then. As explained in detail above, we can't update the OkHttp dependency right now [because of our minimum API level](https://github.com/google/ExoPlayer/pull/6329#issuecomment-525419534) and there are [workarounds if you really want to upgrade](https://github.com/google/ExoPlayer/pull/6329#issuecomment-526522954) in your app.

PR #6347, opened 2019-08-27
Add DownloadHelper.estimateTotalSize() utility

noamtamim:
Helps applications to estimate the total download size bitrate of selected tracks and duration.

An interface `DurationProvider` was added, implemented by all manifest objects, that allows the code in `core` to access the duration fields of the different asset types (DASH, HLS, SmoothStreaming).


library/core/src/main/java/com/google/android/exoplayer2/util/DurationProvider.java
```diff
@@ -0,0 +1,12 @@
+package com.google.android.exoplayer2.util;
+
+import com.google.android.exoplayer2.C;
+
+public interface DurationProvider {
```
tonihei: There should be no need for this interface or any of the manifest changes as you can access `DownloadHelper.MediaPreparer.timelime`. This timeline already contains the duration of all `Timeline.Period`s.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
```diff
@@ -723,6 +727,61 @@ public DownloadRequest getDownloadRequest(String id, @Nullable byte[] data) {
     return new DownloadRequest(id, downloadType, uri, streamKeys, cacheKey, data);
   }
 
+  /**
+   * Try to estimate the total size of chunked downloads. Returns {@link C#LENGTH_UNSET} for
```
tonihei: There is probably no reason to special-case progressive streams because you already allow `C.LENGTH_UNSET` as return value if "not known". Progressive streams can just go through the same code paths allowing for the possibility that a bitrate is set in the Format so that we can actually calculate the total size.

tonihei:
The idea to provide a way to retrieve the total size for downloading before actually starting the download makes sense, but I'm a bit wary that calculating it here using the available information may be wrong or misleading.

This is because the selected tracks don't necessarily correspond to downloadable data streams. This mapping happens when we call `MediaPeriod.getStreamKeys` to identify the streams that we need to load. These streams and the manifest and potential extra data loads to obtain initilization data (e.g. for the DASH segment index) can then be used to calculate a total download size. This functionality is already part of the `Downloader` implementations (specifically `SegmentDownloader`) and it also takes into account that some of the media chunks have already been downloaded. A down-side of the existing implementation is that it doesn't do any estimation if the actual stream length is unknown (for example for text) and it requires extra data loads.

That means we should probably still provide a way to get an estimated download size in `DownloadHelper` even if it may not be 100% correct. To keep it simple, it may be better to keep out hard-coded estimates, for example for text tracks as they are likely to be only a small part of the downloaded data anyway.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
```diff
@@ -723,6 +727,61 @@ public DownloadRequest getDownloadRequest(String id, @Nullable byte[] data) {
...
+
+    long selectedSize = 0;
+
+    for (int periodIndex = 0; periodIndex < this.getPeriodCount(); periodIndex++) {
+
+      final MappingTrackSelector.MappedTrackInfo mappedTrackInfo = mappedTrackInfos[periodIndex];
+      final int rendererCount = mappedTrackInfo.getRendererCount();
+
+      for (int rendererIndex = 0; rendererIndex < rendererCount; rendererIndex++) {
+        final List<TrackSelection> trackSelections = this.getTrackSelections(periodIndex, rendererIndex);
```
tonihei: It's probably easier to use `trackSelectionsByPeriodAndRenderer` directly.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
```diff
@@ -723,6 +727,61 @@ public DownloadRequest getDownloadRequest(String id, @Nullable byte[] data) {
...
+        for (TrackSelection selection : trackSelections) {
+          final Format format = selection.getSelectedFormat();
+
+          int bitrate = format.bitrate;
+          if (bitrate == Format.NO_VALUE) {
```
tonihei: As noted in the generic comment, please leave out unspecified bit rates as it likely only applies to the following cases:
(a) tracks that have very little data compared to audio and video (e.g. text and metadata tracks)
(b) tracks that are actually muxed into other tracks and are already accounted for (e.g. muxed HLS audio tracks). 

Or to put it another way: If at least one of the `Format`s specifies a bitrate, it's very likely that all relevant non-negligible bitrates are specified in other `Format`s as well. After all, it's only an estimate :) 

If you want to add a fixed offset for the text size, you could also do that in your app afterwards using the app-specific knowledge and assumptions you have about the size of the text data. If you need to know the duration of the stream, you can probably add a new method `DownloadHelper.getTimeline()` similiar to the existing `getManifest` method to allow access to duration information.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
noamtamim: The reason for this special case is that progressive streams don't call `prepare()`, and as a result, `assertPreparedWithMedia()` fails. I figured that in this case `LENGTH_UNSET` is better than a crash. But that depends on ExoPlayer's policy regarding this type of violation.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
noamtamim: I agree, but since `DownloadHelper` does not provide the application with a duration, it can't even do the estimation -- even if it does know that the audio bitrate in HLS is 128k, for example.

Maybe an alternative is to add a `getDuration()` method so the app can do the estimation, but even then - the app has to know (or "remember") how many audio tracks were selected for download. Remember this is mainly for HLS with alternate audio streams, where bitrate is not specified (it's not even in the spec).

The trouble with text tracks, BTW, is that they are small (especially if they they are gzip-compressed), and the http request overhead (without http/2) is the most significant factor in their download speed. This is particularly important when the text tracks themselves are chunked. This estimation does not "fix" this problem, but it might allow some mitigation because the app developer can "play" with the given value until it makes sense.

noamtamim: @tonihei your thoughts on adding `DownloadHelper.getDurationUs()` (or `Ms`), please. Note that it will have to iterate through all periods, so maybe it should be cached in a field.

ojw28: An application can probably get the duration fairly easily by calling `getManifest()` and casting to the type of media that it's working with.

noamtamim: I suppose it can, but having to cast is not ideal. And still, it would have to count audio tracks that have no bitrate across all periods. Like I wrote, this is an HLS issue.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
tonihei: Thanks for explaining. Can you use "mediaSource == null" as an alternative check? That's what we use in other places.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
tonihei: > your thoughts on adding DownloadHelper.getDurationUs() (or Ms), please.

Could you just deprecate getManifest and add getTimeline instead? The Timeline contains both the manifest and the duration information. 

> The trouble with text tracks, BTW, is that they are small (especially if they they are gzip-compressed), and the http request overhead (without http/2) is the most significant factor in their download speed. 

That's indeed a problem when it comes to download speed, but it doesn't influence the download size as it is calculated here.

> how many audio tracks were selected for download

I see the issue now. Originally I was assuming that you just download the muxed audio stream and that its bitrate is already part of the muxed stream bitrate. But adding alternative audio streams indeed complicates the matter. 

Could you add a small inner class `interface FormatBitrateEstimator{ int estimateFormatBitrate(Format format); }` that is passed into this method? And then also add a version of the method without parameters which uses the default implementation of `format -> format.bitrate == NO_VALUE ? 0 : format.bitrate`. This ensures that the basic method returns a sensible estimate and for more advanced use cases where you care about the formats without bitrates (as you do), you can use the extended method. Having the interface keeps it flexible to also include metadata tracks etc. Would that work for you?

noamtamim: Implemented all suggestions.

noamtamim:
Implemented all suggestions. Please review again.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
```diff
@@ -1099,4 +1168,21 @@ public void removeEventListener(EventListener eventListener) {
...
     }
   }
+
+  /**
+   * See {@link #estimateTotalSize(FormatBitrateEstimator)} for interface usage.
+   */
+  public interface FormatBitrateEstimator {
+
+    /**
+     * Return the estimated format bitrate, or {@link Format#NO_VALUE} if not known.
```
tonihei, line 1178: I think this interface is easier if you remove the special return value of Format.NO_VALUE.  The default estimator can return 0 instead and the estimateTotalSize doesn't need to check for this condition either.

library/core/src/main/java/com/google/android/exoplayer2/util/DurationProvider.java
tonihei: Can you remove this class? It's not needed anymore I think.

tonihei (commented):
Looks good now. Thanks! Only two smaller things left.

library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadHelper.java
noamtamim: That was my original implementation, but then I thought -- what if someone implementing the interface will just return format.bitrate (which can be NO_VALUE), without handling NO_VALUE as a special case?
For example, this looks like a valid implementation, if you don't pay attention, and it will likely pass code review:

```java
if (format.roleFlags & ROLE_FLAG_COMMENTARY == ROLE_FLAG_COMMENTARY) {
  return 64000;  // we know that commentary tracks have low bitrates
}
// else
return format.bitrate;
```

You'd get NO_VALUE (== -1), multiply it by the duration and accumulate, to get a really low size value. I agree that it's the app's "fault" that it would get invalid values, but IMHO in this case it's better to protect programmers from tricky mistakes.

tonihei: Yes, thanks for the explanation. So it's maybe better to allow NO_VALUE to avoid accidental bugs when this interface is implemented.

noamtamim:
@tonihei will this PR be merged to 2.11.x?

tonihei:
There is still some ongoing discussion about whether we should move some of the logic to the `MediaPeriod` classes in case they can provide better estimates than just using the bitrate (or whether we should just leave it as it is). I'll let you know once we merged it.

tonihei:
@noamtamim  We just discussed this change again and were still unsure about the usefulness of the `FormatBitrateEstimator`. It seems it's only needed for the specific case of adding extra audio or text tracks of demuxed HLS streams. Could you address the following questions so that we get some idea how to proceeed?

- Would the method still be useful for you without the `FormatBitrateEstimator`? It would still work for the main selected stream, but the estimate wouldn't increase even if you add more audio tracks. 
- More generally - if you want to display the download size, wouldn't it make more sense to provide this information from your server directly if available? 
- If you add extra tracks, there must be some call to `DownloadHelper` that adds tracks based on a language (e.g. `addAudioLanguagesToSelection`). That should give you a fairly good idea of the number of extra tracks without iterating over the selected tracks in the end. So without `FormatBitrateEstimator`, you could probably do something like: `estimate = downloadHelper.getBitrateEstimate() + (numberOfAudioLanguages-1) * myCustomAudioBitrate`. Would that work for you?

noamtamim:
I agree - `FormatBitrateEstimator` is only for demuxed audio tracks in HLS, which is one in my original implementation the input was just an `int`.
1. It will be misleading, unless clearly documented. But see 3.
2. From past experience, the server does not always know the size (or, it does not always have the API to return it). But even if it could provide the size to the app, the server definitely doesn't know which tracks were selected for download by the app.
3. That's a valid point, but (a) can you be sure that demuxed audio tracks will never be auto-selected by ExoPlayer, even in the case where the main stream is itself demuxed (video-only, like in DASH) and/or that track has DEFAULT or AUTOSELECT flags? (b) it will mean different logic between DASH and HLS. Except for the selected factory method used for obtaining the helper, everything else is the same.

ojw28:
I'm the main complainant here (sorry). To expand on my objections a little:

* The problem with `FormatBitrateEstimator`, as I see it, is that it actively doesn't make sense for any use case other than demuxed HLS. It doesn't make sense for progressive, or DASH, or SmoothStreaming, or really muxed HLS. Should we be introducing an API onto a general purpose class if it only makes sense for one very specific use case?
* In the problematic case, I think we have other related problems that also arise from demuxed HLS tracks not having bitrate estimates. For example, when we start doing simultaneous audio/video adaptation, it's not going to work unless the bitrates are set properly. If we injected a bitrate estimator directly into the `HlsMediaSource` then we could fix both problems at once by correcting the bitrates at source, whilst also putting the HLS specific API in an HLS specific place.
* Underneath the API, the implementation doesn't make sense for progressive or muxed HLS. It feels like it would be better to push some of the implementation inside the `MediaPeriod` implementations. For example by having `MediaPeriod.getEstimatedSize` return an estimate for the currently selected tracks. This would allow for correct implementations in these cases.

Regarding the "right" way to do this: For the streaming services I've worked on, the server would always provide the exact size in bytes of each stream. Which seems like a much better way of doing things. It's easy to compute this once after transcoding and serve it through an API, and provides an exact answer. It does of course require you to control the serving side, so this API is arguably still useful for cases where you don't ;).

noamtamim:
Hi @ojw28 and sorry for the late response.
All of your points are valid, and there are a few options of moving forward with it.
- Remove the `FormatBitrateEstimator` and declare that this utility doesn't take into account demuxed audio tracks in HLS (or doesn't support HLS at all, to prevent mistakes)
- Implement it in one of the 2 ways you mention, which is obviously a much larger change than intended.
- Provide this function as reference code that can be customized by the app (it should be possible to implement using just the public getters of `DownloadHelper`).

For the "problematic" case of demuxed audio tracks in HLS, I want to add this: HLS in its current form does not allow for a `BANDWIDTH` attribute on alternate audio tracks (unless it has changed in the past 2-3 months). Effectively it means that the player is not supposed to "do ABR" on those tracks - so I'm not sure if your second bullet is relevant.

Note that this is **not** a rare case: it's very common when a stream contains alternative languages. See Apple's guide on that matter:
https://developer.apple.com/documentation/http_live_streaming/example_playlists_for_http_live_streaming/adding_alternate_media_to_a_playlist

jakoss:
Are there any plans moving this forward? We could really use this feature

ojw28:
I think the best way to fix the issue involves a larger change, as alluded to above. I don't particularly want to merge this change in the meantime. It has some issues, also alluded to above. It's also true that:

1. If you do want to use the logic in this pull request, it's easy to replicate it in your app code as a utility function. There's no particular requirement that it be part of the library for you to use it. For this reason, it's not particularly important that we merge this change as an interim solution.
1. Streaming services can (and would do better to) provide their client apps with the exact size in bytes of each stream. It's not that hard (I know of multiple streaming services that do this), and is definitely a better solution. For this reason, working on the larger change is also not a priority at this moment in time.

> For the "problematic" case of demuxed audio tracks in HLS, I want to add this: HLS in its current form does not allow for a BANDWIDTH attribute on alternate audio tracks (unless it has changed in the past 2-3 months). Effectively it means that the player is not supposed to "do ABR" on those tracks - so I'm not sure if your second bullet is relevant.

As a bit of an aside, this is not quite true. It's true that the BANDWIDTH attribute is not specified on alternate audio tracks. However it's also true that the BANDWIDTH attribute on EXT-X-STREAM-INF includes the bandwidth of any alternate audio renditions that it references, and that the EXT-X-STREAM-INF tags define what kind of adaptations are allowed (including for alternate audio tracks). For example suppose the individual bitrates are:

low audio: 10
high audio: 20
low video: 1000
high video: 2000

This HLS master playlist allows for adaptation between four EXT-X-STREAM-INF combinations, which is effectively equivalent to allowing independent adaptation of audio and video:

```
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="low-audio",NAME="",...
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="high-audio",NAME=""...

#EXT-X-STREAM-INF:BANDWIDTH=1010,AUDIO="low-audio"
low-video.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2010,AUDIO="low-audio"
high-video.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=1020,AUDIO="high-audio"
low-video.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2020,AUDIO="high-audio"
high-video.m3u8
```

This HLS master playlist disallows use of low quality audio with high quality video, and vice-versa:

```
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="low-audio",NAME="",...
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="high-audio",NAME=""...

#EXT-X-STREAM-INF:BANDWIDTH=1010,AUDIO="low-audio"
low-video.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2020,AUDIO="high-audio"
high-video.m3u8
```

ojw28:
Edit: Corrected some typos in the response above!

yoobi:
Hello, Do you have a plan to merge this into master?

Koster35:
Hi, I am using the ExoPlayer workmanager extension in my app.

I would like to implement logic to estimate the size of a download using some of the code here for guidance. Is there anyway to access a MediaPreparer's Timeline object prior to this pull-request being merged? 

It does not appear to be possible right now, but I want to make sure.

noamtamim:
This is the code we use from outside. It's similar, but takes duration from the manifest by casting to a specific type. There's probably other ways to do it.
You just need the `DownloadHelper` instance, everything else is taken from there.
DISCLAIMER: it's been a while since I tested this code standalone, use at your own risk.


```java
    private static long estimateTotalSize(DownloadHelper helper, int hlsAudioBitrate) {
        long selectedSize = 0;

        final Object manifest = helper.getManifest();

        final long durationMs;
        if (manifest instanceof DashManifest) {
            final DashManifest dashManifest = (DashManifest) manifest;
            durationMs = dashManifest.durationMs;
        } else if (manifest instanceof HlsManifest) {
            final HlsManifest hlsManifest = (HlsManifest) manifest;
            durationMs = hlsManifest.mediaPlaylist.durationUs / 1000;
        } else if (manifest instanceof SsManifest) {
            durationMs = ((SsManifest) manifest).durationUs / 1000;
        } else {
            durationMs = 0; // FIXME
        }

        for (int pi = 0; pi < helper.getPeriodCount(); pi++) {

            final MappingTrackSelector.MappedTrackInfo mappedTrackInfo = helper.getMappedTrackInfo(pi);
            final int rendererCount = mappedTrackInfo.getRendererCount();

            for (int i = 0; i < rendererCount; i++) {
                final List<TrackSelection> trackSelections = helper.getTrackSelections(pi, i);
                for (TrackSelection selection : trackSelections) {
                    final Format format = selection.getSelectedFormat();

                    int bitrate = format.bitrate;
                    if (bitrate <= 0) {
                        if (format.sampleMimeType != null && format.sampleMimeType.startsWith("audio/")) {
                            bitrate = hlsAudioBitrate;
                        }
                    }
                    selectedSize += (bitrate * durationMs) / 1000 / 8;
                }
            }
        }
        return selectedSize;
    }
```

noamtamim:
@Koster35 please see above.

Koster35:
@noamtamim Thank you, I will try this!

Neustradamus:
Any news on it?

ojw28:
@Neustradamus - You can/should implement this in your own codebase. The code snippet @noamtamim provided above should help. We don't plan to merge this pull request for the reasons mentioned further up this thread.

PR #6379, opened 2019-09-03
return 2 SeekPoints from FlacDecoderJni

wingyippp:
get 2 SeekPoints from the getSeekPosition() in flac_parser.cc

### DIFF
--- a/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
+++ b/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
@@ -223,7 +223,7 @@ import java.nio.ByteBuffer;
    * @return The corresponding position (byte offset) in the flac stream or -1 if the stream doesn't
    * have a seek table.
    */
-  public long getSeekPosition(long timeUs) {
+  public long[] getSeekPosition(long timeUs) {
     return flacGetSeekPosition(nativeDecoderContext, timeUs);
   }
 
@@ -283,7 +283,7 @@ import java.nio.ByteBuffer;
 
   private native long flacGetNextFrameFirstSampleIndex(long context);
 
-  private native long flacGetSeekPosition(long context, long timeUs);
+  private native long[] flacGetSeekPosition(long context, long timeUs);
 
   private native String flacGetStateString(long context);
 
--- a/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
+++ b/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
@@ -19,6 +19,7 @@ import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.ParserException;
 import com.google.android.exoplayer2.extractor.ExtractorInput;
+import com.google.android.exoplayer2.extractor.SeekMap;
 import com.google.android.exoplayer2.util.FlacStreamMetadata;
 import com.google.android.exoplayer2.util.Util;
 import java.io.IOException;
@@ -220,11 +221,12 @@ import java.nio.ByteBuffer;
    * stream.
    *
    * @param timeUs A seek position in microseconds.
-   * @return The corresponding position (byte offset) in the flac stream or -1 if the stream doesn't
+   * @return The corresponding SeekPoints in the flac seek table or
+   * {@link com.google.android.exoplayer2.extractor.SeekPoint#position} set -1 if the stream doesn't
    * have a seek table.
    */
-  public long[] getSeekPosition(long timeUs) {
-    return flacGetSeekPosition(nativeDecoderContext, timeUs);
+  public SeekMap.SeekPoints getSeekPoints(long timeUs) {
+    return flacGetSeekPoints(nativeDecoderContext, timeUs);
   }
 
   public String getStateString() {
@@ -283,7 +285,7 @@ import java.nio.ByteBuffer;
 
   private native long flacGetNextFrameFirstSampleIndex(long context);
 
-  private native long[] flacGetSeekPosition(long context, long timeUs);
+  private native SeekMap.SeekPoints flacGetSeekPoints(long context, long timeUs);
 
   private native String flacGetStateString(long context);
 
--- a/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
+++ b/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
@@ -221,10 +221,10 @@ import java.nio.ByteBuffer;
    * stream.
    *
    * @param timeUs A seek position in microseconds.
-   * @return The corresponding SeekPoints in the flac seek table or
-   * {@link com.google.android.exoplayer2.extractor.SeekPoint#position} set -1 if the stream doesn't
+   * @return The corresponding SeekPoints in the flac seek table or null if the stream doesn't
    * have a seek table.
    */
+  @Nullable
   public SeekMap.SeekPoints getSeekPoints(long timeUs) {
     return flacGetSeekPoints(nativeDecoderContext, timeUs);
   }
--- a/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
+++ b/extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
@@ -29,7 +29,6 @@ import com.google.android.exoplayer2.extractor.ExtractorsFactory;
 import com.google.android.exoplayer2.extractor.Id3Peeker;
 import com.google.android.exoplayer2.extractor.PositionHolder;
 import com.google.android.exoplayer2.extractor.SeekMap;
-import com.google.android.exoplayer2.extractor.SeekPoint;
 import com.google.android.exoplayer2.extractor.TrackOutput;
 import com.google.android.exoplayer2.metadata.Metadata;
 import com.google.android.exoplayer2.metadata.id3.Id3Decoder;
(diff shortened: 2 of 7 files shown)


ojw28:
Please could you provide some additional context on why you are doing this? How far apart are seek points in practice (i.e. does this have a significant benefit)?

wingyippp:
Sure!
For example, a flac audio file.
file size: 30.4MB
bit rate: 946kb/s
sample rate: 48.0 kHz
channels: 2
bit depth: 16 bits
duration: 4min 30s

seekTo timeUs:67000000.
Before this commit, SeekMap.getSeekPoints() returns [[timeUs=59000000, position=6064446]]
In this commit, SeekMap.getSeekPoints() returns [[timeUs=59000000, position=6064446], [timeUs=69000000, position=7321068]]
So in this case, the different between 2 seek points is 10 seconds.

Moreover, if seekMap.getSeekPoints() returns only one seek points, the SeekParameters of MediaPeriod.getAdjustedSeekPositionUs() method is useless. No matter SeekParameters is {EXACT, CLOSEST_SYNC, PREVIOUS_SYNC, NEXT_SYNC}, getAdjustedSeekPositionUs() has only one seek point to choose. So I prefer seekMap.getSeekPoints() can return 2 seek points.

P.S. The audio file is larger than 10MB. I cannot upload here~

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -452,8 +454,21 @@ int64_t FLACParser::getSeekPosition(int64_t timeUs) {
...
+        result[3] = firstFrameOffset + points[i + 1].stream_offset;
+      } else {
+        result[2] = result[0];
+        result[3] = result[1];
+      }
+      return result;
     }
   }
-  return firstFrameOffset;
+  result[0] = timeUs;
```
ojw28: Should this be set to `0`? The same question for `result[2]`? I think `firstFrameOffset` corresponds to `t=0` rather than `timeUs`.

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
```diff
@@ -341,8 +342,12 @@ public boolean isSeekable() {
 
     @Override
     public SeekPoints getSeekPoints(long timeUs) {
-      // TODO: Access the seek table via JNI to return two seek points when appropriate.
-      return new SeekPoints(new SeekPoint(timeUs, decoderJni.getSeekPosition(timeUs)));
+      long[] result = decoderJni.getSeekPosition(timeUs);
+      if (result.length == 4) {
```
ojw28: Isn't the result always of length 4?

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
```diff
@@ -283,7 +283,7 @@ private native int flacDecodeToArray(long context, byte[] outputArray)
 
   private native long flacGetNextFrameFirstSampleIndex(long context);
 
-  private native long flacGetSeekPosition(long context, long timeUs);
+  private native long[] flacGetSeekPosition(long context, long timeUs);
```
ojw28: Can this return `SeekPoints` directly, so the reader doesn't need to understand what this array contains?

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -438,9 +438,11 @@ size_t FLACParser::readBuffer(void *output, size_t output_size) {
   return bufferSize;
 }
 
-int64_t FLACParser::getSeekPosition(int64_t timeUs) {
+int64_t* FLACParser::getSeekPosition(int64_t timeUs) {
```
ojw28: If this is now returning up to two seek positions, it should be called `getSeekPositions` or `getSeekPoints` (probably better if it's going to return a `SeekPoints` object. Similar naming comments apply elsewhere.

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
```diff
@@ -223,7 +223,7 @@ public long getNextFrameFirstSampleIndex() {
    * @return The corresponding position (byte offset) in the flac stream or -1 if the stream doesn't
    * have a seek table.
    */
-  public long getSeekPosition(long timeUs) {
+  public long[] getSeekPosition(long timeUs) {
```
ojw28: The Javadoc for this method needs updating.

ojw28 (commented):
Thanks for the clarifications. The change looks good overall. I've put some comments inline.

extensions/flac/src/main/jni/flac_parser.cc
wingyippp: Because there is no SeekPoints in C/C++ code, to do so I have to create a new class in C/C++. So, I just made it "getSeekPositions" simply.

wingyippp:
Is it Okay to merge? @ojw28 :)

wingyippp:
Is it OK to merge?

ojw28:
Hi! Sorry for the delay; we're a bit backlogged. We'll try and look at this soon (and I think we plan to merge it).

library/core/proguard-rules.txt
```diff
@@ -58,6 +58,16 @@
...
 
+-dontnote com.google.android.exoplayer2.extractor.SeekPoint
+-keepclasseswithmembers class com.google.android.exoplayer2.extractor.SeekPoint {
+  <init>(long, long);
+}
+
+-dontnote com.google.android.exoplayer2.extractor.SeekMap$SeekPoints
+-keepclasseswithmembers class com.google.android.exoplayer2.extractor.SeekMap$SeekPoints {
+  <init>(com.google.android.exoplayer2.extractor.SeekPoint, com.google.android.exoplayer2.extractor.SeekPoint);
+}
```
ojw28: I think this configuration should go in the flac extension rather than the core library, since these only need to be retained if the extension is being used.

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
```diff
@@ -220,11 +221,12 @@ public long getNextFrameFirstSampleIndex() {
    * stream.
    *
    * @param timeUs A seek position in microseconds.
-   * @return The corresponding position (byte offset) in the flac stream or -1 if the stream doesn't
+   * @return The corresponding SeekPoints in the flac seek table or
+   * {@link com.google.android.exoplayer2.extractor.SeekPoint#position} set -1 if the stream doesn't
```
ojw28: It would probably be a lot clearer if this method returned `null` if the stream doesn't have a seek table?

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacDecoderJni.java
```diff
@@ -283,7 +285,7 @@ private native int flacDecodeToArray(long context, byte[] outputArray)
 
   private native long flacGetNextFrameFirstSampleIndex(long context);
 
-  private native long flacGetSeekPosition(long context, long timeUs);
+  private native SeekMap.SeekPoints flacGetSeekPoints(long context, long timeUs);
```
ojw28, line 288: Ditto (return null if the stream does't have a seek table).

extensions/flac/src/main/java/com/google/android/exoplayer2/ext/flac/FlacExtractor.java
```diff
@@ -341,8 +341,7 @@ public boolean isSeekable() {
 
     @Override
     public SeekPoints getSeekPoints(long timeUs) {
-      // TODO: Access the seek table via JNI to return two seek points when appropriate.
-      return new SeekPoints(new SeekPoint(timeUs, decoderJni.getSeekPosition(timeUs)));
+      return decoderJni.getSeekPoints(timeUs);
```
ojw28: This would then return
```
SeekPoints seekPoints = decoderJni.getSeekPoints(timeUs);
return seekPoints == null ? new SeekPoints(SeekPoint.START) : seekPoints;
```

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -438,9 +438,11 @@ size_t FLACParser::readBuffer(void *output, size_t output_size) {
   return bufferSize;
 }
 
-int64_t FLACParser::getSeekPosition(int64_t timeUs) {
+int64_t* FLACParser::getSeekPositions(int64_t timeUs) {
+  int64_t *result = new int64_t[4];
+  memset(result, -1, sizeof(result));
```
ojw28: This code does not compile for me. Please take a look:
```
extensions/flac/src/main/jni/flac_parser.cc:443:29: error: 'memset' call operates on objects of type 'int64_t' (aka 'long long') while the size is based on a different type 'int64_t *' (aka 'long long *') [-Werror,-Wsizeof-pointer-memaccess]
  memset(result, -1, sizeof(result));
         ~~~~~~             ^~~~~~
extensions/flac/src/main/jni/flac_parser.cc:443:29: note: did you mean to dereference the argument to 'sizeof' (and multiply it by the number of elements)?
  memset(result, -1, sizeof(result));
                            ^~~~~~
```

ojw28 (commented):
Please see inline comments (hopefully the final round). Apologies for the delay.

wingyippp:
All are fixed. Please check again. Thank you! @ojw28

extensions/flac/src/main/jni/flac_parser.cc
```diff
@@ -452,8 +453,21 @@ int64_t FLACParser::getSeekPosition(int64_t timeUs) {
   for (unsigned i = mSeekTable->num_points; i > 0; ) {
     i--;
     if (points[i].sample_number <= sample) {
-      return firstFrameOffset + points[i].stream_offset;
+      result[0] = points[i].sample_number / getSampleRate() * 1000000LL;
+      result[1] = firstFrameOffset + points[i].stream_offset;
+      if (i + 1 <= mSeekTable->num_points) {
```
ojw28, line 458: No need to push any further changes as we'll edit this during merge, but just to sanity check my thinking:

* This should be `<` rather than `<=`, right?
* I'm also going to adjust this to use the else path if `points[i].sample_number == sample` (i.e. the exact seek case). Does that make sense?

wingyippp: Are you talking about the Line 455, which is "if (points[i].sample_number <= sample) {"?
If yes, that makes sense.

ojw28: I meant that this line should use `i + 1 < mSeekTable->num_points` rather than `i + 1 <= mSeekTable->num_points`.

The biggest index you can use into `points` is `num_points - 1` (due to 0-indexing), but the current logic allows `num_points` to be used. I was able to make the current implementation return a broken looking seek point by seeking right to the end of a stream as a result.

ojw28:
This is merged. It ended up being turned back into something a bit more like what you had originally, as it went through our internal review process. Apologies for the churn :). Please take a look at the final version in `dev-v2` and check that it works for you.

Thanks a lot for the high quality contribution!

Issue #6366, opened 2019-08-30
Playback stalls forever in buffering with "Music Choice" content tunneling


stevemayhew:
### Content description
The content is a stream generated from Music Choice channels (often found in cable broadcast).  What happens is playback stalls in buffering state, the video renderer is stalled (`Renderer.isReady()` reports false because the sample queue has been completely read) but no more video segments loads are kicked off (`LoadControl` does not request loading).

I've only reproduced this on Broadcom STB hardware (Broadcom BCM7271 based) with tunneling on.

###  Link to test content
Provided by email.

### [REQUIRED] Version of ExoPlayer being used
Dev-v2-2.10.4 (sha d1ac2727a)

### [REQUIRED] Device(s) and version(s) of Android being used
Android P on an Evolution eSTREAM STB.

### [REQUIRED] A full bug report captured from the device
Provided by email.

stevemayhew:
The issue is caused by the stream, but it points out something worth considering in ExoPlayer's buffering state transitions algorithm...

Because the video frame rate is so low the first segment of video is loaded, decoded and buffered in the video MediaCodec, then the renderer is no longer isReady().  Falling out of the loop it would move the playhead forward enough to cause it to need to load more video, however the video never gets a chance to catch up to the audio because the audio renderer is stoped note even a second into playing.

Here's the PTS from the first segments:

AUDIO

```
$ ffprobe -v error -select_streams a:0 -show_entries frame=pkt_pts_time,pkt_dts_time,pkt_pos,pkt_size,pict_type,key_frame,width,height -of csv -i AUDIO_eng-Stereo/CCURStream_mcrockstg0-10_261100133.tsa  | head -10
frame,1,50830.694578,50830.694578,376,268
frame,1,50830.715900,50830.715900,752,251
frame,1,50830.737233,50830.737233,1128,237
frame,1,50830.758567,50830.758567,1504,240
frame,1,50830.779900,50830.779900,1880,275
frame,1,50830.801233,50830.801233,2256,268
frame,1,50830.822567,50830.822567,2632,296
frame,1,50830.843900,50830.843900,3008,242
frame,1,50830.865233,50830.865233,3384,247
frame,1,50830.886567,50830.886567,3760,239
...
[...]

PR #6400, opened 2019-09-05
WIP: Work around and demo for stuck buffering tunneling issue

stevemayhew:
This pull is just  to expose the issue, logging and one workaround/fix for discussion. 

The fix is for the "Playback stalls forever in buffering with "Music Choice" content tunneling.  Issue #6366 

Look at the change in MediaCodecVideoRenderer for the 'workaround/fix'

I will close and re-open a new pull request with a proper base to `dev-v2` and just the final fix.


library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -405,7 +405,11 @@ protected void onPositionReset(long positionUs, boolean joining) throws ExoPlayb
 
   @Override
   public boolean isReady() {
-    if (super.isReady() && (renderedFirstFrame || (dummySurface != null && surface == dummySurface)
+//        if (super.isReady() && (renderedFirstFrame || (dummySurface != null && surface == dummySurface)
+    // This "works around" the problem, even if it is not time to render a frame, and we don't have samples
+    // (because the entire LoadControl limited buffered sample queue is in the decoder), stay 'ready' to render the frame
+    // when it is due.
+    if ((super.isReady() || buffersInCodecCount > 0) && (renderedFirstFrame || (dummySurface != null && surface == dummySurface)
         || getCodec() == null || tunneling)) {
```
stevemayhew, line 413: The 'fix' is to keep reporting ready, even if we have fully buffered samples, as long as buffers are in the decoder that have not been presented to the output.

stevemayhew: The obvious drawback is the main `doSomeWork()` loop will not transition to buffering playback state in the event there are decoded buffers that have not been rendered.

stevemayhew:
I've cleaned the fix up and opened a new pull request.

PR #6407, opened 2019-09-06
Stuck buffering while tunneling.

stevemayhew:
This pull fixes the issue where the video render does not correctly report ready in tunneled mode.  The specific bug is #6366 

I can remove the manifest and network security changes (these allow local http testing on my workstation).  I've provided a stream that reproduces this on streaming devices with Broadcom SOC chips. 

The issue is largely timing related, streams that have very low video frame rate compared to the audio will cause the issue to manifest strait away.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
@@ -34,6 +34,7 @@ import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.ExoPlaybackException;
 import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.FormatHolder;
+import com.google.android.exoplayer2.Renderer;
 import com.google.android.exoplayer2.decoder.CryptoInfo;
 import com.google.android.exoplayer2.decoder.DecoderCounters;
 import com.google.android.exoplayer2.decoder.DecoderInputBuffer;
@@ -1144,6 +1145,19 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
     return outputIndex >= 0;
   }
 
+  /**
+   * Check if the renderer has one or more output frames queued to render.  For non-tunneled
+   * mode MediaCodecRenderer's this is true if the codec has returned a frame ready to render
+   * (that is {@see #hasOutputBuffer() is true}.
+   *
+   * This factors into the ready {@link Renderer#isReady()} decision
+   *
+   * @return
+   */
+  protected boolean hasOutputReady() {
+    return hasOutputBuffer();
+  }
+
   private void resetInputBuffer() {
     inputIndex = C.INDEX_UNSET;
     buffer.data = null;
@@ -1528,7 +1542,7 @@ public abstract class MediaCodecRenderer extends BaseRenderer {
     return inputFormat != null
         && !waitingForKeys
         && (isSourceReady()
-            || hasOutputBuffer()
+            || hasOutputReady()
             || (codecHotswapDeadlineMs != C.TIME_UNSET
                 && SystemClock.elapsedRealtime() < codecHotswapDeadlineMs));
   }
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
@@ -145,6 +145,8 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
   /* package */ @Nullable OnFrameRenderedListenerV23 tunnelingOnFrameRenderedListener;
   @Nullable private VideoFrameMetadataListener frameMetadataListener;
 
+  private long lastInputTimeUs;
+  private long lastOutputTimeUs;
   /**
    * @param context A context.
    * @param mediaCodecSelector A decoder selector.
@@ -231,6 +233,8 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
     frameReleaseTimeHelper = new VideoFrameReleaseTimeHelper(this.context);
     eventDispatcher = new EventDispatcher(eventHandler, eventListener);
     deviceNeedsNoPostProcessWorkaround = deviceNeedsNoPostProcessWorkaround();
+    lastInputTimeUs = C.TIME_UNSET;
+    lastOutputTimeUs = C.TIME_UNSET;
     joiningDeadlineMs = C.TIME_UNSET;
     currentWidth = Format.NO_VALUE;
     currentHeight = Format.NO_VALUE;
@@ -370,6 +374,8 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
     clearRenderedFirstFrame();
     initialPositionUs = C.TIME_UNSET;
     consecutiveDroppedFrameCount = 0;
+    lastInputTimeUs = C.TIME_UNSET;
+    lastOutputTimeUs = C.TIME_UNSET;
     if (joining) {
       setJoiningDeadlineMs();
     } else {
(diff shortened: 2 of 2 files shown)


demos/main/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,7 @@
       android:banner="@drawable/ic_banner"
       android:largeHeap="true"
       android:allowBackup="false"
+      android:networkSecurityConfig="@xml/network_security_config"
       android:name="com.google.android.exoplayer2.demo.DemoApplication"
```
stevemayhew: This commit is just for my testing, I will revert it out of the final pull if needed.

demos/main/src/main/java/com/google/android/exoplayer2/demo/PlayerActivity.java
```diff
@@ -365,6 +368,17 @@ private void initializePlayer() {
       RenderersFactory renderersFactory =
           ((DemoApplication) getApplication()).buildRenderersFactory(preferExtensionDecoders);
 
+      boolean enableTunneling = intent.getBooleanExtra(ENABLE_TUNNELED_PLAYBACK, false);
+      // Get a builder with current parameters then set/clear tunnling based on the intent
```
stevemayhew: Add a simple intent extra (`-ez enable_tunneled_playback true`) that allows you view a URL with tunneling on for platforms that support it.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -979,6 +980,19 @@ private boolean hasOutputBuffer() {
...
+   * (that is {@see #hasOutputBuffer() is true}.
+   *
+   * This factors into the ready {@link Renderer#isReady()} decision
+   *
+   * @return
+   */
+  protected boolean hasOutputReady() {
+    return hasOutputBuffer();
+  }
+
```
stevemayhew, line 1160: Here's the crux of the fix, the comments should spell out the intentions.

stevemayhew: We can review and debate a good name...

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -1362,7 +1376,7 @@ public boolean isReady() {
     return inputFormat != null
         && !waitingForKeys
         && (isSourceReady()
-            || hasOutputBuffer()
+            || hasOutputReady()
```
stevemayhew, line 1545: More descriptive of what the check really is.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -479,6 +479,18 @@ public boolean isReady() {
...
+   * output ready to render whenever we have queued any sample buffers to the codec that
+   * it has not reported as rendered.
+   *
+   * @return
+   */
+  @Override
+  protected boolean hasOutputReady() {
+    return tunneling && buffersInCodecCount > 0 || super.hasOutputReady();
+  }
+
```
stevemayhew, line 428: As the comment says, in tunneling mode once we have queued buffers to the video codec to decode it will render it when the matching audio syncs up.

The bug occurs because the audio track is stopped before the audio matching the video PTS is queued thus stalling the video codec forever.

stevemayhew (commented):
This is clearly a bug (Bug #6366 ) in the tunneling implementation.  Can Andrew or someone else review this  Thanks.

andrewlewis:
Had a quick look. This looks good for API 23 onwards where we can update `buffersInCodecCount` correctly. I'm not certain it works for API 21 though because we don't know when a buffer is rendered so we decrement the counter right after queueing (please let me know if I'm missing something).

stevemayhew:
🙁it’s a damm shame Lollipop still even matters...  Is tunneling even
supported that old?

But, yes, the fix won’t work correctly if the codec doesn’t report rendered
events when in tunneled mode

demos/main/src/main/AndroidManifest.xml
AquilesCanta: Humble suggestion: I think it would be great for the reviewer to split into 2 pull requests:
- One for the Renderer change.
- One for the demo app intent change.

A third one would be allowing clear text in the demo app, but i suspect that's less likely to be accepted, since it was probably an explicit decision not to allow it.

This way, the decision of bringing in the tunneling fix is not affected by the decision of enabling clear text, plus avoids any back-and-forth.

demos/main/src/main/java/com/google/android/exoplayer2/demo/PlayerActivity.java
krocard: A similar option was implemented in 9af8511

Please rebase.

demos/main/src/main/AndroidManifest.xml
stevemayhew: thanks, the demo app change was not really intended to submit.  I'll pull it out.

stevemayhew: This change is now removed from the branch.

stevemayhew:
I pulled out the changes not required for the actual bug fix, sorry for polluting the pull request with these.

andrewlewis:
Thinking about this more, I'm not sure it's safe to assume we get one output buffer per input buffer when using tunneling mode, due to the possibility of the OEM doing frame rate conversion (this wouldn't be valid for non-tunneled playbacks and I think is covered by CTS, but as I understand it one of the reasons that tunneling mode exists is to give flexibility in postprocessing). I will investigate further whether we can avoid using `buffersInCodecCount`.

stevemayhew:
@andrewlewis We are seeing this issue on boxes when there is low bandwidth that delays the audio (so similar to the issue the Music Choice reproduces).   

One of my team members, @dpolyakova is our internal expert on all things Broadcom media stack and has some ideas on this.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -409,7 +415,15 @@ public boolean isReady() {
    */
   @Override
   protected boolean hasOutputReady() {
-    return tunneling && buffersInCodecCount > 0 || super.hasOutputReady();
+    boolean fifoReady = true;
```
dpolyakova, line 418: Here we are trying to use FIFO length as renderer readiness criteria. If the frame rate is low such as 1fps on music channels the length would still be satisfactory and player wouldn't go into buffering state as long as there is a frame queued in the FIFO. If network bandwidth is low the FIFO length will get close to zero before decoder starts stalling. It's important to note that Broadcom's decoder will stall before rendering last few frames so zero FIFO length cannot be used as a criteria. The 333ms value is somewhat experimental. It was chosen to accommodate at least 10 frames of 30fps stream.

dpolyakova:
@googlebot I consent

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
stevemayhew: @andrewlewis @ojw28  Where are you guys at with addressing this issue.  I'm preparing to update our code base to what will likely by your 2.11.5 release.   We would very much like to not keep merging conflicted changes to the `MediaCodecVideoRenderer`, for obvious reasons.

Pretty sure we can send you an STB (Broadcom based) that reproduces the issue #6366  I think I included the URL for a section of music choice content that reproduces the issue 100% of the time

We are also see this, as Dasha explained, in other streams where audio is buffered sufficiently ahead of video.  As I'm sure you guys are aware, Broadcom codecs often require some special spoon feeding.

Thanks again for looking at this with us, please tell me what I need to do to help the process along.

PR #6442, opened 2019-09-19
 allow to disable previous ads before continue watching position

zzzhc:
Skip the first ad will improve retention when users re-play a video which has been played a part of it.

This PR did 3 things:
* adds `continueWatchingPositionMs` parameter in `ImaAdsLoader`, allow the app to disable previous ads before continue watching position
* improve ima-demo MainActivity for testing ad tag samples provided by dfp
* calculate `adGroupIndex` from `AdPod`.`TimeOffset` in `handleAdEvent` method


zzzhc:
@googlebot I signed it!

andrewlewis:
Thanks for the pull request.

- We should allow apps to set whether to play the ad immediately preceding the content start position when creating an `ImaAdsLoader` (this has been requested before). I think it should just be a boolean flag rather than a position that gets passed in, though, because the position would be redundant with the player's content position.
- There's an IMA SDK launch in progress, and once that's finished `AdEvent` handling should be much simpler. I'd like to wait for that before making changes to this code, as modifying it now may be wasted effort.
- The full demo app has a comprehensive list of IMA samples already. I think `imademo` should ideally be quite minimal.

If you'd be open to modifying this pull request to do just the first item in a minimal way that would be great (if not I'm happy to take care of this). Thanks!

zzzhc:
Thank you for your reply. I will try to improve the PR.
do you know when the new IMA SDK will release?

zzzhc:
new PR: https://github.com/google/ExoPlayer/pull/6467

Issue #6501, opened 2019-10-01
[Dash:DefaultDashChunkSource]: Wrong segmentNumShift was calculated in copyWithNewRepresentation


caiiiyua:
### [REQUIRED] Issue description
We are using AWS MediaPackager with Elemental for LIVE streaming, (etc, https://skysport3.akamaized.net/out/v1/686ec493b68042c7b943b4f55c95b867/index.mpd), by default the timeShiftBufferDepth="PT57.000S", we configured a 1800s availability window on the AWS MediaPackager, so that we could get MPD via somehing like https://skysport3.akamaized.net/out/v1/686ec493b68042c7b943b4f55c95b867/index.mpd?start=1569876612, which has timeShiftBufferDepth="PT1800.000S" with the start parameter.
Somehow the AWS MediaPackager is not 100% stable, so when the ExoPlayer request the MPD with the start parameter, suppose it will return all the MPDs with timeShiftBufferDepth="PT1800.000S",  but randomly rarely timeShiftBufferDepth="PT57.000S" instead.

Above is the background the reason lead to the wrong segmentNumShift problem.

**The root cause:**
In DefaultDashChunkSource.copyWithNewRepresentation, it will handle the logic that new MPD manifest file is updated and calculate a newSegmentNumShift for furthermore segNum index calculation in getSegmentUrl, when a 57s MPD updated and then back to 1800s MPD,  copyWithNewRepresentation will go into the overlap case.
For example:
When 57s MPD was updated:
```
10-02 11:19:54.451 24089 24089 D EventLogger: timelineChanged [30.75, 24.83, window=0, period=0, periodCount=1, windowCount=1, reason=DYNAMIC
10-02 11:19:54.451 24089 24089 D EventLogger:   period [?]
10-02 11:19:54.451 24089 24089 D EventLogger:   window [59.98, true, true]
10-02 11:19:54.451 24089 24089 D EventLogger: ]
10-02 11:19:55.652 24089 25278 D EventLogger: getNextChunk: previous: 193603
10-02 11:19:55.652 24089 25278 D EventLogger: getNextChunk: firstAvailableSegmentNum: 193595 lastAvailableSegmentNum: 193604 segmentNum: 193604
10-02 11:19:55.654 24089 25278 D EventLogger: getNextChunk: maxSegmentCount: 1 segmentNum: 193604
10-02 11:20:10.032 24089 25278 D EventLogger: copyWithNewRepresentation: oldIndexSegmentCount: 10
[...]

PR #6502, opened 2019-10-01
Issue: #6501 Wrong segmentNumShift calculated in copyWithNewRepresentation

caiiiyua:
In DefaultDashChunkSource.copyWithNewRepresentation, it will handle the logic that
new MPD manifest file is updated and calculate a newSegmentNumShift for furthermore
segNum index calculation in getSegmentUrl, when a shorter window MPD updated and then
back to a longer window MPD, copyWithNewRepresentation will go into the overlap case
but the new index actually contains the old index.
Fix: https://github.com/google/ExoPlayer/issues/6501

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
@@ -702,12 +702,14 @@ public class DefaultDashChunkSource implements DashChunkSource {
         // There's a gap between the old index and the new one which means we've slipped behind the
         // live window and can't proceed.
         throw new BehindLiveWindowException();
-      } else if (oldIndexStartTimeUs >= newIndexStartTimeUs) {
+      } else if (newIndexStartTimeUs < oldIndexStartTimeUs) {
         // The new index overlaps with (but does not have a start position contained within) the old
         // index. This can only happen if extra segments have been added to the start of the index.
-        // Continue process the next segment as is.
+        newSegmentNumShift -=
+            newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
+                - oldIndexFirstSegmentNum;
       } else {
-        // The new index overlaps with the old one.
+        // The new index overlaps with (and has a start position contained within) the old index.
         newSegmentNumShift +=
             oldIndex.getSegmentNum(newIndexStartTimeUs, newPeriodDurationUs)
                 - newIndexFirstSegmentNum;
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
@@ -686,6 +686,8 @@ public class DefaultDashChunkSource implements DashChunkSource {
             newPeriodDurationUs, newRepresentation, extractorWrapper, segmentNumShift, newIndex);
       }
 
+      long oldIndexFirstSegmentNum = oldIndex.getFirstSegmentNum();
+      long oldIndexStartTimeUs = oldIndex.getTimeUs(oldIndexFirstSegmentNum);
       long oldIndexLastSegmentNum = oldIndex.getFirstSegmentNum() + oldIndexSegmentCount - 1;
       long oldIndexEndTimeUs =
           oldIndex.getTimeUs(oldIndexLastSegmentNum)
@@ -700,8 +702,10 @@ public class DefaultDashChunkSource implements DashChunkSource {
         // There's a gap between the old index and the new one which means we've slipped behind the
         // live window and can't proceed.
         throw new BehindLiveWindowException();
-      } else if (oldIndex.getFirstSegmentNum() >= newIndexFirstSegmentNum) {
-        // The new index contains the old one, continue process the next segment
+      } else if (oldIndexStartTimeUs >= newIndexStartTimeUs) {
+        // The new index overlaps with (but does not have a start position contained within) the old
+        // index. This can only happen if extra segments have been added to the start of the index.
+        // Continue process the next segment as is.
       } else {
         // The new index overlaps with the old one.
         newSegmentNumShift +=
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
@@ -700,6 +700,8 @@ public class DefaultDashChunkSource implements DashChunkSource {
         // There's a gap between the old index and the new one which means we've slipped behind the
         // live window and can't proceed.
         throw new BehindLiveWindowException();
+      } else if (oldIndex.getFirstSegmentNum() >= newIndexFirstSegmentNum) {
+        // The new index contains the old one, continue process the next segment
       } else {
         // The new index overlaps with the old one.
         newSegmentNumShift +=


caiiiyua:
@googlebot I signed it!

caiiiyua:
@googlebot I fixed it.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
```diff
@@ -700,6 +700,8 @@ private RepresentationHolder(
         // There's a gap between the old index and the new one which means we've slipped behind the
         // live window and can't proceed.
         throw new BehindLiveWindowException();
+      } else if (oldIndex.getFirstSegmentNum() >= newIndexFirstSegmentNum) {
```
ojw28: One of the main purposes of this code is to handle first segment numbers not being set correctly in the DASH manifest (setting `segmentNumShift` corrects such cases). So I think this code needs to avoid relying on first segment numbers for deciding what to do, and also handle the case where `newSegmentNumShift` still needs to be adjusted.

Does changing this whole block to the following also work, and if so could we do that instead? Please give it a try (I was unable to do so because the test link provided didn't work for me!)

```
      long oldIndexFirstSegmentNum = oldIndex.getFirstSegmentNum();
      long oldIndexStartTimeUs = newIndex.getTimeUs(oldIndexFirstSegmentNum);
      long oldIndexLastSegmentNum = oldIndex.getFirstSegmentNum() + oldIndexSegmentCount - 1;
      long oldIndexEndTimeUs =
          oldIndex.getTimeUs(oldIndexLastSegmentNum)
              + oldIndex.getDurationUs(oldIndexLastSegmentNum, newPeriodDurationUs);
      long newIndexFirstSegmentNum = newIndex.getFirstSegmentNum();
      long newIndexStartTimeUs = newIndex.getTimeUs(newIndexFirstSegmentNum);
      long newSegmentNumShift = segmentNumShift;
      if (oldIndexEndTimeUs == newIndexStartTimeUs) {
        // The new index continues where the old one ended, with no overlap.
        newSegmentNumShift += oldIndexLastSegmentNum + 1 - newIndexFirstSegmentNum;
      } else if (oldIndexEndTimeUs < newIndexStartTimeUs) {
        // There's a gap between the old index and the new one which means we've slipped behind the
        // live window and can't proceed.
        throw new BehindLiveWindowException();
      } else if (newIndexStartTimeUs < oldIndexStartTimeUs) {
        // The new index overlaps with (but does not have a start position contained within) the old
        // index. This can only happen if extra segments have been added to the start of the index.
        newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
                + newIndexFirstSegmentNum);
      } else {
        // The new index overlaps with (and has a start position contained within) the old index.
        newSegmentNumShift +=
            (oldIndex.getSegmentNum(newIndexStartTimeUs, newPeriodDurationUs)
                - newIndexFirstSegmentNum);
      }
      return new RepresentationHolder(
          newPeriodDurationUs, newRepresentation, extractorWrapper, newSegmentNumShift, newIndex);
```

caiiiyua: The logic you mentioned in the comments of the code looks great, but the calculation seems not correct, I would assume in this case, the newSegmentNumShift should actually as is, because the new index has the correct first segment number, I will try to understand the whole logic then.

The reason the link didn't work is due to the Token in the CustomLicenseServerUrl was expired, it has only 4 hours lifecycle.
Post a new one here: tEbZ6yIgPnzMCMkwPQZScdASQFBKsHAU
And also need to change the "start" parameter in manifest url to a epoch time that more than 30mins from now. 
There was a minor mistake in the new code snippet, should be oldIndex instead of newIndex:
`long oldIndexStartTimeUs = oldIndex.getTimeUs(oldIndexFirstSegmentNum);`

10-04 13:25:03.890 13527 18006 D EventLogger: copyWithNewRepresentation: oldIndex.getFirstSegmentNum(): 222984 newIndexFirstSegmentNum: 222695
10-04 13:25:03.890 13527 18006 D EventLogger: copyWithNewRepresentation: newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs): 222984
10-04 13:25:03.890 13527 18006 D EventLogger: copyWithNewRepresentation: segmentNumShift: 0 newSegmentNumShift: -445679
10-04 13:25:03.891 13527 18006 D EventLogger: getNextChunk 2: firstAvailableSegmentNum: -222984 lastAvailableSegmentNum: -222684 segmentNum: 222994
10-04 13:25:03.892 13527 13527 D EventLogger: timelineChanged [46.28, 1775.89, window=0, period=0, periodCount=1, windowCount=1, reason=DYNAMIC
10-04 13:25:03.892 13527 13527 D EventLogger:   period [?]
10-04 13:25:03.892 13527 13527 D EventLogger:   window [1805.99, true, true]
10-04 13:25:03.892 13527 13527 D EventLogger: ]

So in the end, the firstAvailableSegmentNum and lastAvailableSegmentNum would be wrong.

caiiiyua: How about I changed the condition to the below:
`} else if (oldIndexStartTimeUs >= newIndexStartTimeUs) {`
where oldIndexStartTimeUs = oldIndex.getTimeUs(oldIndexFirstSegmentNum);

Will update the PR

caiiiyua: PR updated.

ojw28: > I would assume in this case, the newSegmentNumShift should actually as is, because the new index has the correct first segment number

As per my comment above, one of the purposes of this code is to correctly calculate an offset when the first segment number isn't correct. You're right that my proposed logic is wrong, but I think it's easy to fix it. Where I put:
```
newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
                + newIndexFirstSegmentNum);
```
I should have put:
```
newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
                + oldIndexFirstSegmentNum);
```
If the segment numbers are already correct in the manifest then `newSegmentNumShift` should always remain unchanged. The difference between my suggestion and yours is that I think you're assuming this is the only case, where-as we also want to handle the case where it's not.

Please could you try testing with this solution to see whether it works? It's hard for me to test given only a four hour window from each time you provide a token. I've provided the updated complete code block below
```
      long oldIndexFirstSegmentNum = oldIndex.getFirstSegmentNum();
      long oldIndexStartTimeUs = newIndex.getTimeUs(oldIndexFirstSegmentNum);
      long oldIndexLastSegmentNum = oldIndex.getFirstSegmentNum() + oldIndexSegmentCount - 1;
      long oldIndexEndTimeUs =
          oldIndex.getTimeUs(oldIndexLastSegmentNum)
              + oldIndex.getDurationUs(oldIndexLastSegmentNum, newPeriodDurationUs);
      long newIndexFirstSegmentNum = newIndex.getFirstSegmentNum();
      long newIndexStartTimeUs = newIndex.getTimeUs(newIndexFirstSegmentNum);
      long newSegmentNumShift = segmentNumShift;
      if (oldIndexEndTimeUs == newIndexStartTimeUs) {
        // The new index continues where the old one ended, with no overlap.
        newSegmentNumShift += oldIndexLastSegmentNum + 1 - newIndexFirstSegmentNum;
      } else if (oldIndexEndTimeUs < newIndexStartTimeUs) {
        // There's a gap between the old index and the new one which means we've slipped behind the
        // live window and can't proceed.
        throw new BehindLiveWindowException();
      } else if (newIndexStartTimeUs < oldIndexStartTimeUs) {
        // The new index overlaps with (but does not have a start position contained within) the old
        // index. This can only happen if extra segments have been added to the start of the index.
        newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
                + oldIndexFirstSegmentNum);
      } else {
        // The new index overlaps with (and has a start position contained within) the old index.
        newSegmentNumShift +=
            (oldIndex.getSegmentNum(newIndexStartTimeUs, newPeriodDurationUs)
                - newIndexFirstSegmentNum);
      }
      return new RepresentationHolder(
          newPeriodDurationUs, newRepresentation, extractorWrapper, newSegmentNumShift, newIndex);
```

ojw28:
See inline comment. Thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
caiiiyua: The following logic still not correct, newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs) will return oldIndexFirstSegmentNum, + oldIndexFirstSegmentNum is not the offset of old and new index.
`newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
                + oldIndexFirstSegmentNum);`
From EventLog we will get:
10-07 10:29:10.157 15784 16203 D EventLogger: copyWithNewRepresentation: oldIndex.getFirstSegmentNum(): 264425 newIndexFirstSegmentNum: 264135
10-07 10:29:10.157 15784 16203 D EventLogger: copyWithNewRepresentation: segmentNumShift: 0 newSegmentNumShift: -528850

After understanding this part of the code, I realized that the original code is trying to find the offset of firstSegmentNum between oldIndex and newIndex when overlaps.
So I would change the fix as below:
```
 else if (newIndexStartTimeUs < oldIndexStartTimeUs) {
        // The new index overlaps with (but does not have a start position contained within) the old
        // index. This can only happen if extra segments have been added to the start of the index.
        newSegmentNumShift -=
            (newIndex.getSegmentNum(oldIndexEndTimeUs, newPeriodDurationUs)
                - oldIndexLastSegmentNum);
      }
```
So when newStartTime < oldStartTime, calculate the newSegmentNumShift by checking the offset between the oldEndTime in newIndex and oldLastSegmentNum itself, or the offset between the oldStartTime in newindex and oldFirstSegmentNum itself, both of the logic work.

Will update PR again.

Thanks a lot.

ojw28: > After understanding this part of the code, I realized that the original code is trying to find the offset of firstSegmentNum between oldIndex and newIndex when overlaps.

Yes. It's just trying to find the offset between the old and new indices. It doesn't matter which segment it's looking at to do this; the offset should always be the same.

> So I would change the fix as below

It seems nicer to use the start of the first segment rather than the end of the last segment. The way rounding works is quite unclear for the latter, and you may end up with an off-by-one error. I think your solution is equivalent to this, which is my solution but with a `+` updated to a `-`:

```
newSegmentNumShift -=
    (newIndex.getSegmentNum(oldIndexStartTimeUs, newPeriodDurationUs)
        - oldIndexFirstSegment;
```

Does that work? If so, let's go with that. Sorry for the churn!

ojw28:
See inline comment [here](https://github.com/google/ExoPlayer/pull/6502/files/c2d9960a6e3a6a790a11a3842e72f8cda5d6e946#r331958889). Thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
caiiiyua: Yep, using start time is also good.
I could confirm that the solution fixed the problem.
I also have updated the PR.
Great thanks :-D

ojw28:
Cool; thanks for your patience (sorry for my bad maths :)). We'll get this merged shortly.

PR #6535, opened 2019-10-11
Add content descriptions to playback controls

dturner:
I have reused the descriptions from `strings.xml`, however, there are 2 descriptions which need creating for "repeat" and "shuffle". If someone could let me know how to add these new strings, along with their translations I will add them as well.

### DIFF
--- a/library/ui/src/main/res/layout/exo_playback_control_view.xml
+++ b/library/ui/src/main/res/layout/exo_playback_control_view.xml
@@ -31,30 +31,39 @@
     android:orientation="horizontal">
 
     <ImageButton android:id="@id/exo_prev"
+      android:contentDescription="@string/exo_controls_previous_description"
       style="@style/ExoMediaButton.Previous"/>
 
     <ImageButton android:id="@id/exo_rew"
+      android:contentDescription="@string/exo_controls_rewind_description"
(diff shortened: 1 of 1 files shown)


library/ui/src/main/res/layout/exo_playback_control_view.xml
```diff
@@ -31,30 +31,39 @@
     android:orientation="horizontal">
 
     <ImageButton android:id="@id/exo_prev"
+      android:contentDescription="@string/exo_controls_previous_description"
```
ojw28, line 34: These are already added via the styles.

ojw28:
See inline comment. The repeat and shuffle ones are set in code because the description needs to be different based on the current button state (there are already strings - `exo_controls_repeat_*_description` and `exo_controls_shuffle_*_description`).

library/ui/src/main/res/layout/exo_playback_control_view.xml
dturner: Right. Good to know - thanks. This came from Android Studio's linter....I will file a bug. 

image

dturner:
Thanks for the review. Sorry for the noise. Android Studio bug filed: 

https://b.corp.google.com/issues/142636050

PR #6578, opened 2019-10-23
Fix incorrect anamorphic handling in Leanback extension.

bennettpeter:
Anamorphic videos were being displayed with incorrect aspect ratio as pixel shape
was not taken into account. This fix adjusts the video size based on pixel shape.

The problem can be seen and reproduced using the sample at https://github.com/android/tv-samples/tree/master/Leanback , adjusting the URLS to point to an anamorphic video. I used a 16x9 video that had 720x480 resolution (standard DVD resolution).


krocard:
Seems good to me, but I would like @ojw28 review as he wrote that code last year.
Also, do we have any anamorphic video content in the demo app ? If not, I can add some.

ojw28:
We do have some anamorphic content (at least internally). But our demo app doesn't use the leanback extension. I can get this merged; thanks!

ojw28:
Note: Contributions need to go into the `dev-v2` branch, rather than `release-v2`. Given this is a trivial change, I'll just merge the equivalent directly into `dev-v2`, and close this PR. Thanks!

Issue #6320, opened 2019-08-19
SSA/ASS subtitles - Overlapping start/end times and position tag is not handled


XBigTK13X:
### Use case description
When playing videos files with SSA/ASS subtitle tracks, styled subtitles are mashed together with conversation. For example, many animated Japanese television shows have subtitles that float on screen to translate things that aren't part of speech. In ExoPlayer, this results in the subtitles flipping back and forth between conversation and the translated symbols. This leads to neither being on screen long enough to be read.

### Proposed solution
Integrate [libass](https://github.com/libass/libass) as a supported subtitle renderer

### Alternatives considered
The only alternative would be rolling a new renderer from scratch for ExoPlayer. All major media players are currently using this library. I did look through other issues on this project about SSA/ASS subtitles, but none of them lead to a request for integrating this library as a solution.

tonihei:
I had a quick look at the library and from what I see it produces overlay images including the subtitles. Is that correct? I'm afraid that's not that easy to support in ExoPlayer, because you'd need some form of special renderer that draws the overlay. See #5860 for details around writing such a renderer. It would also require an extra surface and at least some extra layer in the UI module to actually draw the overlay. 

Depending on how it looks like, we could probably accept pull requests for an extension that integrates this library and provides the necessary integration layers. But we should definitely keep the basic text-only subtitle extraction for SSA/ASS we currently have. The problem you are describing sounds more like a bug in our implementation and it would be great if you can provide example media and some guidance on the expected result, so that we can fix our support for the text-only version.

XBigTK13X:
Thank you for the feedback. That points me in the right direction. I will try adding an alternative SSA renderer locally to the ExoPlayer source to see if that solves the problem for my project.

Here is a snippet from a subtitle file that renders incorrectly in the current ExoPlayer.

```
[Script Info]
; Script generated by Aegisub 3.2.2
; http://www.aegisub.org/
Title: ExportedSubs
ScriptType: v4.00+
WrapStyle: 0
PlayResX: 1280
PlayResY: 720
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709

[Aegisub Project Garbage]
Last Style Storage: Aikatsu 1
Audio File: yuru05_premux.mkv
Video File: yuru05_premux.mkv
Keyframes File: yuru05_premux_keyframes.log
Video AR Mode: 4
Video AR Value: 1.777778
Video Zoom Percent: 0.875000
Scroll Position: 1422
[...]

PR #6595, opened 2019-10-29
SSA/ASS subtitles - Overlapping start/end times and position tag is not handled

szaboa:
Pull request for #6320.

### DIFF
--- a/Project Default.xml
+++ b/Project Default.xml
@@ -1,0 +1,5 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+  </profile>
+</component>
--- a/Project Default.xml
+++ b/Project Default.xml
@@ -1,5 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <profile version="1.0">
-    <option name="myName" value="Project Default" />
-  </profile>
-</component>
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -226,44 +226,68 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
       cue = new Cue(text);
     }
 
-    int startTimeIndex = insertToCueTimes(cueTimesUs, startTimeUs);
+    int startTimeIndex = 0;
+    boolean startTimeFound = false;
+    // Search the insertion index for startTimeUs in cueTimesUs
+    for (int i = cueTimesUs.size() - 1; i >= 0; i--) {
+      if (cueTimesUs.get(i) == startTimeUs) {
+        startTimeIndex = i;
+        startTimeFound = true;
+        break;
+      }
 
-    List<Cue> startCueList = new ArrayList<>();
-    if (startTimeIndex != 0) {
-      startCueList.addAll(cues.get(startTimeIndex - 1));
-    }
-    cues.add(startTimeIndex, startCueList);
-
-    if (endTimeUs != C.TIME_UNSET) {
-      int endTimeIndex = insertToCueTimes(cueTimesUs, endTimeUs);
-      List<Cue> endList = new ArrayList<>(cues.get(endTimeIndex - 1));
-      cues.add(endTimeIndex, endList);
-
-      int i = startTimeIndex;
-      do {
-        cues.get(i).add(cue);
-        i++;
-      } while (i != endTimeIndex);
-    }
-  }
-
-  /**
-   * Insert the given cue time into the given array keeping the array sorted.
-   *
-   * @param cueTimes The array with sorted cue times
-   * @param timeUs The cue time to be inserted
-   * @return The index where the cue time was inserted
-   */
-  private static int insertToCueTimes(List<Long> cueTimes, long timeUs) {
-    for (int i = cueTimes.size() - 1; i >= 0; i--) {
-      if (cueTimes.get(i) <= timeUs) {
-        cueTimes.add(i + 1, timeUs);
-        return i + 1;
+      if (cueTimesUs.get(i) < startTimeUs) {
+        startTimeIndex = i + 1;
+        break;
       }
     }
 
-    cueTimes.add(0, timeUs);
-    return 0;
+    if (startTimeIndex == 0) {
+      // Handle first cue
+      cueTimesUs.add(startTimeIndex, startTimeUs);
+      cues.add(startTimeIndex, new ArrayList<>());
+    } else {
+      if (!startTimeFound) {
+        // Add the startTimeUs only if it wasn't found in cueTimesUs
+        cueTimesUs.add(startTimeIndex, startTimeUs);
+        // Copy over cues from left
+        List<Cue> startCueList = new ArrayList<>(cues.get(startTimeIndex - 1));
+        cues.add(startTimeIndex, startCueList);
+      }
+    }
+
+    int endTimeIndex = 0;
+    if (endTimeUs != C.TIME_UNSET) {
+      boolean endTimeFound = false;
+
+      // Search the insertion index for endTimeUs in cueTimesUs
+      for (int i = cueTimesUs.size() - 1; i >= 0; i--) {
+        if (cueTimesUs.get(i) == endTimeUs) {
+          endTimeIndex = i;
+          endTimeFound = true;
+          break;
+        }
+
+        if (cueTimesUs.get(i) < endTimeUs) {
+          endTimeIndex = i + 1;
+          break;
+        }
+      }
+
+      if (!endTimeFound) {
+        // Add the endTimeUs only if it wasn't found in cueTimesUs
+        cueTimesUs.add(endTimeIndex, endTimeUs);
+        // Copy over cues from left
+        cues.add(endTimeIndex, new ArrayList<>(cues.get(endTimeIndex - 1)));
+      }
+    }
+
+    // Iterate on cues from startTimeIndex until endTimeIndex, add the current cue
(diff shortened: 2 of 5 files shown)


library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -196,16 +204,67 @@ private void parseDialogueLine(String dialogueLine, List<Cue> cues, LongArray cu
...
+      List<Cue> endList = new ArrayList<>(cues.get(endTimeIndex - 1));
+      cues.add(endTimeIndex, endList);
+
+      int i = startTimeIndex;
+      do {
+        cues.get(i).add(cue);
+        i++;
+      } while (i != endTimeIndex);
+    }
+  }
```
icbaker: I could be wrong, but I'm not convinced this correctly handles multiple cues that have the **same** start or end time.
e.g.

```
[3, 5] -> "A"
[4, 5] -> "B"
```

We'll insert the first one, after which we have: `cueTimesUs=[3, 5]`, `startTimeIndex=0` & `endTimeIndex=1` and `cues=[["A"], []]`.

Then insert the second one: `cueTimesUs=[3, 4, 5, 5]`, `startTimeIndex=1` & `endTimeIndex=3` and `cues=[["A"], ["A", "B"], ["B"], []]`

Whereas I think with this input we'd want these lists: `cueTimesUs=[3, 4, 5]` and `cues=[["A"], ["A", "B"], []]`

It also took quite a lot of thought for me to follow that through, especially with the multiple mutations to the `cue` list on each iteration of the outer loop.

Note that we have this same overlapping challenge in the webvtt package and we actually solve it in a slightly different way, by more lazily evaluating `Subtitle#getCues` (every call to that iterates over all the subtitles we have) [1]. I chatted a bit to the team, and that seems unfortunately inefficient, so I think it makes sense to keep the logic here and not copy webvtt, but maybe we can correct it and make it a bit easier to follow.

My suggestion is to get rid of `insertToCueTimes()` and do something more like (I haven't tested this, it might have other problems...):

* (binary?) search for `startTimeUs` in `cueTimesUs`
    * if `startTimeUs` is already there, then get the matching list (by index) from `cues` and add `cue` to it.
    * else insert `startTimeUs` to `cueTimesUs` and insert a new matching list to `cues` (containing all the Cues from `index - 1` plus `cue`).
* Walk through `cueTimesUs`, adding `cue` to every entry matching entry in `cues` until you find a time that's either equal to or greater than `endTimeUs` (mostly your existing do/while loop)
  * On each step, store a reference to the matching list of cues **before** you add `cue`. (This reference should also store the list from the `else` in the first bullet before `cue` is added)
  * If the time you stopped on is equal to `endTimeUs`, then do nothing (the cues list already has the correct 'end' value, right?)
  * If it's greater, then insert a new cues list equal to the most recent list you stored at the top of this sub-section.


[1]
https://github.com/google/ExoPlayer/blob/41b3fc110178e9a5c969a0102c5b074189065273/library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttSubtitle.java#L75

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -226,4 +285,15 @@ public static long parseTimecodeUs(String timeString) {
     return timestampUs;
   }
 
+  @Nullable
+  public static Pair<Float, Float> parsePosition(String line){
```
icbaker: Might make more sense to use android.graphics.PointF here because it's a little less general/ambiguous than Pair<Float, Float>

Also avoids auto-boxing

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -98,6 +100,12 @@ protected Subtitle decode(byte[] bytes, int length, boolean reset) {
   private void parseHeader(ParsableByteArray data) {
     String currentLine;
     while ((currentLine = data.readLine()) != null) {
+      if (currentLine.startsWith("PlayResX:")) {
+        playResX = Integer.valueOf(currentLine.substring(9).trim());
```
icbaker: I think it's a *tiny* bit clearer to use the string literal again instead of a 'magic' length (saves a future reader manually counting the number of characters in "PlayResX:" :))

`playResX = Integer.valueOf(currentLine.substring("PlayResX:".length()).trim());`

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -196,16 +204,67 @@ private void parseDialogueLine(String dialogueLine, List<Cue> cues, LongArray cu
       }
     }
 
+    // Parse \pos{x,y} attribute
```
icbaker: nit: I'd move this comment to the javadoc of `parsePosition`

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -50,6 +53,9 @@
   private int formatEndIndex;
   private int formatTextIndex;
 
+  private int playResX;
+  private int playResY;
```
icbaker: It's probably safer to explicitly initialise these to an invalid value to clearly indicate 'unset', because 0 seems like a potentially genuine value we could see in a subtitle file? And then also update the comparison below on L216.

We have `C.LENGTH_UNSET` which I think would work well.
https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/C.java

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaSubtitle.java
```diff
@@ -28,45 +28,44 @@
...
-    return cueTimesUs[index];
+    Assertions.checkArgument(index < cueTimesUs.size());
+    return cueTimesUs.get(index);
   }
 
   @Override
   public List<Cue> getCues(long timeUs) {
     int index = Util.binarySearchFloor(cueTimesUs, timeUs, true, false);
-    if (index == -1 || cues[index] == Cue.EMPTY) {
+    if (index == -1 || cues.get(index).isEmpty()) {
```
icbaker: I think you can get rid of the empty check, since we'll just return **the** empty list below anyway (right?)

icbaker (changes requested):
Thanks for the contribution!

The position stuff looks good, just a few small comments.

I've added a larger comment about the general approach to the handling of overlapping start/end times. It's an awkward bit of algorithm/logic to get right (and in the future if we need to implement it a third time we might try and abstract it into a common utility class that handles all the overlapping resolution etc.).

On the subject of testing:
I think it probably makes sense to test at the level of SsaDecoder (rather than directly testing SsaSubtitle for example). It's probably worth adding my "equal start/end times" case as one of your tests :) As well as nested subtitles too:

```
[3, 7] -> "A"
[4, 5] -> "B"
```

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaSubtitle.java
szaboa: Yes, that's right.

Project Default.xml
```diff
@@ -0,0 +1,5 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+  </profile>
+</component>
```
szaboa: Deleted in commit fb2a70291847115cee2554c45cbc553aec9d78d6 :)

szaboa:
Applied the suggested changes (thank you!) except the one with same start or end time. I'll follow up on that tomorrow :)

szaboa:
I've pushed the latest changes.

Testing:
Couldn't test the code strictly on decoder level, as the `parseDialogueLine` is private, that's why I verified the cue times and length of cues from the decoded subtitle. Also needed to correct the "no endtime" test case, with this overlapping mechanism that use case behaves differently from now on (it carries over the previous subtitles). It is ok like this?

Overlapping challenge:

> I could be wrong, but I'm not convinced this correctly handles multiple cues that have the same start or end time.

Yes, the previous implementation resulted redundant cues and cueTimes in this case, visually it was fine though.

I agree that we shouldn't add redundant things, and the algorithm you've described is more clear and bit more optimal (no need to search ahead where endTimeUs fits).

I've tried to implement it following that approach (a lot of trial and error) but couldn't really succeed because of the corner cases (e.g. handling first cue, endTimeUs is greater than all of the times, same startTime/endTime, storing reference of cues - to later add it - means we need to inspect the next+1 cueTime, not just the next one etc.) so decided to go back to the first approach and just correct the same startTime/endTime problem. 

What do you think?

icbaker:
Looks good, thanks! I'll work on getting this merged.

szaboa:
Great, let me know if any further changes are needed.

icbaker:
Just to keep you posted, I haven't forgotten about this :)

It turns out it's a little tricky to merge this while also supporting the 'blank' end timecode behaviour currently in SsaDecoder (where the intention is that the line appears only until the next line...). I've chatted with the team, and we're likely to remove the blank end timecode 'feature' as it's not really supported by the spec afaict.

Once that support is removed, I'll be able to merge this more easily.

szaboa:
Sure, there's no hurry :)

MurtadhaS:
Hi guys, any news regarding this PR?

icbaker:
Just merged it in to dev-v2 (with fairly significant changes, but the functionality originally proposed should all be there).

Issue #4352, opened 2018-06-08
ANR on player release


aarondib:
### Issue description
According to log reports from users, after upgrade to r2.8.1 (from r2.6.1), on about 1% of app sessions we have ANR on SimpleExoPlayer.release().
Never seen this issue (reports) on versions <=r2.6.1.

### Reproduction steps
Play some stream and then call SimpleExoPlayer.release(), called in main thread.

### Link to test content
Unknown for a while.

### Version of ExoPlayer being used
r2.8.1.

### Device(s) and version(s) of Android being used
Different devices and Android versions: Nvidia Shield TV, Xiaomi Mi-Box, etc.

### A full bug report captured from the device
```
"main" tid=1 Waiting

"main" prio=5 tid=1 Waiting | group="main" sCount=1 dsCount=0 obj=0x762777f0 self=0x7527e95a00 | sysTid=32258 nice=-10 cgrp=default sched=0/0 handle=0x752beaea98 | state=S schedstat=( 264497782805 29710697109 1267932 ) utm=22158 stm=4291 core=7 HZ=100 | stack=0x7ff8a69000-0x7ff8a6b000 stackSize=8MB | held mutexes=
at java.lang.Object.wait! (Native method)
- waiting on <0x0225fb39> (a com.google.android.exoplayer2.ExoPlayerImplInternal)
at com.google.android.exoplayer2.ExoPlayerImplInternal.release (unavailable)
- locked <0x0225fb39> (a com.google.android.exoplayer2.ExoPlayerImplInternal)
com.google.android.exoplayer2.ExoPlayerImpl.release (unavailable)
com.google.android.exoplayer2.SimpleExoPlayer.release (unavailable)
...
```

andrewlewis:
- It would be useful to see a bug report so we can check what the playback thread is doing (it is responsible for notifying the thread waiting in `release` that resources have been returned to the system). If you have one, please could you attach it here or email it to `dev.exoplayer@gmail.com` with subject "Issue #4352"?
- Do you have any information on whether all devices are affected equally, or certain devices are affected more than others?
- To help narrow down which changes between 2.6.1 and 2.8.1 might have caused this, it would be good to know if the app is doing anything that would exercise recently added functionality like: setting/clearing the player's surface during playback, or using blocking message sending (we couldn't see any issues with this, but thought it was worth checking just in case). Thanks!

ojw28:
[...]

PR #6603, opened 2019-10-31
Fix for ANR on release (at least one case)

stevemayhew:
This fixes an ANR we are seeing on Player.release().  The problem is the `MediaCodec` listener callbacks hold a lock that blocks any other message to `MediaCodec`'s handler.  This can cause a call into `MediaCodec` to not complete

The bug number is: https://github.com/google/ExoPlayer/issues/4352

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
@@ -563,6 +563,9 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
     clearReportedVideoSize();
     clearRenderedFirstFrame();
     frameReleaseTimeHelper.disable();
+    if (tunnelingOnFrameRenderedListener != null && getCodec() != null) {
+      tunnelingOnFrameRenderedListener.destroyHandler(getCodec());
+    }
     tunnelingOnFrameRenderedListener = null;
     try {
       super.onDisabled();
@@ -687,6 +690,9 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
     }
     codec.configure(mediaFormat, surface, crypto, 0);
     if (Util.SDK_INT >= 23 && tunneling) {
+      if (tunnelingOnFrameRenderedListener != null) {
+        tunnelingOnFrameRenderedListener.destroyHandler(codec);
+      }
       tunnelingOnFrameRenderedListener = new OnFrameRenderedListenerV23(codec);
     }
   }
@@ -1228,6 +1234,9 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
       MediaCodec codec = getCodec();
       // If codec is null then the listener will be instantiated in configureCodec.
       if (codec != null) {
+        if (tunnelingOnFrameRenderedListener != null) {
+          tunnelingOnFrameRenderedListener.destroyHandler(codec);
+        }
         tunnelingOnFrameRenderedListener = new OnFrameRenderedListenerV23(codec);
       }
     }
@@ -1808,8 +1817,23 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
   @TargetApi(23)
   private final class OnFrameRenderedListenerV23 implements MediaCodec.OnFrameRenderedListener {
 
+    private final Handler handler;
+
     private OnFrameRenderedListenerV23(MediaCodec codec) {
-      codec.setOnFrameRenderedListener(/* listener= */ this, Util.createHandler());
+      handler = new Handler();
+      codec.setOnFrameRenderedListener(/* listener= */ this, handler);
+    }
+
+    /**
+     * Cleanup any messages pending for this listener before removing references to it
+     *
+     * @param codec optional, if the codec is still around it should be passed here
+     */
+    void destroyHandler(@Nullable MediaCodec codec) {
+      if (codec != null) {
+        codec.setOnFrameRenderedListener(null, null);
+      }
+      handler.removeCallbacksAndMessages(null);
     }
 
     @Override
@@ -1818,12 +1842,32 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
         // Stale event.
         return;
       }
+
+      // Work around bug in MediaCodec that causes deadlocks if you call back a
+      // MediaCodec API that requires MediaCodec to use it's Looper, as it calls this
+      // listener method holding locks.  This was fixed in:
+      //  https://android-review.googlesource.com/1156807
+      //
+      // The work around simply queues the processing to a subsequent message, where
+      // the lock will not be held.
+      //
+      if (Util.SDK_INT < 30) {
+        handler.post(() -> {
+          if (this == tunnelingOnFrameRenderedListener) {   // event not stale
+            handleFrameRendered(presentationTimeUs);
+          }
+        });
+      } else {
+        handleFrameRendered(presentationTimeUs);
+      }
+    }
+
+    private void handleFrameRendered(long presentationTimeUs) {
       if (presentationTimeUs == TUNNELING_EOS_PRESENTATION_TIME_US) {
         onProcessedTunneledEndOfStream();
       } else {
         onProcessedTunneledBuffer(presentationTimeUs);
       }
     }
-
   }
 }
--- a/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
@@ -158,7 +158,7 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
 
   private boolean tunneling;
   private int tunnelingAudioSessionId;
-  /* package */ OnFrameRenderedListenerV23 tunnelingOnFrameRenderedListener;
+  /* package */ @Nullable OnFrameRenderedListenerV23 tunnelingOnFrameRenderedListener;
 
   private long lastInputTimeUs;
   private long outputStreamOffsetUs;
@@ -563,10 +563,7 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
     clearReportedVideoSize();
     clearRenderedFirstFrame();
     frameReleaseTimeHelper.disable();
-    if (tunnelingOnFrameRenderedListener != null) {
-      tunnelingOnFrameRenderedListener.destroyHandler(getCodec());
-    }
-    tunnelingOnFrameRenderedListener = null;
+    destroyOnFrameRenderListener();
     try {
       super.onDisabled();
     } finally {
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1750,7 +1753,8 @@ public void onFrameRendered(MediaCodec codec, long presentationTimeUs, long nano
         // Stale event.
         return;
       }
-      onProcessedTunneledBuffer(presentationTimeUs);
+
+      handler.post(() -> onProcessedTunneledBuffer(presentationTimeUs));
     }
```
stevemayhew, line 1884: Might be good for MediaCodecRenderer to have an actual Handler that we can post messages too, in the event there are other things like this...   Note, some concern this could queue lots of messages as well.

The Handler is of course on the player's thread so it shares the message queue with `ExoPlayerImplInternal`

Everything else but the code on line 938 above is safe to do in `onProcessTunnelBuffer()` in this same handler `dispatchMessage()`.   It's the calls back into `MediaCodec` that need to be avoided (deferred to a subsequent `Message` that doesn't hold MediaCodec locks.

tonihei:
Kevin, could you have a look at this PR to see if it makes sense?

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1740,8 +1740,11 @@ public CodecMaxValues(int width, int height, int inputSize) {
   @TargetApi(23)
   private final class OnFrameRenderedListenerV23 implements MediaCodec.OnFrameRenderedListener {
 
+    protected Handler handler;
```
krocard: private final

tobyworks:
Just leaving a comment here because i want to follow this PR. Currently we are being bothered with some sporadic ANR's. And i very well believe this could fix it.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1750,7 +1753,8 @@ public void onFrameRendered(MediaCodec codec, long presentationTimeUs, long nano
         // Stale event.
         return;
       }
```
krocard, line 1865: This stall event detection needs to be done in the lambda.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1750,7 +1753,8 @@ public void onFrameRendered(MediaCodec codec, long presentationTimeUs, long nano
...
-      onProcessedTunneledBuffer(presentationTimeUs);
+
+      handler.post(() -> onProcessedTunneledBuffer(presentationTimeUs));
```
krocard: A comment should be added that the codec is not reentrant from a callback.

krocard (changes requested):
The CL needs to be rebased.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
tonihei: This is also fixed in the platform code now. So this workaround can be behind an "if(Util.SDK_INT < 30)" to avoid the post on upcoming Android versions.

krocard: For reference, it was fixed by https://android-review.googlesource.com/1156807.

stevemayhew:
That’s validating...  can add the link to that in a comment so we can
remove the workaround when the older versions of Android finally age our

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
krocard: You have marked this done, have you have not change that code. Have you forgotten to upload a patch?

krocard (changes requested):

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -506,6 +506,9 @@ protected void onDisabled() {
     clearReportedVideoSize();
     clearRenderedFirstFrame();
     frameReleaseTimeHelper.disable();
+    if (tunnelingOnFrameRenderedListener != null && getCodec() != null) {
+      tunnelingOnFrameRenderedListener.destroyHandler(getCodec());
+    }
     tunnelingOnFrameRenderedListener = null;
     try {
```
stevemayhew: Explicitly cleanup the previous handler, and any messages it might have that keep references to the handler itself and implicitly to the outer class.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1740,8 +1749,23 @@ public CodecMaxValues(int width, int height, int inputSize) {
...
+     * Cleanup any messages pending for this listener before removing references to it
+     *
+     * @param codec optional, if the codec is still around it should be passed here
+     */
+    void destroyHandler(@Nullable MediaCodec codec) {
+      if (codec != null) {
+        codec.setOnFrameRenderedListener(null, null);
+      }
+      handler.removeCallbacksAndMessages(null);
     }
```
stevemayhew: I added in the stale check, however this code will prevent you from ever getting messages after the handler reference is dropped (assuming you call it first).

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1750,7 +1774,24 @@ public void onFrameRendered(MediaCodec codec, long presentationTimeUs, long nano
...
+      // MediaCodec API that requires MediaCodec to use it's Looper, as it calls this
+      // listener method holding locks.  This was fixed in:
+      //  https://android-review.googlesource.com/1156807
+      //
+      // The work around simply queues the processing to a subsequent message, where
+      // the lock will not be held.
+      //
+      if (Util.SDK_INT < 30) {
+        handler.post(() -> {
+          if (this == tunnelingOnFrameRenderedListener) {   // event not stale
```
stevemayhew, line 1877: Here's the stale detect, and comment explaining why we have to do this at all.  I'll rebase in the next commit and pick up the changes from 5579cc89c

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
stevemayhew: @krocard I've put the check for listener changed (stale event), but also cleaned up the handler to avoid leaks and prevent the stale event from ever happening...  If you agree we can remove the stale check, but kind of hard to ever test.

stevemayhew (commented):
We are running with this change in our released code now, will of course let you know of any problems we have.  We have stopped seeing the ANR :-)

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -513,6 +513,9 @@ protected void onDisabled() {
     clearReportedVideoSize();
     clearRenderedFirstFrame();
     frameReleaseTimeHelper.disable();
+    if (tunnelingOnFrameRenderedListener != null && getCodec() != null) {
```
bjitivo: Probably should call destroyHandler even if codec is null ...

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1172,6 +1178,9 @@ private void clearRenderedFirstFrame() {
       MediaCodec codec = getCodec();
       // If codec is null then the listener will be instantiated in configureCodec.
       if (codec != null) {
+        if (tunnelingOnFrameRenderedListener != null) {
```
bjitivo: Have to do the destroy unconditionally to protect against a queued render that will run and see the new listener.

bjitivo: I mean do the destroy whether or not there is a codec.

stevemayhew: Thanks Bryan, made this fix in commit 43700ea

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -1172,6 +1178,9 @@ private void clearRenderedFirstFrame() {
...
+          tunnelingOnFrameRenderedListener.destroyHandler(codec);
+        }
         tunnelingOnFrameRenderedListener = new OnFrameRenderedListenerV23(codec);
```
krocard: If there is a listener installed, this leads to 2 calls to calls to `codec.setOnFrameRenderedListener`. One with null, the next one with the new listener.
This leads to disabling codec timestamp reporting then reenabling it immediately. Which is a costly operation as far as I can tell.
Thus I think that destoryHandler should not setOnFrameRenderedListener.
How about creating a function to install a new Listener:
```
MediaCodec.OnFrameRenderedListener oldListener = tunnelingOnFrameRenderedListener;
tunnelingOnFrameRenderedListener = new OnFrameRenderedListenerV23(codec);
if (oldListener != null) {
  oldListener.destroyHandler()
}
```
Maybe add a OnFrameRenderedListenerV23.disable method to codec.setOnFrameRenderedListener(null, null); in disable(). Although I'm not sure it is necessary.

krocard (changes requested):

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
krocard: If you prefer, I can merge your PR as is and do a patch with this optimization over it.

krocard: @stevemayhew, let me know if you would prefer do this change yourself or I can go along and do it on top of yours. We would like to merge your PR. Thanks

stevemayhew: Removing the old handler is just an optimization.  The check for stale events is sufficient.  I will add the function as you describe and the disable() function but not call it for now with a nite with your comments as to why.  Should have this today for you to review

stevemayhew: This commit 72f2e7d4fa6a29aa0cf6bc3a1bf2a8738b4f8ac0 adds the method and removes the double call to `codec.setOnFrameRenderedListener` if the codec is not being disabled.

krocard (approved):

koAlech:
We're also getting these ANRs in production. Any estimate when this PR be merged?

krocard:
Sorry for the delay, I am working on it right now. A few changes were suggested by the team but this PR will be part of the next minor release.
Note however that this change only solves a deadlock in tunneling mode during `OnFrameRenderedListener.onFrameRendered` calls.
Other deadlocks have been observed during `Player.release()` and are not solved by this PR.

stevemayhew:
Pretty special edge case solved by this pull, you must be running codec in
tunneling mode to see it.

Best to provide a bugreport with the ANR. One of the ExoPlayer devs might
no best if the signature is unique then

PR #6650, opened 2019-11-14
Detect Dolby Vision profile 7

stanionascu:
In official documentation dvProfile 7 uses dvhe as the codec type.


stanionascu:
@googlebot I signed it!

Issue #6676, opened 2019-11-20
Improve support for multiple metadata tracks


phhusson:
### [REQUIRED] Searched documentation and issues
I've browsed https://github.com/google/exoplayer/issues?utf8=%E2%9C%93&q=is%3Aissue+metadata

### [REQUIRED] Question
As far as I understand, ExoPlayer currently supports only one metadata track at the same time.
My usecase is to have EIT and TDT DVB tracks (respectively EPG and current time) reported back to the application through `onMetadata`. So what I want is to select all metadata tracks, instantiate related decoders (based on various mime-types), and have all those metadata sent to a single MetadataOutput.
I don't know if there are cases where we don't want all tracks (there is the case when there are multiple PMTs in the stream, but that's already not well handled by ExoPlayer anyway)
EIT and TDT are just examples, DVB has many other metadata on various tracks. (AIT, BAT, NIT, ...)

What is the recommendation to do that?
Currently i'm thinking of instantiating multiple MetadataRenderer (though it needs to be a constant number, that'll need to be magically set), and changing selectTracks to allow assigning multiple renderers, but that's not really clean...

AquilesCanta:
As things are now you'll need one Metadata renderer per metadata track you want to handle. The number of renderers depends on the formats you want to handle. For example, if you need to handle ID3 and SCTE35, you'll need to inject two renderers. You will also need to implement track selection so that each of the tracks gets mapped to a different metadata renderer.

In the case of PSI and TS tables in general, I think you can just inject custom payload readers (See [`TsPayloadReader.Factory`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.html)) but keep in mind that those payload readers run on a worker thread, so keep cross-thread communication in mind. 

I'll use this issue to track better support for multiple metadata tracks.

tonihei:
For the point of using multiple renderers and making it simpler to set them up: There is also #6589 where this question has been asked for audio renderers and #6652 for video renderers (duped to the other one).

tonihei:
This requested enhancement was implemented by #6678, so closing this issue. If there is something left to implement, please file a new issue for tracking. Thanks!

Issue #6589, opened 2019-10-27
Simplify/allow selection of multiple renderers of the same type in DefaultTrackSelector.


mrj976:
I want play multi sound in same time(simultaneously) in exoplayer but it only play mediaSource1 and mediaSource2. I want play all mediaSources at the same time(simultaneously) 

My code with exoplayer is:

> 
     Uri uri1 = Uri.parse(videopath);
     Uri uri2 = Uri.parse(audio_a.getAbsolutePath());
     Uri uri3 = Uri.parse(audio_b.getAbsolutePath());
     Uri uri4 = Uri.parse(audio_c.getAbsolutePath());

    MediaSource mediaSource1 = buildMediaSource(uri1);
    MediaSource mediaSource2 = buildMediaSource(uri2);
    MediaSource mediaSource3 = buildMediaSource(uri3);
    MediaSource mediaSource4 = buildMediaSource(uri4);

    MergingMediaSource mediaSource = new MergingMediaSource(mediaSource1, mediaSource2, mediaSource3, mediaSource4);

    exoPlayer.prepare(mediaSource, false, false);

> 

please help me
thanks

tonihei:
Multiple streams with the same type of media require multiple renderers to play them at the same time. By default, ExoPlayer's `DefaultRenderersFactory` provides one `MediaCodecAudioRenderer` and potentially some extension audio renderers. So it's actually surprising that two of the audio files play in parallel already.

You could in theory provide multiple `MediaCodecAudioRenderers` in a `RenderersFactory` you pass in to `ExoPlayerFactory`. However, audio renderers also function as the media clock and the player fails if multiple media clocks are provided. You can circumvent this problem by subclassing `MediaCodecAudioRenderer` and returning null from `getMediaClock`. I haven't tested this, but this would probably allow the parallel playback. Note however, that there won't be any time synchronization between the renderers and the audio files may drift apart.

tonihei:
If your app is doing nothing else with ExoPlayer and all the files are local, you can also consider using [`SoundPool`](https://developer.android.com/reference/android/media/SoundPool.html) without any ExoPlayer integration.

mrj976:
> Multiple streams with the same type of media require multiple renderers to play them at the same time. By default, ExoPlayer's `DefaultRenderersFactory` provides one `MediaCodecAudioRenderer` and potentially some extension audio renderers. So it's actually surprising that two of the audio files play in parallel already.
> 
[...]

PR #6678, opened 2019-11-20
Allow multiple metadata tracks, assuming multiple metadata renderers

phhusson:
Test done:
- Extend DefaultRenderersFactory to return 3 MetadataRenderer
- in TsExtractor create two metadata tracks
- Check in `onMetadata` that metadata from both tracks are received

My test has two metadata tracks, 4 audio tracks, two audio renderers (mediacodec & ffmpeg).
The result is that two metadata tracks have been selected, the third metadatarenderer doesn't select any track, only one audio renderer has been enabled, and has selected a track.

Cf bug #6676


AquilesCanta:
Assigning @tonihei in the context of #6589. Please feel free to reassign if you don't think it's relevant.

tonihei:
Thanks for the pull request!  And sorry for the review delay :) 

It looks as if we can merge this change (probably with minor modifications).

Issue #2863, opened 2017-05-25
Expose peak bitrate in Format when available (as well as average bitrate)


juechemparathy:
Apple spec talks about having average bandwidth and peak bandwidth in master playlist as below -
'You should include the average bit rate as well as the peak bit rate. This is especially important if your peak is more than 20% different from your average. Specifying the average allows the client to make more intelligent decisions about which variant to play.'
Ref - 
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#page-28
https://developer.apple.com/library/content/technotes/tn2224/_index.html#//apple_ref/doc/uid/DTS40009745-CH1-BITRATERECOMMENDATIONS

Exoplayer seems to be expecting only average Bandwidth in the master playlist as the regex in HLSplaylistParser is "BANDWIDTH=(\d+)\b". 

For supporting VBR we need a) regex fix to extract average bitrate and peak bitrate, b) data holder in Format object for peak bitrate

Sample snippet from a master .m3u8 file - 
#EXT-X-STREAM-INF:PROGRAM-ID=1,**BANDWIDTH**=2095000,SUBTITLES="subs",**AVERAGE-BANDWIDTH**=1945000,RESOLUTION=1280x720

I am looking at Exo v2.3.1.
Thanks, _jue

ojw28:
The change above updates ExoPlayer to use `AVERAGE-BITRATE` to populate `Format.bitrate` when it's available, since that's what `Format.bitrate` is documented to contain.

It's somewhat unclear to me that there's a need to also include the peak bitrate in the `Format` object. Unless people are going to write ABR algorithms that make use of it, the need to include it is pretty hypothetical. If you do have a concrete near term plan to make use of it, please could you clarify how you intend to use it (particularly given you don't know where the peaks are).

juechemparathy:
@ojw28 - Yes, we intend to use it in our custom ABR algorithm(esp with fmp4) and is also considered as a useful input in our VBR instrumentation data. Thanks.

NayaneshGupte:
@ojw28 any idea when is this planned tentatively ?

ojw28:
[...]

PR #6706, opened 2019-11-27
Expose `peakBitrate` property in `Format`

julioz:
Fixes https://github.com/google/ExoPlayer/issues/2863

Exposes `peakBitrate` value when both `BANDWIDTH` and `AVERAGE-BANDWIDTH` attributes are available.
The `bitrate` property will always carry the average bandwidth of the `Format` instance, while `peakBitrate`
will carry the value for the respective attribute, when present; according to
tools.ietf.org/html/draft-pantos-http-live-streaming-23#page-28


library/hls/src/test/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylistParserTest.java
```diff
@@ -236,14 +236,17 @@ public void testParseMasterPlaylist() throws IOException {
...
     HlsMasterPlaylist masterPlaylist =
         parseMasterPlaylist(PLAYLIST_URI, PLAYLIST_WITH_AVG_BANDWIDTH);
 
     List<HlsMasterPlaylist.Variant> variants = masterPlaylist.variants;
 
     assertThat(variants.get(0).format.bitrate).isEqualTo(1280000);
+    assertThat(variants.get(0).format.peakBitrate).isEqualTo(Format.NO_VALUE);
+
     assertThat(variants.get(1).format.bitrate).isEqualTo(1270000);
+    assertThat(variants.get(1).format.peakBitrate).isEqualTo(1280000);
```
julioz, line 249: This is the example case for the presence of the attribute.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/playlist/HlsMasterPlaylistParserTest.java
```diff
@@ -236,14 +236,17 @@ public void testParseMasterPlaylist() throws IOException {
   }
 
   @Test
-  public void testMasterPlaylistWithBandwdithAverage() throws IOException {
+  public void testMasterPlaylistWithBandwidthAverage() throws IOException {
```
julioz, line 239: Fixed typo.

julioz:
When running locally, some `Extractor` tests fail; I can't figure out why. I assume it's one of the (innumerous) `Format` constructors (or factory methods) that take in `bitrate` and didn't take in `peakBitrate`. Please advise before we merge this (or; feel free to fix and commit on top of this branch).

icbaker:
Mind taking a look Santiago?

AquilesCanta:
I think this is for @ojw28 to look at.

AquilesCanta:
On my side, I think it's unfortunate we are using the `Format.bitrate` field for both `BANDWIDTH` and `AVERAGE-BANDWIDTH`. Adding the peak bitrate adds a bit of confusion to the mix.

Aside, I think it would be helpful for us to know why/how having both average and peak bitrates could be of use.

julioz:
@AquilesCanta Please check the referred issue #2863 for context.

AquilesCanta:
My bad, didn't see the link. However, the referenced issue makes the same question I made above. And an answer is still not available. It's worth looking into the issue first.

EDIT: @ojw28 removed the need more info tag from the issue, so I'm assuming he understands the use case. Thanks for your patience!

tonihei:
I think the peak bitrate is preferable for adaptive track selection to be on the safe side, especially when the peak is much higher than the average. Otherwise, the ABR algorithm may run into issues during complex scenes.  

So, if available, we should probably just populate the `Format` bitrate by BANDWIDTH in all cases?

julioz:
@tonihei That is for sure. The `BANDWIDTH` attribute is [_mandatory_](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#page-28), while the `AVERAGE-BANDWIDTH` attribute is optional.
The change in https://github.com/google/ExoPlayer/commit/2795269daee2270fc2bda1512cefd86ef021f905 exposes the `AVERAGE-BANDWIDTH` value overriding the `BANDWIDTH` one, but that means we _hide away_ the latter. For advanced ABR algorithms that might need to introspect into that value, this becomes an issue; which this PR fixes.

Let me know if hiding one of the values is _by design_. If that's the case, we should document it publicly somewhere, I assume others might hit the same limitation in the future.

AquilesCanta:
To provide some context: Back when AVERAGE-BANDWIDTH was added to the HLS spec, preferring AVERAGE-BANDWIDTH (which is option) over BANDWIDTH (which is mandatory) was a design decision. @tonihei, please feel free to take whichever approach you think is most suitable for ABR.

ojw28:
> I think the peak bitrate is preferable for adaptive track selection to be on the safe side, especially when the peak is much higher than the average. Otherwise, the ABR algorithm may run into issues during complex scenes.

Presumably using both is preferable, else there wouldn't be value in them both existing. I agree that if we were to use just one, then peak bitrate is preferable, and so https://github.com/google/ExoPlayer/commit/53b52e784c2a7902b634348a2edfe81306318770 was probably a mistake.

I think we should expose both in `Format`, and use peak bitrate for ABR decisions for as long as we're only using one of them. I can sort this out whilst merging this pull request. I will separately propose a small fix to use peak bitrate ahead of merging the whole change.

ojw28:
For what it's worth, the DASH specification gives a more nuanced definition of bitrate, that seems much more useful than either "peak" or "average". I guess it's closer to "peak" in practice:

> Consider a hypothetical constant bitrate channel of bandwidth with the value of this attribute in bits per second (bps). Then, if the Representation is continuously delivered at this bitrate, starting at any SAP that is indicated either by `@startWithSAP` or by any Segment Index box, a client can be assured of having enough data for continuous playout providing playout begins after `@minBufferTime * @bandwidth` bits have been received (i.e. at time `@minBufferTime` after the first bit is received).

Meanwhile the [HLS authoring specification](https://developer.apple.com/documentation/http_live_streaming/hls_authoring_specification_for_apple_devices) seems to allow you to specify average and peak bitrates that are wrong by up to 10%, whilst referencing another section that doesn't exist. Why?

> 1.26. For VOD content, the average segment bit rate MUST be within 10% of the AVERAGE-BANDWIDTH attribute. (See Declared versus measured values of bandwidths.)
> 1.27. For VOD content, the measured peak bit rate MUST be within 10% of the BANDWIDTH attribute.

julioz:
@ojw28 @marcbaechinger It seems like we have opted for exposing the peak bitrate when available, super-seeding the avg bitrate data, which hinders more advanced ABR algorithms.
As this PR starts to get conflicts with `dev-v2`, I'm tempted to close it. As a side-note, it might be worth extracting some of the constructor calls with default values from many test suites into a test fixture class that can provide sensible values and still builders for further specialization if need be in tests. I didn't do it as part of this PR since I couldn't find other fixture classes in the codebase, but would gladly introduce this pattern as it has proven to be useful for us at SoundCloud.

Let me know what you think.

ojw28:
Please leave it open. We'll resolve the conflict when we merge. Thanks!

ojw28:
We need to make some bigger changes to land this, which we're working on now. These changes will land separately of this pull request, so I'll go ahead and close it. Please follow the tracking bug for updates. Thanks!

PR #6724, opened 2019-12-04
Introduce edgeLayout to keep outline color

nnoury:
Fix outline color style in subtitles

When a ForegroundColorSpan changes the foreground color, it is also applied
to the outline painter. In order to keep the correct color, one needs to
filter out theses span. We do this with a new cue that is our text source
for the specific edgeLayout.

Test: Manual check - Subtitle view can show custom color subtitles from specific Subtitle
Renderer and outline is shown correctly using user defined color.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitlePainter.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitlePainter.java
@@ -33,6 +33,7 @@ import android.text.TextPaint;
 import android.text.TextUtils;
 import android.text.style.AbsoluteSizeSpan;
 import android.text.style.BackgroundColorSpan;
+import android.text.style.ForegroundColorSpan;
 import android.text.style.RelativeSizeSpan;
 import android.util.DisplayMetrics;
 import androidx.annotation.Nullable;
@@ -98,6 +99,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 
   // Derived drawing variables.
   private @MonotonicNonNull StaticLayout textLayout;
+  private @MonotonicNonNull StaticLayout edgeLayout;
   private int textLeft;
   private int textTop;
   private int textPaddingX;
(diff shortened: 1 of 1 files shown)


nnoury:
@googlebot I signed it!

ojw28:
@icbaker - Please take a look. Thanks!

icbaker:
Thanks for the contribution!

I'm afraid I'm not 100% clear what bug/problem you're solving here - would it be possible to provide repro instructions and/or test media that demonstrates the problem?

You can email test media to dev.exoplayer@gmail.com if you're unable to share it directly on github - make sure the subject is `Issue #6724`.

nnoury:
To reproduce, and see that the outline is painted with the foreground color:

either manually change user captionning style
Launch Captionning manager,
``` am start -a android.settings.CAPTIONING_SETTINGS ```

switch caption style to custom (last row radio button),
then select Edge Type -> Outline
on tv, enable then disable Show background to workaround an initialization value bug in leanback, on phones select no background
then back back back

or you can programmatically do (my codebase is in Kotlin)
```
    val captionStyle = CaptionStyleCompat(Color.WHITE, Color.TRANSPARENT, Color.TRANSPARENT,
            EDGE_TYPE_OUTLINE, Color.BLACK, Typeface.DEFAULT)
    subtitleView.setStyle(captionStyle)
```
in library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java:449

Test in exoplayer demo in Subtitles -> WebVTT with the following patch applied:

```
diff --git a/demos/main/src/main/assets/media.exolist.json b/demos/main/src/main/assets/media.exolist.json
index 06f063b1c..2faba848b 100644
--- a/demos/main/src/main/assets/media.exolist.json
+++ b/demos/main/src/main/assets/media.exolist.json
@@ -601,6 +601,13 @@
   {
     "name": "Subtitles",
     "samples": [
+        {
+        "name": "WebVTT",
+        "uri": "https://nnoury.github.io/sintel-short.mp4",
+        "subtitle_uri": "https://nnoury.github.io/sintel-en.vtt",
+        "subtitle_mime_type": "text/vtt",
+      },
       {
         "name": "TTML",
         "uri": "https://html5demos.com/assets/dizzy.mp4",
```

**Comment by nnoury** [comment]
Before apply MR:
Screenshot_1575988264
After:
Screenshot_1575988331

**Comment by icbaker** [comment]
Thanks for the clear repro instructions & screenshots - I can see the problem you're fixing here.

I think what we've got so far isn't quite right - or rather doesn't interact well with existing decisions. With your change and a non-transparent `backgroundColor` It seems the background is drawn above the outline & below the text - you can see in the screenshot a bit of the w's edge sticking out the side of the box. Seems like the background should be **behind** the outline?

```java
CaptionStyleCompat outlineWithBackgroundNoWindow =
          new CaptionStyleCompat(
              /* foregroundColor= */ Color.WHITE,
              /* backgroundColor= */ Color.CYAN,
              /* windowColor= */ Color.TRANSPARENT,
              CaptionStyleCompat.EDGE_TYPE_OUTLINE,
              /* edgeColor= */ Color.BLACK,
              /* typeface= */ null);
subtitleView.setStyle(outlineWithBackgroundNoWindow);
```

Screenshot_20191211-165827

Can you have a go at seeing if you can get the edge to appear in front of the background?

I think it's happening because the `backgroundColor` is rendered using a `BackgroundColorSpan` on `cueText`. I guess this should be on `cueTextEdge` (only) if `edgeType` is `OUTLINE` or `RAISED`.

This might be a bit tricky in the current code layout - if there's an obvious way to reshuffle things a bit to make it much better feel free to have a go.

Arctic2333 (approved):

ojw28:
Removing labels that might cause the bot to close this. @nnoury - Are you going to be able to take a look at the comments from @icbaker above?

phhusson:
Quick update: we initially thought that the issue raised would need to add yet another painter, and that it would raise other issues, but we checked again and now it seems pretty straightforward.

We'll try reporting back soon.

nnoury:
@icbaker @phhusson and I rewrote the background color logic to take into account various painter configurations when using outline : one apply the background only to the first pass painter when the painting is done in two passes

Issue #6798, opened 2019-12-23
Stuck downloads when changing network conditions


daveoxley:
### Issue description
DownloadService race condition causing downloads to not resume when network requirements are met after previously not being met.

The issue appears to be that `DownloadManagerHelper.onDownloadChanged` is called before the new service is attached to the `DownloadManagerHelper`. Therefore `DownloadService.notifyDownloadChanged` is never called and `startForeground` is never called on the new service. The app process is then killed after roughly 10 seconds as the scheduler has been cancelled.

### Reproduction steps
Using the exoplayer app:
- Start downloading some content
- Click back to exit the app. Downloading notification should be shown.
- Turn off network. Downloading notification should be hidden.
- Wait approximately 2 minutes.
- Turn on network. Downloading notification is not reshown when it should be.

### Link to test content
Tested using `WV: Clear SD & HD (MP4,H264)` in the sample app.

### Version of ExoPlayer being used
2.10.7. Also reproduced the issue on 2.11.1.

### Device(s) and version(s) of Android being used
Pixel 3XL Android 10 (QQ1A.191205.008)

ojw28:
This is probably a duplicate of https://github.com/google/ExoPlayer/issues/6733. Marking as such, but please do test with the `dev-v2` branch to confirm that the issue is fixed.

daveoxley:
@ojw28 Unfortunately this issue still exists on the latest dev-v2 branch (Tested on 826083db923f26f2000ca42f7b54b9531726d713). The below patch fixes the issue and I'll put together a better change and submit a pr later. Any chance you could reopen this issue please?

```
diff --git a/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java b/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
index db10517b6..8c4220c35 100644
--- a/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadService.java
@@ -32,6 +32,8 @@ import com.google.android.exoplayer2.util.Assertions;
 import com.google.android.exoplayer2.util.Log;
 import com.google.android.exoplayer2.util.NotificationUtil;
 import com.google.android.exoplayer2.util.Util;
+
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 
@@ -856,6 +858,7 @@ public abstract class DownloadService extends Service {
[...]

PR #6828, opened 2020-01-05
Requeue MSG_DOWNLOAD_UPDATE when received before the service is started.

daveoxley:
This fixes the case where downloads get stuck when network requirements are met after previously not being met. It resends the update messages once the service is attached to the helper. Also added logging when this occurs:
```
2020-01-05 11:55:45.391 3204-3204/com.google.android.exoplayer2.demo D/DownloadService: onDownloadChanged with null service. Queuing to resend message later.
2020-01-05 11:55:45.415 3204-3204/com.google.android.exoplayer2.demo D/DownloadService: 1 messages to resend.
```

Fixes https://github.com/google/ExoPlayer/issues/6798

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadManager.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/offline/DownloadManager.java
@@ -484,6 +484,13 @@ public final class DownloadManager {
     }
   }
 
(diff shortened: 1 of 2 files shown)


ojw28:
We fixed the ref'd issue with a different set of changes, so this is no longer required.

PR #6836, opened 2020-01-06
Optimize AC-4 code and add related test cases

ybai001:
- Optimize Mp4Extractor for AC-4
- Optimize FragmentedMp4Extractor for AC-4
- Add test case for AC-4 in MP4
- Add test case for AC-4 in Fragmented MP4


ybai001:
As suggested in [this pull request](https://github.com/google/ExoPlayer/pull/6797), I submitted this new request at first.

library/core/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
```diff
@@ -42,4 +42,9 @@ public void testMp4SampleWithSlowMotionMetadata() throws Exception {
   public void testMp4SampleWithMdatTooLong() throws Exception {
     ExtractorAsserts.assertBehavior(Mp4Extractor::new, "mp4/sample_mdat_too_long.mp4");
   }
+
+  @Test
+  public void testMp4SampleWithAc4Track() throws Exception {
```
ojw28, line 47: This test fails. Interestingly, it fails with `Mp4Extractor` as it was before this change as well, so it's great these tests are being added! The problem is that `Mp4Extractor` doesn't recover properly from input read failures at certain positions, and ends up outputting incorrect samples. No need to modify the PR; I'll push a fix in as part of the merge.

ojw28:
We will get this merged in the next day or two. Thanks!

Issue #2953, opened 2017-06-14
TTML: Support referential styling


evandipietro:
### Issue description
**Observed**: MP4-Encapsulated TTML subtitles default to top position (with `positionAnchor` defaulting to Float.MIN_VALUE)
**Expected**: Subtitles appear at the bottom of the screen with a valid position

### Reproduction steps
Using a dash steam w/ MP4-Encapsulated TTML subtitles, enable subtitles. Subtitles appear at the top.

### Link to test content
Cannot provide (legal reasons)

### Version of ExoPlayer being used
2.4.1 (Was not an issue in 2.4.0)

### Device(s) and version(s) of Android being used
**Device:** FireTV  
**OS:** Fire OS v5.2.4.1 (Android v5.1.1)

ojw28:
Are you sure they're not TTML captions? I don't think we made any changes to WebVTT captioning between 2.4.0 and 2.4.1.

evandipietro:
@ojw28 My mistake, it is TTML. Updating the issue.

ojw28:
We fixed TTML positioning in 2.4.1 (at least in theory). It's quite possible that your TTML caption files are actually specifying that the captions should be rendered at the top of the screen (or that the implicitly specify this by not specifying something else), and that what the previous behavior was incorrect.

There's nothing we can do to look at this without test content (or at least the TTML part of it).

soutua:
We also encountered this problem with the ExoPlayer 2.4.1/2.4.2. The subtitles we use do not define the alignment so the subtitles are displayed at the top of the screen. Would it be possible to add a way to change the default position from top to bottom, in case the position isn't defined in the subtitles?

ojw28:
We need to see sample TTML files to investigate this properly.

zharf:
I extracted a snippet from our ttml stream (and swapped out the texts just in case) and attached it here. My initial investigation suggests that ExoPlayer doesn't support [(chained?) referential styling](https://www.w3.org/TR/ttaf1-dfxp/#semantics-style-association-referential) as described by TTML

sidenote: I work with @soutua 

[ttml.zip](https://github.com/google/ExoPlayer/files/1077720/ttml.zip)

ojw28:
@zharf - Yes, that looks like the problem for your case. Thanks for the sample. We should fix that, after which you'll get proper positioning as is defined in your TTML file.
[...]

PR #6860, opened 2020-01-10
Use previous TtmlRegion when no origin and extent are present on current TtmlRegion

xirac:
I have some DASH streams containing ttml subtitles with image bitmap where it's sometime flickering one the screen:
The subtitle appears shortly for a very short time, then disappear, and then appear again before being replaced by the next subtitle.
Each time I see this behaviour, I see the following logs into logcat:
`Ignoring region without an extent`
`Ignoring region without an origin`

I had a look into the code and saw that it's related to this opened issue:
https://github.com/google/ExoPlayer/issues/2953

I tried to uncomment the "TODO" in order to default line/position/width/height as already present in dev-v2 but it is leading to image subtitles being stretched on the whole screen.

I tried to reuse previously decoded TtmlRegion origin and extent and I get the image subtitle rendered on screen as expected.

Not sure this is the right way to fix the issue I'm experiencing.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlDecoder.java
@@ -91,6 +91,7 @@ public final class TtmlDecoder extends SimpleSubtitleDecoder {
       new CellResolution(/* columns= */ 32, /* rows= */ 15);
 
   private final XmlPullParserFactory xmlParserFactory;
+  private TtmlRegion previousTtmlRegion = null;
 
   public TtmlDecoder() {
     super("TtmlDecoder");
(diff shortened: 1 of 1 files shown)


icbaker:
Thanks for the contribution! It's not clear to me that using values from the previously-parsed region is in-line with the TTML spec though - unless I'm missing something? I'd expect the behaviour to be pretty independent of parsing order.
https://www.w3.org/TR/2018/REC-ttml2-20181108/

https://github.com/google/ExoPlayer/issues/2953 is about [referential styling](https://www.w3.org/TR/2018/REC-ttml2-20181108/#semantics-style-association-referential) - is your TTML file using this?

It may be worth filing a new 'bug' issue with a TTML file attached that reproduces your problem. Then I can take a look to understand a) why this PR fixes the problem for you and b) if there's a way to solve your issue that's compatible with the TTML spec.

xirac:
> Thanks for the contribution! It's not clear to me that using values from the previously-parsed region is in-line with the TTML spec though - unless I'm missing something? I'd expect the behaviour to be pretty independent of parsing order.
> https://www.w3.org/TR/2018/REC-ttml2-20181108/
> 
> #2953 is about [referential styling](https://www.w3.org/TR/2018/REC-ttml2-20181108/#semantics-style-association-referential) - is your TTML file using this?
> 
> It may be worth filing a new 'bug' issue with a TTML file attached that reproduces your problem. Then I can take a look to understand a) why this PR fixes the problem for you and b) if there's a way to solve your issue that's compatible with the TTML spec.

Thanks for your comment.
I'm still investigating at the TTML file side. I'll update or close this PR depending on my findings.

xirac:
Actually, the issue is on TTML file side. Closing this PR.

Issue #6853, opened 2020-01-08
Add error response field to InvalidResponseCodeException


chrisfillmore:
### [REQUIRED] Use case description
I would like to include the HTTP response body in an InvalidResponseCodeException, so that I can read the response.

In particular, our license server sends back specific error codes which we need to triage.

### Proposed solution
Add a new field to InvalidResponseCodeException which stores the value returned from `HttpURLConnection#getErrorStream()` (or string equivalent).

### Alternatives considered
We could make our license request without using Exo's APIs. But I think making the change in Exo would be useful for everyone, and backwards-compatible.

I can make a pull request, with your support.

ojw28:
https://github.com/google/ExoPlayer/issues/3920 was requesting this as well. 

@chrisfillmore - Is there a reason to use 403 rather than 200 for this case (see the issue ref'd above for more information)?
@jrocharodrigues - Did you implement this already, or did you do it differently?

chrisfillmore:
In our particular case we get HTTP 400 with an error code in the response body.

It's not feasible for us to change the HTTP status from the license proxy; this would have a far-reaching impact on other client platforms we support. But I also don't think that 200 is an appropriate status for this response. The server wants to indicate the request is bad, hence 400.

Is there a particular reason you'd prefer not to add this field? I'll admit, to me this seems a straightforward change with obvious benefit, without breaking anything. Am I missing something?

tonihei:
For the record, this feature request is already tracked by #5399, so I think this can be marked as a duplicate.

ojw28:
> Is there a particular reason you'd prefer not to add this field? I'll admit, to me this seems a straightforward change with obvious benefit, without breaking anything. Am I missing something?

Agreed it's not particularly difficult. There's a weird case in which there's an error reading the body though. What would you end up throwing in that case? To throw the `InvalidResponseCodeException` without the body in this case would be misleading, because calling code might be expecting the body to be present and you're not indicating that a failure occurred when reading it. Putting another exception inside the `InvalidResponseCodeException` to convey that failure would work, but is pretty weird and over complicated. Throwing the error that occurred when reading the body directly means you don't convey the response code to the caller in this case, but I suspect is probably the right thing to do.

jrocharodrigues:
@ojw28 Hi,
Yes i've implemented it. 
I've created a custom HttpDataSource with a custom HttpDataSourceException that includes the reponse code headers and responseBody.


```

  /**
[...]

PR #6861, opened 2020-01-10
Add responseBody field to InvalidResponseCodeException

chrisfillmore:
Issue #6853

Previously the body of an HTTP response was not included in an
InvalidResponseCodeException. This change adds a field to store
the response body, and updates DefaultHttpDataSource to populate
this value.

### DIFF
--- a/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
+++ b/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
@@ -443,16 +443,42 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
       throw new OpenException(new InterruptedIOException(e), dataSpec, Status.INVALID);
     }
 
+    // Calculate the content length.
+    if (!isCompressed(responseInfo)) {
+      if (dataSpec.length != C.LENGTH_UNSET) {
+        bytesRemaining = dataSpec.length;
+      } else {
+        bytesRemaining = getContentLength(responseInfo);
+      }
+    } else {
+      // If the response is compressed then the content length will be that of the compressed data
+      // which isn't what we want. Always use the dataSpec length in this case.
+      bytesRemaining = dataSpec.length;
+    }
+
     // Check for a valid response code.
     UrlResponseInfo responseInfo = Assertions.checkNotNull(this.responseInfo);
     int responseCode = responseInfo.getHttpStatusCode();
     if (responseCode < 200 || responseCode > 299) {
+      byte[] responseBody = null;
+
+      if (bytesRemaining > 0) {
+        int responseBodyLength = (int)bytesRemaining;
+        responseBody = new byte[responseBodyLength];
+        try {
+          readIntoByteArray(responseBody, 0, responseBodyLength);
+        } catch (HttpDataSourceException e) {
+          throw new OpenException(e, dataSpec, Status.INVALID);
+        }
+      }
+
       InvalidResponseCodeException exception =
           new InvalidResponseCodeException(
               responseCode,
               responseInfo.getHttpStatusText(),
               responseInfo.getAllHeaders(),
-              dataSpec);
+              dataSpec,
+              responseBody);
       if (responseCode == 416) {
         exception.initCause(new DataSourceException(DataSourceException.POSITION_OUT_OF_RANGE));
       }
@@ -474,19 +500,6 @@ public class CronetDataSource extends BaseDataSource implements HttpDataSource {
     // requested position.
     bytesToSkip = responseCode == 200 && dataSpec.position != 0 ? dataSpec.position : 0;
 
-    // Calculate the content length.
-    if (!isCompressed(responseInfo)) {
-      if (dataSpec.length != C.LENGTH_UNSET) {
-        bytesRemaining = dataSpec.length;
-      } else {
-        bytesRemaining = getContentLength(responseInfo);
-      }
-    } else {
-      // If the response is compressed then the content length will be that of the compressed data
-      // which isn't what we want. Always use the dataSpec length in this case.
-      bytesRemaining = dataSpec.length;
-    }
-
     opened = true;
     transferStarted(dataSpec);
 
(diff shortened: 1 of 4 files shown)


chrisfillmore:
I haven't done any testing yet. Also, looking at DefaultHttpDataSourceTest, I see there's only one test. Please advise if you would like me to write a test to cover this new behaviour. Thanks.

julioz:
Please update the extension DataSource's too ([OkHttp](https://github.com/google/ExoPlayer/blob/dev-v2/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java#L236), [Cronet](https://github.com/google/ExoPlayer/blob/dev-v2/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java#L451)) so other callers can also benefit from the ability to inspect the error response body.

chrisfillmore:
@julioz is this possible for CronetDataSource? It seems response body is not exposed by `UrlResponseInfo`. Do you have a recommendation here?

I can make the change for OkHttp. Thanks.

ojw28:
> @julioz is this possible for CronetDataSource? It seems response body is not exposed by UrlResponseInfo. Do you have a recommendation here?

I didn't try, but I'd guess you read it in the same way as the response body in the case of success. Try using `CronetDataSource.readInternal`, although note that you'll have to be sure to keep calling it until you've read everything.

julioz:
@chrisfillmore I haven't looked in detail, but I believe it is similar to how the success response is read via reading the byte stream.
Maybe @ojw28 can chime in and point the way, but at least by having the default and okhttp data sources exposing the body we're likely covering most use cases :+1:

chrisfillmore:
Here's what it would look like in Cronet:

```
if (responseCode < 200 || responseCode > 299) {
      ByteBuffer responseBuffer = ByteBuffer.allocateDirect((int)responseInfo.getReceivedByteCount());
      try {
        readInternal(responseBuffer);
      } catch (HttpDataSourceException e) {
        throw new OpenException(e, dataSpec, Status.INVALID);
      }

      InvalidResponseCodeException exception =
          new InvalidResponseCodeException(
              responseCode,
              responseInfo.getHttpStatusText(),
              responseInfo.getAllHeaders(),
              dataSpec,
              responseBuffer.array());
      
```
I am concerned about any side effects from `UrlRequestCallback` when calling `readInternal`.

Cronet is not (yet) one of our use cases. I'm inclined not to add this change to CronetDataSource at this time, since I don't think I can adequately certify that it works (at least, in the time I have).

@ojw28 would it be acceptable to add this to DefaultHttpDataSource and OkHttpDataSource, without CronetDataSource? For now anyway.

ojw28:
> Cronet is not (yet) one of our use cases. I'm inclined not to add this change to CronetDataSource at this time, since I don't think I can adequately certify that it works (at least, in the time I have). @ojw28 would it be acceptable to add this to DefaultHttpDataSource and OkHttpDataSource, without CronetDataSource? For now anyway.

If the field is being added then it seems quite confusing to not populate it for one implementation. So I think it does need doing. I assume you've done some local testing to see that your changes in the Default and OkHttp implementations work correctly. It should easy to swap in the Cronet implementation and do the same local testing with that one as well? If you could do that and add the Cronet change to the pull request, that would be great.

If your concern is more around subtle issues that local testing might not catch, then please include the change in the pull request regardless. All pull requests go through a fairly detailed internal code review process before they're merged. So whilst you should obviously do some local testing and not send something that just doesn't work at all, you can send something in the knowledge that there's an additional safety net that has a good chance of catching any remaining issues. We can also modify the change during that review process, including (in the worst case!) omitting the Cronet part if we're not confident either :).

chrisfillmore:
Sorry for the delay; I will be following up with this. I just had a few priorities to deal with this week, and I'll get back to it next week.

chrisfillmore:
@google-oss-bot please keep the issue open, thanks :)

(I was out sick most of last week, so sorry googoobot)

ojw28:
@marcbaechinger - Could we exclude Pull Requests from the Google OSS Bot? In most cases it's unwanted, and in any case the volume of PRs is low enough we can trivially do it manually.

chrisfillmore:
I have confirmed, through local testing, that DefaultHttpDataSource and OkHttpDataSource work as expected. Cronet testing is ongoing; not immediately working as I expect, so I will take another look tomorrow.

Also I see I have conflicts now. I will resolve those too.

chrisfillmore:
I'm on vacation for a week as of tomorrow. I resolved the merge conflicts but I'm still working on Cronet. I'll get back to this next week.

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -443,16 +443,42 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
       throw new OpenException(new InterruptedIOException(e), dataSpec, Status.INVALID);
     }
 
+    // Calculate the content length.
+    if (!isCompressed(responseInfo)) {
```
chrisfillmore, line 447: This block was moved up here so that we could know the length of the response. Besides being moved, the block is unchanged.

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -568,6 +581,43 @@ public int read(byte[] buffer, int offset, int readLength) throws HttpDataSource
...
+
+  /** Returns current {@link UrlResponseInfo}. May be null if the data source is not opened. */
+  @Nullable
+  protected UrlResponseInfo getCurrentUrlResponseInfo() {
+    return responseInfo;
+  }
+
+  // Internal methods.
+
+  private int readIntoByteBuffer(ByteBuffer buffer) throws HttpDataSourceException {
```
chrisfillmore: This diff kind of sucks.

~None of the above methods have changed (except `read(ByteBuffer buffer)`. The change here is that there is a new private method called `readIntoByteBuffer(ByteBuffer buffer)`, which is used in two places: `read(ByteBuffer buffer)` and in `open()` (to read the error response body).~

~`readIntoByteBuffer(ByteBuffer buffer)` is the same as `read(ByteBuffer buffer)` used to be, except without the assertion that the request has been opened.~

Disregard the above, I've made an update which makes more sense, I think. See new comments.

chrisfillmore:
I have added logic to CronetDataSource to read the response body. Please review. Thanks!

extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java
```diff
@@ -682,6 +651,50 @@ protected UrlResponseInfo getCurrentUrlResponseInfo() {
 
   // Internal methods.
 
+  private int readIntoByteArray(byte[] buffer, int offset, int readLength) throws HttpDataSourceException {
```
chrisfillmore, line 658: This new method is just the body of `read(byte[] buffer, int offset, int readLength)`, without the assertion, which remains in the public read method.

This is now called from `open()` to get the response body.

icbaker:
Thanks for the contribution!

> Also, looking at DefaultHttpDataSourceTest, I see there's only one test. Please advise if you would like me to write a test to cover this new behaviour. Thanks.

Adding some tests for this would be great! I think it probably makes sense to add a new test method to `DefaultHttpDataSourceTest`, the existing method in there is already doing too much imo. Something like `DefaultHttpDataSourceTest#open_invalidResponseCode()` and just assert that all the fields in the `InvalidResponseCodeException` are as you'd expect. Same for `OkHttpDataSourceTest`.

For `CronetDataSourceTest` you can probably extend `testRequestOpenValidatesStatusCode` rather than writing new method. In fact it would be great if you could take a look at this method anyway since when I patch this PR & run it the test hangs forever, blocked on this line:
https://github.com/google/ExoPlayer/blob/ab21f885d42209d0c65a70fc93dee26c96af6f84/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java#L761

I haven't spent much time getting to the bottom of what's going on - but it's probably useful to be aware that `ConditionVariable#block(int)` doesn't timeout correctly under Robolectric (both `com.google.android.exoplayer2.util.ConditionVariable` and `android.os.ConditionVariable` because both rely on `android.os.SystemClock.elapsedRealtime()` which doesn't advance "normally" in Robolectric).

chrisfillmore:
Thanks for the feedback. I am not ignoring this PR, just need to complete some other work first, and I will revisit this (later this week or next week).

PR #6882, opened 2020-01-16
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


ojw28:
It's probably easier for us to simply not accept pull requests upgrading the wrapper, and do it ourselves.

PR #6922, opened 2020-01-28
Add minimal support of DVB AIT

phhusson:
This is used by broadcast channels to provide interactive contents, through HbbTV.

There are many more informations available in this TS section, though this is enough for a minimal HbbTV implementation.


phhusson:
@googlebot I consent.

phhusson:
Test stream: https://treble.phh.me/test-ait.ts (this link won't live very long, don't use the link for permanent tests)

ojw28:
@icbaker - Could you take a look at this, since it falls under metadata?

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
```diff
@@ -0,0 +1,195 @@
...
+    private Metadata parseAit(ParsableByteArray sectionData) {
+        int tmp;
+
+        int tableId = sectionData.readUnsignedByte();
+
+        tmp = sectionData.readUnsignedShort();
+        int endOfSection = sectionData.getPosition() + (tmp & 4095) - 4 /* Ignore leading CRC */;
+
+        tmp = sectionData.readUnsignedShort();
+        int applicationType = tmp & 0x7fff;
```
icbaker: Because a lot of the values don't fall on clean byte boundaries, but they are mostly numeric, i think you might be better off using `ParsableBitArray` instead of `ParsableByteArray`

Then you can read and skip bits and it should be super-clear that you've implemented the spec.

It makes reading strings a bit more awkward, but you can use `Util.fromUtf8Bytes` to help I suspect (or look into adding string reading methods to `ParsableBitArray`.

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
```diff
@@ -0,0 +1,195 @@
...
+import com.google.android.exoplayer2.metadata.MetadataInputBuffer;
+import com.google.android.exoplayer2.util.ParsableByteArray;
+import com.google.android.exoplayer2.util.TimestampAdjuster;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+
+public class AitDecoder implements MetadataDecoder {
+    // Specification of AIT can be found in 5.3.4 of TS 102 809 v1.1.1
```
icbaker: If this is meant to be explaining the `DESCRIPTOR_TRANSPORT_PROTOCOL` constant, it seems like 5.3.6 is a more precise reference?

I think we should add similar references to the other constants too to point to the relevant sections of the spec

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
```diff
@@ -0,0 +1,195 @@
...
+                int type = sectionData.readUnsignedByte();
+                int l = sectionData.readUnsignedByte();
+                int positionOfNextSection2 = sectionData.getPosition() + l;
+
+                if(type == DESCRIPTOR_TRANSPORT_PROTOCOL) {
+                    int protocolId = sectionData.readUnsignedShort();
+                    int label = sectionData.readUnsignedByte();
+
+                    if(protocolId == TRANSPORT_PROTOCOL_HTTP) {
+                        while (sectionData.getPosition() < positionOfNextSection2) {
```
icbaker: instead of using a `2` suffix, can we call this `positionOfNextSelector` (and also rename `positionOfNextSection` to `positionOfNextDescriptor` above)

Same also for any of the `l` or `len` variables, especially if their scope is quite large, could we qualify them with what they're the length of?

(this is assuming I've correctly matched up your code with the names used in the spec! Please correct me if I'm making nonsense suggestions)

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
```diff
@@ -0,0 +1,195 @@
...
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+    };
+
+    static String readDvbString(ParsableByteArray data, int length) {
```
icbaker: Is this method used? I couldn't see a reference to it

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
```diff
@@ -176,11 +182,41 @@ public TsPayloadReader createPayloadReader(int streamType, EsInfo esInfo) {
...
         return new PesReader(
             new DvbSubtitleReader(esInfo.dvbSubtitleInfos));
+      case TsExtractor.TS_STREAM_TYPE_AIT:
+        return new SectionReader(new SectionPassthrough(MimeTypes.APPLICATION_AIT));
       default:
         return null;
     }
   }
 
+  public class SectionPassthrough implements SectionPayloadReader {
```
icbaker, line 202: Can you add some documentation about what this class is for/what it does?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
```diff
@@ -176,11 +182,41 @@ public TsPayloadReader createPayloadReader(int streamType, EsInfo esInfo) {
...
+    private TimestampAdjuster timestampAdjuster = null;
+    private final String mimeType;
+    private TrackOutput output;
```
icbaker: Can this be `final`?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -96,6 +96,9 @@
   public static final int TS_STREAM_TYPE_SPLICE_INFO = 0x86;
   public static final int TS_STREAM_TYPE_DVBSUBS = 0x59;
 
+  //Those are special IDs, which don't have actual TS definitions
```
icbaker, line 99: Where does the value come from? Can you reference a spec or something?

icbaker (changes requested):
Could you also please look at adding some tests for AitDecoder?
There's some other metadata decoder tests in the `icy` and `scte35` packages for inspiration.

If you have a (short & small!) TS stream with an AIT that can be included in the PR you can also add it to TsExtractorTest which will add some nice robustness. This might be a bit awkward to get working, so feel free to reach out for help if it's not working as you'd expect.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
phhusson: No, this value is magic value that is beyond max standard value of TS_STREAM_TYPE which is one byte.
It is meant to carry stream type for types not defined in the spec.

Do you have a recommendation to handle that?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
phhusson: I can't see how, since output can be created only in init() function not in constructor.
Looking at other Readers (Ac3Reader), I see that trackouput is `@MonotonicNonNull` which I'll add.

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
phhusson: Correct, it isn't used.
That was used by some dead code (there are many informations not parsed by this decoder) (namely to get the "Application name")

This will be removed, thanks

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
phhusson: Right, that's indeed a mess, I've refactored this.
FWIW I don't end with the same renaming as you.
Hopefully I added enough comments and reference to the standard so we can converge :-)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DefaultTsPayloadReaderFactory.java
phhusson: Added

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
phhusson: Well no I really meant to link this for global AIT documentation
I updated the documentation structure to make more sense

library/core/src/main/java/com/google/android/exoplayer2/metadata/dvbsi/AitDecoder.java
phhusson: Done. I ended up adding `readString()` to ParsableBitArray

phhusson:
There I did the few bits of refactoring.
I don't know if you prefer that I squash all my patches, or I fix your remarks commit by commit?

I'll now add tests.

phhusson:
I'm not totally sure what test you want to see in TsExtractor.
I can add a testAitSample that has same code as testSample, on a 30kB sample which includes AIT, SDT, EIT and DSM-CC, does that good for you?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
AquilesCanta: Sorry to jump in: You can avoid the public magic value by adding a boolean in StreamInfo saying "hasApplicationSignallingDescriptor" which is true if and only if one of the descrptor uses descriptor tag 0x6F. Alternatively, you could include in stream info a list / array of all descriptor tags associated with the program, and then checking for that in the TsPayloadReaderFactory. The boolean is probably just simpler.

icbaker:
Looks good, thanks for making the changes! I'll work on getting this merged.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
phhusson: For what it's worth, Teletext tracks are also detected by a specific descriptor in PMT
(see https://www.etsi.org/deliver/etsi_EN/300400_300499/300472/01.03.01_60/en_300472v010301p.pdf section 4)

So the boolean way doesn't work much. I'm not a fan of including all descriptors tags.
We could have an `enum ExtraType { Nope, Application, Teletext, }` but I'm not much of a fan either.

Now, in Teletext's case, descriptors contain useful information to send to Teletext{Reader,Decoder} (namely the page of the subtitle and the language). At the moment we transmit it by creating a new TelextSubtitleInfo storage-class (which inherits SubtitleInfo, and we changed DvbSubtitleInfo inherits SubtitleInfo as well). And we gave teletext track 0x102 id

I don't have strong opinion on how to handle that, though at the moment I still prefer leaving those "internal" IDs.

phhusson:
Thanks!

Issue #4362, opened 2018-06-09
Broadcast of Intent ANR


needz:
### Issue description
Since ExoPlayer version 2.8.* I am receiving a lot of ANRs in the Google Play Console.
`Broadcast of Intent { act=android.intent.action.TIME_TICK flg=0x50000014 (has extras) }`
Seems to have something in common with https://github.com/google/ExoPlayer/issues/4352 , because there are ANRs when releasing the player.

### Reproduction steps
I cannot reproduce it, seems like it depends on the type of stream.

### Version of ExoPlayer being used
ExoPlayer 2.8.1

### Device(s) and version(s) of Android being used
Sony BRAVIA 4K GB (BRAVIA_ATV2), 2560MB RAM, Android 7.0
Bouygues Telecom Bbox Miami (HMB4213H), 2048MB RAM, Android 5.1
Xiaomi Mi Box (once), 2048MB RAM, Android 6.0
TIM TIMvision (dwt765ti), 2048MB RAM, Android 7.0
Philips 4K LED TV powered by Android (PH7M_EU_5596), 2560MB RAM, Android 7.0
etc.

### A full bug report captured from the device

```
"main" prio=5 tid=1 Waiting
  | group="main" sCount=1 dsCount=0 obj=0x741a8668 self=0xa9f85400
  | sysTid=722 nice=0 cgrp=default sched=0/0 handle=0xacc19534
  | state=S schedstat=( 4495701312 1963517425 12473 ) utm=381 stm=68 core=3 HZ=100
  | stack=0xbe03b000-0xbe03d000 stackSize=8MB
  | held mutexes=
  at java.lang.Object.wait! (Native method)
- waiting on <0x0443c7f6> (a com.google.android.exoplayer2.z)
  at com.google.android.exoplayer2.PlayerMessage.blockUntilDelivered (PlayerMessage.java:282)
  at com.google.android.exoplayer2.SimpleExoPlayer.setVideoSurfaceInternal (SimpleExoPlayer.java:962)
  at com.google.android.exoplayer2.SimpleExoPlayer.access$102 (SimpleExoPlayer.java:59)
  at com.google.android.exoplayer2.SimpleExoPlayer.access$200 (SimpleExoPlayer.java:59)
  at com.google.android.exoplayer2.SimpleExoPlayer.access$302 (SimpleExoPlayer.java:59)
  at com.google.android.exoplayer2.SimpleExoPlayer.access$802 (SimpleExoPlayer.java:59)
[...]

PR #6932, opened 2020-01-30
issue #4352 resolved using handler on handling messages

kashu1209:


ojw28:
If not blocking where we currently block was a valid solution to the referenced issue, we'd just remove the `blockUntilDelivered()` calls. As per the issue itself, this is not a valid fix.

kashu1209:
You guys were using blockUntilDelivered()
But I have a test phone which stuck there 50% of times and when i did my
solution.
It didnt get an ANR.
In tried it for Nougat,OREO,Pie,10
All are working fine.
Please Can you refer me a link which can help me understand the issue
better.
As I was getting error similar to issue #4362
Which is closed by saying duplicate of issue #4352.

ojw28:
This is already explained by the issue you're referencing as being fixed by this pull request. For example see https://github.com/google/ExoPlayer/issues/4352#issuecomment-416580300.

PR #6994, opened 2020-02-19
use swresample instead of avresample

xufuji456:
Hello, ExoPlayer developers. I found that avresample module of FFmpeg has deprecated. Therefore, this commit is using swresample instead of avresample.


andrewlewis:
Thanks for the pull request. Please could you send this to the branch called `dev-v2`?

xufuji456:
@andrewlewis Ok, I send this commit to branch of "dev-v2".

PR #6999, opened 2020-02-19
use swresample instead of avresample

xufuji456:
Hello, ExoPlayer developers. I found that avresample module of FFmpeg has deprecated. Therefore, this commit is using swresample instead of avresample, and we could reduce the library size of avresample.


andrewlewis:
Thanks for sending this. Changes look good and I'll take care of updating the build configuration and references to avresample in logging when merging it.

xufuji456:
You are welcome. It's my pleasure to do something for open source project.

Issue #855, opened 2015-10-09
MSS Live Playback : Not according to Spec


Manjunath-N:
Hello,

 In MSS live, playback is achieved by refreshing the manifest, which is not according to the Specification( [MS-SSTR] — v20120705 -Smooth Streaming Protocol Specification). 
According to the spec the TFXD and TFRF boxes should be used to get the new fragment info and the same should be used to update the manifest.

Can you please let us know why the manifest was refreshed every time instead of parsing the boxes?

Thanks
Manjunath N

ojw28:
No particular reason. We should move things to be more compliant with the spec.

orangeotc:
Digging back some old issue, is there any way to change the behavior from the exoplayer and request the MSS manifest only once at playback startup?

PR #7005, opened 2020-02-21
Parse TFRF for continuous SmoothStreaming segments refresh

phhusson:
This fixes issues #855

This is needed in our use-case because we have a backend for which the access to the manifest requires a token that is valid only for one minute.

I have tested this with http://essdrmproducts-dev.insidesecure.com:9090/live/clear.isml/manifest
I added logs in HttpDataSource to log every access to manifest and confirmed it is now only accessed once, while it was recurrent before.

Points that are unclear to me:
- When testing on http://essdrmproducts-dev.insidesecure.com:9090/live/clear.isml/manifest I can see that Exoplayer tries to download chunks ahead of their time, so the server returns few 412 before the chunks are actually available. This doesn't happen prior to the patch. Since I don't change the chunk download algorithm I'm not sure why this happen.
- I'm using System.currentTimeMillis() to expire old chunks. I think this is the sane way to do, but i'm not totally convinced

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
@@ -65,6 +65,10 @@ public class FragmentedMp4Extractor implements Extractor {
   public static final ExtractorsFactory FACTORY =
       () -> new Extractor[] {new FragmentedMp4Extractor()};
 
+  public interface SsAtomCallback {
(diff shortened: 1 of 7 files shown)


guidofranceschini:
Hi phhusson ,
I am not among the contributors, but I can probably explain why you observe a few 412.
Within a manifest you can get information only about chunks that are fully available.
Instead. from the boxes within a chunk you can determine the timestamp and hence the URL of the next chunk, but this "next chunk" is not necessarily available yet. You have probably updated the code in a way that it now knows about the "next chunk" before it is actually available.

PR #7029, opened 2020-02-28
Throw exception for timestamp issues that would stall playback

stevemayhew:
ExoPlayer can hang forever when presented with content with unmarked discontinuities.  This change causes a playback error instead, reporting the errant segment[s].

Note, recovery would of course be the preferred course of action, hopefully this pull request can server as a place for discussing what that might look like.


stevemayhew:
Closed, will re-open without polluting commits

Issue #7030, opened 2020-02-28
Improve robustness when adaptive playbacks encounter unexpected discontinuities


stevemayhew:
### Issue description
ExoPlayer freezes playback in `Player.STATE_BUFFERING` forever when faced with un-marked discontinuities in the stream.

Other players (Safari/QuickTime, VisualOn, Hls.js) *recover* from this to varying degrees (Safari looses A/V sync, not desirable, but at least the evidence is more clear the problem is with the stream not the player.

I'm not suggesting ExoPlayer's design of insisting on streams that conform to the HLS Draft is incorrect, just that best effort should be made to report the playback error without freezes.

### Reproduction steps
Play the sample HLS stream through the Verizon commercial, starting at 20 seconds in, ExoPlayer freezes the last frame in the OR scene before the Cheerios commercial.  Note, the stream has continuity issues around the two ad insertions, that is clear.

##### Actual Behavior
ExoPlayer freezes indefinitely in `Player.STATE_BUFFERING` with plenty of buffered segments.

##### Expected Behavior
In order of decreasing desirability, one of the following:

1. Recover from the error by introducing a discontinuity internally (split the segment and create a new [TimestampAdjuster](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/util/TimestampAdjuster.html) linked to the media time)
1. Detect the error during playback (one or more `Renderer` will be stuck with [Renderer.isReady()](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/Renderer.html#isReady--) false but playback is fully buffered.  
1. Detect the error at load and throw a load exception (simplest course of action, this is in the pull request)

### Link to test content
The content was mailed to dev.exoplayer@gmail.com

### A full bug report captured from the device
Again emailed to dev.exoplayer@gmail.com 

### Version of ExoPlayer being used
Version 2.10.4, with the pull request code.  Note multiple versions of ExoPlayer all reproduce this.
The pull request for the suggested fix is: https://github.com/google/ExoPlayer/pull/7034

### Device(s) and version(s) of Android being used

Multiple devices, the bug-report is from an Arris vip6102w running Android P.

phhusson:
This seems to be related to #6983

stevemayhew:
Most likely Pierre..  We have seen origin servers where the packager runs wild and creates an extremely long segment cause the TS to run on forever and throw this exact same OOM exceptions.

[...]

PR #7034, opened 2020-02-29
Fail segment load on gross PTS deviation

stevemayhew:
If the ExoPlayer encouters an unreported PTS discontinuity in the media that exceeds segment size the player will ultimately freeze forever (with one of the renderers starved for samples in the buffering state, even though LoadControl reports fully bufferd).

This change detects the issue and fails the segment load.

Note, other players (ffplay, Apple's AVPlayer, etc) 'recover' from these streams, at least playing the stream perphaps with A/V sync issues.  These kind of stream issues regulary occur in broadcast streams where affiliate perform ad insertion on the fly.

Issue: #7030


library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
```diff
@@ -673,6 +676,23 @@ private synchronized void commitSample(
     largestQueuedTimestampUs = Math.max(largestQueuedTimestampUs, timeUs);
 
     int relativeEndIndex = getRelativeIndex(length);
+
+    // Check for obvious discontinuities in the content
+    if ( length > 0 && (sampleFlags & C.BUFFER_FLAG_KEY_FRAME) == 1) {
```
ojw28: How is this change going to work with subtitle and metadata tracks, which are often quite sparse? A 10 second gap is not unexpected, and probably occurs quite regularly.

In general, I worry that this kind of change solves one problem (for media that's actually broken) whilst risking introducing new problems (for media that's actually valid). It's not even obviously wrong for a video track to have a >10 second gap between frames, for some more unusual cases. For example if a video track is being used for album artwork.

stevemayhew:
I agree with you, a hard and fast 10s is not a good number, and certainly
not for non video tracks.  You can see from the sample video (broadcast of
the US ProBowl) it’s poorly handled ad insertions often. One other thought
is catching in the `Extractor`. Doing nothing (ExoPlayer freezes then), is
not an option for our customers.  They simply blame it on ExoPlayer (“other
players work”) making it difficult for us to convince them the origin needs
to fix the stream metadata

Perhaps some average and check if a sample deviates more than a standard
deviation from it

ojw28:
Presumably you can point to Safari losing A/V sync as a second player that's not correctly handling this type of stream... It seems somewhat hard to argue that it's ExoPlayer's fault when it's only "some other players work", and also when the streams are clearly in violation of the spec ;).

I agree we should try and improve the current behaviour, but we need to come up with a way of doing so that clearly and obviously doesn't negatively impact on any other case (and where that can be determined just by looking at the code change). So I think we need to think a bit about what the best way of doing that is.

library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
stevemayhew: We have sample streams from MusicChoice which we support that we can try, but I'm not a fan of hard-coded values.  The goal is to detect intra-Extractor deviations, so we could check against and average value, this dos need more thought.

stevemayhew:
I've updated to match the design in the bug, we are running with this change back-ported to current release-v2 to test it.  See comments in the commits for details.

stevemayhew (commented):
I will squash out this commit (468c200) in the final pull branch, it reverts the changes to the SampleQueued.  Executing the check in the SampleQueue itself is not a good idea, as it has no context to the source time ranges.  See the added class `HlsCheckedSampleQueue`

library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
```diff
@@ -480,7 +480,7 @@ public final void sampleData(ParsableByteArray buffer, int length) {
   }
 
   @Override
-  public final void sampleMetadata(
+  public void sampleMetadata(
```
stevemayhew, line 491: Allow subclass.

library/core/src/main/java/com/google/android/exoplayer2/source/UnreportedDiscontinuityException.java
```diff
@@ -11,16 +12,15 @@
...
   /**
    * Consturct the exception
    *
    * @param timesUs last timestamp before attempted commit of the deviant sample
-   * @param sourceId source (e.g. HLS segment) id of the segment with the issue
+   * @param uri uri of the segment with the unreported discontinuity
    */
-  public UnreportedDiscontinuityException(long timesUs, int sourceId) {
+  public UnreportedDiscontinuityException(long timesUs, Uri uri) {
+    super("Unreported discontinuity timeMs: " + C.usToMs(timesUs) + " in URI: " + uri);
```
stevemayhew: Pass the actual segment's URI, much more useful.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsCheckedSampleQueue.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.android.exoplayer2.C;
+import com.google.android.exoplayer2.drm.DrmSessionManager;
+import com.google.android.exoplayer2.source.SampleQueue;
+import com.google.android.exoplayer2.source.UnreportedDiscontinuityException;
+import com.google.android.exoplayer2.upstream.Allocator;
+import com.google.android.exoplayer2.util.Log;
+
+public class HlsCheckedSampleQueue extends SampleQueue {
+  private static final String TAG = "HlsCheckedSampleQueue";
+
```
stevemayhew: I will add a detailed class comment when this is finished.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1340,7 +1341,7 @@ private static DummyTrackOutput createDummyTrackOutput(int id, int type) {
     return new DummyTrackOutput();
   }
 
-  private static final class FormatAdjustingSampleQueue extends SampleQueue {
+  private static final class FormatAdjustingSampleQueue extends HlsCheckedSampleQueue {
```
ojw28: Can we just rename this to `HlsSampleQueue` and have it do both jobs, rather than creating another layer in the class hierarchy?

library/core/src/main/java/com/google/android/exoplayer2/source/UnreportedDiscontinuityException.java
```diff
@@ -0,0 +1,26 @@
...
+import android.net.Uri;
+import com.google.android.exoplayer2.C;
+
+/**
+ * Thrown from the loader thread when an attempt is made to commit a sample that is far
+ * deviant from the expected sequence of timestamps in the SampleStream.   This is likely
+ * caused by a discontinuity in a segment that was not split and reported by metadata in
+ * an HLS (EXT-X-DISCONTINUITY) or DASH stream.
+ */
+public class UnreportedDiscontinuityException extends RuntimeException {
```
ojw28: IMO make it extend `IOException` so you don't have to catch and re-throw.

The naming is also a bit confusing, because the exception is by definition reporting the discontinuity :). Maybe call it `UnexpectedDiscontinuityException`.

Also make it final.

library/core/src/main/java/com/google/android/exoplayer2/source/UnreportedDiscontinuityException.java
```diff
@@ -0,0 +1,26 @@
+package com.google.android.exoplayer2.source;
```
ojw28: Please add the copyright header (copy from another file + update the year)

library/core/src/main/java/com/google/android/exoplayer2/source/UnreportedDiscontinuityException.java
```diff
@@ -0,0 +1,26 @@
...
+ * an HLS (EXT-X-DISCONTINUITY) or DASH stream.
+ */
+public class UnreportedDiscontinuityException extends RuntimeException {
+
+  public final long timesUs;
+
+  /**
+   * Consturct the exception
+   *
+   * @param timesUs last timestamp before attempted commit of the deviant sample
```
ojw28: It would be nice to capture both timestamps: `lastSampleTimeUs` and `nextSampleTimeUs`. And to pass the whole `DataSpec` rather than just the `Uri`.

(Note that some HLS streams will use the same `Uri` for every segment, so at least knowing the range is pretty essential)

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsCheckedSampleQueue.java
```diff
@@ -0,0 +1,65 @@
...
+    this.highestTimeUs = (long) (chunk.endTimeUs + tolerance);
+    this.chunk = chunk;
+    loggedFirst = false;
+  }
+
+
+  @Override
+  public void sampleMetadata(long timeUs, int flags, int size, int offset,  @Nullable CryptoData cryptoData) {
+    if (lowestTimeUs != C.TIME_UNSET && timeUs < lowestTimeUs && ! loggedFirst) {
+      Log.d(TAG, "sampleMetadata() - committed timeUs: " + timeUs + " is " + C.usToMs(lowestTimeUs - timeUs) + "ms less then segment start time.  chunk: " + chunk.dataSpec.uri);
```
ojw28: Please can we remove all the logging.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsCheckedSampleQueue.java
```diff
@@ -0,0 +1,65 @@
...
+      loggedFirst = true;
+    }
+    if (lowestTimeUs != C.TIME_UNSET && timeUs < (lowestTimeUs - C.msToUs(50_000))) {
```
ojw28: Where is the magic 50 coming from? Shouldn't we just subtract `tolerance` from lowestTimeUs` in `setCurrentLoadingChunk`?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsCheckedSampleQueue.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private HlsMediaChunk chunk;
+  private boolean loggedFirst = false;
+
+  HlsCheckedSampleQueue(Allocator allocator, DrmSessionManager drmSessionManager) {
+    super(allocator, drmSessionManager);
+  }
+
+  void setCurrentLoadingChunk(HlsMediaChunk chunk) {
+    double tolerance = (chunk.endTimeUs - chunk.startTimeUs) * 0.1;
```
ojw28: Is there any reason not to just use a constant (and quite large) value here, like 5 seconds?

library/core/src/main/java/com/google/android/exoplayer2/source/UnreportedDiscontinuityException.java
stevemayhew: Excellent idea.  Yes, `lastSampleTimeUs` will give us an idea how far into the issue occurs segment it is (`lastSampleTimeUs` - `lastSampleTimeUs`).

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleQueue.java
```diff
@@ -0,0 +1,159 @@
...
+      }
+    }
+    return new Metadata(newMetadataEntries);
+  }
+
+  @Override
+  public void sampleMetadata(long timeUs, int flags, int size, int offset,  @Nullable CryptoData cryptoData) {
+    // TODO - chunkless prepare, sampleQueue list is not yet initialized for first chunk
+//    Assertions.checkNotNull(chunk, "sampleMetadata without a MediaChunk?");
+    if (chunk == null) {
```
stevemayhew: Had to replace this with null check, the initial segment loads are started before `HlsSampleStreamWrapper#createSampleQueue()` is called so need a way to get the loading chunk in that method and set it in the `HlsSampleQueue`

Or we ignore checking the first segment loads.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1343,82 +1343,6 @@ private static DummyTrackOutput createDummyTrackOutput(int id, int type) {
     return new DummyTrackOutput();
   }
 
-  private static final class FormatAdjustingSampleQueue extends HlsCheckedSampleQueue {
-
-    private final Map<String, DrmInitData> overridingDrmInitData;
```
stevemayhew: Combined into `HlsSampleQueue`

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -379,10 +379,9 @@ private void feedDataToExtractor(
...
           result = extractor.read(input, DUMMY_POSITION_HOLDER);
         }
-      } catch (UnreportedDiscontinuityException e) {
-        Log.d(TAG, "Unreported discontinuity at timeUs: "+ e.timesUs + " uri: " + dataSpec.uri);
-        throw new IOException("Timestamp error", e);
-        
+      } catch (UnexpectedDiscontinuityException e) {
+        Log.d(TAG, "UnexpectedDiscontinuityException - recovering by discarding balance of segment", e);
+        throw new IOException("load aborted for segment - " + e.dataSpec + " unexpected discontinuity", e);
       } finally {
```
stevemayhew: `UnexpectedDiscontinuityException` needs to stay a `RuntimeException` until signatures in the class hierarchy for `TrackOutput#sampleMetadata()` are changed to include a declared exception.  Best to keep this in a separate change to avoid having to merge/rebase lots of files while this branch is open.

stevemayhew:
@ojw28 I rebased to dev-v2 and addressed (hopefully) all your review comments,  feel free to be picky, I'm happy to accommodate.

Currently I'm playing broadcasts streams of major network from multiple affiliates and multiple origin servers we support to make sure they playback normally.  The pro bowl sample I attached to the bug fails with a playback error rather then hanging using the current state of this branch.

I'm working on the recovery code now (looks like this will involve reseting the timestamp adjustment and discarding the balance of the segment).  Also welcome more discussion as to if this is a load error or some other event.  My personal feeling is treat it like any other load error, so:

1. the `HlsMediaChunk#feedDataToExtractor()` would setup any precursor to possible recovery (timestamp adjustment, etc.)
1. catch exception, log and re-throw so error handler would get it
1. if `LoadErrorAction DONT_RETRY` is returned, playback continues with balance of segment discarded.

The player UI's all have the option of implementing `LoadErrorHandlingPolicy` and choosing how to deal with this.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
ojw28: Can you just make it an `IOException`, to avoid needing to catch and re-throw?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleQueue.java
```diff
@@ -0,0 +1,159 @@
...
+ * </ul>
+ *
+ * The timestamp check verifies that the adjusted sample time (via {@link TimestampAdjuster}) does not
+ * fall outside of a set percentage ({@link #MAX_TIMESTAMP_DEVIATION_PERCENTAGE}) of the time
+ * boundaries of the segment as expressed by the segment duration ((@link HlsMediaChunk#endTimeUs} -
+ * {@link HlsMediaChunk#startTimeUs}).  This is loosely mandated by the Pantos spec and checked by
+ * Apple's mediastreamvalidator.
+ *
+ */
+public class HlsSampleQueue extends SampleQueue {
```
ojw28: Can this stay where `FormatAdjustingSampleQueue` was (i.e., as an inner class)? We definitely don't want it to be part of the ExoPlayer library's actual API surface, and I don't think enough is really being added to justify splitting it into its own file.

library/core/src/main/java/com/google/android/exoplayer2/source/UnexpectedDiscontinuityException.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /** The errant timestamp
+   */
+  public final long deviantSampleTimeUs;
+
+  /** The source of the samples that resulted in this error
+   */
+  public final DataSpec dataSpec;
+
+  /** The timeUs that the source of the samples starts (from HLS metadata)
```
ojw28, line 50: The Javadoc for this class seems confused about whether it's an HLS specific exception or not. Can we either move it into the HLS package and make it HLS specific, or remove specific references to HLS from the documentation?

ojw28 (commented):
Looking good!

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleQueue.java
stevemayhew: IMHO, breaking up big classes is a good thing (less conflicted merges, etc), I can make it package private and exclude it from the JavaDoc, but I do think it is a good idea to make it a standalone class.  Pretty much equivalent from the VM's view to a static inner class.

If I can exclude it from the package classes are you ok with this?  Otherwise, sure I'll move it to an inner class.

ojw28: It's not a great idea to split it out whilst simultaneously modifying it, because it makes the change harder to review (or look at later to see what changed). If it's to be split out, that should IMO be a separate change after this one, that's a pure no-op move rather than one that also changes behavior.

library/core/src/main/java/com/google/android/exoplayer2/source/UnexpectedDiscontinuityException.java
stevemayhew: The same problem could exist in DASH (at least as I understand it), the origin must split the timeline into multiple periods if the timestamps wrap or jump (https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#timing-sampletimeline) I don't have a sample DASH playlist that breaks this rule.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleQueue.java
stevemayhew: Ah. Yes, I agree 100%, I"ll move it back and reduce the diffs.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1337,26 +1340,57 @@ private static boolean formatsMatch(Format manifestFormat, Format sampleFormat)
     return true;
   }
 
-  private static DummyTrackOutput createDummyTrackOutput(int id, int type) {
-    Log.w(TAG, "Unmapped track with id " + id + " of type " + type);
-    return new DummyTrackOutput();
-  }
```
stevemayhew, line 1343: Patch diff moved this to after the original `FormatAdjustingSampleQueue` class.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1337,26 +1340,57 @@ private static boolean formatsMatch(Format manifestFormat, Format sampleFormat)
...
+    /**
+     * largest timestamp deviation from the segment time bounds expressed as a percentage of
+     * the segment duration.
+     */
+    private static double MAX_TIMESTAMP_DEVIATION_PERCENTAGE = 0.50;
+
+    private long lowestTimeUs = C.TIME_UNSET;
+    private long highestTimeUs = C.TIME_UNSET;
 
-  private static final class FormatAdjustingSampleQueue extends SampleQueue {
```
stevemayhew: Rename `FormatAdjustingSampleQueue` to `HlsSampleQueue` and add checks for timestamp discontinuity.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1337,26 +1340,57 @@ private static boolean formatsMatch(Format manifestFormat, Format sampleFormat)
...
-    public FormatAdjustingSampleQueue(
-        Allocator allocator,
-        DrmSessionManager<?> drmSessionManager,
-        MediaSourceEventDispatcher eventDispatcher,
-        Map<String, DrmInitData> overridingDrmInitData) {
+    private HlsSampleQueue(Allocator allocator,
+            DrmSessionManager<?> drmSessionManager,
+            MediaSourceEventDispatcher eventDispatcher,
+            Map<String, DrmInitData> overridingDrmInitData) {
       super(allocator, drmSessionManager, eventDispatcher);
```
stevemayhew, line 1387: spacing change obfuscates the simple rename.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleQueue.java
stevemayhew: @ojw28 In this commit ccf2ba3e1 I moved the class back as an inner class and tried to remove the diffs.   Once we finish the pull up and agree it is merge ready I will rebase out the intermediate states to further reduce the changes to `HlsSampleStreamWrapper` so the y are clear.

Hope this helps.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
stevemayhew: I still need to catch in order to get to the timestamp adjuster (and/or add metadata to the segment to discard the time period and recover).  I can probably do this on my own fork if no one else would want it (my preference always is ExoPlayer proper to take the changes, note if by default we fail playback anything done here to facilitate recovery will not matter).

This latest checkin (4701a6b39081cccaab202effb6cf59de6d4510b2) make the exception extend IOException (much nicer, I agree, then a runtime exception).   It took a while to make and run the tests, so hopefully I'll not have to rebase it much (Most of it is of course pretty rote to review and in the extractor library)

stevemayhew:
Thanks @ojw28 for reviewing this, I know it has become a pretty large change.  This will make ExoPlayer less vulnerable to infinite stalls.  

Our team deals with lots of live streams from a growing variety of origin servers, each with their own "creative" solution to the discontinuity problem.  The more we can at least detect and better recover or at least report these errors the better our user experience is.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1416,6 +1450,32 @@ private Metadata getAdjustedMetadata(@Nullable Metadata metadata) {
...
+    @Override
+    public void sampleMetadata(
+        long timeUs,
+        @C.BufferFlags int flags,
+        int size,
+        int offset,
+        @Nullable CryptoData cryptoData) throws IOException {
+      // TODO - chunkless prepare, sampleQueue list is not yet initialized for first chunk
+  //    Assertions.checkNotNull(chunk, "sampleMetadata without a MediaChunk?");
+      if (chunk == null) {
```
ojw28, line 1464: I don't think I understand when this can happen, and the TODO above didn't help me figure it out. Could you elaborate?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -1337,26 +1340,57 @@ private static boolean formatsMatch(Format manifestFormat, Format sampleFormat)
...
+            DrmSessionManager<?> drmSessionManager,
+            MediaSourceEventDispatcher eventDispatcher,
+            Map<String, DrmInitData> overridingDrmInitData) {
       super(allocator, drmSessionManager, eventDispatcher);
       this.overridingDrmInitData = overridingDrmInitData;
     }
 
-    public void setDrmInitData(@Nullable DrmInitData drmInitData) {
+    void setCurrentLoadingChunk(HlsMediaChunk chunk) {
+      double tolerance = (chunk.endTimeUs - chunk.startTimeUs) * MAX_TIMESTAMP_DEVIATION_PERCENTAGE;
```
ojw28: For small segments this can end up allowing quite a small tolerance. I wonder whether it's worth adding a minimum tolerance (e.g., 5 seconds) as well, and `Math.max`-ing the two? I kind of wonder whether just a fixed tolerance (e.g., 10 seconds) might be easier to reason above, since I'm not sure you'd necessarily want to allow a really large deviation for long segments either.

ojw28 (commented):
Sorry for the delay. See inline comments.

I didn't realize having the exception extend IOException would require so many files to be touched! I'll probably put it back to a runtime exception as you had it (probably `IllegalStateException`) when I merge this, to avoid that. No need to do it on your side though.

stevemayhew:
> Sorry for the delay. See inline comments.
> 
> I didn't realize having the exception extend IOException would require so many files to be touched! I'll probably put it back to a runtime exception as you had it (probably `IllegalStateException`) when I merge this, to avoid that. No need to do it on your side though.

Not a problem,  I pretty much guessed you would feel this way... So I purposely made the commit so I can back it out (branched before) so I'll remove it and change to `IllegalStateException`

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: I agree.   I have increased this in my own branch and testing to a larger value,  this is particularly evident for i-Frame only which can be of extremely short times.  The `Math.max()` is a great way to do this.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: The first segments to load when it is not [chunk-less prepare mode](https://medium.com/google-exoplayer/faster-hls-preparation-f6611aa15ea6).  This is because until the prepare is complete the `SampleStream` is not added to the `HlsSampleStreamWrapper.sampleQueues[]` array and the `HlsSampleStreamWrapper.init()` method does not get called until the load has already started.  So samples can be committed before the current loading `MediaChunk` is set.  

I'm working out how architecturally to make this happen, so the TODO is for me.

ojw28:
> Not a problem, I pretty much guessed you would feel this way... So I purposely made the commit so I can back it out (branched before) so I'll remove it and change to IllegalStateException

Thanks. Please could you push this, and we'll look at getting this merged.

stevemayhew:
@ojw28 the force push eliminated the commit 4701a6b (changing from `RuntimeException`) so all the methods do not need signature change.  Also, took your suggestions in the latest commit 4859207.  I've updated our internal release branch to match these changes to test them out.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: See commit 4859207 changed to make max the tolerance against a min possible value.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
ojw28: Sorry, I still don't understand this. It looks to me like `HlsSampleStreamWrapper.init()` is always called before a chunk starts to load. Can you actually reproduce this not happening, and if so, how?

It's also unclear whether this is about chunk-less or non-chunk-less preparation. The comment says one thing, but the conversation says another. Please clarify; thanks!

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -379,6 +382,9 @@ private void feedDataToExtractor(
...
         nextLoadPosition = (int) (input.getPosition() - dataSpec.position);
       }
+    } catch (UnexpectedDiscontinuityException e) {
+      Log.d(TAG, "UnexpectedDiscontinuityException - recovering by discarding balance of segment", e);
```
ojw28, line 386: What does "discarding balance of segment" mean? Does this actually handle recovery in its current form? I'm not sure how it does, if so?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
ojw28: I see how this happens now.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
stevemayhew: The balance of the chunk's samples simply follow the new PTS sequence, so without a fresh timestamp adjuster, they are useless.

With the last checkin the exception is reported as a playback error  (normal error handler in `Loader` wraps it in an `UnexpectedLoaderException` so `DefaultLoadErrorHandlingPolicy.getRetryDelayMsFor()` will see an `UnexpectedLoaderException` then `Loader.DONT_RETRY_FATAL` is returned so the main playback thread is notified)

I'm working on a partial recovery, which would involve skipping at least to the next chunk, like a seek outside of the `SampleQueue`  range (reseting timestampAdjuster).  This works manually if you add code to 'ignore' the error, like:

````java
  @Override
  public LoadErrorAction onLoadError(
      Chunk loadable,
      long elapsedRealtimeMs,
      long loadDurationMs,
      IOException error,
      int errorCount) {
...
    if (error.getCause() instanceof UnexpectedDiscontinuityException) {
      loadErrorAction = Loader.DONT_RETRY;    // Ignore the balance of this chunk
    }

    eventDispatcher.loadError(
        loadable.dataSpec,
...
    return loadErrorAction;
  }
````

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
ojw28: I fixed this as part of the merge.

ojw28:
I merged a variant of most of this change, in https://github.com/google/ExoPlayer/commit/8991586c3b9b17a71214dfdeeaf679c66565a591. Note that:

* I found a way to make the exception extend `IOException` and not have to modify all the other files (it's super hacky, but still marginally better than a `RuntimeException` that then needs to be wrapped to re-throw.
* Actual throwing of the exception is commented out with a TODO. If you could send another pull request that enables it and also enables recovery at the same time, I think that would be ideal.
* I scoped things so that this is limited to HLS for now.

So I think next steps are to figure out how to make recovery work. Thanks!

stevemayhew:
Thanks, this is great.  I can pull this into a new branch from `dev-v2` then and add the recovery.   My thinking is to do it in the demo player in an `LoadErrorPolicy` (by skipping past the bad content, as the seek resets the timestamp adjustment)

What does this mean?
> I scoped things so that this is limited to HLS for now.

Anything special I need to do to include the change in the scope in building the demo?

ojw28:
I was hoping it would be possible just to do a sensible recovery inside the HLS module, rather than having to get things like `LoadErrorPolicy` involved.

> What does this mean?

It just means that the exception was moved to the HLS module and made package private there. Which makes sense if recovery can be done as per my comment above. If it needs to happen somewhere else the exception might need moving and making public.

stevemayhew:
@ojw28 maybe..  I think you'll find the same issue is possible in live playback in DASH, as I mentioned in the pull request.

ojw28:
I don't want to encourage bad serving implementations by handling spec violations that aren't already proven to be deeply rooted in existing solutions. I don't think we've seen evidence that this is significant problem in DASH solutions, and so I don't think we have evidence that this will ever be required for DASH.

I also suspect that it's much harder to splice random things together in an FMP4 container and still produce a segment that's vaguely close to being valid, so it may be that this makes it less likely for DASH solutions to end up with this kind of problem.

PR #7051, opened 2020-03-05
Rendererd buffer count when tunneling enabled

Cizor:
Increasing buffer count when tunneling enabled, for video tracks.
Audio buffer count works fine for both modes.


stevemayhew:
Amrit, yes this is the correct thing to do, this will allow using the AnalyticsListener without otherwise modifying ExoPlayer core code. You might want to enter a bug explaining why this is wrong and referencing it here.

andrewlewis:
Thanks. Looks good but I will probably tweak it to increment the counter in onProcessedTunneledBuffer, so that we don't treat the EOS buffer as a rendered output buffer and to give an approximate value for pre API 23 (rather than having it be stuck at zero). Let me know if you see any issues with that change.

andrewlewis:
Don't worry about updating the pull request -- I will make those changes on our side and this will get marked as merged. Thanks again.

Cizor:
Thanks @andrewlewis! I found no issues with your mentioned change.

andrewlewis:
This was merged in 26d3914e43bac7491f0caa1f9ffbce1ebd6b5924 (this PR should have been marked as merged automatically -- not sure why that didn't work but I will close it manually). Thanks for the contribution.

Issue #6907, opened 2020-01-23
DASH: Keep using redirected URL after 3xx response code


jzhoubupt:
### [REQUIRED] Issue description
When play a DASH Live stream with the MPD file configured as permanently re-direct(http status code 301/308) ExoPlayer still use the original MPD URI to re-fresh, which doesn't follow DASH IF IOP recommendation.

### [REQUIRED] Reproduction steps
Use the standard Exo Demo App play a DASH LIVE stream with the MPD file configured as permanently redirect to another URI. The actual ExoPlayer behaves in following sequence:

1. Make HTTP Request with URI "http://xxx/content/index.mpd"
2. Server returns http status code 308 with re-directed URI to "http://xxx/content/sessionID/index.mpd"
3. ExoPlayer network module will use the new address to download mpd which is all correct
... after a few seconds to re-fresh the manifest file
4. Make HTTP Request with URI "http://xxx/content/index.mpd"
...

The step 4 i'm afraid not following the HTTP re-direct and not follow the DASH IF IOP 4.3  chapter 3.2.15.3. Client Requirements and Recommendations:
> If the HTTP request results in an HTTP redirect using a 3xx response code, the redirected URL replaces the original manifest URL

The source code in DashMediaSource::onManifestLoadCompleted is the place to handle this logic.
BTW, testing with DASH.js and Shaka-Player the re-direct handling is more aligned with DASH IOP.
I'm not quite sure if this is a BUG from your point of view. I open this as a bug mainly referring to DASH IOP spec.

### [REQUIRED] Link to test content
The issue is found in production with commercial CDN and dynamic session management. Some CDN open a new session when client first requests MPD file and send back http re-direct to a new URI with sessionID there. When client doesn't follow the re-direct spec it causes CDN opening too many session and failed to server segment request.

### [REQUIRED] A full bug report captured from the device
The issue is very clear and i belie you all understand the scenario.

### [REQUIRED] Version of ExoPlayer being used
Test with 2.9.3 and 2.11.1 get the same behavior.

### [REQUIRED] Device(s) and version(s) of Android being used
Issue can be observed from any devices.

icbaker:
Thanks for the report, I'll let Olly take a look as he knows the most about DASH.

tonihei:
We already have a pending internal feature request for this but didn't consider it as high-priority so far because repeated redirects still work. I'll mark this as an enhancement as well to track it publicly.

jzhoubupt:
[...]

PR #7056, opened 2020-03-06
stick to new dash manifest uri after http 302 redirection

davibe:
Related to discussion in #6907


davibe:
@googlebot I signed it!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaSource.java
```diff
@@ -818,6 +818,12 @@ protected void releaseSourceInternal() {
         }
       }
     }
+    boolean isRedirect = !loadable.getUri().equals(this.manifestUri);
```
ojw28, line 805: As per my comment in https://github.com/google/ExoPlayer/issues/6907, this change will need updating to let `manifest.location` take priority. You should also do the `isSameUriInstance` check as above. So I think you probably want to replace L810 to L826 with something that looks more like (disclaimer: untested):

```
Uri newManifestUri = manifest.location != null ? manifest.location : loadable.getUri();
synchronized (manifestUriLock) {
  // This condition checks that replaceManifestUri wasn't called between the start and end of
  // this load. If it was, we ignore newManifestUri and prefer the manual replacement.
  @SuppressWarnings("ReferenceEquality")
  boolean isSameUriInstance = loadable.dataSpec.uri == manifestUri;
  if (isSameUriInstance) {
    manifestUri = newManifestUri;
  }
}
```

ojw28:
As per our contribution guidelines, pull requests should be made to the `dev-v2` branch. Please close this one and send a new one to that branch instead. See also the inline comment on the code. Thanks!

PR #7064, opened 2020-03-09
stick to new dash manifest uri after http 302 redirection

davibe:
Related to discussion in #6907


library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaSource.java
```diff
@@ -791,6 +791,18 @@ protected void releaseSourceInternal() {
     manifestLoadPending &= manifest.dynamic;
     manifestLoadStartTimestampMs = elapsedRealtimeMs - loadDurationMs;
     manifestLoadEndTimestampMs = elapsedRealtimeMs;
+    boolean isRedirect = !loadable.getUri().equals(this.manifestUri);
+    if (isRedirect) {
```
ojw28: The problem with having this as a separate block is that L811 will always evaluate to false if a redirect has occurred, which means that the location element wont be given priority. I think you need a single block similar to the suggested code snippet I put on the previous pull request.

davibe: Right.

Issue #7038, opened 2020-03-02
Video player gets stuck when stating a video from pre-defined (known) position on video containing multiple mid breaks.


nishantkp:
### Issue description

Stating a video from pre-defined position on video with multiple mid-rolls will get stuck. Either it will not play content (or get stuck on content) or will get stuck on pod and doesn't resume content anymore.

Code used for stating a video from known position,

```java
 player.prepare(mediaSource, false, false);
 player.seekTo(35000);
```

`AdsLoader.AdsLoadedListener` will fire `CONTENT_RESUME_REQUESTED` and `CONTENT_PAUSE_REQUESTED` events but it doesn't do anything. 
If player is playing an Ad and when all the ads are completed in pod `CONTENT_RESUME_REQUESTED` gets fired but content is not resuming. Same goes for Ads, `CONTENT_PAUSE_REQUESTED` gets fired and player will not play ads and it looks like player is frizzed.

### Reproduction steps
Play a video for `VMAP pre-roll single ad, mid-roll standard pods with 5 ads every 10 seconds`

##### NOTE : 
This happens only on video with multiple mid-rolls. Any video which has single mid-roll will be fine. You can check  `VMAP pre-roll single ad, mid-roll optimized pod with 3 ads, post-roll single ad` and it will work just fine.

### Link to test content

Reproducible on demo-app

### A full bug report captured from the device
```java
I/ExoPlayerImpl: Init b233575 [ExoPlayerLib/2.11.1] [gts210vewifi, SM-T813, samsung, 24]
2020-03-02 11:05:37.625 32452-32452/com.google.android.exoplayer2.demo D/EventLogger: state [eventTime=0.00, mediaPos=0.00, window=0, true, BUFFERING]
2020-03-02 11:05:37.626 32452-32452/com.google.android.exoplayer2.demo D/EventLogger: seekStarted [eventTime=0.01, mediaPos=0.00, window=0]
2020-03-02 11:05:37.626 32452-32452/com.google.android.exoplayer2.demo D/EventLogger: positionDiscontinuity [eventTime=0.01, mediaPos=35.00, window=0, SEEK]
2020-03-02 11:05:37.652 32452-32452/com.google.android.exoplayer2.demo D/ViewRootImpl@e14de57[PlayerActivity]: ThreadedRenderer.create() translucent=false
2020-03-02 11:05:37.663 32452-32452/com.google.android.exoplayer2.demo D/InputTransport: Input channel constructed: fd=190
[...]

PR #7074, opened 2020-03-12
keep indices from offsetting when playing from a midroll

TakuSemba:
## Overview

When playing a second mid-roll or any mid-roll after the second mid-roll with an initial seek position, ExoPlayer gets stuck.
The reason is that I think ImaAdsLoader calculate `adGroupIndex` from `podIndex + podIndexOffset` but podIndex always return the correct pod index (even when playback starts over multiple mid points) so no need to shift indicies by podIndexOffset.

https://github.com/google/ExoPlayer/blob/release-v2/extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/ImaAdsLoader.java#L1120

You can reproduce the stuck just playing `VMAP pre-roll single ad, mid-roll standard pods with 5 ads every 10 seconds for 1:40, post-roll single ad` on demo app with an initial seek position 25 * 1000L ms. (which should play the second mid-roll and play content subsequently)

```java
  private void initializePlayer() {
    ...
    player.prepare();
    player.seekTo(25 * 1000);
    ...
}
```

After applying my changes, the stuck will be gone and the ad will be played correctly.

### DIFF
--- a/extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/ImaAdsLoader.java
+++ b/extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/ImaAdsLoader.java
@@ -1059,15 +1059,12 @@ public final class ImaAdsLoader
 
     // IMA indexes any remaining midroll ad pods from 1. A preroll (if present) has index 0.
     // Store an index offset as we want to index all ads (including skipped ones) from 0.
-    if (adGroupIndexForPosition == 0 && adGroupTimesUs[0] == 0) {
-      // We are playing a preroll.
-      podIndexOffset = 0;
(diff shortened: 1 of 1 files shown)


andrewlewis:
> podIndex always return the correct pod index (even when playback starts over multiple mid points) so no need to shift indicies by podIndexOffset.

I think this is a change of behavior due to recent changes in the JavaScript part of the IMA SDK, and @donato may be working on restoring the old behavior now?

TakuSemba:
Hi, @andrewlewis.

> @donato may be working on restoring the old behavior now?

I'm closing this since the behavior was restored.

https://github.com/google/ExoPlayer/issues/7038

Issue #7078, opened 2020-03-13
OfflineLicenseHelper does not have customizable DRM session.


gregmcox:
Hello,
Currently, I have the need to customize the load error handling policy for an offline license request. Since the DefaultDrmSession is created within the helper, this is not really possible. I was wondering the rationale behind not exposing the DrmSession during the creation of OfflineLicenseHelper.

I'm in a use-case where I'm unable to modify the base ExoPlayer libraries and extending OfflineLicenseHelper is not possible as it's a final class.

Please let me know if it's a good idea and I'll submit a pull request.

Thanks!

Documentation:
https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/drm/OfflineLicenseHelper.html

tonihei:
There is no particular reason for that. The LoadErrorHandlingPolicy in DrmSessionManager is relatively new, so that's why this wasn't asked before. 

I'll update the class to accept a customized DefaultDrmSessionManager.

gregmcox:
@tonihei, I've created the pull request. Please let me know if it's sufficient.

tonihei:
Thanks for the pull request - as mentioned above I already had a pending change that fixes the problem on our dev branch, so I'll go ahead and merge this one instead. Please give it a try once ready to see if it fixes your issue.

gregmcox:
Thanks for the fast turn around! Could you reference your commit here? I'm not seeing it on the dev-v2 branch.

PR #7082, opened 2020-03-13
Added load error handling policy to OfflineLicenseHelper.

gregmcox:
This change is in reference to:
https://github.com/google/ExoPlayer/issues/7078

Adds the ability to use a load error handling policy to OfflineLicenseHelper. Since the DRM session is being built within the OfflineLicenseHelper, I've exposed the policy to the end user, similar to that of the KeyRequestParameters.

Added the NonNull annotation since there is an assertion within setLoadErrorHandlingPolicy.

Please let me know if there is anything else.

Thanks!


gregmcox:
@googlebot I signed it!

gregmcox:
@googlebot I fixed it.

PR #7083, opened 2020-03-13
Added load error handling policy to OfflineLicenseHelper.

gregmcox:
This change is in reference to:
#7078

Adds the ability to use a load error handling policy to OfflineLicenseHelper. Since the DRM session is being built within the OfflineLicenseHelper, I've exposed the policy to the end user, similar to that of the KeyRequestParameters.

Added the NonNull annotation since there is an assertion within setLoadErrorHandlingPolicy.

Please let me know if there is anything else.

Thanks!


ojw28:
Thanks for your contribution! I think @tonihei already has a change that'll allow this, so I suggest you follow up on the issue. I'll close this for now because it's sent to the wrong branch (it should be sent to the `dev-v2` branch as per our contribution guidelines), so if we do want it it'll still be necessary to resend it to the `dev-v2` branch instead.

PR #7098, opened 2020-03-17
Update media timestamp regex to allow negative values

matamegger:
The RegEx for matching the `MPEGTS` in an `X-TIMESTAMP-MAP` header of a WebVtt file is extended to allow negative values.

### Reference
- The 2017 [spec](https://tools.ietf.org/html/rfc8216#section-3.5) 
- The 2019 edition of the [spec](https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-05#section-3.5) (still a draft, but Apple calls it 2nd Edition already)

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
@@ -49,7 +49,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
 public final class WebvttExtractor implements Extractor {
 
   private static final Pattern LOCAL_TIMESTAMP = Pattern.compile("LOCAL:([^,]+)");
-  private static final Pattern MEDIA_TIMESTAMP = Pattern.compile("MPEGTS:(\\d+)");
+  private static final Pattern MEDIA_TIMESTAMP = Pattern.compile("MPEGTS:(-?\\d+)");
   private static final int HEADER_MIN_LENGTH = 6 /* "WEBVTT" */;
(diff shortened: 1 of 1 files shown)


ojw28:
@matamegger - I've no particular objection to this change, but it seems very weird that `<MPEG-2 time>` can be negative, given MPEG-2 PES timestamps are by definition unsigned. I can't see anything in the ref'd specifications suggesting that negative timestamps are allowed in this case, which I think I'd expect to be explicit given the default understanding would be that they can only be positive. So I suspect it's not actually valid.

Thoughts? Have you confirmed that other clients successfully handle this case?

matamegger:
Agree with the fact that a MPEG-2 PES may be not negative per definition. In the 2nd Edition of the spec although they seem to distance from PES, and even altered the relevant text and format definition
> This header maps WebVTT cue timestamps to media timestamps
   in other Renditions of the Variant Stream.  Its format is:
>
>   X-TIMESTAMP-MAP=LOCAL:\<cue time>,MPEGTS:\<media time>
   e.g., X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000
>
>   indicating the media time to which the cue time MUST be mapped.

Further `AVPlayer` seems to handle negative `MPEGTS` values without any problem.

That said, I am not 100% sure on the validity by reading the spec, however `AVPlayer` as a product of Apple and Apple being the main driver of HLS, it seemed like we could create "feature parity" here.

ojw28:
Thanks for the clarification. I agree we can merge it. If we consider negative timestamps as non-spec compliant then the change is a no-op for compliant media, and it isn't going to make handling of non-compliant media any worse.

PR #7099, opened 2020-03-17
Extend FrameworkMediaDrm to support PSSHv1 on Amazon FireTV devices

matamegger:
Extend `FrameworkMediaDrm` to support PSSHv1 on Amazon FireTV Gen1, FireTV Gen2 as well as Fire Stick Gen1 and Fire Stick Gen2.

There exists already a workaround for Amazon devices in `FrameworkMediaDrm` regarding PlayReady DRM data. This workaround is extended to also cover Widevine DRM and the issue with PSSHv1 data.

Without this PR the initial key request fails on named devices 
```
E/MediaDrm-JNI( 9635): Illegal state exception: Failed to get key request: DRM vendor-defined error: -2998 (-2998)\
E/ExoPlayerImplInternal( 9635): Playback error.\
E/ExoPlayerImplInternal( 9635): com.google.android.exoplayer2.ExoPlaybackException: com.google.android.exoplayer2.drm.DrmSession$DrmSessionException: android.media.MediaDrm$MediaDrmStateException: Failed to get key request: DRM vendor-defined error: -2998\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.maybeInitCodec(MediaCodecRenderer.java:465)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.reinitializeCodec(MediaCodecRenderer.java:1261)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.onInputFormatChanged(MediaCodecRenderer.java:1111)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.audio.MediaCodecAudioRenderer.onInputFormatChanged(MediaCodecAudioRenderer.java:412)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.render(MediaCodecRenderer.java:647)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:529)\
E/ExoPlayerImplInternal( 9635): 	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:300)\
E/ExoPlayerImplInternal( 9635): 	at android.os.Handler.dispatchMessage(Handler.java:98)\
E/ExoPlayerImplInternal( 9635): 	at android.os.Looper.loop(Looper.java:135)\
E/ExoPlayerImplInternal( 9635): 	at android.os.HandlerThread.run(HandlerThread.java:61)\
```

## Notes
The change will result in parsing the `initData` independent of the actual used PSSH version. Let me know if you would prefer an explicit version check.
But PSSHv0 streams still play back as expected.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
@@ -358,13 +358,15 @@ public final class FrameworkMediaDrm implements ExoMediaDrm<FrameworkMediaCrypto
     }
 
     // Prior to L the Widevine CDM required data to be extracted from the PSSH atom. Some Amazon
-    // devices also required data to be extracted from the PSSH atom for PlayReady.
+    // devices also required data to be extracted from the PSSH atom for PlayReady and Widevine
+    // with PSSHv1 boxes.
     if ((Util.SDK_INT < 21 && C.WIDEVINE_UUID.equals(uuid))
-        || (C.PLAYREADY_UUID.equals(uuid)
+        || ((C.PLAYREADY_UUID.equals(uuid) || C.WIDEVINE_UUID.equals(uuid))
             && "Amazon".equals(Util.MANUFACTURER)
             && ("AFTB".equals(Util.MODEL) // Fire TV Gen 1
                 || "AFTS".equals(Util.MODEL) // Fire TV Gen 2
-                || "AFTM".equals(Util.MODEL)))) { // Fire TV Stick Gen 1
+                || "AFTM".equals(Util.MODEL) // Fire TV Stick Gen 1
+                || "AFTT".equals(Util.MODEL)))) { // Fire TV Stick Gen 2
       byte[] psshData = PsshAtomUtil.parseSchemeSpecificData(initData, uuid);
       if (psshData != null) {
         // Extraction succeeded, so return the extracted data.


ojw28:
If you want to use PSSH V1 boxes, then I think you should also be providing PSSH V0 boxes for backward compatibility. Else any device that doesn't know how to handle PSSH V1 boxes will have the same problem (i.e., I doubt this is specific to Amazon devices). Note that we automatically prefer V0 boxes on older devices for this exact reason:

https://github.com/google/ExoPlayer/blob/r2.11.3/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java#L310

ojw28:
What API levels are the affected devices running by the way? I think we should merge something to make this case more likely to work (in addition to the suggestion of including V0 PSSH atoms in the media itself). I'm not sure exactly what the best change for us to make is though :)...

matamegger:
Thanks for the input!
According to Amazon the devices are on API level 22 
(See https://developer.amazon.com/docs/fire-tv/identify-amazon-fire-tv-devices.html)

ojw28:
I would hypothesise that PSSH V1 boxes wont work on any device running API level 22 (or earlier). The code I referenced above suggests that PSSH V1 boxes are only understood correctly from API level 23. So:

* I think media should also include PSSH V0 boxes, as mentioned above. This is generally required for compatibility with older devices/software that do not know how the structure of PSSH V1 boxes.

To better handle media that doesn't do this:

1. For Widevine, we might want to simply extend the data extraction logic that we always apply for API levels earlier than 21 to also apply on API levels 21 and 22 as well. @jt64 - Can you see any disadvantages of doing this (for context, see the block of code being modified by this pull request - it's basically pulling out the Widevine specific data from the PSSH, and only feeding that to the CDM rather than feeding the entire PSSH box)?
1. For PlayReady, it's unclear to me what we should do. We can definitely extend the Amazon specific workaround to include additional devices. @jt64 - Is the platform code that interprets the PSSH box scheme specific, or common? If it was common on API level 23 and earlier, I would conclude it should be safe for us to do the same thing for PlayReady as for Widevine across all devices?

ojw28:
@jt64 - Could you comment on this? Thanks!

matamegger:
@ojw28 Do you need any further action from my side?

jt64:
> I would hypothesise that PSSH V1 boxes wont work on any device running API level 22 (or earlier).

Correct, at least for widevine

> Can you see any disadvantages of doing this (for context, see the block of code being modified by this pull request - it's basically pulling out the Widevine specific data from the PSSH, and only feeding that to the CDM rather than feeding the entire PSSH box)?

That will be okay

> Is the platform code that interprets the PSSH box scheme specific, or common? If it was common on API level 23 and earlier, I would conclude it should be safe for us to do the same thing for PlayReady as for Widevine across all devices?

The PSSH box processing is scheme-specific, it is not common.

ojw28:
Thanks @jt64! @matamegger - Nothing additional needed from your side.

ojw28:
We merged a modified version of this change, matching my comments above. You can view it here: https://github.com/google/ExoPlayer/commit/918172cca7b1731cd90517a3a88a2994b0310b8c. Please give it a test!

matamegger:
Very well! Thanks for the update! 🙂

PR #7111, opened 2020-03-19
Calculate frame rate for Matroska

Nemiroff:
Added calculation of the frame rate for Matroska


Nemiroff:
> @googlebot I signed it!

ojw28:
I don't think it's accurate to say that this is a frame-rate. It might be equal to the frame-rate in some (possibly many) cases, but I don't think it's true in the general case.

Nemiroff:
defaultSampleDurationNs is Number of nanoseconds (not scaled via TimestampScale) per frame
1000000000ns = 1 second

ojw28:
> Number of nanoseconds (not scaled via TimestampScale) per frame **('frame' in the Matroska sense - one Element put into a (Simple)Block).**

This is discussed in some detail on [this page](https://gitlab.com/mbunkus/mkvtoolnix/-/wikis/Wrong-frame-rate-displayed).

PR #7115, opened 2020-03-19
Use AUD in hevc ts when available 

### LINKED ISSUE #7113: Dynamic HDR SEI messages in h265 bitstreams do not reach media Codec when played from .ts
### [REQUIRED] Issue description
In the 2.11.1 version and all tested previous versions, Dynamic HDR SEI messages in an h265 bitstream (such as defined in e.g.  ETSI TS 103 433) 
do not all reach the MediaCodec properly when the bitstream is packed in a .ts container. This poses a problem for streams that intend to 
indicate video decoder meta data through SEI such as dynamic HDR systems.

It seems related to the current H265Reader.java implementation.

Since all of these dynamic HDR streams I have encountered do use AUD's, the simplest modification seems be to use these when available in a stream
in the sample reader. Will open a pull request to share this modification.

It is also observed that without the AUD modification some of the .ts content does not play smoothly. On Qualcomm SD855-based devices this also 
triggers mediacodec errors logged in the logcat at the time of the hiccup. 

"03-16 16:37:22.195   947 21162 E OMX-VDEC-1080P: Unable to convey fps info to driver, performance might be affected"

Unfortunately the content with the the hiccup is not content that we own or are allowed to distribute.

### [REQUIRED] Reproduction steps
Play Hevc Dynamic HDR .ts video with exoplayer

### [REQUIRED] Link to test content
The content was mailed to dev.exoplayer@gmail.com

### [REQUIRED] A full bug report captured from the device
No applicable for the given content

### [REQUIRED] Version of ExoPlayer being used
2.11.1 ,2.10.5 and 2.10.2 

### [REQUIRED] Device(s) and version(s) of Android being used
Samsung Tab S6 (Android 9)

dbPhilips:
Add support for using AUD on HEVC TS to fix #7113

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/H265Reader.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/H265Reader.java
@@ -47,6 +47,7 @@ public final class H265Reader implements ElementaryStreamReader {
   private static final int VPS_NUT = 32;
   private static final int SPS_NUT = 33;
   private static final int PPS_NUT = 34;
+  private static final int AUD_NUT = 35;
   private static final int PREFIX_SEI_NUT = 39;
   private static final int SUFFIX_SEI_NUT = 40;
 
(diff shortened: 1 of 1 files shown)


ojw28:
@AquilesCanta - Could you take a look at this?

ojw28:
Thanks for your contribution. We will merge a different fix for this today, and update https://github.com/google/ExoPlayer/issues/7113 to explain why and provide a bit more detail.

Issue #7119, opened 2020-03-21
Could we avoid default method interfaces in Java at least in internal implementations?


dryganets:
I have noticed that the default methods got higher and higher adoption in the ExoPlayer library.

The problem is that not every build toolchain supports it equally well. We have some issues using BUCK for example. 

While most of the issues could be work-arounded by just overriding the default implementation, some of the cases can't be fixed that easy.

Example of that is DUMMY DrmFactory which is always created by default.

I understand that code looks a bit better with default implementation, but NoOp implementation should be able to fix it as well without burden of de-sugaring and extra build infrastructure issues.

What do you think about banning default methods completely?
Would you approve a PR which removes at least internal usage of the default implementation?

ojw28:
Default methods aren't just about making code look nicer. They add significant additional flexibility because it becomes possible to add to an interface without breaking all implementations of that interface (you can achieve something similar with an abstract class, but that approach has the significant limitation of not allowing multiple inheritance).

* Please could you point us at a specific issue or issues on the BUCK issue tracker related to default methods, so we can better understand the problems you're running into? Without knowledge of the specifics, it's not really possible for us to have an informed opinion on your particular case.
* If any other developers have feedback on this, to give us a better overall picture, that would be appreciated. What other toolchains are impacted, and why? What adoption do these toolchains have? Why aren't the toolchains being fixed to support (not particularly new) language features?

> What do you think about banning default methods completely? Would you approve a PR which removes at least internal usage of the default implementation?

This issue does not yet make a good case for doing this. We would need a lot more specifics to even form an opinion, starting with those requested above.

dryganets:
For my particular case, I'd rather make a small PR to the ExoPlayer.

We are using an older version of BUCK for the developer builds, we are working on upgrade in the mean time.

I understand the value of defaults methods, just saying that there is a cost for some users of the library. I agree that everyone that uses modern toolchains like d8 should be able to deal with it.

If somebody else has the similar issues he probably chime in into the conversation, I've seen a number of AbstractMethod errors complains in the bug tracker, but I guess most of the cases were related to miss-configuration fo the gradle build-script.

ojw28:
[...]

PR #7139, opened 2020-03-25
DUMMY drm manager overrides default methods.

dryganets:
DUMMY drm manager is internal class which is created for every HLS/DASH session.

It addresses this issue: https://github.com/google/ExoPlayer/issues/7119.

It is impossible to work-around AbstractMethodError issue in toolchains without java 8 de-sugaring support.

It could be addressed in many different ways, pick the one which suits the ExoPlayer the best.

Out of blue:
- Allow passing the implementation to Builder constructor, so external developer will be able to override the methods
- Remove the defaults from this particular interface
- Implement all methods for DUMMY

It would be wonderful if you could publish 2.11.4 with one of this fixes applied.


dryganets:
@googlebot I signed it!

PR #7184, opened 2020-04-04
Add HLS support for "stpp.*" codec support for SMPTE-TT fmp4 subtitle…

stevemayhew:
We will enter a bug for this shortly. 

ExoPlayer needs a codec to decide among WEBVTT and TTML decoder mimeType.
Apple describes IMSC1 in MP4 in
[RFC-8216 Section 3.6](https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-04#section-3.6).

The DASH manifest specifies the SMPTE-TT captions in the codecs in the manifest
(from W3C [TTML Profiles for Internet Media Subtitles and Captions 1.1](https://www.w3.org/TR/ttml-imsc1.1/#general-0).
DASH just doesn't require the rendition linking, but HLS does.

Apple implies the CODECS attribute of the variant needs to be do this.  That is
with SHOULD and MAY language to imply the codec to use for it in the
[Authoring Guidelines](https://developer.apple.com/documentation/http_live_streaming/hls_authoring_specification_for_apple_devices)

This change defaults to WebVTT if no codec is specifed (same as current behavior) otherwise it picks it from the variants
referencing the media.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
@@ -462,7 +462,7 @@ public final class HlsPlaylistParser implements ParsingLoadable.Parser<HlsPlayli
           variant = getVariantWithSubtitleGroup(variants, groupId);
           if (variant != null) {
             @Nullable
-            String codecs = Util.getCodecsOfType(variant.format.codecs, C.TRACK_TYPE_AUDIO);
+            String codecs = Util.getCodecsOfType(variant.format.codecs, C.TRACK_TYPE_TEXT);
             sampleMimeType = MimeTypes.getMediaMimeType(codecs);
             if (sampleMimeType == null) {
               sampleMimeType = MimeTypes.TEXT_VTT;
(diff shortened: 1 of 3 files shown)


jparekh-tivo:
Linked issue is - https://github.com/google/ExoPlayer/issues/7185

ojw28:
Please could you get rid of the 7 unnecessary files being added as part of this pull request?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.java
```diff
@@ -458,7 +458,18 @@ private static HlsMasterPlaylist parseMasterPlaylist(LineIterator iterator, Stri
           }
           break;
         case TYPE_SUBTITLES:
-          formatBuilder.setSampleMimeType(MimeTypes.TEXT_VTT).setMetadata(metadata);
+          sampleMimeType = MimeTypes.TEXT_VTT;
```
stevemayhew, line 461: Default is original behavior (assume VTT)...

stevemayhew:
@ojw28 Sorry, there was also an error in the merge from our commit (that is release branch based) that I fixed (track type).  The test case passes and covers the default case (select VTT if no matching codecs in the variant CODEC) works fine.   We could also add the explicit CODEC of "wvtt", I did not see any reference to this though outside of Apple guidelines.

PR #7199, opened 2020-04-06
Fix for issue #7181 stuck CEA-608 captions

### LINKED ISSUE #7181: CEA-608 caption stuck on screen for several minutes
#### Summary 
CEA-608 caption stuck on screen for several minutes (mostly  during commercial breaks)

#### Root Cause
On certain Live TV channels the Caption from Live TV program stays on screen throughout the commercial. The commercials in such case don’t have the caption. But the the outdated caption from the program stays on screen.
 
The root cause is that the caption provider is not inserting the control words such as End_Of_Caption(EOC 14 2F), Erase_Displayed_Memory(EDM 14 2C), Eras_NonDisplay_Memory(ENM 14 2E) on CEA-608 caption services. Since the commercials don’t have caption, the previously displayed captions are not erased. 
 
CEA-608 spec Annex C.9 recommends Automatic Caption Erasure (Preferred).  FCC rules describe events, other than reception of commands intended to erase, which trigger the erasure of caption display and memory. These include the viewer switching receiver channel, data channel, or the selected field; loss of valid data; or the viewer selecting a non-captioning receiver function. 
CEA-608 Annex C.9 prefers If no data are received for the selected caption channel within a given time, the decoder should automatically erase the caption. Such erasure may supersede the intentions of the caption service providers and institute one maximum display time for all captioning services. If such a timeout is deemed necessary, however, the time limit should be no less than 16 seconds.

### Proposed Solution
Pull request is in progress.  We propose adding a check in [Decoder.dequeueOutputBuffer()](https://github.com/google/ExoPlayer/blob/49910fe72579c4e55e14b47b0cf3d3ced9474219/library/core/src/main/java/com/google/android/exoplayer2/text/cea/CeaDecoder.java#L91) that will track the time since the last erasure control word was processed and if it exceeds the 16 second time out, generate an erasure sequence (Cue?)

### Reproduction steps
Using the sample HLS stream, play to the first commercial break 30 seconds into the stream. Observe the caption "A Nationwide ..." stays on the screen during the entire ad series

### Link to test content

Content is broadcast from cable, so link was sent to exoplayer dev email

N/A

2.11.4 

Tested and reproduced on Arris and Amino AndroidTV devices running Android Pie.  Reproduction is 100%.

stevemayhew:
This pull request proposes a fix for issue #7181 by clearing the caption after a time delay as outlined in CEA-608 spec Annex C.9

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -1041,7 +1041,8 @@ public final class Cea608Decoder extends CeaDecoder {
         if (timeElapsed >= VALID_DATA_CHANNEL_TIMEOUT_MS) {
           // Force erase captions. There might be stale captions stuck on screen.
           // (CEA-608 Annex C.9)
-          flush();
+          cues = Collections.emptyList();
+          ccTimeOutCounter = C.TIME_UNSET;
         }
       }
   }
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -31,6 +31,7 @@ import com.google.android.exoplayer2.text.Subtitle;
 import com.google.android.exoplayer2.text.SubtitleDecoder;
 import com.google.android.exoplayer2.text.SubtitleInputBuffer;
 import com.google.android.exoplayer2.util.Assertions;
+import com.google.android.exoplayer2.util.Clock;
 import com.google.android.exoplayer2.util.Log;
 import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.ParsableByteArray;
@@ -260,18 +261,23 @@ public final class Cea608Decoder extends CeaDecoder {
   // Static counter to keep track of last CC rendered. This is used to force erase the caption when
   // the stream does not explicitly send control codes to remove caption as specified by
   // CEA-608 Annex C.9
-  private long ccTimeOutCounter = C.TIME_UNSET;
+  private long lastCueUpdateMs = C.TIME_UNSET;
   private boolean captionEraseCommandSeen = false;
   // CEA-608 Annex C.9 propose that if no data are received for the selected caption channel within
   // a given time, the decoder should automatically erase the caption. The time limit should be no
   // less than 16 seconds
-  public static final int VALID_DATA_CHANNEL_TIMEOUT_MS = 16000;
 
-  public Cea608Decoder(String mimeType, int accessibilityChannel) {
+  // This value is set in the constructor. The automatic erasure is disabled when this value is 0
+  private long validDataChannelTimeoutMs = 0;
+  private Clock clock;
+
+  public Cea608Decoder(String mimeType, int accessibilityChannel, long timeoutMs, Clock clock) {
     ccData = new ParsableByteArray();
     cueBuilders = new ArrayList<>();
     currentCueBuilder = new CueBuilder(CC_MODE_UNKNOWN, DEFAULT_CAPTIONS_ROW_COUNT);
     currentChannel = NTSC_CC_CHANNEL_1;
+    validDataChannelTimeoutMs = timeoutMs;
+    this.clock = clock;
     packetLength = MimeTypes.APPLICATION_MP4CEA608.equals(mimeType) ? 2 : 3;
     switch (accessibilityChannel) {
       case 1:
@@ -320,7 +326,7 @@ public final class Cea608Decoder extends CeaDecoder {
     repeatableControlCc2 = 0;
     currentChannel = NTSC_CC_CHANNEL_1;
     isInCaptionService = true;
-    ccTimeOutCounter = C.TIME_UNSET;
+    lastCueUpdateMs = C.TIME_UNSET;
   }
 
   @Override
@@ -433,8 +439,8 @@ public final class Cea608Decoder extends CeaDecoder {
     if (captionDataProcessed) {
       if (captionMode == CC_MODE_ROLL_UP || captionMode == CC_MODE_PAINT_ON) {
         cues = getDisplayCues();
-        if (!captionEraseCommandSeen) {
-          ccTimeOutCounter = System.currentTimeMillis();
+        if ((validDataChannelTimeoutMs != 0) && !captionEraseCommandSeen) {
+          lastCueUpdateMs = clock.elapsedRealtime();
         }
       }
     }
@@ -1036,13 +1042,14 @@ public final class Cea608Decoder extends CeaDecoder {
 
   protected void clearStuckCaptions()
   {
-      if (ccTimeOutCounter != C.TIME_UNSET) {
-        long timeElapsed = System.currentTimeMillis() - ccTimeOutCounter;
-        if (timeElapsed >= VALID_DATA_CHANNEL_TIMEOUT_MS) {
+      if ((validDataChannelTimeoutMs != 0) &&
+              (lastCueUpdateMs != C.TIME_UNSET)) {
+        long timeElapsed = clock.elapsedRealtime() - lastCueUpdateMs;
+        if (timeElapsed >= validDataChannelTimeoutMs) {
           // Force erase captions. There might be stale captions stuck on screen.
           // (CEA-608 Annex C.9)
           cues = Collections.emptyList();
-          ccTimeOutCounter = C.TIME_UNSET;
+          lastCueUpdateMs = C.TIME_UNSET;
         }
       }
   }
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
@@ -257,6 +257,16 @@ public final class Cea608Decoder extends CeaDecoder {
   // service bytes and drops the rest.
   private boolean isInCaptionService;
 
+  // Static counter to keep track of last CC rendered. This is used to force erase the caption when
+  // the stream does not explicitly send control codes to remove caption as specified by
+  // CEA-608 Annex C.9
+  private long ccTimeOutCounter = C.TIME_UNSET;
+  private boolean captionEraseCommandSeen = false;
+  // CEA-608 Annex C.9 propose that if no data are received for the selected caption channel within
+  // a given time, the decoder should automatically erase the caption. The time limit should be no
+  // less than 16 seconds
+  public static final int VALID_DATA_CHANNEL_TIMEOUT_MS = 16000;
+
   public Cea608Decoder(String mimeType, int accessibilityChannel) {
     ccData = new ParsableByteArray();
     cueBuilders = new ArrayList<>();
@@ -310,6 +320,7 @@ public final class Cea608Decoder extends CeaDecoder {
     repeatableControlCc2 = 0;
     currentChannel = NTSC_CC_CHANNEL_1;
     isInCaptionService = true;
+    ccTimeOutCounter = C.TIME_UNSET;
   }
 
   @Override
@@ -334,6 +345,7 @@ public final class Cea608Decoder extends CeaDecoder {
     ByteBuffer subtitleData = Assertions.checkNotNull(inputBuffer.data);
     ccData.reset(subtitleData.array(), subtitleData.limit());
     boolean captionDataProcessed = false;
+    captionEraseCommandSeen = false;
     while (ccData.bytesLeft() >= packetLength) {
       byte ccHeader = packetLength == 2 ? CC_IMPLICIT_DATA_HEADER
           : (byte) ccData.readUnsignedByte();
@@ -343,7 +355,6 @@ public final class Cea608Decoder extends CeaDecoder {
       // TODO: We're currently ignoring the top 5 marker bits, which should all be 1s according
       // to the CEA-608 specification. We need to determine if the data should be handled
       // differently when that is not the case.
-
       if ((ccHeader & CC_TYPE_FLAG) != 0) {
         // Do not process anything that is not part of the 608 byte stream.
         continue;
@@ -353,7 +364,6 @@ public final class Cea608Decoder extends CeaDecoder {
         // Do not process packets not within the selected field.
         continue;
       }
-
       // Strip the parity bit from each byte to get CC data.
       byte ccData1 = (byte) (ccByte1 & 0x7F);
       byte ccData2 = (byte) (ccByte2 & 0x7F);
@@ -423,6 +433,9 @@ public final class Cea608Decoder extends CeaDecoder {
     if (captionDataProcessed) {
       if (captionMode == CC_MODE_ROLL_UP || captionMode == CC_MODE_PAINT_ON) {
         cues = getDisplayCues();
+        if (!captionEraseCommandSeen) {
+          ccTimeOutCounter = System.currentTimeMillis();
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


sneelavara:
@googlebot I fixed it.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -257,6 +257,16 @@
...
 
+  // Static counter to keep track of last CC rendered. This is used to force erase the caption when
+  // the stream does not explicitly send control codes to remove caption as specified by
+  // CEA-608 Annex C.9
+  private long ccTimeOutCounter = C.TIME_UNSET;
+  private boolean captionEraseCommandSeen = false;
+  // CEA-608 Annex C.9 propose that if no data are received for the selected caption channel within
+  // a given time, the decoder should automatically erase the caption. The time limit should be no
+  // less than 16 seconds
+  public static final int VALID_DATA_CHANNEL_TIMEOUT_MS = 16000;
```
zsmatyas: What happens when there is, for example, a song playing with captions "♫♫". Those captions are usually shown for a many seconds. Are those repeatedly sent, processed and kept on the screen correctly?

When restarting the timer, have you considered special cases, like:
- receiving a duplicate of a previous command
- receiving byte to a CEA608 mode that is not part of the captions, like TEXT or RDS mode
- receiving an invalid command
- when process_cc_data_flag is set to false
- receiving a value with incorrect CRC
- receiving bytes that are marked not to process.
- receiving bytes to a different channel, not the currently selected ones (like receiving CC3 bytes, but nothing for CC1)
- in ROLL_UP mode, when there is no one talking for a prolonged period of time, so no roll event happens.

Did you have a chance to run certification test cases with the change? Like the Sarnoff test cases for CEA608 and CEA708?

sneelavara: Thank you @zsmatyas.

I don’t see the captions "♫♫”. Its sent once followed by null bytes (0x0, 0x0) and EOC. In this fix, the max displayed period is 16 seconds. If EOC arrives before the timeout, the caption is erased and timer is stopped. Otherwise the caption is erased after 16 second timeout.

Timer is started in Cea608Decoder.decode() function when the flag captionDataProcessed is set and caption mode is ROLL_UP and PAINT_ON

> When restarting the timer, 
> * receiving a duplicate of a previous command

Cea608Decoder ignores duplicate commands in isRepeatedCommand(). In such case the timer continues to be in its current state. I.e. if already started it continues, otherwise timer is not started on duplicate command

> * receiving byte to a CEA608 mode that is not part of the captions, like TEXT or RDS mode

Timer is started only in ROLL_UP and PAINT_ON mode.

> * receiving an invalid command

I believe as per CEA-608, there can’t be an invalid control word. Annex D.2 mentions - “Encoder manufacturers shall not insert these invalid control codes on line 21, field 1, as services designed for line 21, field 2, use these codes as valid data”. I don’t see any check done for invalid command in CeaDecoder path. So, for any reason if Cea608Decoer gets invalid command code the timer is started as it is done on normal Control codes. By the way, timer is not started on control words such as End_Of_Caption(EOC 14 2F), Erase_Displayed_Memory(EDM 14 2C), Eras_NonDisplay_Memory(ENM 14 2E) since these commands erase the captions.

> * when process_cc_data_flag is set to false

This is filtered at UserDataReader. If process_cc_data_flag is not set, the consumeCcData() function discards the ccDataBuffer. So the Cea608Decoder won’t see those triplets and the timer is not started

> * receiving a value with incorrect CRC

Please correct me in case I am missing something here. The cc_data syntax doesn’t describe CRC. I don’t seem to find CRC in CEA-608 either. 

> * receiving bytes that are marked not to process.

Could you please elaborate on this. Are you referring to cc_valid flag? If cc_valid flag is not set the, the timer is not started on reception of such triplets.  

> * receiving bytes to a different channel, not the currently selected ones (like receiving CC3 bytes, but nothing for CC1)

Timer is not started in such case.
> * in ROLL_UP mode, when there is no one talking for a prolonged period of time, so no roll event happens.

The previously displayed caption is erased after 16 seconds timeout if no roll event or no valid data is received (Captions are erased after 16 seconds even if null bytes (0x0, 0x0) are seen for 16 seconds). 

No, I did not run the certification tests with this change. I will have to check the availability of Sarnoff streams. 

Thanks,
 Sadashiva N.

icbaker:
> @googlebot I fixed it.

There are some commits authored by `sneelavara@SJC-1SNEELAVA-M.local` which doesn't have an associated CLA. Can you either add that address to your existing CLA or rebase the commits to use your other email address? I think that will then resolve this.

sneelavara:
@googlebot I fixed it

Thank you @icbaker.  Rebased the commits to sneelavara@gmail.com. Hope that resolves this. issue.

icbaker:
Thanks!

I think you now just need to do what the robot says in https://github.com/google/ExoPlayer/pull/7199#issuecomment-610507573 and reply with a message that **only** contains `@googlebot I consent.`.

sneelavara:
@googlebot I consent.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
icbaker: Thanks for your comments @zsmatyas - as you've contributed a fair amount of the CEA support to ExoPlayer it's helpful to hear your opinions.

From my reading of the spec this is an optional feature.

It seems there is a non-zero risk that this inadvertently removes a subtitle that was intended to be shown for longer than 16s. The benefit of this change is that invalid streams that are missing the End_Of_Caption control word don't have their subtitles stuck on screen for longer than intended.

Currently I'm inclined to keep the current behaviour and reject this PR, since it seems better for invalid data to result in "stuck" rendering than valid data to result in "early-removal".

> No, I did not run the certification tests with this change. I will have to check the availability of Sarnoff streams.

If you're able to do that it would be excellent. Unfortunately we don't have access to these test streams.

stevemayhew: Thanks Ian (@icbaker) Sada (@sneelavara) works with me here at TiVo, we of course have the Sarnaoff transport stream files for testing our STB's, we just need to work out how to get access to use these for this testing.  

BTW,  Since TiVo delivers broadcast content we have been mandated by the FCC to provide CC support since day one.  This clear logic has been in our STB's for years.  Very grateful for @zsmatyas insight on this, we will work to make this bullet proof before we ask you @icbaker to make it merge ready.

zsmatyas: I had the privilege of debugging many complaints for broken captions. My impression was that content providers (various live channels creating content or only subtitles to existing content) around North America seem to interpret the CEA608 captions in different ways - as originally it was targeting over-the-air analogue content and was not prepared for the use cases we have today.
I have met many issues that are caused by undefined values, unexpected ordering, ambiguous wording in the standard and similar - and the issue is not our interpretation of the expectation, but the interpretation of the content providers or the tools they use.

I agree that the commercials not having any subtitle commands interrupting normal content with correct captions cause issues and the "stuck" lines are annoying, but we could not propose a way that has more benefit than side effects and will correctly work with the thousands of untested channels we do not have access to at the moment. For example, if we test it and fix all issues with the current user base, some future application released in a county not yet tested but also using CEA608 still might have unexpected issues.

The Sarnoff test cases are very useful because they exercise the functionality in many unexpected ways. One of them was, for example, "flashing" caused by switching the visible and invisible buffers without any other commands received. This proposed code change might regress such test cases as it clears the caption buffers when the test case expects them not to be modified (only the visibility is flipped). 

One more use case to consider, when the normal content is interrupted by ads without captions: when returning to the normal content, the buffers for pending (but not visible) captions should have the content from before the advertisements were started, otherwise the first set of captions will not be present and the state of the caption decoder might be altered. But this is a general issue with today's solution of injecting commercials into content previously correctly working with 608 captions - you need to consider the timing and the state of the decoder. If the content provider does that correctly, we might expect them to start with a "clear" command.

The proposed change might probably be useful for many developers. But in general, I would suggest not merging this change without a flag to enable / disable the proposed "clear" functionality as most probably it should not be applied universally. One path forward could be have this feature turned off by default, so developer need to opt-in for the effects.

stevemayhew: @zsmatyas I'm sure.  Captions are often problematic.  The use case we see this most on is live authored streams (news channels like CNN etc).  We certainly welcome your help for a best solution to this, looks like you're living this problem too.

icbaker: We discussed this in the team, and decided we'd like to accept the PR as the default behaviour, but provide developers a way to tweak it/turn it off.

I think the easiest way to do that is to add a `long timeoutMs` constructor parameter to `Cea608Decoder`, and set it to 16000 in `SubtitleDecoderFactory.DEFAULT`. Passing zero to this parameter should disable the timeout.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
```diff
@@ -423,6 +433,9 @@ protected void decode(SubtitleInputBuffer inputBuffer) {
     if (captionDataProcessed) {
       if (captionMode == CC_MODE_ROLL_UP || captionMode == CC_MODE_PAINT_ON) {
         cues = getDisplayCues();
+        if (!captionEraseCommandSeen) {
+          ccTimeOutCounter = System.currentTimeMillis();
```
icbaker: I think `ccTimeOutCounter` isn't quite the right name for this variable - it's not counting the timeout but rather recording the start of the timer - i.e. the last time we received visible cue data.

Maybe rename it to `lastCueUpdateMs` or something?

Also: Can you pass a `com.google.android.exoplayer.util.Clock` into the constructor and use that here rather than statically accessing the system clock? It will make it easier to test (if/when we write tests for this class...)

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
```diff
@@ -1296,6 +1296,14 @@ public static int getArgbColorFromCeaColor(int red, int green, int blue, int opa
     }
   }
 
+  protected void clearStuckCaptions()
+  {
+    // Do nothing for CEA-708.
+    // As per spec CEA-708 Caption text sequences shall be terminated by either the start of a new
+    // DTVCC Command, or with an ASCII ETX (End of Text) (0x03) character when no other DTVCC
+    // Commands follow.
+  }
```
icbaker, line 1305: I'm not a fan of having to add a method to CeaDecoder that only makes sense for Cea608.

I wonder (I haven't tested this) if you could add this functionality to `isNewSubtitleDataAvailable()`: If the timeout has expired then set `cues` to `emptyList` - would that work?

icbaker (changes requested):

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
icbaker: Ah: And I think we should use `SystemClock.elapsedRealtime()` rather than `System.currentTimeMillis()` just so we don't need to worry about the clock changing underneath us.

More info on the difference:
https://developer.android.com/reference/android/os/SystemClock

stevemayhew: Yes, you mean use `Clock.elapsedRealtime()` right?  The testable [Clock](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/util/Clock.html) interface?  Which I agree, is a great idea.

icbaker: Sorry - yeah.

We should: a) use Clock and b) use `elapsedRealtime()`

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
stevemayhew: I thought Sada (@sneelavara) had found that there was a similar issue and fix in CEA-708?

Also, you should know, in our internal review, there was some debate over if it was best to depend on the fact that `dequeueOutputBuffer()` is called every render cycle to run the timer.  Probably not an issue, but I've quoted that review comment in the code here: https://github.com/google/ExoPlayer/pull/7199/files/697dafd4cf2766f53e64b92078b5643c6e2f5e14#r406338568

library/core/src/main/java/com/google/android/exoplayer2/text/cea/CeaDecoder.java
```diff
@@ -97,6 +97,8 @@ public SubtitleOutputBuffer dequeueOutputBuffer() throws SubtitleDecoderExceptio
     if (availableOutputBuffers.isEmpty()) {
       return null;
     }
+    // check if 608 decoder needs to clean up the stale caption
+    clearStuckCaptions();
     // iterate through all available input buffers whose timestamps are less than or equal
```
stevemayhew, line 102: Comment from our internal review, about where to put the check function ('clearStuckCaptions()`):

>  @bjitivo(Bryan Ischo @ TiVo Corp.) 22 hours ago
> I don't know this code well but the way I read it, the only time we're going to check on whether or not to clear stuck captions is when a non-empty output buffer has arrived. But wouldn't that clear captions anyway?
> 
> And if cc are stuck on the screen and no non-empty output buffers are not coming, then this code path will not be entered and the cc will not be cleared.
> 
> Why aren't you just resetting a 16 second timer whenever cc is displayed and then clearing captions when the 16 second timer goes off (which implies no captioning commands for 16 seconds)?

>   @stevemayhew
> stevemayhew Pending
> Using timers in a handler is problematic,  but it is pretty unclear that dequeueOutputBuffer() is called every call to TextRenderer.render(). The render() method is called every 16ms for all ExoPlayer renderers, so it is probably clearer to either put it in setPosition() or even better add an abstract method to the SubtitleDecoder (checkStuckCaptions() with a default that does nothing and override that, call it from the render() method in TextRenderer

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
sneelavara: @icbaker isNewSubtitleDataAvailable() is called when there is something to decode on input buffer. That may not be called regularly and we may skip checking for time expiry.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java
sneelavara: @icbaker Thank you for the feed back. Modified 
- Cea608Decoder to use com.google.android.exoplayer.util.Clock.
- The timeout is. now passed to  Cea608Decoder constructor
- The ccTimeOutCounter is changes lastCueUpdateMs

sneelavara: @icbaker Please let us know if there is any other changes needed

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
stevemayhew: @icbaker using `isNewSubtitleDataAvailable()` only works if there is an input buffer, we need the timeout check called most very render() cycle.  So `dequeueOutputBuffer()` is the best method we have without adding a new method.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/CeaDecoder.java
stevemayhew: again, `dequeueOutputBuffer()` is the best, unless we want to add methods to `CeaDecoder`

icbaker:
Thanks for the changes - I'll work on getting this merged.

icbaker:
This is merged into dev-v2, but the git history got a bit messed so I'm having to manually close it as github hasn't detected it as "merged".

PR #7227, opened 2020-04-13
Saving subtitle track name

vadimdolgachev:
Some files have several subtitles in the same language, such as forced and regular subtitles. In this case, a name is added for the subtitle track.
This code allows to save the name of the subtitle track.
The subtitle name and language name can also be seen in the VLC player.
Sample video with name subtitle track: https://drive.google.com/file/d/1WoWcL9B4cVd1fsDEAtYpXDmG0cMF0F_3/view?usp=sharing

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/Format.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/Format.java
@@ -626,6 +626,28 @@ public final class Format implements Parcelable {
       @Nullable DrmInitData drmInitData) {
     return createTextSampleFormat(
         id,
+        /* label= */ null,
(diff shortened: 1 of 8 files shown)


ojw28:
As per our [contribution guidelines](https://github.com/google/ExoPlayer/blob/release-v2/CONTRIBUTING.md), you need to send pull requests to our `dev-v2` branch.

Closing this one since it's to the wrong branch. I think we'd be happy to merge an equivalent change to `dev-v2`, so please do send one. Note that `Format.Builder` has been added on that branch, and likely makes this change quite a bit smaller. Thanks!

Issue #7231, opened 2020-04-14
Transition delay in exoplayer playlist


Clement-Jean:
Working with playlist on the ExoPlayer, I managed the setting of the playlist by following the documentation.

```
val firstSource = ProgressiveMediaSource.Factory(...).createMediaSource(firstVideoUri);
val secondSource = ProgressiveMediaSource.Factory(...).createMediaSource(secondVideoUri);

val playlist = ConcatenatingMediaSource(firstSource, secondSource);

player.prepare(playlist)
```

But I noticed that the transition between video is instantaneous. However, I would like to add a waiting stage between each video, exactly like youtube does (let's say 5 sec).

Is it already handled by the Exoplayer itself or not? If not what is a clean way to solve the problem ?

Originally posted on [StackOverflow](https://stackoverflow.com/questions/61207783/transition-delay-in-exoplayer-playlist)

marcbaechinger:
A simple thing which comes to my mind is using a [`SilenceMediaSource`](https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/source/SilenceMediaSource.java) which you can put in between the actual tracks.  You can then listen to `onPositionDiscontinuity()` to trigger actions like putting a view on top when the silence media source starts and remove the view again when the silence ends.

tonihei:
Another option that is only available on the dev branch so far is to tell the player to pause at the end of each media item with `player.setPauseAtEndOfMediaItems(true)`. And as soon as the player is paused (listen to `onPlayWhenReadyChanged` with reason `PLAY_WHEN_READY_CHANGE_REASON_END_OF_MEDIA_ITEM`), you can start your waiting UI animation (or whatever you are planning to do with this stage), and after 5 seconds you can call `player.play()` again to continue with the next item.

Clement-Jean:
@tonihei @marcbaechinger  Do you know when this setPauseAtEndOfMediaItems(true) feature will be available?

tonihei:
We can't give exact release dates yet, but probably within the next month or two depending on how fast we can cut a release branch with a sensible stable state.

Clement-Jean:
Is there a specific reason why the user can't set a tag for a `SilenceMediaSource` ? If not I'm willing to do a pull request for something like:

``
SilenceMediaSource.Factory().setDuration(...).setTag(...).build()
``

Because in my use case, I'm relying on the tag to know the id of my current playing video.

[...]

PR #7241, opened 2020-04-16
Origin/dev silencemedia factory

Clement-Jean:
## What

Added a Factory to SilenceMediaSource to be able to create it like the other MediaSource. Also added a tag in the class so that users can get more information on this object.

## Why

The idea comes from the following [issue](https://github.com/google/ExoPlayer/issues/7231). In the use case, the user would like to create a transition between two medias in a playlist. To do that we could use a SilenceMediaSource with a tag in order to know more about the media in the `onPositionDiscontinuity()` method. The user would only have to call `player.getCurrentTag()`.


Clement-Jean:
@googlebot I signed it!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

icbaker:
The diff in this PR looks enormous (426 files & 200 commits) - it seems you might not have merged/rebased correctly and a lot of irrelevant changes (by exoplayer-team) are showing up.

Can you fix it so that the diff only shows the files you intend to change?

Clement-Jean:
@icbaker sorry, I apparently did a mistake. Do you have any clue how to change this in this pull request ?

marcbaechinger (commented):
In `prepareSourceInternal(@Nullable TransferListener mediaTransferListener)` you need to pass the `tag` to the constructor of the `SinglePeriodTimeline` as the last parameter. This makes the tag available in the timeline.

Clement-Jean:
How people are supposed to know that ?

marcbaechinger (commented):
Developers who implement the `MediaSource` interface know they have to provide a timeline with the information of the media source. The java doc of the `MediaSource` interface tells them to provide the timeline.

library/core/src/main/java/com/google/android/exoplayer2/source/SilenceMediaSource.java
```diff
@@ -33,6 +33,30 @@
 /** Media source with a single period consisting of silent raw audio of a given duration. */
 public final class SilenceMediaSource extends BaseMediaSource {
 
+  public static final class Factory {
```
marcbaechinger, line 37: Please provide a short java doc for public constructors and methods which are not overridden. Here and below.

library/core/src/main/java/com/google/android/exoplayer2/source/SilenceMediaSource.java
```diff
@@ -33,6 +33,30 @@
...
+
+    @Nullable private Object tag;
+    private long duration;
+    private boolean isCreateCalled;
+
+    public SilenceMediaSource.Factory setTag(Object tag) {
+      Assertions.checkState(!isCreateCalled);
```
marcbaechinger, line 52: We made factories of other media sources re-usable. So create can be called multiple times.  I'd probably remove this for consistency.

Clement-Jean: I just followed how the ProgressiveMediaSource.Factory is implemented. So should I remove the boolean and the assertion ?

icbaker:
> @icbaker sorry, I apparently did a mistake. Do you have any clue how to change this in this pull request ?

I think you might have started from the `release-v2` branch but proposed this PR back to the `dev-v2` branch. We do all our development on the `dev-v2` branch, but cherrypick across to `release-v2` to support the point releases - all the spurious commits in this PR seem to be from `release-v2`.

You could try rebasing your current branch onto the head of `dev-v2`. It might work, or it might be a disaster.

If it's a disaster, I'd suggest starting again from `dev-v2`, creating a new branch, and then cherrypicking your changes across from this branch. You'll probably have to then submit a new PR, but that's fine we can just abandon this one.

Clement-Jean:
@icbaker  I can just close this PR and restart right ? I'm just trying to finish my work with @marcbaechinger

icbaker:
> @icbaker I can just close this PR and restart right ? I'm just trying to finish my work with @marcbaechinger

Yeah, that should be totally fine. @marcbaechinger will understand - if you address his comments so far and raise a new PR, we'll be able to follow along. Thanks!

PR #7244, opened 2020-04-16
Cancel hls chunk download and discard upstream

tvarga-dss:
This PR is building on top of https://github.com/google/ExoPlayer/pull/6321 by adding the ability to cancel a chunk download.


tvarga-dss:
Had a bit of trouble with how my local checkout was setup, but I think now this PR should have only the relevant commits.

tvarga-dss:
Saw your comment at at https://github.com/google/ExoPlayer/pull/6321#issuecomment-614637660 @stevemayhew I didn't see any `ConcurrentModificationException` since the merge with `dev-2`.

I'd still much appreciate your input on the proposed changes in this pull request.

tonihei:
CLA is fine for Tamás. The bot is missing consent for the diffbase author, but this will go away eventually.

stevemayhew:
@tvarga-dss Thanks for taking this on an rebasing my pull to latest `dev-v2`!   We have lots of live streams with multiple bitrates at 720p / 1080, we are looking at this for faster stream switch.

The ConcurrentModificationException happened with the `mediaChunks` list in `HlssampleStreamWrapper`, I wish I had saved the stack traces but I didn't.   I think in the latest `dev-v2` the way the `HlsMediaChunk` calls back to the `HlsSampleStreamWrapper`  (method used to be `init()` has changed in a way so that the `Loader` thread does not touch the `mediaChunks` lists.  If so, that should fix the issue.

@ojw28 / @icbaker at any rate, I'll work with @tvarga-dss to finish testing this pull request and close my request so there is no confusion.

Again thanks @tvarga-dss  for taking this on!

stevemayhew:
@tonihei @tvarga-dss  I've branched this pull (https://github.com/TiVo/ExoPlayer/tree/testing-cancel-hls-chunk-download-and-discard-upstream) and added code to the sample player to aggressively exercise the code.  

In testing with a live stream here (which has a distant CDN so it frequently changes bandwidth), I was not able to reproduce the `ConcurrentModificationException` (I suspect this is because of the change to eliminate the callback from Loader thread to `HlsSampleStreamWrapper.init()`)

I was able to reproduce the following exception:

```
04-27 10:12:39.175 15032 15032 E EventLogger:   com.google.android.exoplayer2.ExoPlaybackException: Unexpected runtime error
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:490)
04-27 10:12:39.175 15032 15032 E EventLogger:       at android.os.Handler.dispatchMessage(Handler.java:102)
04-27 10:12:39.175 15032 15032 E EventLogger:       at android.os.Looper.loop(Looper.java:193)
04-27 10:12:39.175 15032 15032 E EventLogger:       at android.os.HandlerThread.run(HandlerThread.java:65)
04-27 10:12:39.175 15032 15032 E EventLogger:   Caused by: java.lang.IndexOutOfBoundsException: off=-106, len=2965 out of bounds (size=65536)
04-27 10:12:39.175 15032 15032 E EventLogger:       at java.nio.Buffer.checkBounds(Buffer.java:587)
04-27 10:12:39.175 15032 15032 E EventLogger:       at java.nio.DirectByteBuffer.put(DirectByteBuffer.java:285)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.source.SampleDataQueue.readData(SampleDataQueue.java:309)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.source.SampleDataQueue.readToBuffer(SampleDataQueue.java:147)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.source.SampleQueue.read(SampleQueue.java:340)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.source.hls.HlsSampleStreamWrapper.readData(HlsSampleStreamWrapper.java:565)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.source.hls.HlsSampleStream.readData(HlsSampleStream.java:79)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.BaseRenderer.readSource(BaseRenderer.java:354)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.feedInputBuffer(MediaCodecRenderer.java:1225)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.mediacodec.MediaCodecRenderer.render(MediaCodecRenderer.java:794)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:812)
04-27 10:12:39.175 15032 15032 E EventLogger:       at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:401)
```

I'm investigating why this occurs now, will keep the branch updated and this pull request with my findings.

stevemayhew:
@tvarga-dss Could you please rebase this to latest `dev-v2`.  Some conflicts to resolve.

library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
```diff
@@ -767,7 +767,8 @@ private synchronized boolean attemptSplice(long timeUs) {
 
   private long discardUpstreamSampleMetadata(int discardFromIndex) {
     int discardCount = getWriteIndex() - discardFromIndex;
-    Assertions.checkArgument(0 <= discardCount && discardCount <= (length - readPosition));
+    Assertions.checkArgument(0 <= discardCount);
+    discardCount = getMaximumDiscardCount(discardCount);
```
tonihei: This change implies that you tried to discard samples from before the read position. This should be prevented because it leaves the sample queue in a non-recoverable state. Please see my other comment in HlsSampleStreamWrapper to that regard.

library/core/src/main/java/com/google/android/exoplayer2/upstream/Loader.java
```diff
@@ -59,6 +59,11 @@ public UnexpectedLoaderException(Throwable cause) {
      */
     void cancelLoad();
 
+    /**
+     * @return true if the load has been cancelled
+     */
+    boolean isLoadCanceled();
```
tonihei: I think it would be clearer to track this on the calling side where this information is needed. It directly corresponds to a call to cancelLoad(), so there is no ambiguity. It also prevents the problem that every Loadable needs to implement this method.

library/core/src/main/java/com/google/android/exoplayer2/upstream/Loader.java
```diff
@@ -261,6 +267,17 @@ public void cancelLoading() {
     Assertions.checkStateNotNull(currentTask).cancel(false);
   }
 
+  /**
+   * Interrupt the current load later discarding samples too.
+   *
+   * @throws IllegalStateException If the loader is not currently loading.
+   */
+  public void interruptLoading() {
```
tonihei: The same comment as above applies to this method and all the interrupted flag plumbing to the callback. You can as well set an isInterrupted flag in HlsSampleStreamWrapper locally when calling this method and save yourself all these changes. Just looking at the Loader class, it's very unclear how interrupt differs from normal cancelation and it doesn't really make sense to introduce a new method just to describe a property of the cancelation.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsInterruptStateDelegate.java
```diff
@@ -0,0 +1,28 @@
+package com.google.android.exoplayer2.source.hls;
+
+import androidx.annotation.NonNull;
+
+/**
+ * Delegate setting and querying the hls chunk download interrupt state
+ */
+public interface HlsInterruptStateDelegate {
```
tonihei: I think passing this object around makes it hard to reason about when it can be set and it's also hard to explain the threading implications of this method. The final decision to discard upstream data should come from TrackSelection.evaluateQueueSize and there is in my opinion no need for an additional side path. So you can just remove this object and use the return value of HlsChunkSource.getPreferredQueueSize to discard data and cancel the current load if needed.

Note: This means you can still use this delegate class in your code if needed. Just pass it to your TrackSelection class and use it in there. If yo are not currently using a custom TrackSelection class, you can also subclass our default AdaptiveTrackSelection and just override evaluateQueueSize.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -291,13 +299,39 @@ public boolean isLoadCompleted() {
...
+  }
+
+  /**
+   * Set the index of the first sample written to a sample queue from this media chunk,
+   * indexed by the caller's stream index for the sample queue
+   *
+   * @param streamIndex - caller's index for the {@link SampleQueue}
+   * @param firstSampleIndex - index value to store
+   */
+  void setFirstSampleIndex(int streamIndex, int firstSampleIndex) {
```
tonihei, line 318: There was a pending discussion on the previous PR about concurrent modifications and potential threading issues. Just wanted to add some explanations for why I think this is not an issue here.

This method (setFirstSampleIndex) may be called from the playback thread (=the thread that handles reevaluateBuffer) or the loading thread, depending on whether we create a new sample stream or keep using an existing one. The two setter calls are mutually exclusive - they never happen for the same chunk. So this shouldn't be an issue.

The reading method (getFirstSampleIndexes) is called from the playback thread, but is guaranteed to be called after all setters have finished. So this shouldn't pose a problem either.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +704,83 @@ public boolean isLoading() {
 
   @Override
   public void reevaluateBuffer(long positionUs) {
-    // Do nothing.
+    if (hlsInterruptStateDelegate != null && REQUESTED.equals(hlsInterruptStateDelegate.getCancelState()) && loader.isLoading()) {
+      hlsInterruptStateDelegate.setCancelState(TRIGGERED);
+      loader.interruptLoading();
+      return;
+    }
+    if (loader.isLoading() || isPendingReset()) {
```
tonihei: If isPendingReset() is true, then you shouldn't need to reevaluate the buffer in any case, including your new interrupt case. Also, this code should also probably check for loader.hasFatalError() to avoid any unnecessary work.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +704,83 @@ public boolean isLoading() {
...
+
+    int currentQueueSize = mediaChunks.size();
+    int preferredQueueSize = chunkSource.getPreferredQueueSize(positionUs, readOnlyMediaChunks);
+    if (currentQueueSize <= preferredQueueSize) {
+      return;
+    }
+
+    int newQueueSize = currentQueueSize;
+    for (int i = preferredQueueSize; i < currentQueueSize; i++) {
+      if (!haveReadFromMediaChunk(i)) {
```
tonihei, line 726: This condition is important to see if we need to reset the sample streams or can continue filling them. In your version with load cancellation, you don't currently check for this condition, so I assume you want to consider cancellation where the rendering side already started reading the chunk. An example where this may easily happen is the first chunk. If you already loaded any samples from the first chunk, then it is likely the renderer started reading them.

If you want to discard chunks that have already been read from, you need to set the pendingResetPositionUs to the position where playback should continue from (most likely lastSeekPositionUs or the start of the chunk) and clear the list of media chunks. This ensures the renderers get reset correctly. Something like that is already done in upstreamDiscard for the empty queue case, but needs to be extended to cover all cases where we discard chunks that have been read from. If you do this, then you also don't need to change the SampleQueue implementation.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -732,7 +814,10 @@ public void onLoadCompleted(Chunk loadable, long elapsedRealtimeMs, long loadDur
 
   @Override
   public void onLoadCanceled(
-      Chunk loadable, long elapsedRealtimeMs, long loadDurationMs, boolean released) {
+      Chunk loadable, long elapsedRealtimeMs, long loadDurationMs, boolean released, boolean interrupted) {
+    if (interrupted) {
+      discardAfterInterrupt();
```
tonihei: The onLoadCanceled method does 3 things:
 1. Report event to eventDispatcher.
 2. Update sample queues as needed to continue loading afterwards
 3. Trigger or request to continue loading.
And it used for 2 main cases now: 
 A. Cancelation after seek or disabling tracks.  -> clear sample queues completely.
 B. Cancelation after reevaluateBuffer.  -> discard only pending load upstream.

Could you make sure both cases (A and B) are handled in the same way in the same order? At least step 1 can share the same code. Step 2 is intentionally different and this is where I would expect the "if(interrupted) { doB() } else { doA(); } code. And step 3 can be shared again.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -696,9 +704,83 @@ public boolean isLoading() {
 
   @Override
   public void reevaluateBuffer(long positionUs) {
-    // Do nothing.
+    if (hlsInterruptStateDelegate != null && REQUESTED.equals(hlsInterruptStateDelegate.getCancelState()) && loader.isLoading()) {
+      hlsInterruptStateDelegate.setCancelState(TRIGGERED);
```
tonihei: Related to my other comment about not needing this delegate:
You can query chunkSource.getPreferredQueueSize as below. And only if you need to discard samples, then you can check if you are currently loading. If yes, cancel the load and set your isInterrupted flag. If no, discard immediately.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -760,6 +852,28 @@ public void onLoadCanceled(
     }
   }
 
+  private void discardAfterInterrupt() {
+    int lastChunkIndex = mediaChunks.size() - 1;
+    int preferredQueueSize = lastChunkIndex;
+    for (int i = lastChunkIndex; i >= 0; i--) {
```
tonihei: This logic always calculates preferredQueueSize = lastChunkIndex I think, because there can only be one canceled chunk. 

If you implement my other comment about moving all logic inside getPreferredQueueSize/evaluateQueueSize, then you can just return queueSize-1 if you want to discard the last, currently loading chunk only.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -851,11 +965,16 @@ private void initMediaChunkLoad(HlsMediaChunk chunk) {
     upstreamTrackFormat = chunk.trackFormat;
     pendingResetPositionUs = C.TIME_UNSET;
     mediaChunks.add(chunk);
-
     chunk.init(this);
-    for (HlsSampleQueue sampleQueue : sampleQueues) {
+    HlsMediaChunk loadingChunk = findChunkMatching(chunk.uid);
```
tonihei: It looks to me as if findChunkMatching will always return the same chunk as the one passed in to this method? Guess it was needed before rebasing to the latest dev branch code state.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -951,6 +1070,10 @@ private SampleQueue createSampleQueue(int id, int type) {
     sampleQueueTrackIds = Arrays.copyOf(sampleQueueTrackIds, trackCount + 1);
     sampleQueueTrackIds[trackCount] = id;
     sampleQueues = Util.nullSafeArrayAppend(sampleQueues, sampleQueue);
+    HlsMediaChunk mediaChunk = findChunkMatching(sourceChunk.uid);
```
tonihei: Again, I think the returned media chunk should be identical to sourceChunk, so no need to call this method. Also means that findChunkMatching can probably be removed.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -951,6 +1070,10 @@ private SampleQueue createSampleQueue(int id, int type) {
...
+    if (mediaChunk != null) {
+      mediaChunk.setFirstSampleIndex(trackCount, 0);
```
tonihei: Maybe better to call sampleQueue.getWriteIndex() here as well to make the intention clear even though it's always 0 at this point.

tonihei (commented):
I think the overall approach is right. Most comments are code simplifications to make sure the PR just touches the classes it needs to touch and changes are more local to where they are needed. 

I'll try and merge the parts from #6321 first to make sure the PR stays relatively small and easy to review. After that's done, you probably need to rebase as there will be some internal code style cleanups etc, or adjustments from internal code reviews.

library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
stevemayhew: Could this cause the `java.lang.IndexOutOfBoundsException` with `SampleDataQueue` I saw in my testing?

BTW the ConcurrentModificationException from my pull request.  @tvarga-dss feel free to collapse out any of my commits in your re-base to simplicity this, I just want to see the future supported ASAP.  If you are unable to work on this please let us know.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
stevemayhew: Yes, the non-chunkless prepare state causes this method to be called from the loader thread (the Extractor).   The `sampleQueue` array itself is accessed from both the player thread and loader threads because of non-chunkless prepare. 

The `ConcurrentModificationException` was because of the way `Extractor` init worked in pre-2.11.x versions, it looks like this is no longer so in the current `dev-v2`

library/core/src/main/java/com/google/android/exoplayer2/source/SampleQueue.java
tonihei: > Could this cause the java.lang.IndexOutOfBoundsException with SampleDataQueue I saw in my testing?

Sounds like a likely reason, yes.

> BTW the ConcurrentModificationException from my pull request.

I added some explanations further down why I think access to the first sample indices is safe. But I only now realized this is not the concurrent modification you were talking about. :)  I'll check against your comments made in https://github.com/google/ExoPlayer/pull/6321#issuecomment-570344663 to see if this can still happen.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsInterruptStateDelegate.java
tvarga-dss: We are indeed subclassing `AdaptiveTrackSelection` already, I'll take another look at simplifying things as per your suggestions.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
tvarga-dss: I have done most of the cleanup/simplification you've suggested, except for this one:


> If you want to discard chunks that have already been read from, you need to set the pendingResetPositionUs to the position where playback should continue from (most likely lastSeekPositionUs or the start of the chunk) and clear the list of media chunks. This ensures the renderers get reset correctly. Something like that is already done in upstreamDiscard for the empty queue case, but needs to be extended to cover all cases where we discard chunks that have been read from. If you do this, then you also don't need to change the SampleQueue implementation.

This is something I still have to look at, but maybe it's fine for now to only discard chunks that have not been read from.

Do let me know what you think @tonihei

stevemayhew: @tvarga-dss discarding chunks that have just started playing would be nice (with a 6 second chunk size doing so might avoid up to 6 seconds of poor quality content playing), this of course would need to be balanced against the chance it would cause a stall (very likely).

So, IMHO, start small and only cancel loading chunks that are not currently being rendered.  We saw benefit to this change even without the load cancel, as it is easily possible (in the non-live case at least) for 10s of seconds of poor quality content to be buffered.

Once you have that, I'll re-test to make sure the assert in the sample queue read (https://github.com/google/ExoPlayer/pull/7244#issuecomment-620119891)  has been fixed

tonihei: I agree with Steve's comment - canceling chunks we started reading seems very useful, especially at the beginning of playback. However, it's fine to merge this first and add support as a second step.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -753,13 +837,23 @@ public void onLoadCanceled(
...
+        resetSampleQueues();
+      }
       if (enabledTrackGroupCount > 0) {
         callback.onContinueLoadingRequested(this);
       }
     }
   }
 
+  private void discardAfterInterrupt() {
+    int preferredQueueSize = mediaChunks.size() - 1;
```
tonihei: Might be better to save the value that was returned from chunkSource.getPreferredQueueSize and use it here.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -733,6 +814,9 @@ public void onLoadCompleted(Chunk loadable, long elapsedRealtimeMs, long loadDur
   @Override
   public void onLoadCanceled(
       Chunk loadable, long elapsedRealtimeMs, long loadDurationMs, boolean released) {
+    if (interrupted) {
+      discardAfterInterrupt();
```
tonihei: Can you just move this into the "if (interrupted)" condition below? Seems more logical to have it all in one place.

tonihei (commented):
Thanks for doing all the simplifications! Looks good now except for the preferredQueueSize that needs to be propagated correctly.

tonihei:
Thanks, I'll mark the issue as "should merge" now and start the internal review process.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
```diff
@@ -291,6 +299,27 @@ public boolean isLoadCompleted() {
...
```
tonihei, line 318: We removed this method in the end because the values can generated locally by passing in the SampleQueues to the init() method. 

The case where a sample queue is created by this chunk can actually be removed entirely because if a sample queue doesn't exist yet, we know that it is created by this chunk (or even a later one), so the index to discard from is always 0.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -146,6 +146,8 @@
   private @MonotonicNonNull Format upstreamTrackFormat;
   @Nullable private Format downstreamTrackFormat;
   private boolean released;
+  private boolean interrupted;
```
tonihei, line 149: I removed this boolean and "merged" it with the other variable by using POSITION_UNSET if no cancellation is pending.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -853,8 +944,10 @@ private void initMediaChunkLoad(HlsMediaChunk chunk) {
     mediaChunks.add(chunk);
 
     chunk.init(this);
-    for (HlsSampleQueue sampleQueue : sampleQueues) {
+    for (int i = 0; i < sampleQueues.length; i++) {
+      HlsSampleQueue sampleQueue = sampleQueues[i];
       sampleQueue.setSourceChunk(chunk);
+      chunk.setFirstSampleIndex(i, sampleQueue.getWriteIndex());
```
tonihei, line 950: There is actually an edge case where chunk.shouldSpliceIn is true in which we can't use the write index as it is. This happens if we switch playlists for example. I changed the code to use the firstSampleIndices from the previous chunk instead to be on the safe side.

tonihei (commented):
Internal review is done. I added some comments to highlight changes we made fyi.

tonihei:
@tvarga-dss @stevemayhew 
For your information, we've updated the code slightly since the PR has been merged to fix subtle bugs and to improve integration. In particular:
1. We split `TrackSelection.evaluateQueueSize` into `evaluateQueueSize` (just for discarding) and `shouldCancelChunkLoad` (just for cancellation). This allows the `TrackSelection` to more clearly handle both cases separately if needed, for example to cancel without discarding.  (https://github.com/google/ExoPlayer/commit/95b61eb835e884678daef6e938c4bb0175489c2d)
2. We removed the immediate discard after cancellation because `evaluateQueueSize` will be called anyway before loading continues (https://github.com/google/ExoPlayer/commit/7bc5fa855f774ce198e23a455fec751b7743486f). This also fixed a bug that was reported in #7549 and #7580. The bug was caused by saving the discard-from chunk index and then waiting for the chunk cancellation. If the cancellation process took too long and the decoders starting reading the chunk that is about to be discarded, the reported error occurred.  
3. We disallowed discarding chunks that have been spliced in because it's currently impossible to restore the previous state (https://github.com/google/ExoPlayer/commit/2e749f70aeb98269fad65643fe1435e7b632cdb1). Splicing-in may happen, e.g., when changing playlists. At the same time, we also optimized the splicing so that it doesn't happen for well-aligned segments across playlists with independent segments. That way this new restriction doesn't affect well-formed media. (https://github.com/google/ExoPlayer/commit/5c52915f0c50330085caefb38146a33bff032cbd).

PR #7247, opened 2020-04-16
Add saving subtitle track name for MatroskaExtractor

vadimdolgachev:
Some files have several subtitles in the same language, such as forced and regular subtitles. In this case, a name is added for the subtitle track.
This code allows to save the name of the subtitle track. DefaultTrackNameProvider consider the name for TRACK_TYPE_TEXT track type.   
The subtitle name and language name in the pair can also be seen in VLC player.
Sample video with subtitle track name: https://drive.google.com/file/d/1WoWcL9B4cVd1fsDEAtYpXDmG0cMF0F_3/view?usp=sharing

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
@@ -2139,6 +2139,7 @@ public class MatroskaExtractor implements Extractor {
       Format format =
           formatBuilder
               .setId(trackId)
+              .setLabel(name)
               .setSampleMimeType(mimeType)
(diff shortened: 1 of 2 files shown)


icbaker:
Thanks for the contribution! Using the `name`/`title` from Matroska files to populate `Format.label` looks good.

But the changes to `DefaultTrackNameProvider` cause us to surface an un-localisable string (directly from the media file) in an otherwise localised context. It's also not clear the concatenating language & track name always makes sense - especially as in some cases the track name might contain the language so it would be duplicated.

So I'm going to merge only the changes to `MatroskaExtractor` - it should be possible for you to keep the track name behaviour in your own app by providing a custom `TrackNameProvider`. Hope that's OK!

vadimdolgachev:
Good. Thanks!

Issue #7303, opened 2020-04-28
DefaultTimeBar ignores touch transformations


AChep:
https://github.com/google/ExoPlayer/blob/7d3f54a375fac04a746ca76a8f2e1ad32c8b45b2/library/ui/src/main/java/com/google/android/exoplayer2/ui/DefaultTimeBar.java#L757

Steps to reproduce:
1. Create `PlayerView` with a default time bar.
2. Create a container that transforms every touch event by translating it by 100 px.
3. `DefaultTimeBar` ignores the translation.

Steps to fix:
I believe that simply replacing the:
```
  private Point resolveRelativeTouchPosition(MotionEvent motionEvent) {
    getLocationOnScreen(locationOnScreen);
    touchPosition.set(
        ((int) motionEvent.getRawX()) - locationOnScreen[0],
        ((int) motionEvent.getRawY()) - locationOnScreen[1]);
    return touchPosition;
  }
```

with 
```
  private Point resolveRelativeTouchPosition(MotionEvent motionEvent) {
    touchPosition.set(
        (int) motionEvent.getX(),
        (int) motionEvent.getY());
    return touchPosition;
  }
```
should do the trick, but I don't know what was the reasoning behind obtaining the touch position manually.

ojw28:
I'm not sure what the reason is either. It's existed since the code was originally written, and wasn't commented on during internal code review. I assume you tested your pull request and it works, in which case I'm happy to merge.

AChep:
@ojw28 let me double check it again.

AChep:
@ojw28 everything seem to work fine. Also removed no longer being used `locationOnScreen ` variable.

ojw28:
https://github.com/google/ExoPlayer/pull/7304 was merged.

PR #7304, opened 2020-04-28
Fix DefaultTimeBar ignores touch transformations #7303

AChep:


AChep:
@googlebot I signed it!

PR #7316, opened 2020-04-30
Change needed to prevent bandwidth based adaptation from selecting iFrame only

stevemayhew:
Thank you all for merging the iFrame only support pull request.  I apologize for not rebasing it to fewer commits first, as it has grown old.  So the single commit 88de774 was great,I like the idea to put the iFrames in the same with the other variants.   This change was @ojw28's design, I just missed making them the same list.

This version of my changes will work just fine, however I would strongly recommend you include the commit in this pull (already rebased to dev-v2).   This will prevent i-Frame tracks from being selected at 1x forward speed.

I can rebase again and factor in the TODO (basically only needed if there are multiple iFrame only tracks).  Let me know.

We are re-testing our code with the latest `dev-v2` and found this.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -540,7 +540,17 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
   @SuppressWarnings("unused")
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
-    return Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    boolean isIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) != 0;
+    boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    if (Math.abs(playbackSpeed) > 6.0f) {
+      canSelect = isIframeOnly;   // TODO factor in playback speed...
+    } else {
+      canSelect = ! isIframeOnly && canSelect;
+    }
+    return canSelect;
+
   }
 
   /**


library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
```diff
@@ -540,7 +540,17 @@ public int evaluateQueueSize(long playbackPositionUs, List<? extends MediaChunk>
   @SuppressWarnings("unused")
   protected boolean canSelectFormat(
       Format format, int trackBitrate, float playbackSpeed, long effectiveBitrate) {
-    return Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    boolean isIframeOnly = (format.roleFlags & C.ROLE_FLAG_TRICK_PLAY) != 0;
+    boolean canSelect = Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate;
+
+    if (Math.abs(playbackSpeed) > 6.0f) {
+      canSelect = isIframeOnly;   // TODO factor in playback speed...
```
stevemayhew, line 548: For the TODO, I would change this code to:

```java
    if (Math.abs(playbackSpeed) > 1.0f) {
        canSelect = isIframeOnly && Math.round(trackBitrate * playbackSpeed) <= effectiveBitrate);
    }
```

The other alternative is to modify the track selection's TrackGroups to include only i-Frame tracks on speed change over 1.0.

ojw28:
I think this is specific to how you happen to be implementing trick play (using playback speed). There are other trick play scenarios where the player is paused and trick play is performed by repeated seeking. So I think this needs more thought.

Note that you trick play tracks are *never* part of track selections in the first place, even statically, with the code in `dev-v2`. They will only be selected if you specifically override the track selection to include them.

You should be able to easily extend `AdaptiveTrackSelection` to do what you need, in the meantime.

stevemayhew:
Thanks, I’ll look in AdaptiveTrackSelection for this exclusion logic.  Just
wanted to be sure I didn’t contribute to adding a bug to ExoPlayer 😀


On Fri, May 1, 2020 at 5:06 AM Oliver Woodman <notifications@github.com>
wrote:

> I think this is specific to how you happen to be implementing trick play
> (using playback speed). There are other trick play scenarios where the
> player is paused and trick play is performed by repeated seeking. So I
> think this needs more thought.
>
Yes

Note that you trick play tracks are *never* part of track selections in the

ojw28:
The exclusion logic is in `DefaultTrackSelector`.

stevemayhew:
Thanks

PR #7324, opened 2020-05-04
[MOD] Use Cronet from Play Services as default instead of using embedded one. It lets to reduce APK size ~8MB. Fixed CronetDataSourceTest, as UrlResponseInfoImpl is not available with play-services-cronet:17.0.0.

tpiDev:


ojw28:
Related [Internal ref: b/149941833]

noamtamim:
@ojw28 Do you now recommend using Cronet instead of URLConnection by default? 
Do you know what Youtube is using?

ojw28:
I think it's generally preferable to use Cronet or OkHttp over the platform URLConnection, if only because it's going to better isolate you from possible fragmentation. In general, I suspect that Cronet is the best choice for many apps, but I don't think we (the ExoPlayer team) take a particularly strong view on it. Which network stack you should use probably depends on app specific factors. For example which protocol you're using, what other networking tasks your app needs to do, how sensitive you are to APK size, and whether your app is likely to be used in markets where Google Play Services is not widely supported.

I believe YouTube use Cronet and QUIC in most circumstances, but I'm not 100% certain of that.

PR #7370, opened 2020-05-13
Enable embedded CEA-708 support

jruesga:
Currently, DashMediaPeriod only takes into account as CEA-608 accessibility tags as embedded closed captions tracks CEA-608. CEA-708 closed captions format is parsed when is present on its own AdaptationSet, but not when is embedded as an accessibility tag in a video AdaptationSet.          

Embedded CEA-708 support is added by parsing accessibility tags like the example below:

```
<Accessibility schemeIdUri="urn:scte:dash:cc:cea-708:2015" value="1=lang:eng;2=lang:deu"/>
<Accessibility schemeIdUri="urn:scte:dash:cc:cea-708:2015" value="1=lang:eng;2=lang:eng,war:1,er:1"/>
```

Creates a new CEA-708 track for accessibility tags with schemeIdUri `urn:scte:dash:cc:cea-708:2015` and extract accessibilityChannel and language from value attribute.

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
@@ -69,9 +69,6 @@ import org.checkerframework.checker.nullness.compatqual.NullableType;
         SequenceableLoader.Callback<ChunkSampleStream<DashChunkSource>>,
         ChunkSampleStream.ReleaseCallback<DashChunkSource> {
 
-  private enum CC_TYPE {
-    CEA608, CEA708
-  }
   private static final Pattern CEA608_SERVICE_DESCRIPTOR_REGEX = Pattern.compile("CC([1-4])=(.+)");
   private static final Pattern CEA708_SERVICE_DESCRIPTOR_REGEX = Pattern.compile("([1-4])=lang:(\\w+)(,.+)?");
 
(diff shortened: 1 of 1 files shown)


jruesga:
@googlebot I signed it!

ojw28:
I'm not sure why the CLA check is unhappy. Marking with the `cla: yes` label based on manual verification.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
```diff
@@ -69,7 +69,11 @@
         SequenceableLoader.Callback<ChunkSampleStream<DashChunkSource>>,
         ChunkSampleStream.ReleaseCallback<DashChunkSource> {
 
+  private enum CC_TYPE {
+    CEA608, CEA708
+  }
   private static final Pattern CEA608_SERVICE_DESCRIPTOR_REGEX = Pattern.compile("CC([1-4])=(.+)");
+  private static final Pattern CEA708_SERVICE_DESCRIPTOR_REGEX = Pattern.compile("([1-4])=lang:(\\w+)(,.+)?");
```
ojw28, line 73: What is the purpose of the `(,.+)?` part of the REGEX? Does it work exactly the same without it?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
```diff
@@ -824,59 +831,96 @@ private static boolean hasEventMessageTrack(List<AdaptationSet> adaptationSets,
     return false;
   }
 
-  private static Format[] getCea608TrackFormats(
+  private static Format[] getClosedCaptionsTrackFormats(
```
ojw28, line 831: The use of the enum through this code is IMO making it quite a bit harder to read than it needs to be. Can't you just do something like (warning, untested):

```

  private static Format[] getClosedCaptionTrackFormats(
      List<AdaptationSet> adaptationSets, int[] adaptationSetIndices) {
    for (int i : adaptationSetIndices) {
      AdaptationSet adaptationSet = adaptationSets.get(i);
      List<Descriptor> descriptors = adaptationSets.get(i).accessibilityDescriptors;
      for (int j = 0; j < descriptors.size(); j++) {
        Descriptor descriptor = descriptors.get(j);
        if ("urn:scte:dash:cc:cea-608:2015".equals(descriptor.schemeIdUri)) {
          Format cea608Format =
              new Format.Builder()
                  .setSampleMimeType(MimeTypes.APPLICATION_CEA608)
                  .setId(adaptationSet.id + ":cea608")
                  .build();
          return parseClosedCaptionDescriptor(
              descriptor, CEA608_SERVICE_DESCRIPTOR_REGEX, cea608Format);
        } else if ("urn:scte:dash:cc:cea-708:2015".equals(descriptor.schemeIdUri)) {
          Format cea708Format =
              new Format.Builder()
                  .setSampleMimeType(MimeTypes.APPLICATION_CEA708)
                  .setId(adaptationSet.id + ":cea708")
                  .build();
          return parseClosedCaptionDescriptor(
              descriptor, CEA708_SERVICE_DESCRIPTOR_REGEX, cea708Format);
        }
      }
    }
    return new Format[0];
  }

  private static Format[] parseClosedCaptionDescriptor(
      Descriptor descriptor, Pattern serviceDescriptorRegex, Format baseFormat) {
    @Nullable String value = descriptor.value;
    if (value == null) {
      // There are embedded closed caption tracks, but service information is not declared.
      return new Format[] {baseFormat};
    }
    String[] services = Util.split(value, ";");
    Format[] formats = new Format[services.length];
    for (int i = 0; i < services.length; i++) {
      Matcher matcher = serviceDescriptorRegex.matcher(services[i]);
      if (!matcher.matches()) {
        // If we can't parse service information for all services, assume a single track.
        return new Format[] {baseFormat};
      }
      int accessibilityChannel = Integer.parseInt(matcher.group(1));
      formats[i] =
          baseFormat
              .buildUpon()
              .setId(baseFormat.id + ":" + accessibilityChannel)
              .setAccessibilityChannel(accessibilityChannel)
              .setLanguage(matcher.group(2))
              .build();
    }
    return formats;
  }
```

and get rid of the enum and all the little helper methods.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
```diff
@@ -824,59 +831,96 @@ private static boolean hasEventMessageTrack(List<AdaptationSet> adaptationSets,
...
-            formats[k] =
-                buildCea608TrackFormat(
-                    adaptationSet.id,
-                    /* language= */ matcher.group(2),
-                    /* accessibilityChannel= */ Integer.parseInt(matcher.group(1)));
-          }
-          return formats;
-        }
+      for (Descriptor descriptor : descriptors) {
+        return parseClosedCaptionsDescriptor(adaptationSet.id, descriptor);
```
ojw28: Might be irrelevant if you can go with my proposal above, but this shouldn't return if `parsedClosedCaptionsDescriptor` returned `null`. Else parsing will fail if there's a descriptor for something other than 608/708 before the descriptor you're looking for.

ojw28 (commented):
Please take a look at the comments. Thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
jruesga: The value can contain other information (like easy reader, ...) 

`1=lang:eng;2=lang:eng,war:1,er:1`

The regular expresion try to gets only the lang part of the property

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
jruesga: @ojw28 Thanks for the review and suggestions. It seems ok to me.

What about if we instead added all the ClosedCaptions formats and not only the first found in the AdaptationSet? Something like:

```
  private static Format[] getClosedCaptionsTrackFormats(
      List<AdaptationSet> adaptationSets, int[] adaptationSetIndices) {
    for (int i : adaptationSetIndices) {
      AdaptationSet adaptationSet = adaptationSets.get(i);
      List<Descriptor> descriptors = adaptationSets.get(i).accessibilityDescriptors;
      List<Format> closedCaptionFormats = new ArrayList<>();
      for (int j = 0; j < descriptors.size(); j++) {
        Descriptor descriptor = descriptors.get(j);
        if ("urn:scte:dash:cc:cea-608:2015".equals(descriptor.schemeIdUri)) {
          Format cea608Format =
              new Format.Builder()
                  .setSampleMimeType(MimeTypes.APPLICATION_CEA608)
                  .setId(adaptationSet.id + ":cea608")
                  .build();
          Format[] cea608Formats = parseClosedCaptionsDescriptor(
              descriptor, CEA608_SERVICE_DESCRIPTOR_REGEX, cea608Format);
          closedCaptionFormats.addAll(Arrays.asList(cea608Formats));
        } else if ("urn:scte:dash:cc:cea-708:2015".equals(descriptor.schemeIdUri)) {
          Format cea708Format =
              new Format.Builder()
                  .setSampleMimeType(MimeTypes.APPLICATION_CEA708)
                  .setId(adaptationSet.id + ":cea708")
                  .build();
          Format[] cea708Formats = parseClosedCaptionsDescriptor(
              descriptor, CEA708_SERVICE_DESCRIPTOR_REGEX, cea708Format);
          closedCaptionFormats.addAll(Arrays.asList(cea708Formats));
        }
      }
      return closedCaptionFormats.toArray(new Format[0]);
    }
    return new Format[0];
  }

```

Let me know what you think and can provide whatever of the above solutions you prefer.

jruesga: Otherwise, maybe it could be better to provide 708 format over 608 since is a new format?

jruesga: Done

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
jruesga: Done

ojw28:
I'm not sure why it's not been picked up, but this was merged in https://github.com/google/ExoPlayer/commit/3db703a983dce46dcf4d4e0a11c6e4c644386a9f. Thanks!

PR #7395, opened 2020-05-19
Adds storage not low requirement

jdegroot-dss:
Added monitoring storage levels in RequirementsWatcher
Added dependency on extension-workmanager to the demo app to be able to test with WorkManagerScheduler
Added getSupportedRequirements method to Scheduler interface
Implemented getSupportedRequirements for schedulers

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
@@ -116,6 +116,14 @@ public final class PlatformScheduler implements Scheduler {
     } else if (requirements.isNetworkRequired()) {
       builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
     }
+    if (requirements.isStorageNotLowRequired()) {
+      if (Util.SDK_INT < 26) {
+        Log.w(TAG, "Storage not low requirement not supported on the PlatformScheduler"
+            + "on API below 26. Requirement not set.");
+      } else {
+        builder.setRequiresStorageNotLow(true);
+      }
+    }
     builder.setRequiresDeviceIdle(requirements.isIdleRequired());
     builder.setRequiresCharging(requirements.isChargingRequired());
     builder.setPersisted(true);
--- a/library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
@@ -86,6 +86,21 @@ public final class PlatformScheduler implements Scheduler {
     return true;
   }
 
+  @Override
+  public Requirements getSupportedRequirements(Requirements requirements) {
+    Requirements supportedRequirements = requirements;
+    if (Util.SDK_INT < 26) {
+      if (requirements.isStorageNotLowRequired()) {
+        Log.w(TAG, "Storage not low requirement not supported on the PlatformScheduler"
+            + "on API below 26. Requirement removed.");
+        int newRequirements =
+            supportedRequirements.getRequirements() ^ Requirements.DEVICE_STORAGE_NOT_LOW;
+        supportedRequirements = new Requirements(newRequirements);
+      }
+    }
+    return supportedRequirements;
+  }
+
   // @RequiresPermission constructor annotation should ensure the permission is present.
   @SuppressWarnings("MissingPermission")
   private static JobInfo buildJobInfo(
--- a/library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
@@ -39,7 +39,11 @@ public final class Requirements implements Parcelable {
 
   /**
    * Requirement flags. Possible flag values are {@link #NETWORK}, {@link #NETWORK_UNMETERED},
-   * {@link #DEVICE_IDLE} and {@link #DEVICE_CHARGING}.
+   * {@link #DEVICE_IDLE}, {@link #DEVICE_CHARGING} and {@link #DEVICE_STORAGE_NOT_LOW}.
+   *
+   * Note that {@link #DEVICE_STORAGE_NOT_LOW} only works when downloading to internal storage.
+   * Setting this requirement when downloading to external storage will actually monitor internal
+   * storage and can lead to unexpected results.
    */
   @Documented
   @Retention(RetentionPolicy.SOURCE)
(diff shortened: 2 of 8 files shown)


library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
```diff
@@ -162,6 +171,34 @@ private boolean isDeviceIdle(Context context) {
...
   }
 
+  /**
+   * Implementation taken from the the WorkManager source.
+   * @see <a href="https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/work/workmanager/src/main/java/androidx/work/impl/constraints/trackers/StorageNotLowTracker.java">StorageNotLowTracker</a>
+   */
+  private boolean isStorageNotLow(Context context) {
+    IntentFilter intentFilter = new IntentFilter();
+    intentFilter.addAction(Intent.ACTION_DEVICE_STORAGE_OK);
+    intentFilter.addAction(Intent.ACTION_DEVICE_STORAGE_LOW);
```
Vibragimov, line 184: @jdegroot-dss It works only for internal storage. When downloading to external storage, this method of determining is useless and even false. If there is no space in the internal storage, we cannot download to the external storage when this requirement is enabled

Vibragimov (changes requested):

library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
ojw28: @jdegroot-dss Does this affect your intended usage at all, or is it fine for your use case? It still seems OK to support this to me, although perhaps this should be made clearer in the documentation. If someone is placing content onto external storage they can just not use the feature.

If it's still useful for you, please just confirm this here and I'll look at getting it merged in. Thanks!

jdegroot-dss: Thanks for the heads up @vigneshvg, I was unaware of that limitation. 

@ojw28 I don't think this will affect our use case and it would still be useful to have. Agreed with updating the documentation. Do we want to restrict the usage of the requirement if downloads are going to external storage?

ojw28: I think just making it clear in the documentation is enough.

library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
```diff
@@ -162,6 +175,34 @@ private boolean isDeviceIdle(Context context) {
...
+    Intent intent = context.registerReceiver(null, intentFilter);
+    if (intent == null || intent.getAction() == null) {
+      // ACTION_DEVICE_STORAGE_LOW is a sticky broadcast that is removed when sufficient
+      // storage is available again.  ACTION_DEVICE_STORAGE_OK is not sticky.  So if we
+      // don't receive anything here, we can assume that the storage state is okay.
+      return true;
+    } else {
+      switch (intent.getAction()) {
+        case Intent.ACTION_DEVICE_STORAGE_OK:
+          return true;
```
ojw28: I guess based on the comment above that you copied this code from elsewhere, but if `ACTION_DEVICE_STORAGE_OK` is not sticky, how does this case ever occur :)? Seems to me we can completely remove `ACTION_DEVICE_STORAGE_OK` from this case without any change in behavior?

library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
```diff
@@ -162,6 +175,34 @@ private boolean isDeviceIdle(Context context) {
...
+  /**
+   * Implementation taken from the the WorkManager source.
+   * @see <a href="https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/work/workmanager/src/main/java/androidx/work/impl/constraints/trackers/StorageNotLowTracker.java">StorageNotLowTracker</a>
+   */
+  private boolean isStorageNotLow(Context context) {
+    IntentFilter intentFilter = new IntentFilter();
+    intentFilter.addAction(Intent.ACTION_DEVICE_STORAGE_OK);
+    intentFilter.addAction(Intent.ACTION_DEVICE_STORAGE_LOW);
+    Intent intent = context.registerReceiver(null, intentFilter);
+    if (intent == null || intent.getAction() == null) {
```
ojw28: Do you understand what the purpose of the `intent.getAction() == null` check is here?

ojw28: Ignore this; figured it out!

library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
ojw28: As a summary, I wonder whether this entire method can just be:
```
@Nullable Intent intent =
    context.registerReceiver(null, new IntentFilter(Intent.ACTION_DEVICE_STORAGE_LOW));
return intent == null;
```

library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
```diff
@@ -86,6 +86,21 @@ public boolean cancel() {
     return true;
   }
 
+  @Override
+  public Requirements getSupportedRequirements(Requirements requirements) {
+    Requirements supportedRequirements = requirements;
+    if (Util.SDK_INT < 26) {
```
ojw28, line 92: Hm. If it's supported for API level 26 and above, it needs wiring up in the `buildJobInfo` method below, doesn't it?

ojw28: No need to fix this. I can do it when I merge. Just confirming that the comment is correct is fine :).

library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java
ojw28: No need to fix this. I can do it when I merge. Although please check whether the proposed alternative is sufficient!

jdegroot-dss: You are correct! Tested this and it works as expected.

library/core/src/main/java/com/google/android/exoplayer2/scheduler/PlatformScheduler.java
jdegroot-dss: Good catch, that is correct.

ojw28:
I'm not sure why our tooling didn't mark this as merged, but it was (with some non-functional changes) in https://github.com/google/ExoPlayer/commit/496a315d917be2ed1a419aabc3820f663f12a4c7. Thanks!

PR #7422, opened 2020-05-24
Assume default 4G bitrate for 5G

noamtamim:
There's no data for 5G, but it's likely that 5G speeds will be closer to 4G than to Wifi. 

According to the bitrate tables in `DefaultBandwidthMeter`, in 109 out of 238 countries, 4G is faster than Wifi; In 64 of them the difference is 1200kbps or more - in favor of 4G.

CSV with data from the tables:
https://gist.github.com/noamtamim/64b9a85338bff5c22b609f591b77deb2


tonihei:
Thanks! I'll merge that soon.

Issue #7462, opened 2020-06-03
Subsample offset and cue timestamp parsing for WebVTT fails when PTS wraps.


stevemayhew:
### Issue description
ExoPlayer does not handle PTS wrap that is not propagated to the Cue timestamps, that is the timestamps continue in time past the max representable as a PTS value.  This is allowed by the Pantos spec as:

```
 When synchronizing WebVTT with PES timestamps, clients SHOULD account
   for cases where the 33-bit PES timestamps have wrapped and the WebVTT
   cue times have not.  When the PES timestamp wraps, the WebVTT segment
   SHOULD have a X-TIMESTAMP-MAP header that maps the current WebVTT
   time to the new (low valued) PES timestamp.
```

Note [this commit](https://github.com/google/ExoPlayer/commit/fed2a1a6eaf8025bc6ff8ff92c78ce7dcabc17d1)  to fix Issue #2928 does not fix this issue, as all the cue timestamps have values that are monotonically increasing after the PTS wrap, not just the first cue timestamp.

### Reproduction steps
The demo app reproduces this, I've sent an S3 bucket with a snap shot of a live stream that has run long enough for PTS to wrap.

### Link to test content
I will email the link to dev.exoplayer@gmail.com. 

### Pull Request
I've submitted a request that fixes the issue, we of course needed this quickly for a customer.  We are more then happy to add more unit tests around this change and  help regress other bugs that may or may not be related or affected by this [Bug 609].

### [REQUIRED] A full bug report captured from the device
There is nothing relevant in the logs.  The captions simply are queue to the sample queue with values so out of scope that playback quickly goes to hell.

### [REQUIRED] Version of ExoPlayer being used
2.11.4

### [REQUIRED] Device(s) and version(s) of Android being used
Reproduced on an Arris STB on Android Pie

andrewlewis:
@icbaker @AquilesCanta Please could you take a look?

AquilesCanta:
I thought ExoPlayer already supported this case, actually. I can take a quick look.

icbaker:
@AquilesCanta and I have taken a look at this. I'm not convinced the source media is valid/correct in this case. There's 2 (quite different) questions:

1. Should WebVTT headers be allowed on the same line as the `WEBVTT` magic string?
2. Should the `LOCAL` value of `X-TIMESTAMP-MAP` wrap when the `MPEGTS` value does?

### WebVTT header parsing

[...]

PR #7464, opened 2020-06-03
WebVTT parsing issues.

stevemayhew:
This pull fixes issue #7462  Cue timestamps do not wrap with PTS wrap in the stream


stevemayhew:
There are UT that fail, will look into these and fix them first.

stevemayhew:
We are working on an extractor test to verify the logic in `WebvttExtractor.processSample()`

All the core and hls tests now run, including the one just added for timestamp wrap detection.

library/common/src/main/java/com/google/android/exoplayer2/util/TimestampAdjuster.java
```diff
@@ -33,7 +33,7 @@
    * The value one greater than the largest representable (33 bit) MPEG-2 TS 90 kHz clock
    * presentation timestamp.
    */
-  private static final long MAX_PTS_PLUS_ONE = 0x200000000L;
+  public static final long MAX_PTS_PLUS_ONE = 0x200000000L;
```
stevemayhew: Can keep this private and move `adjustCueForPtsWrap(long)` into this class, pretty much a personal preference choice I could go either way.

library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttParserUtil.java
```diff
@@ -52,8 +54,16 @@ public static void validateWebvttHeaderLine(ParsableByteArray input) throws Pars
...
+    @Nullable  String line = input.readLine();
+    boolean isValidHeader = line != null && line.startsWith(WEBVTT_HEADER);
+    if (isValidHeader) {
+      // According to w3c spec, TIMESTAMP-MAP header can be on same line as WEBVTT header.
+      int headerStartAt = line.indexOf(TIMESTAMP_MAP_HEADER);
+      if (headerStartAt != -1) {
+        input.setPosition(headerStartAt);   // set so next readLine() reads TIMESTAMP header
+      }
+    }
+    return isValidHeader;
```
stevemayhew: It seemed best to keep all the logic dealing with the WEBVTT_HEADER here, Pantos is ambiguous as to termination, the w3c spec explicitly says it can be followed by a space or tab.   This code handles space, tab or newline termination.

library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttParserUtil.java
```diff
@@ -74,7 +84,21 @@ public static long parseTimestampUs(String timestamp) throws NumberFormatExcepti
...
+   * <a href="https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-04#section-3.5">section 3.5 </a>) the
+   * Cue timestamp can be non-monotonic, that is it <b>does not</b> wrap with the PTS.
+   *
+   * @param webvttCueTimestampMs - the timestamp in ms parsed from the VTT cue (non-monotonic)
+   * @return cue time, in ms adjusted for any PTS wrap.
+   */
+  private static long adjustCueForPtsWrap(long webvttCueTimestampMs) {
+    long cueAsPts = (webvttCueTimestampMs * 90) % (TimestampAdjuster.MAX_PTS_PLUS_ONE - 1);
+    return TimestampAdjuster.ptsToUs(cueAsPts);
   }
```
stevemayhew: All WebVTT timestamps pass through this logic, including the LOCAL time, so that should cover all the possible interpretations of Section 3.5 last paragraph.  Note, if the timestamp is less then 1 day, 2:30:43.717677 (33 bit PTS wrap max value), then this code becomes a noop.  If it's over that, then the actual value is the remainder (line 100).

library/core/src/test/java/com/google/android/exoplayer2/text/webvtt/WebvttDecoderTest.java
```diff
@@ -90,6 +92,24 @@ public void decodeTypical() throws Exception {
...
+    assertThat(subtitle.getEventTime(1)).isEqualTo(5_524_062_222L);
+
+    assertThat(subtitle.getEventTime(2)).isEqualTo(5_524_822_222L);
+    assertThat(subtitle.getEventTime(3)).isEqualTo(5_526_941_222L);
+
+    assertThat(subtitle.getEventTime(4)).isEqualTo(5_527_000_000L);
+    assertThat(subtitle.getEventTime(5)).isEqualTo(5_527_005_000L);
+
+  }
+
```
stevemayhew, line 112: This test covers `adjustCueForPtsWrap()` for wrapped or un-wrapped timestamps (unwrapped is probably covered by other tests.  I can remove that if people prefer?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
```diff
@@ -160,9 +160,8 @@ private void processSample() throws ParserException {
         vttTimestampUs =
             WebvttParserUtil.parseTimestampUs(
                 Assertions.checkNotNull(localTimestampMatcher.group(1)));
-        tsTimestampUs =
-            TimestampAdjuster.ptsToUs(
-                Long.parseLong(Assertions.checkNotNull(mediaTimestampMatcher.group(1))));
+        tsTimestampPts =
+                Long.parseLong(Assertions.checkNotNull(mediaTimestampMatcher.group(1)));
       }
```
stevemayhew: The MPEGTS values needs to stay a PTS un-molested.  You'll see why below on line 183 below.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
```diff
@@ -174,11 +173,25 @@ private void processSample() throws ParserException {
...
     }
 
+    // Compute subsampleOffsetUs, the time offset of the Cue timestamps in this extracted
+    // segment to media time.  This is describe in Pantos Section 3.5 WebVTT
+    //  (https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-04#section-3.5)
+    //
+
+    // Scale the MPEGTS "media time" (default is 0 if no TIMESTAMP-MAP) to ExoPlayer media time timebase
+    //
+    long mediaTimeOfSegmentStartUs = timestampAdjuster.adjustTsTimestamp(tsTimestampPts);
```
stevemayhew: First adjust the only PTS value we have, either the default of 0 (almost no one that has synchronized VTT to a live MPEG stream is able to do this) or the value from MPEGTS attribute.   This value is normalized to sample time (ExoPlayer's normalized time across all elementary streams).

We can debate a better name, this is basically the time for the start of the segment the VTT file covers.  Pantos calls this media time, but in ExoPlayer this is sample time right?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
```diff
@@ -174,11 +173,25 @@ private void processSample() throws ParserException {
...
+
+    // The "LOCAL" time (if any) set time the start time of this segment in Cue timestamp time
+    // base, time 0.  The delta from the adjusted PTS is the time to offset samples times with
+    //
+    long subsampleOffsetUs = mediaTimeOfSegmentStartUs - vttTimestampUs;
+
```
stevemayhew: The renderer adds this to the Cue timestamp to covert the Cue timestamp to media time, so we subtract the segment start time stamp, aka vttTimestampUs (maybe a different name would make this more clear)

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
```diff
@@ -174,11 +173,25 @@ private void processSample() throws ParserException {
...
+    // parse and output the first queue.
     long firstCueTimeUs =
         WebvttParserUtil.parseTimestampUs(Assertions.checkNotNull(cueHeaderMatcher.group(1)));
-    long sampleTimeUs = timestampAdjuster.adjustTsTimestamp(
-        TimestampAdjuster.usToPts(firstCueTimeUs + tsTimestampUs - vttTimestampUs));
-    long subsampleOffsetUs = sampleTimeUs - firstCueTimeUs;
+    long sampleTimeUs = firstCueTimeUs + subsampleOffsetUs;
+
```
stevemayhew: A further trip through the `adjustTsTimestamp()` will not work here.  The Cue's sample time is always simply it's timestamp + `subsampleOffsetUs` (already normalized to media time)

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/WebvttExtractorTest.java
```diff
@@ -30,6 +34,19 @@
...
+    // Test file has map with X-TIMESTAMP-MAP=LOCAL:01:32:02.778,MPEGTS:497050076  So, local time
+    // is offset -622us from PTS.
+    //
+    WebvttExtractor extractor = new WebvttExtractor(null,
+        new TimestampAdjuster(TimestampAdjuster.ptsToUs(497050076)));
+
+    FakeExtractorOutput output = TestUtil.extractAllSamplesFromFile(extractor, ApplicationProvider.getApplicationContext(), "webvtt/wrapped_pts");
+
+    output.assertOutput(ApplicationProvider.getApplicationContext(), "webvtt/wrapped_pts_golden_extractor_output");
+  }
```
stevemayhew, line 48: We can add more test data, not sure if it was ok to depend on the `testdata` module here.

testdata/src/test/assets/webvtt/wrapped_pts_golden_extractor_output
```diff
@@ -0,0 +1,16 @@
...
+  getPosition(0) = [[timeUs=0, position=0]]
+numberOfTracks = 1
+track 0:
+  total output bytes = 669
+  sample count = 1
+  format 0:
+    sampleMimeType = text/vtt
+    subsampleOffsetUs = 622
+  sample 0:
+    time = 5522778844
```
stevemayhew: This is correct for the first sample, the first time stamp is: 266:39:21.999, which unwrapped is 5522778222, or 1:32:02.778222.

Add the offset:

````
>>> print 5522778222 + 622
5522778844
````

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
icbaker: We shouldn't pass this through the stateful `timestampAdjuster.adjustTsTimestamp()` method it updates internal state in `TimestampAdjuster` assuming this is the 'next' sample timestamp, and it might not be a sample timestamp at all (see comment below about how this is just a mapping, it might not relate to the cue timestamps at all).

The only values that should be passed through `adjustTsTimestamp()` are times that correspond to actual samples to be displayed. This is distinct from the static methods on `TimestampAdjuster` which don't update any internal state. I think that's why the existing code is careful to use the static methods until initialising `sampleTimeUs` below.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java
icbaker: `vttTimestampUs` is not the segment start time stamp. It's only valid for mapping directly to the MPEGTS value from the X-TIMESTAMP-MAP header (`tsTimestampsPts`). Paragraph 3.5 says:
"The cue timestamp in the LOCAL attribute MAY fall outside the range of time covered by the segment." So we can't assume it indicates the start of the segment.

stevemayhew:
@icbaker  are you proposing this commit ee0c6224 with this pull request or instead of it?  

I'm fine to keep the \t vs \n bug fix ourselves in our own branch and remove it from the pull request (FWIW, this is a major origin vendor that presents the header in this way, they argue that Safari plays the content) we will keep trying to convince them of the error in their ways, thanks for your support with this with data from the specs.

I'll rebase this pull then to include your fix and my test cases (which I"ll change to fix with your change).  I'm playing this branch against a live feed of the channel, will let you all know if there are further issues.

icbaker:
> @icbaker are you proposing this commit [ee0c622](https://github.com/google/ExoPlayer/commit/ee0c6224af554bd076fcc91bc7d7b7d1cf7853d5) with this pull request or instead of it?

I'm proposing "instead of", or at least instead of the changes in `WebvttExtractor`. I have added a test to `WebvttExtractorTest` in a follow-up commit (just pushed) - but it doesn't cover all the way to rendering.

> I'm fine to keep the \t vs \n bug fix ourselves in our own branch and remove it from the pull request (FWIW, this is a major origin vendor that presents the header in this way, they argue that Safari plays the content) we will keep trying to convince them of the error in their ways, thanks for your support with this with data from the specs.

Have you shown categorically that Safari is parsing the `X-TIMESTAMP-MAP`? Because the example you provided to us only has that 622µs difference between TS and WebVTT time, I'm not sure how you could tell the difference between parsing it and ignoring it (and assuming 0:0) - it would only shift the subtitles by 622µs which I suspect is undetectable.

(maybe the offset isn't always 622µs on this channel, and the sample you provided just happens to have that)

For now I think we should keep our parsing as-is (require the header on a new line), though I'm happy if you'd like to file a separate issue to track this and others can reply there if they start seeing more content that looks like this out in the wild.

FWIW you can probably carry a simpler patch in your own branch by just adding `webvttData.setPosition(0);` after this line:
https://github.com/google/ExoPlayer/blob/b1e56304a1fda8075fc637074927c0886f49fdf1/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java#L141

And then change L151 to `.contains("X-TIMESTAMP-MAP")`:
https://github.com/google/ExoPlayer/blob/b1e56304a1fda8075fc637074927c0886f49fdf1/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/WebvttExtractor.java#L151

This just rewinds to look at the `WEBVTT` line in the first iteration of the loop.

> I'll rebase this pull then to include your fix and my test cases (which I"ll change to fix with your change). I'm playing this branch against a live feed of the channel, will let you all know if there are further issues.

Thanks! Let me know if you think we need more test coverage in `WebvttExtractorTest` - I suspect what I've added is good enough for now - but if you've got a test that is more end-to-end that would be interesting.

stevemayhew:
@ojw28  and @icbaker  Please let me know if you want any of the unit tests.  I am working on a test for the balance of the samples (only the first sample is dealt with by WebvttExtractorTest), but if you are working on this internally I'll stop.  

We are working to increase the unit test coverage (our internal organization has coverage goals).

I'm perfectly fine if my commits are cleaned up before merging in, When it creates conflicts in the pull request though I'd appreciate a heads up as to what I should do with the pull request, I don't want to spend time on work that is duplicated.

stevemayhew:
> > @icbaker are you proposing this commit [ee0c622](https://github.com/google/ExoPlayer/commit/ee0c6224af554bd076fcc91bc7d7b7d1cf7853d5) with this pull request or instead of it?
> 
> I'm proposing "instead of", or at least instead of the changes in `WebvttExtractor`. I have added a test to `WebvttExtractorTest` in a follow-up commit (just pushed) - but it doesn't cover all the way to rendering.
> 

Ok.. I'll shoot a note here before I work on any other changes.  I'm testing re-based with your changes on the channel here.  I reverted my `WebvttExtractorTest`

> > I'm fine to keep the \t vs \n bug fix ourselves in our own branch and remove it from the pull request (FWIW, this is a major origin vendor that presents the header in this way, they argue that Safari plays the content) we will keep trying to convince them of the error in their ways, thanks for your support with this with data from the specs.
> 
> Have you shown categorically that Safari is parsing the `X-TIMESTAMP-MAP`? 

That is a great point.   Yes, I can modify locally to shift it many seconds and maybe prove Safari does not parse it, that will help convince the origin vendor they need to fix it.  Meanwhile we will keep that change out.

> For now I think we should keep our parsing as-is (require the header on a new line), though I'm happy if you'd like to file a separate issue 

Yes, agreed..  Hopefully we can prove Safari ignores it too.

stevemayhew:
@icbaker  Good call.. I'll file the bug back to the origin vendor.  Safari ignores the malformed WEBVTT header.   Thanks so much for the background from the specs, that will help.

If you've got a handle on doing test cases and all I can close this pull request and let you continue.  Please let me know if there is anything else we can help with.

icbaker:
> @icbaker Good call.. I'll file the bug back to the origin vendor. Safari ignores the malformed WEBVTT header. Thanks so much for the background from the specs, that will help.

Interesting! Thanks for checking Safari with a bigger offset.

> If you've got a handle on doing test cases and all I can close this pull request and let you continue. Please let me know if there is anything else we can help with.

I'm planning to leave the testing as-is (i.e. https://github.com/google/ExoPlayer/commit/b7486f4883384ab564531db2a5303e33dd26d3d9, checking the output of WebvttExtractor). It's not perfect, as you say ideally we'd check on the rendering side, but writing tests of that scope is hard (harder than it should be probably) and I don't have the bandwidth to dedicate to working on it right now I'm afraid.

So I think you can probably close this.

stevemayhew:
We’re fine doing a test case on the render side, my organization is motivated to increase test coverage and it’s always good to have test cases around code that is complicated to understand and possibly will get refactored.

Will rebase rest of the changes out then, or just open a new pull 


From Steve's iPhone

stevemayhew:
I’ll open a fresh pull request with the test cases when they are ready, everything substantial in this branch is “merged”.  Thanks @icbaker

PR #7497, opened 2020-06-14
Added Display Orientation SEI parsing for h264

tunjid:
This PR allows for setting a flag to wait for and parse the display orientation of a video from the Display Orientation SEI before setting the video format.

From D.1.27 in ITU-T Recommendation H.264 (2014 onwards): Display orientation SEI message syntax

diosei

Details on the parameters can be found in ITU-T Recommendation H.264: D.2.27 Display orientation SEI message semantics

This is useful for HLS streams generated from FLV streams pushed over RTMP encoded by the Android `MediaCodec` class so the publisher does not have to manually rotate the image while streaming.

I wanted to add a test for this, but the format and generation of the dump files used to test `sample.ts` is not readily apparent.


tunjid:
Alternatively, if `TrackOutput.format(Format format)` is fairly indempotent, if the display orientation changes from one frame to the next, perhaps `format` should be called again with the updated format.

ojw28:
@tunjid - Outputting the updated format sounds like a better idea to me, primarily because it avoids the need to add a special flag.

tunjid:
@ojw28 Calling format multiple times on the `TrackOutput` causes the video to freeze while testing. My test bitstream currently appends display orientation SEI every 150 frames for a 30 FPS video. Are there any limitations or rules to the `format()` method I should be aware of?

tunjid:
The log output:

```
25750-4388 D/SurfaceUtils: disconnecting from surface 0xeb619c98, reason disconnectFromSurface
2020-06-15 21:04:37.283 25750-5940 I/OMXClient: IOmx service obtained
2020-06-15 21:04:37.286 25750-5939 D/SurfaceUtils: connecting to surface 0xeb61de98, reason connectToSurface
2020-06-15 21:04:37.287 25750-5939 I/MediaCodec: [OMX.google.h264.decoder] setting surface generation to 26368005
2020-06-15 21:04:37.288 25750-5939 D/SurfaceUtils: disconnecting from surface 0xeb61de98, reason connectToSurface(reconnect)
2020-06-15 21:04:37.288 25750-5939 D/SurfaceUtils: connecting to surface 0xeb61de98, reason connectToSurface(reconnect)
2020-06-15 21:04:37.290 25750-5940 E/ACodec: [OMX.google.h264.decoder] setPortMode on output to DynamicANWBuffer failed w/ err -1010
2020-06-15 21:04:37.294 25750-5940 I/ACodec: codec does not support config priority (err -1010)
2020-06-15 21:04:37.317 25750-5939 D/MediaCodec: [OMX.google.h264.decoder] setting dataspace on output surface to #103
```

It did unfreeze after a while with the proper orientation, maybe I was being too aggressive while testing?

ojw28:
Hm, thinking about this a bit more, perhaps that suggestion just doesn't work very well. Can we rewind a bit and discuss the actual use case. I'm struggling to parse this sentence:

> This is useful for HLS streams generated from FLV streams pushed over RTMP encoded by the Android MediaCodec class

Could you explain a bit more about the use case, and why it needs to be this complicated? It seems like a pretty niche case for adding this plumbing through two different extractors.

> My test bitstream currently appends display orientation SEI every 150 frames for a 30 FPS video.

Where exactly are you inserting the SEI messages? My reading of the spec is that you're pretty-much required to insert one just before every IDR frame. I.e., keyframes would probably have a NALU sequence like: ```AUD,SPS,PPS,SEI[display_orientation],IDR```. Is this where you're inserting them, and are you doing this for every IDR?

tunjid:
@ojw28 

I'm muxing H264 and AAC tags from Android's Media Codec to FLV for RTMP streaming. 

The video pipeline is:

Android Camera2 input surface -> Android Media Codec -> FLVMuxer

To make sure the video orientation is right, I'm using the Display Orientation SEI to let the player know how much to rotate the video, especially if it was recorded in portrait.

In my particular case, the server will ingest the RTMP stream and produce HLS streams for iOS and Android to consume, I only added it to the FLV VideoTagReader for completeness;  my test setup for example consumes the FLV stream directly from a local RTMP server with no HLS conversion.

For production, I do add it like you have it above `AUD,SPS,PPS,SEI[display_orientation],IDR` and it works great. For testing however I simply was curious how often the output can be changed and the idempotency of calling it, so I put it in randomly in front of any IDR and non-IDR Access units to stress test, hence my asking if I was being too aggressive with it.

ojw28:
Does HLS playback on iOS respect the display orientation when you put it directly into the H264 bitstream, as you're doing?

Could you also provide a link to a working HLS stream that you've produced using this approach, for us to test with?

tunjid:
@ojw28 

I'm only aware of the Android side of things at the moment unfortunately. Ditto on the HLS stream as we use a third party vendor for it. I'll attach a raw h264 bitstream that has the display orientation SEI in it exactly as it will exist in the FLV Video Tag. The SEI byte sequence is `06 2f 03 10 00 09` followed by a RBSP stop bit in the byte `0x80`. The Entire sequence is `SPS, PPS, SEI [display_orientation], IDR, nonIDR...` the latter repeated for a few frames.

[display_orientation_sei.zip](https://github.com/google/ExoPlayer/files/4787443/display_orientation_sei.zip)

 
I do know however that rotating H264 video via metadata in the bitstream is admittedly  niche, `ffplay` ad VLC both don't do it despite parsing the SEI, they only support container metadata rotation. Makes some sense as Display orientation SEI is fairly recent addition, and it mobile was their use case for it.

tunjid:
I tried folding some logic into `SeiReader` as it's aptly named to parse SEI in general, but since I need to update the `TrackOutput` with the video `Format`, the changes required were too much.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/FormatHolder.java
```diff
@@ -0,0 +1,61 @@
...
+      format = newFormat;
+      output.format(newFormat);
+    }
+  }
+
+  /**
+   * Updates the {@link FormatHolder#output} with the {@link Format} derived from the passed in
+   * orientationData
+   */
+  public void update(DisplayOrientationSeiReader.DisplayOrientationData orientationData) {
```
tunjid, line 54: This may be better generically as a `Function<Format, Format>` and an implementation made as a static method reference in `DisplayOrientationData.update(Format format)`

tunjid:
@ojw28 is there anything I should change, or changes coming through the pipeline that would make this easier to merge?

ojw28:
We need to have a think about this on our side.

icbaker:

Issue #7544, opened 2020-06-25
Measure framerate and 99/90/50 percentiles in tunneling mode


changxiangzhong:
### [REQUIRED] Searched documentation and issues
- Official MediaCodec API
- ExoPlayer developer documents/javadoc
- ExoPlayer videos on Youbute
- ExoPlayer source code
- #4088

### [REQUIRED] Question
How to measure framerate, 99th/90th/50th percentiles of FPS on both tunneling and non-tunneling usecases. Ticket #4088 recommended to use the `releaseTimeNs` in the following segment to calculate framerate.
```  
protected void renderOutputBufferV21(// inside MediaCodecVideoRenderer
      MediaCodec codec, int index, long presentationTimeUs, long releaseTimeNs) {
```

It is viable solution for **non-tunneling** mode, but not for **tunneling** mode. In tunneling mode, such method is not invoked. 

So we are trying to find another solution. The following is our proposal to record the time that video buffer being rendered on the screen. We are aware of it's impossible to get the real timing. But we want to get the information as accurate as possible, OFC, for **both** tunneling and non-tunneling mode. 
```
@@ -1018,6 +1019,8 @@ public class MediaCodecVideoRenderer extends MediaCodecRenderer {
   @CallSuper
   @Override
   protected void onProcessedOutputBuffer(long presentationTimeUs) {
      // It's very simple, just save the time to an array, and the consumer would take it 
      // and reset the renderTimeIdx
+    decoderCounters.renderTime[decoderCounters.renderTimeIdx++] = System.nanoTime();
     if (!tunneling) {
       buffersInCodecCount--;
     }
```

### [REQUIRED] Question 1
Is this feature in the road map? We are willing to develop this tiny feature and contribute to the community.

changxiangzhong:
After digging a bit more into `DebugMediaCodecVideoRenderer` it seems like the method parameter `presentationTimeUs` indicates the actual buffer presentation time **both** with tunneling and w/o tunneling. Can you pls clarify this?
```
    @Override
    protected void onProcessedOutputBuffer(long presentationTimeUs) {
      super.onProcessedOutputBuffer(presentationTimeUs);
      bufferCount++;
      long expectedTimestampUs = dequeueTimestamp();
      if (expectedTimestampUs != presentationTimeUs) {
        throw new IllegalStateException("Expected to dequeue video buffer with presentation "
            + "timestamp: " + expectedTimestampUs + ". Instead got: " + presentationTimeUs
[...]

Issue #7100, opened 2020-03-17
Count droppedFrames in tunneling mode


Cizor:
###  Searched documentation and issues
1. Exoplayer Issues
2. https://medium.com/google-exoplayer/tunneled-video-playback-in-exoplayer-84f084a8094d

### [REQUIRED] Question
How to count dropped frames count when tunneling mode is enabled?
For non tunneling mode I am using decoderCounters.droppedBufferCount but code in MediaCodecVideoRenderer doesn't seem to handle it for tunneling mode. Please advise.

### Link to test content
Content used from Demo app

Cizor:
I am not very sure about it but since decoding of video is not in exoplayer code, it cannot determine if a frame is late, very late or should be dropped when tunneling mode is enabled

krocard:
The tunneling decoder reports rendered frames with the callback `MediaCodec.OnFrameRenderedListener.onFrameRendered` which calls `onProcessedTunneledBuffer` which updates `renderedOutputBufferCount` but not the drop counter.
`onProcessedTunneledBuffer` could try to guess if a frame was not rendered by comparing the received timestamps to the expected ones.

As the tunneling decoders are not required to call `onFrameRendered` for every frame (and I have not tested if they do either). So it's not sure if the resulting metrics will be reliable.

Unfortunately I will not access to a tunneling setup until the London office reopens (COVID-19 confinement).

krocard:
@Cizor, changxiangzhong in #7544 has done some testing regarding the tunneling `onFrameRendered ` callback and it seems that the callback is reliable enough to detect frame drops, at least for their device.

I would encourage you to try the same methodology if possible to validate that it successfully detects drop frames for your devices.

changxiangzhong:
@krocard I've found in the `DebugMediaCodecVideoRenderer::onProcessedOutputBuffer()` in *dev-v2* branch, there are some logic regarding frame drop detecting. Source code is as following. I think we can reuse and modify it to detect frame drops.
```Java
    @Override
    protected void onProcessedOutputBuffer(long presentationTimeUs) {
      super.onProcessedOutputBuffer(presentationTimeUs);
      bufferCount++;
      long expectedTimestampUs = dequeueTimestamp();
      if (expectedTimestampUs != presentationTimeUs) {
        throw new IllegalStateException("Expected to dequeue video buffer with presentation "
            + "timestamp: " + expectedTimestampUs + ". Instead got: " + presentationTimeUs
[...]

PR #7604, opened 2020-07-07
Measure frame drops in tunneling mode

changxiangzhong:
A discussion can be found [here](https://github.com/google/ExoPlayer/issues/7544).

There are two major components here. (I like the 1st name, but not the 2nd, feel free to suggest).
- `FpsMediaCodecVideoRenderer4Tunneling` - This is very similar `DebugMediaCodecVideoRenderer`, basically it records the expected PTS of each frame while feeding it into the `MediaCodec::inputBuffer`. And compare it after it's rendered to determine whether there's a frame drop.
- `FpsMediaCodecVideoRenderer4Tunneling2` - Is inspired by @krocard in the discussion. Basically we just log the PTS after the buffer has been rendered/released. One issue here is we can't know the actual frame-rate. So we are using some awk to guess the expected frame rate. The following is my script to **filter out frames takes  45+ ms**.
```bash
chang@p300a:~/Desktop$ adb logcat com.canaldigital.ngp:I  *:S FPS:V |grep -v "===>" | \
awk '/pts/{if($10-t > 45 && $10-t < 100000 || $10 - t < 0) printf "%d\t%d\t%d\t%d\n", $10, $10-t, $13-$10, $13-c} {t=$10; c=$13}'
```

Regarding `FpsMediaCodecVideoRenderer4Tunneling2`, due to the fact that we are **guessing** the FPS, we can't report the frame drop through `updateDroppedBufferCounters()`. But we can use this method to verify item 1

The following is a screen shot output by item 2
Screenshot from 2020-07-07 23-21-00

The following is a screen shot output by item 1. The result matches perfectly.
Screenshot from 2020-07-07 23-21-24

OFC we can remove code about item2, it's just test code. I haven't figured out how to make use of it as testcases. Any suggestions would be appreciated.

Test
Test platform: Canal Digital OnePlace - a derivation from BCM banff
How to introduce video jank: Open the Control UI, switch the focus as fast as you can to trigger the animation and stress the box. This is a very effective way to introduce video jank.

Thanks a lot.


changxiangzhong:
@krocard Any progress on this MR?

krocard:
@christosts @tonihei, this PR felt between the cracks. Drop frame measurement in tunneling has been requested multiple time. The approach had been discussed in https://github.com/google/ExoPlayer/issues/7100#issuecomment-653938758.

tonihei:
I'd suggest we close this PR and only leave #7100 to track the feature request. The reason is that we can't take this PR as it is anyway, because it adds new special renderers with additional threads just to count dropped frames (which is definitely too much customization and computation for this purpose). #7100 links to this PR, so future readers can look up the idea still.

PR #7625, opened 2020-07-14
Make it possible to override DefaultAudioSink

henrikpersson:
My ultimate goal is to be able to access the `AudioTrack` inside the `DefaultAudioSink`, so I can check the audio routing via [OnRoutingChangedListener](https://developer.android.com/reference/android/media/AudioTrack#addOnRoutingChangedListener(android.media.AudioTrack.OnRoutingChangedListener,%20android.os.Handler))/[getRoutedDevice](https://developer.android.com/reference/android/media/AudioTrack#getRoutedDevice()).

However I started small to see if there's any plans to support this at all, or a better way it could be implemented?


henrikpersson:
@googlebot I signed it!

ojw28:
Something equivalent to this was committed in https://github.com/google/ExoPlayer/commit/363a2a3b45488c53f12c8382d3c6a8b41ac5866f#diff-73b20e1913ef54168a567fac9f548c7b, so I don't think we need this PR any more. Please let us know if that change doesn't do what you need. Thanks!

Issue #7267, opened 2020-04-20
Read data from CodecPrivate when playing AVC or HEVC


steffenmanden:
**[REQUIRED] Use case description**

It has recently been decided in the mkv standard to add support for extra data in the codecprivate area when playing AVC or HEVC content.

This allows Dolby Vision data, as well as MVC data, to be properly read during playback and thus enable Dolby Vision support and more.

**Proposed solution**
It would be great it Exoplayer could be made possible to playback using this data allowing monitors to utilize Dolby Vision data.

The create of MakeMKV has allowed contact to him for easy explaination of how the data is stored and implemented

**Alternatives considered**
Workarounds are currently used through mp4 and ts files - however they lack a lot of features regarding subtitles, chapters and more! Directly play from the most common format (mkv) would make it super easy to use for the average user.

kim-vde:
Is it standard to have Dolby Vision streams inside MKV? It is not listed in the Matroska [codec IDs](https://matroska.org/technical/specs/codecid/index.html).

steffenmanden:
It's not in the standard definition. Dolby Vision has not previously been in the standard.

This implementation extends the format to allow for Dolby Vision within this format. Currently solutions (mp4 and ts) however have the standard, however are limited in a lot of features.

So although its not a part of the standard of mkv, it is an extention of the standard which will improve the player capabilities in a massive way without going out of bounds in a massive way.

MKV is after all the most used format :)

kim-vde:
Ok. I have flagged the issue as a low-priority enhancement for now.  Thanks for reporting it.

steffenmanden:
Thanks :) - referenced the MakeMKV creator to comment in here, as he can give more details as to how the format was extended! Hopefully he will comment

MikeChenMM:
I'm author of MakeMKV. For the record, I 'm  willing to provide any help to make this happen.

Now, few points:
"Matroska standard" is de-facto non-existent thing and is very outdated. :(
Dolby Vision is not a separate codec, but rather a (HEVC) stream extension - DV data (EL+RPU) are encoded in separate NAL units, according to Dolby spec.
The presence of DV data can be checked by several different methods - most obvious are:
(1) DV descriptor in codecprivate data (there is an "standard" of storing additional elements in codec private for AVC and HEVC streams, that was used for ages since 3D MVC video) or 
(2)presence of DV RPU unit in first frame.

[...]

PR #7667, opened 2020-07-24
MKV Dolby Vision Support

willtrking:
This adds in support for reading Dolby Vision data from MKV files per the modifications to the Matroska spec here https://github.com/cellar-wg/matroska-specification/pull/390

This should address https://github.com/google/ExoPlayer/issues/7267


steffenmanden:
Just for the sake of convenience, here is some test material with DV profile 5 and 8 (video only, no audio in the files):

https://www.makemkv.com/download/dvtest/

willtrking:
Great stuff @steffenmanden! I tested locally and ExoPlayer properly extracts the DV info. I'm working on cutting them down to the first few frames to be of a smaller file size so I can included them in this PR as test media.

willtrking:
During the process of generating small enough MKV test media for this PR, I determined I needed to start from shorter mp4 test media as there isn't a current MKV editor that will properly handle Dolby Vision. I also noticed that ExoPlayer currently lacks Dolby Vision mp4 tests, even though it support Dolby Vision within mp4's already.

I generated short mp4 test media from Dolby's public Dolby Vision sample's using Dolby's public mp4 muxer/demuxer and have included tests to validate them in this PR as well. If you take a look at the `.dump` files, you will see the dolby vision codec / mime properly detected. These files also play properly in the real world using the ExoPlayer demo application on an nvidia shield.

Will update again when MKV test media is included.

Dolby's public Dolby Vision samples [can be found here](https://developer.dolby.com/tools-media/sample-media/video-streams/dolby-vision-streams/)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -155,8 +156,13 @@
   private static final int ID_BLOCK = 0xA1;
   private static final int ID_BLOCK_DURATION = 0x9B;
   private static final int ID_BLOCK_ADDITIONS = 0x75A1;
+  private static final int ID_BLOCK_ADDITION_MAPPING = 0x41E4;
   private static final int ID_BLOCK_MORE = 0xA6;
   private static final int ID_BLOCK_ADD_ID = 0xEE;
+  private static final int ID_BLOCK_ADD_ID_VALUE = 0x41F0;
+  private static final int ID_BLOCK_ADD_ID_NAME = 0x41A4;
```
kim-vde: Is it necessary to parse the value and name? Can't you just remove them from the list so that they are skipped?

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
```diff
@@ -74,4 +74,14 @@ public void mp4SampleWithEac3Track() throws Exception {
   public void mp4SampleWithEac3jocTrack() throws Exception {
     ExtractorAsserts.assertBehavior(Mp4Extractor::new, "mp4/sample_eac3joc.mp4", simulationConfig);
   }
+
+  @Test
+  public void mp4SampleWithProfile5DolbyVision() throws Exception {
+    ExtractorAsserts.assertBehavior(Mp4Extractor::new, "mp4/dvhe_05_09.mp4", simulationConfig);
```
kim-vde: Could you reduce the size of the test file to less than a second (these files are more than 2 MB large while other MP4 test files are generally under 200 KB)?

Ideally, it would be easier to move the new MP4 tests to a new PR.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -231,6 +237,18 @@
    */
   private static final int BLOCK_ADDITIONAL_ID_VP9_ITU_T_35 = 4;
 
+  /**
+   * Dolby Vision configuration for profiles <= 7
+   * https://www.matroska.org/technical/codec_specs.html
+   */
+  private static final int BLOCK_ADDITIONAL_ID_DVCC = 0x64766343;
```
kim-vde: Replace BLOCK_ADDITIONAL_ID_ prefix by BLOCK_ADD_ID_TYPE to distinguish between the BlockAddId inside Cluster element and the BlockAddIdType inside the Track element? Or is there a reason to use the same prefix?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
```diff
@@ -1933,6 +1978,9 @@ public void outputPendingSampleMetadata(Track track) {
     public TrackOutput output;
     public int nalUnitLengthFieldLength;
 
+    // Block additional state
+    private int blockAdditionalId;
```
kim-vde: Rename to blockAddIdType to match the specs?

kim-vde (commented):
Thanks for this PR! Unit tests for MKV would really be helpful. Hope you manage to generate test content :)

steffenmanden:
just wanted to note that mkvtoolnix now supports DV data and therefore mkv testdata should be possible to make for the testcase

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
willtrking: Removed

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
willtrking: Good point, renamed

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java
willtrking: Renamed

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
willtrking: I've reduced the file size of the mp4s. I decided the keep them in the same PR as these mp4's are intrinsically tied to the MKV's I'm generating (they are the source for the MKVs).

In reducing the file size the dolby vision profile changed from dvhe.05.09 and hev1.08.09 to  dvhe.05.05 and hev1.08.05 respectively. From what I've seen, these are less common profiles.

I was unable to generate even a 1 second clip that was under 2MB that maintained dvhe.05.09 and hev1.08.09. The tools provided by Dolby (found [here](https://github.com/DolbyLaboratories/dlb_mp4base) and [here](https://github.com/DolbyLaboratories/dlb_mp4demux)) do not provide functionality to generate a clip less then 1 second. The only publicly available Dolby Vision source material of multiple profiles is only available at the link in my above comment, and other tools (e.g. ffmpeg) do not currently preserve Dolby Vision metadata when changing the quality of the material. This has restricted me to the high-quality files Dolby provides, which means I must compromise when generating clips (filesize vs. common profile).

kim-vde: I think it is fine to have less common profiles here, as there is nothing profile specific in the code added in the extractor.

willtrking:
@kim-vde Apologies for the delay. MKV test files have been added

kim-vde:
Thanks @willtrking. Do you confirm that the test assets have been generated from public content and can therefore be open-sourced?

cannikin:
MediaInfo will be adding support for recognizing Dolby Vision inside MKV files in their next release: https://github.com/MediaArea/MediaInfo/issues/478#issuecomment-677483293

Very excited for this PR as it's the missing piece of the Dolby Vision playback puzzle! Thanks so much @willtrking!

willtrking:
Yes @kim-vde, as mentioned these were generated from Dolby's public samples at the link below

https://developer.dolby.com/tools-media/sample-media/video-streams/dolby-vision-streams/

kim-vde:
Hi @willtrking,

I am afraid that it is not sufficient for the samples to be public in order to use them as test contents. My previous message was incorrect, sorry for that.

We need to make sure that we are allowed by Dolby to modify the samples and add them to our tests. I don't think it is the case here (see the [Terms of use](https://www.dolby.com/about/legal/terms-of-use/)). Maybe you know more about this?

An alternative would be to use content that you own.

cannikin:
The problem is that not just anyone can create Dolby Vision clips in something like iMovie—you need to [apply for and purchase a plugin from Dolby](https://professional.dolby.com/licensing/apply-for-a-professional-license/) in order to create Dolby Vision-enabled content and add it into a professional video editing suite, like [Davinci Resolve](https://www.blackmagicdesign.com/products/davinciresolve/). I imagine not many people outside of professional film editors have access to these tools.

@willtrking I'm not familiar enough with the data structure of these video clips...is it possible to just munge the actual video content with random characters, but keep the DV metadata intact?

If not, is there anything we can do for now to get the code merged until we can find public domain clips @kim-vde? Can @willtrking remove the tests that depend on the embedded video clips until we can replace them?

willtrking:
I see @kim-vde 

Unfortunately I'm not really able to produce dolby vision content at home, at least easily or cheaply.

I found open source dolby vision content from Netflix at the link below, but this content uses a version of the Creative Commons license that specifically states that any content that is altered may not be shared, and the content would need to be altered to make it small enough to be a reasonable test.

https://opencontent.netflix.com/

I could simply remove the test media from the PR if that would be acceptable.

andrewlewis:
@ybai001 Do you have any minimal Dolby Vision media we could add to the project to use for testing? Thanks!

willtrking:
Munging the data is an interesting idea @cannikin, it may actually be possible to do using dolby's [mp4 demuxer](https://github.com/DolbyLaboratories/dlb_mp4demux). 

If this worked, it would result in a final video final that contains a partial piece of the original Dolby Vision metadata from the dolby vision data linked above, plus a video I created.

The file would, of course, still contain some amount of data that is covered by Dolby's Terms of Use as I cannot produce the Dolby Vision metadata on my own.

I'm not enough of a licensing expert to know if that would be acceptable @kim-vde

steffenmanden:
As far as i can see, no terms are listed for the LG demo files:

https://4kmedia.org/lg-earth-dolby-vision-uhd-4k-demo/

willtrking:
@steffenmanden to my knowledge, that source (4kmedia.org) is not associated with LG, which would mean the content may have terms located elsewhere

cannikin:
Another thing I thought about was reaching out to users in a thread like this https://liftgammagain.com/forum/index.php?threads/dolby-vision-and-textless.14828/ and see if someone that already has a license can help us out by generating some clips and giving them to us under something like the MIT license.

steffenmanden:
LG support says we are allowed to use it and also shorten the video. Got screens and will get an email
Screenshot_20200825-193049_Chrome

cannikin:
Here's a post over at makemkv.com where someone created a 10 second DV file from scratch: https://makemkv.com/forum/viewtopic.php?p=95224&sid=7675e2b1d2b2d4e8f6ece2fd45ef0af1#p95224 They're not 100% sure it works because they don't have a device capable of playing DV, but it sounds like this PR should be able to test it!

kim-vde:
I merged the PR without the tests.

The samples from LG cannot be used because we need a clear and specific license.

And even if we find content that’s clearly licensed, we would probably need to change the structure of our code in order to include this license, which seems too much overhead.

So I suggest we only add tests if the content is owned by the PR writer or is derived from content that is already in our test samples. Dolby Vision tests would still be very helpful (for MKV or for MP4) so, if you are able to generate such content, please open a new PR to add the tests.

willtrking:
Thank you @kim-vde !

kim-vde:
@willtrking I found some test samples that can be used ([Dolby Vision.zip](https://github.com/google/ExoPlayer/files/5157780/Dolby.Vision.zip)). Could you make one of them shorter, transcode it to Matroska and add MP4 and Matroska tests (or explain me how to do it)? Thanks!

Issue #7709, opened 2020-08-04
Support custom track ordering in TrackSelectionView / TrackSelectionDialogBuilder


yoobi:
### [REQUIRED] Searched documentation and issues

I've looked around stackoverfow and multiple forums but they just use `TrackSelectionDialog` without sorting the format

### [REQUIRED] Question

How can I sort the format inside `TrackSelectionDialog` ? I can't modify the videos as I'm just picking up the link from a WebService. Here is a [screen](https://imgur.com/a/ETbg84d) of what I'm describing.

my `TrackSelectionDialog` displays the format in the following order: 480p, 720p, 1080p, 360p, 256p and I'd like to sort the format in the following order: 256p, 360p, 480p, 720p, 1080p.

This is my code:

```
trackSelectionDialogBuilder = TrackSelectionDialogBuilder(context, context.getString(R.string.qualitySelector), trackSelector, videoRenderer)
trackSelectionDialogBuilder.setTrackNameProvider {
    // Override function getTrackName
    context.getString(R.string.exo_track_resolution_pixel, it.height)
}
return trackSelectionDialogBuilder.build()
```

ojw28:
There's no way to do this currently. I'll mark it as an enhancement.

In the meantime, it should be relatively easy for you to fork `TrackSelectionDialogBuilder` and `TrackSelectionView` into your own codebase, at which point you'll be able to customize them to exactly suit your use case.

yoobi:
So you mean I should copy your `TrackSelectionDialogBuilder` and `TrackSelectionView` and change them to add this `sortBy` method ?

ojw28:
We will use this issue to track adding support directly to the library, but in the meantime, yes, that's the easiest thing for you to do. You could also send a pull request directly to the project, if you're sufficiently motivated to do so!

ojw28:
Pull request was merged. Thanks!

PR #7715, opened 2020-08-05
Adding ascending sort of TrackGroup

yoobi:
This enhancement match with this issue https://github.com/google/ExoPlayer/issues/7709

This is my first time contributing to a project, feel free to tell me if I did something wrong.

I'm comparing the bitrate of the `TrackGroup` as it works for both Audio and Video.

I've also notice that the original `TrackGroup` from `trackGroups.get(groupIndex);` is always 0, maybe I could delete the first loop over the `groupIndex`

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
@@ -59,6 +59,7 @@ public final class TrackSelectionDialogBuilder {
   private boolean showDisableOption;
   @Nullable private TrackNameProvider trackNameProvider;
   private boolean isDisabled;
+  private boolean sortTrackAscending = false;
   private List<SelectionOverride> overrides;
 
   /**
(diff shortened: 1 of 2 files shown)


yoobi:
@googlebot I signed it!

yoobi:
It seems I've made a mistake, when selecting the format (ie: 250p) it is not switching to it but to another one.

ojw28:
In terms of making this API generally useful, could you design it so that the caller can pass any `Comparator<Format>`, rather than just providing a boolean option to enable one specific type of sorting? This would also be useful for other use cases, for example sorting text tracks based on their language.

yoobi:
You're right, I'll use a `Comparator<Format>` instead, problem is I don't see where to do the sort.
My previous problem was when I changed the order in `TrackSelectionView` the array was sorted and then displayed in the UI, but behind when I was clicking on a `Format` it was using the old array.
Or Should I pass the `Comparator` to `MappedTrackInfo` and make the sorting there ?

yoobi:
I've been doing the sorting (inside `MappingTrackSelector`) but the issue is once it is all sorted, it no longer goes into `onTracksChanged`, and I've got no error. Do you have any idea why?

ojw28:
Obsoleted by https://github.com/google/ExoPlayer/pull/7798

Issue #7747, opened 2020-08-12
Need help about pushing my code to repository


msilucifer:
I made a customized engine for Australian Subtitles in streaming mode.
I want to share this code but I am not sure how to push it.
When i try to push my code it says the following.

```
remote: Permission to google/ExoPlayer.git denied to msilucifer.
fatal: unable to access 'https://github.com/google/ExoPlayer.git/': The requested URL returned error: 403
```

Waiting for your helps.
Best

zsmatyas:
I think you need to fork your own copy of the exoplayer repository. You will have full access to that copy. Then you push the changes to that copy. After this in the stock exoplayer repository, you can create a pull request between the two.

msilucifer:
Thanks, @zsmatyas .

But I need to push my code to one of new branches in Google's Exoplayer.
Not my repository.

And how did others create their branch in Exoplayer repository? I can't make mine there

Waiting for your helps.
Regards.

andrewlewis:
Forking and pushing changes to the fork sounds right as a way to share the code.

For contributing this to ExoPlayer please see #7746.

msilucifer:
Thanks for your help.

Issue #7746, opened 2020-08-12
Australian Subtitle is not showing for livestream


MrtDolphin0128:
Subtitles are not visible for streaming videos.

andrewlewis:
Could you provide some details about how this is related to #7747 and #7749? We need the information requested in the [new issue template], so please fill that out.

Setting aside the lack of information about what the enhancement is, the pull request currently has a few issues that mean we couldn't merge it (for example, it commits native binaries into the core module, and seems to add a lot of debug logging).

MrtDolphin0128:
It is for the case when we try to play UDP stream,
In this case subtitles are not showing.
Anyway i made custom engine to make Australian subtitles visible but I am not sure how to push my code to google repository.
Can you help me with this?
Best

andrewlewis:
Please fill out the [new issue template](https://github.com/google/ExoPlayer/blob/release-v2/.github/ISSUE_TEMPLATE/feature_request.md). For example, it would be useful to know what specification the subtitles follow and/or what specific features are missing, and how libcce was built in your pull request.

MrtDolphin0128:
libcce is from https://github.com/CCExtractor/ccextractor .
I built .so files from it and used it in my customized engine.
So can you please help me to push my code?

andrewlewis:
> So can you please help me to push my code?

The first step is to fill out the [new issue template](https://github.com/google/ExoPlayer/blob/release-v2/.github/ISSUE_TEMPLATE/feature_request.md) as I've requested already twice. Please don't ignore this request this time. Make sure to include what specification the subtitles follow and/or what specific features are missing.

Then we will review the information and decide whether to add this functionality to ExoPlayer. I'm afraid we can't promise to include this, because whenever we merge new functionality we need to spend resources reviewing and maintaining it. If we decide that it's worth it we'll take a look at the pull request. I guess this may be related to the discussion in #7220. To set expectations, I'd be surprised if we can merge this quickly given the pull request uses an external dependency and is non-trivial.

@icbaker Ian, do you have any guidance on whether this is worth adding and if so how it should be implemented? Thanks.

icbaker:
I think we need the info Andrew has requested from the [issue template](https://github.com/google/ExoPlayer/blob/release-v2/.github/ISSUE_TEMPLATE/feature_request.md) in order to properly evaluate the feature request.

[...]

Issue #7220, opened 2020-04-11
Add support for DVB EBU Teletext 


a4x4kiwi:
### [REQUIRED] Searched documentation and issues
Have checked documentation and see classes relating to subtitles, but would like example stringing the necessary pieces together.

### [REQUIRED] Question
Does anyone have an example of how to display subtitles from a UDP TS DVB stream? 
Any assistance would be most helpful.

Thanks in advance.

AquilesCanta:
Please try your stream in the demo app. You need to ensure the subtitles are being recognised by the player, and that the text track you want is selected. It's easy to confirm both by looking at the logcat.

For an example of how subtitles work end to end (the format shouldn't make a difference from the app's perspective), you should follow the demo app. If you are running into any specific issue, please share the elements required in the issue template (please include a media sample, and a bugreport) and we can address the more specific issues.

LRTNZ:
I also have been working on a media player project utilising UDP TS streams, and was wondering if I could get some help about how to get the subtitles embedded in these files to be played back?

I have tried finding documentation on how to do it, but I have been able to find no clear information on what exactly needs to be done. Yes, I have looked into the demo app, but could not find anything clearly showing how to do it.
If something could be put together by someone knowledgable on the workings of Exoplayer, based on the media samples below, I am sure many people will be very happy, based on the number of accounts I have come across online of people trying to achieve this. I for one, certainly would be, as it would solve a decent problem I am facing in the app that I am developing.

https://drive.google.com/open?id=1Q9n5KY-WSodNbVIyUntgz_jKzgR9GGQs
https://drive.google.com/open?id=1eTFX0KpQUog9bI5LQOe23K6B_FkgpErr

a4x4kiwi:
Thanks  AquilesCanta.
The subtitles are recognised but labelled as 'unknown' by the player. When 'unknown' is selected subtitles are not displayed as I presume they should be. This is an Australian DVB-T  channel.

I am not sure if this is a bug or not at this point.

The 2 example streams posted by LRTNZ (Also Australian TV) contain subtitles playable in VLC.
I have streamed them to the demo Exoplayer app (using http) with the same result. (They play audio and video, but no subtitles)

Do I need to formulate a subtitle URI somehow perhaps?

Below is the Logcat when playing a multicast stream.
[Exoplayer Logcat.txt](https://github.com/google/ExoPlayer/files/4478292/Exoplayer.Logcat.txt)

Thanks for any assistance.

[...]

PR #7749, opened 2020-08-13
Aus subtitle

MrtDolphin0128:
Customized Engine for Australian Subtitles in UPD streaming mode


andrewlewis:
If you just want to share this code somewhere, I suggest making a fork and pushing your changes there (#7747).

If the goal is to contribute it to ExoPlayer we need more information (#7746) and it also looks like this was probably sent from the wrong branch or something else has gone wrong given that a lot of other changes are included in the pull request in addition to https://github.com/google/ExoPlayer/pull/7749/commits/44d2cb83ae798077aaecad5ce7650b9fae36c195.

icbaker:
Moving the implementation/PR-specific discussion here, to avoid spamming #7220.

From https://github.com/google/ExoPlayer/issues/7746#issuecomment-673926423: 
> The library calls upon an external dependency, CCExtractor https://github.com/CCExtractor/ccextractor , per the advise by @icpaker.

To clarify, I advised you **not** to use CCExtractor and to implement this support in pure Java instead (https://github.com/google/ExoPlayer/issues/7746#issuecomment-673387453):
>  I think we're unlikely to accept a PR that implements this using CCExtractor when we weigh up the maintenance cost we'd be introducing against the benefits it brings.

To reiterate my statement from https://github.com/google/ExoPlayer/issues/7746#issuecomment-673425769, we will accept:

1. A high quality PR implementing this support in pure Java, with no external dependencies. Or,
2. A clear, reasoned argument with plenty of technical detail that convinces us a pure Java implementation for this is impossible, and a high quality PR adding an extension to support this using native dependencies (see the [vp9 extension](https://github.com/google/ExoPlayer/tree/release-v2/extensions/vp9) as an example).
    * Note: It's going to be challenging to convince us of this, since we already have pure Java support for what I believe are very similar subtitle formats embedded in MPEG transport streams (e.g. CEA-608).

If you want to add support for this format to ExoPlayer I'd strongly urge you to follow path 1 and do it in pure Java. If you're planning to take path 2 I'd suggest you convince us with the argument **before** spending any more time on coding/implementation, because as I say it's going to be hard to convince us.

a4x4kiwi:
@icbaker What can we do to progress the Australian Subtitles. @MrtDolphin0128 is developing this for us. Do you have any guidance to him to make the code compliant to assist with progressing the code into the project?

icbaker:
I'm going to close this as we're not going to accept it in its current form. Please re-open (or file a new one) with an approach that follows either of my two suggestions above.

PR #7798, opened 2020-08-24
dev: adding sorting for TrackSelectionDialog and TrackSelectionDialog…

yoobi:
This enhancement match with this issue https://github.com/google/ExoPlayer/issues/7709

This is my first time contributing to a project, feel free to tell me if I did something wrong.

Developers need to add their `comparator` which is type of `Comparator<Format>` and it'll sort it the way they implement it

PS: I got an issue with my repo and had to delete it, that's why there is 2 merge request, feel free to close/delete this one: https://github.com/google/ExoPlayer/pull/7715


library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
```diff
@@ -60,6 +62,7 @@
   @Nullable private TrackNameProvider trackNameProvider;
   private boolean isDisabled;
   private List<SelectionOverride> overrides;
+  private Comparator<Format> comparator;
```
ojw28: Let's add an `@Nullable` annotation to this.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
```diff
@@ -195,6 +198,12 @@ public TrackSelectionDialogBuilder setShowDisableOption(boolean showDisableOptio
     return this;
   }
 
+  public void setComparator(Comparator<Format> comparator) {
+    if(this.comparator != comparator) {
```
ojw28: There's no real benefit to doing this check, so you can get rid of it and just do `this.comparator = comparator` directly).

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
```diff
@@ -203,11 +208,13 @@ public void init(
       int rendererIndex,
       boolean isDisabled,
       List<SelectionOverride> overrides,
-      @Nullable TrackSelectionListener listener) {
+      @Nullable TrackSelectionListener listener,
+      @Nullable Comparator<Format> comparator) {
```
ojw28, line 213: Add `@param` Javadoc above for this.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
```diff
@@ -251,12 +258,13 @@ private void updateViews() {
     defaultView.setEnabled(true);
 
     trackGroups = mappedTrackInfo.getTrackGroups(rendererIndex);
+    sortedTrackGroups = sortedTrackGroups != null ? sortedTrackGroups : initSortedTrackGroups(trackGroups);
```
ojw28: I don't think this should be avoiding re-generating `sortedTrackGroups` if it's non-null. The `mappedTrackInfo` may have changed since this method was called, for example due to a call to `init`. Hence you always need to re-generate the `sortedTrackGroups` again.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
```diff
@@ -367,6 +380,48 @@ private void onTrackViewClicked(View view) {
     }
   }
 
+  private TrackGroupArray initSortedTrackGroups(TrackGroupArray trackGroups) {
```
ojw28, line 384: I don't think this method does quite the right thing. In particular, suppose that there are multiple groups in the `TrackGroupArray`. L394 below rewrites it so there's only a single group. Doesn't this cause all other groups to be lost?

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
```diff
@@ -367,6 +380,48 @@ private void onTrackViewClicked(View view) {
...
+    TrackGroupArray trackGroupArray = trackGroups;
+    if(comparator != null) {
+      TrackGroupArray trackGroupsArray = mappedTrackInfo.getTrackGroups(rendererIndex);
```
ojw28: I think this might be the same as the input argument to the method.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
```diff
@@ -367,6 +380,48 @@ private void onTrackViewClicked(View view) {
...
+   * @param trackIndex which index of the initial array
+   * @return index of the sorted array that correspond to the same element in the initial array
+   */
+  private int getSortedIndexFromInitialTrackGroup(int groupIndex, int trackIndex) {
+    int sortedTrackIndex = trackIndex;
+    if(sortedTrackGroups != trackGroups) {
+      Format selectedFormat = sortedTrackGroups.get(rendererIndex).getFormat(trackIndex);
+      int trackHash = selectedFormat.hashCode();
+      for (int formatIndex = 0; formatIndex < trackGroups.get(groupIndex).length; formatIndex++) {
+        if(trackGroups.get(groupIndex).getFormat(formatIndex).hashCode() == trackHash) {
```
ojw28: You can't rely on hash codes to determine equality. More generally though, I don't think it needs to be this complicated. There's a `TrackGroup.indexOf(Format)` method that you can use. I think you can probably replace the whole `for` loop with a call to this method.

ojw28 (commented):
I put some comments inline. For the demo app changes, I think we probably don't want to take those, since it makes the demo app quite a bit more complicated. Feel free to leave them in for now if they're useful for your own testing, but let's remove them when we merge.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
yoobi: You're totally right, I've changed for `indexOf` and it's working the same without the `for` loop

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionView.java
yoobi: > L394 below rewrites it so there's only a single group. Doesn't this cause all other groups to be lost?

Oh I see what you mean, I haven't been able to find a `TrackGroupArray` that has more than one group.
Yes this needs slight modification, currently it only saves the last group, I'll change it so it keeps all the groups and sort them according to the `comparator`

yoobi: > L394 below rewrites it so there's only a single group. Doesn't this cause all other groups to be lost?

Oh I see what you mean, I haven't been able to find a `TrackGroupArray` that has more than one group.
Yes this needs slight modification, currently it saves only the last group, I'll change it so it keeps all the groups and sort them according to the `comparator`

yoobi:
Everything should be okay now, I'm only concern if for some reason a `TrackGroup` has a length of 0, then there will be some issue, I haven't been able to find a `TrackGroupArray` with an empty `TrackGroup` in the demo app though

Issue #7803, opened 2020-08-26
Change theme or UI element of TrackSelectionDialogBuilder


yoobi:
### [REQUIRED] Searched documentation and issues
Stackoverflow and issues in this repo

### [REQUIRED] Question
I'd like to know if it's possible to change the theme of TrackSelectionDialogBuilder, because I can't find any method to do so.

How can I set a theme for my trackSelectionDialogBuilder?

Thanks

AquilesCanta:
@tonihei, mind answering this?

yoobi:
I've already made a PR, https://github.com/google/ExoPlayer/pull/7804

tonihei:
Thanks, just reviewed it :)

tonihei:
PR got merged, so I'll close this issue.

PR #7804, opened 2020-08-26
Add setting theme for TrackSelectionDialogBuilder

yoobi:
This PR is in relation with https://github.com/google/ExoPlayer/issues/7803

After searching a bit I found an easy way to add theme to the TrackSelectionDialogBuilder and thought some people might need it as well

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
@@ -24,6 +24,7 @@ import android.content.DialogInterface;
 import android.view.LayoutInflater;
 import android.view.View;
 import androidx.annotation.Nullable;
+import androidx.annotation.StyleRes;
 import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector.SelectionOverride;
@@ -49,6 +50,7 @@ public final class TrackSelectionDialogBuilder {
   }
 
   private final Context context;
+  @StyleRes private int themeResId;
   private final CharSequence title;
   private final MappedTrackInfo mappedTrackInfo;
   private final int rendererIndex;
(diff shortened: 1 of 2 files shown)


library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
```diff
@@ -238,8 +245,8 @@ private Dialog buildForAndroidX() {
       // the APK size even with shrinking. See https://issuetracker.google.com/161514204.
       // LINT.IfChange
       Class<?> builderClazz = Class.forName("androidx.appcompat.app.AlertDialog$Builder");
-      Constructor<?> builderConstructor = builderClazz.getConstructor(Context.class);
-      Object builder = builderConstructor.newInstance(context);
+      Constructor<?> builderConstructor = builderClazz.getConstructor(Context.class, int.class);
```
tonihei, line 253: Could you also update the proguard rules in library/ui/proguard-rules.txt accordingly? Otherwise the AndroidX constructor will be removed by proguard.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
```diff
@@ -121,6 +123,11 @@ public TrackSelectionDialogBuilder(
                     newOverrides.isEmpty() ? null : newOverrides.get(0)));
   }
 
+  public TrackSelectionDialogBuilder setTheme(int themeResId) {
```
tonihei: Can you add `@StyleRes` to the parameter too?
And also add a JavaDoc: "Sets the resource ID of the theme used to inflate this dialog."

tonihei (commented):
Thanks for the PR! Looks good in general, just some minor comments.

library/ui/src/main/java/com/google/android/exoplayer2/ui/TrackSelectionDialogBuilder.java
yoobi: I've updated `library/ui/proguard-rules.txt` and it should be good.

PR #7814, opened 2020-08-27
Authorize multiple preferred language and text on track selector

zubcoco:


zubcoco:
@googlebot I signed it!

christosts:
Hi @zubcoco 

Can you describe the use-case that requires this change? For example, is it that your app can specify multiple languages in the UI? Just trying to understand if this the optimal change for your use-case.

zubcoco:
Hello,
We are streaming live channels and each one has a default language.
In the UI, the user can specify one preferred language and the second one could be the channel language.

tonihei:
Just for further clarification to check if I understand your use case correctly:

You have a media stream of a known original language, let's say French (FR). Your user specifies a preferred language, e.g. Russian (RU).

If you want to select an *audio track*, it should be sufficient to use `setPreferredAudioLanguage("RU")`. 
- If the media contains ["FR", "RU"], Russian will be selected. 
- If the media contains something like ["FR", "DE"] or just ["FR"], French will be selected because it's the original audio track and first in the list. 
- if the media contains ["DE", "FR"] though, German would be selected. This case could benefit from a second preferred language. However, I'd say the media is slightly broken in this case because if no one sets a preference, media players will always play the first audio track by default. Alternatively, some streaming types (e.g. DASH) can define a "default" audio track to clearly indicate which one should be played when nothing else is specified. 

If you want to select a *text track*, `setPreferredTextLanguage("RU")` means that only Russian text tracks are selected:
- Media contains ["FR", "RU"] -> Russian text is shown.
- Media contains ["FR"], ["FR", "DE"] or ["DE", "FR"] -> no text is shown unless the French text track is marked as "default" or "forced". If your intention is to always show text (in either Russian or French), providing a second preferred language makes sense. What kind of streaming format are you using (DASH, HLS, SmoothStreaming)? And do you control the media creation? Because if the media requires to always show a text track, then it should probably specify the "default" or "forced" flags (depending on the streaming format). 

Could you clarify how your use case fits into the example above? We are not opposed to adding this feature, it would just be good to understand what requirements you have exactly to ensure we integrate it correctly.

zubcoco:
In my case I have an italien stream which has english or italien languages and the order is ["EN", "IT"].
And in my case the user preferred language is french. So I put setPreferredAudioLanguage("FR").
So I want to select italien language on this channel with setPreferredAudioLanguage(new String {"FR", "IT"}).

tonihei:
Thanks for the example, so this is the case where I said above _"This case could benefit from a second preferred language. However, I'd say the media is slightly broken because if no one sets a preference, media players will always play the first audio track by default. Alternatively, some streaming types (e.g. DASH) can define a "default" audio track to clearly indicate which one should be played when nothing else is specified."_

Do you control the creation of the media to set a default flag on the Italian track? This would be the correct solution to fix the media so that it plays correctly on all media players without an additional config.

zubcoco:
I don't control the creation of the media. So I can't set the default flag on the italien track :/

tonihei:
I guess it makes sense to provide the option to override the broken default. Another potential use case I could think of is an app where users can intentionally specify multiple language preferences so that the matching audio/text track is selected.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/TrackSelectionParameters.java
```diff
@@ -222,13 +238,13 @@ public static TrackSelectionParameters getDefaults(Context context) {
    * {@code null} selects the default track, or the first track if there's no default. The default
    * value is {@code null}.
    */
-  @Nullable public final String preferredAudioLanguage;
+  public final String[] preferredAudioLanguage;
```
tonihei: Can you make this an ImmutableList to avoid changes after the Parameter object was created? Same below for the text languages.

Also, could you add an "s" to both parameter names to make it "languages"?

library/core/src/main/java/com/google/android/exoplayer2/trackselection/TrackSelectionParameters.java
```diff
@@ -86,6 +86,14 @@ public Builder() {
...
      */
     public Builder setPreferredAudioLanguage(@Nullable String preferredAudioLanguage) {
+      if (preferredAudioLanguage == null) {
+        return setPreferredAudioLanguage(new String[0]);
+      } else {
+        return setPreferredAudioLanguage(new String[] { preferredAudioLanguage });
+      }
+    }
+
+    public Builder setPreferredAudioLanguage(String[] preferredAudioLanguage) {
```
tonihei: For the setters it's probably best to use "String..." as type to allow passing in a literal list, like `setPreferredAudioLanguages("FR", "IT")`. If you do this, your own code above can be simplified because you no longer need to create an array.

library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java
```diff
@@ -2619,11 +2625,16 @@ public AudioTrackScore(Format format, Parameters parameters, @Capabilities int f
...
-              format,
-              parameters.preferredAudioLanguage,
-              /* allowUndeterminedFormatLanguage= */ false);
+      int bestLanguageScore = 0;
+      for (int i = 0; i < parameters.preferredAudioLanguage.length; i++) {
+        int score = getFormatLanguageScore(
+            format,
+            parameters.preferredAudioLanguage[i],
+            /* allowUndeterminedFormatLanguage= */ false);
+        score = 1000 * score + parameters.preferredAudioLanguage.length - i; // Priorise the first items in array
```
tonihei: Can you use the same pattern as below for the system locale languages that works without magic score calculation? You can save the best score so far and replace the best match only if you find a better score.

tonihei:
Thanks, we should be able to merge this. I'll do some further style clean up when importing.

Issue #6166, opened 2019-07-11
Allow specifying per-item request headers (for playback and downloads)


chanbbong:
I have different cookies for each video.
so each time new video is loaded, I set DownloadManager with new cookie.
```defaultHttpDataSourceFactory.getDefaultRequestProperties().set("Cookie", cookieValue);```
however, I can only download video for the first time.
I think DownloadService seems to call DownloadManager only when DownloadService is started for the first time.
How do I download video with each different cookie?

ojw28:
I don't think we currently support this (you can of course implement your own download functionality, rather than relying on the out-of-the-box components that we provide). Marking this as an enhancement.

ojw28:
@marcbaechinger - We should consider whether we want to allow request headers to be specified in `MediaItem`, and (eventually) whether `DownloadRequest` should include a `MediaItem` rather than some of its current fields.

msesma:
Same problem here. Even if I don't allow the user to download two videos al the same time, the DownloadService only request DownloadManager once, so the second video will use the DownloadManger of the first one and fail because of wrong cookie.

BarryFruitman:
+1 for specifying headers in a `DownloadRequest` or `MediaItem`. Headers are used pretty commonly in authentication schemas and it would be useful to be able to specify them on a per-request basis.

NathanSass:
+1 I would also like this feature

kschults:
Hi @ojw28, I'm about to start looking into contributing a fix for this, but wanted to check in first with a couple of questions.

1) Are you already working on this? Or do you already have thoughts about what will be involved in fixing it?

2) Is there anything I need to know about developing on this project, aside from what's in `CONTRIBUTING.md`?

ojw28:
> Are you already working on this?

No

> Do you already have thoughts about what will be involved in fixing it?

Although not directly what you're trying to achieve, I think a good first step would be to add `requestHeaders` to `MediaItem.PlaybackProperties`, and use them in the various `MediaSource` implementations for playback. It doesn't really make logical sense to support something for downloads and not for playback, so I think this should be done first even if you don't need it yourself.

Then for downloads I think you'll need to:

1. Add a `requestHeaders` field to `DownloadRequest`
1. Upgrade the database logic in `DefaultDownloadIndex` to support persisting the headers
1. Use the headers in the various `Downloader` implementations

[...]

PR #7852, opened 2020-09-03
6166 request headers

kschults:


kschults:
@ojw28 Re: https://github.com/google/ExoPlayer/issues/6166

I've stubbed out the first part - adding the headers to `MediaItem` - but it feels very awkward to have to pass them around everywhere, especially when it's already possible to achieve this for playback by using a `HttpDataSourceFactory` that provides the headers.

I went through the HLS playback flow, and was able to get it working, but wanted to stop and check in before I started in on the other types of media source. I

kschults:
@googlebot I signed it!

kschults:
@ojw28 I'm struggling with the CLA here as well. My company has signed the CLA and added my email to our google group, and I edited the commits to use the same email, but it's still not recognizing it. It's also not showing when I go to https://cla.developers.google.com/clas.

christosts:
assigning to @ojw28 who is also handling #6166

ojw28:
I can't find any record of your company having a signed Google CLA, assuming your company is the one corresponding to the email address on your commits. Are you sure they've done so correctly?

Do you know when they signed it? If it's very recent then it's possible that it's still pending. If it was a while ago, is it possible that they signed the Android CLA rather than the Google CLA? In that case I think your company would need to sign the Google CLA. As far as I'm aware, the Android CLA is no longer a thing.

kschults:
@ojw28 I'll double-check with them, but they claimed to have signed it ~2-3 weeks ago.

kschults:
@googlebot I signed it!

kschults:
@ojw28 Thanks for checking on your end. We had missed an email to our authorized signer that required more action. Looks like we're all set now!

kschults:
I think I've added the headers everywhere for playback. There's a couple places where I'm not sure how to handle it, though. I've tried to add and update tests as I find them, but I definitely might have missed some.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ParsingLoadable.java
```diff
@@ -65,7 +66,8 @@
    */
   public static <T> T load(DataSource dataSource, Parser<? extends T> parser, Uri uri, int type)
       throws IOException {
-    ParsingLoadable<T> loadable = new ParsingLoadable<>(dataSource, uri, type, parser);
+    // TODO: What headers should be used here? This is only ever called from a test
```
kschults: Here's one place I wasn't sure what the purpose of it was, so I wasn't sure how to handle it.

kschults:
@ojw28 We'd love to get this into a future release, especially with the follow-up to add it to downloads. Is there anything I can do to help move this along?

library/core/src/main/java/com/google/android/exoplayer2/upstream/ParsingLoadable.java
```diff
@@ -106,13 +108,19 @@
   /**
    * @param dataSource A {@link DataSource} to use when loading the data.
    * @param uri The {@link Uri} from which the object should be loaded.
+   * @param headers Headers to be added when loading the data
    * @param type See {@link #type}.
    * @param parser Parses the object from the response.
    */
-  public ParsingLoadable(DataSource dataSource, Uri uri, int type, Parser<? extends T> parser) {
+  public ParsingLoadable(DataSource dataSource, Uri uri, Map<String, String> headers, int type,
```
ojw28: I don't think you need to change this. Where headers are needed a caller can just use the constructor below that takes a `DataSpec`. You may need to change some callers to do this.

library/core/src/main/java/com/google/android/exoplayer2/upstream/ParsingLoadable.java
ojw28: I don't think you need to do anything here. The constructor that takes a `DataSpec` can be used if headers need to be specified (see below).

ojw28 (commented):
Sorry for the delay on this. I took a look and left a few inline comments, but more generally, I'm a bit concerned that this is a partial solution to a problem, which isn't going to work in some cases. In particular:

* For adaptive content, it's unclear to which requests headers should really be attached. What if you only want to attach headers to the manifest requests and not to the segment requests? Is it possible that it might be actively a bad idea to attach the headers to segment requests, particularly if they're not going over HTTPS and/or are going to edge caches that you don't control. I'm not sure, but this feels like an assumption that might be incorrect for some cases.
* It doesn't handle the need to refresh header content (e.g., authorization tokens).

It seems this request adds an easier to use API for simple use cases, but it then creates two ways of doing the same thing. It's also somewhat unclear how we'd go about generalizing the new proposal to actually meet the demands of the cases outlined above, without things getting very messy. Any thoughts?

kschults:
Thank you for the comments!

I'm not sure what the right answer is for headers on segments vs manifests. For my use case, I want the same headers on all of the requests, but I guess I can see how that might not always be the case. I'm not sure how one would handle the use case of wanting different headers for the two.

I get the concern of having two different ways to do something. The issue with providing headers via the `DataSource` is that it doesn't translate into downloads, which is the end goal. Do you have thoughts about a different way to provide headers for downloads that could let us avoid having to have two different ways to add headers?

> It doesn't handle the need to refresh header content (e.g., authorization tokens).

By this, do you mean needing to update one of the headers at some point after playback has started? I suppose in that case we could add a way for a `MediaSource` to update part or all of its `MediaItem`? I'm less certain about how that would work, though.

Perhaps that's something that could be addressed in the future if the need arises?

kschults:
Updated the code @ojw28.

giangpham2811 (approved):

kschults:
@tinyboyX Thanks for the approval. There were some outstanding discussion items with @ojw28 - should we try to resolve those before merging, or have you discussed this with him?

kschults:
@ojw28 I'd like to be able to move forward and add the corresponding headers to downloads as well. I know you had some concerns about not being able to distinguish between wanting to add headers to the playlist URLs vs the segment URLs, but I think it's better than not being able to add headers at all. Any further thoughts or suggestions?

kschults:
Just a heads up that I'll be offline until at least late fall, and won't be able to respond or move forward on this

DoctorDooom:
Hallo, are there any updates regarding this MR ? Please merge it as soon as possible since it is required feature

marcbaechinger:
I don't think this PR is pursued any further I'm afraid. This repository is deprecated. The code moved to https://github.com/androidx/media with a new package name.

icbaker:
If someone is considering porting this PR to the media3 repo:

[Olly's concerns above]( https://github.com/google/ExoPlayer/pull/7852#pullrequestreview-522090961) about a single `MediaItem` mapping to multiple requests (in the case of adaptive content, or sideloaded subtitles), and not knowing whether all headers should be added to all requests, does suggest to me that perhaps we shouldn't add headers directly to `MediaItem`, and instead we should add some wiring that allows a `DataSource` to have a richer understanding of the origin of the request. A `ResolvingDataSource` could then add the appropriate headers for e.g. a manifest request for a particular `MediaItem`.

This also solves the 'refreshing headers' case, because the headers are resolved at request time instead of being hard-coded on the `MediaItem` when it's first passed to the player (or at least, they **can be** resolved late like this).

I'm not yet sure how to reconcile this with the download case.

I'm not sure exactly what this wiring looks like - it's possible we could just add the whole 'originating `MediaItem` instance' as a field to `DataSpec`, and maybe also a `C.DataType` field (we already [basically have this part for HLS](https://github.com/androidx/media/blob/release/libraries/exoplayer_hls/src/main/java/androidx/media3/exoplayer/hls/HlsDataSourceFactory.java)).

I would encourage someone looking to get this functionality merged to kick off a discussion on https://github.com/google/ExoPlayer/issues/6166 before sending a PR, as it's probably quicker to start by iterating on a design discussion instead of all the real code.

(To clarify: issues in this tracker are still active, it's only PRs can no longer be merged from here, so no need to file a new copy of https://github.com/google/ExoPlayer/issues/6166 on the media3 tracker)

DoctorDooom:
Thank you, @marcbaechinger, for your answer is there in Media 3 Repo any possibility to do what this feature does?

DoctorDooom:
@icbaker : I would work on it on the new Repo media 3, but let me first gather some info about the possibilities and i will open discussion there.

icbaker:
> i will open discussion there.

Please continue any discussion about possible solutions on https://github.com/google/ExoPlayer/issues/6166, and don't file a duplicate issue in the media3 repo - that will help to keep the discussion in one place. When you're ready to file a PR, that does need to be made on the media3 repo.

icbaker:

PR #7922, opened 2020-09-16
fix durationUs error caused by different track timescale

XiaoLongTX:
I have a video that contains two tracks. The timescale of one track is 1000, the timescale of the other track is 48000, and the movieTimescale is 1000. As a result, the video time I get is 48 times the real video time.


XiaoLongTX:
@googlebot I consent.

PR #7923, opened 2020-09-16
fix durationUs error caused by different track timescale

XiaoLongTX:
I have a video that contains two tracks. The timescale of one track is 1000, the timescale of the other track is 48000, and the movieTimescale is 1000. As a result, the video time I get is 48 times the real video time.


icbaker:
@kim-vde PTAL

kim-vde (commented):
The duration field comes either from the mehd box or from the tkhd box. In both cases, the spec indicates that it is given in the timescale indicated in the mvhd box (and not the mdhd box). This means that the code was correct.

The input file probably does not follow the spec. If you want us to have a closer look, please open an issue with the media file included and a reference to this PR for context.

Thanks.

PR #7931, opened 2020-09-16
FlacStreamMetadata: set bitrate and PCM encoding for Format

wnoun:
the extractor of Flac extension sets these values, the core extractor should have the same behaviour.


kim-vde (commented):
These fields are unset on purpose. The 2 FlacExtractors output different types of data: the extension extractor decodes the FLAC data in the extractor and outputs PCM data, while the core extractor outputs FLAC samples to be decoded by the renders.

The format bitrates refer to the bitrates of the encoded samples. It is fine to set them to the decoded bitrate for the extension extractor, as samples are not encoded. It would be incorrect for encoded samples though. Similarly, the PCM encoding should only be set for PCM audio.

PR #7935, opened 2020-09-17
remove statusListener after ended session

TakuSemba:
Hi, I've noticed that `SessionAvailabilityListener.onCastSessionUnavailable` is not called after `castPlayer.release()` is called.
As the current session would be destroyed at `release()`, I think it would be better `onCastSessionUnavailable` is notified.

Our use case is that we are trying to start casting on `onCastSessionAvailable`, and switch back to playing media locally on a device when `onCastSessionUnavailable`. However, I can't find the right time to start playing locally after session is ended at `release()`.
`SessionManagerListener` could be used, but I would rather use `SessionAvailabilityListener` because listener would be notified after remoteMediaClient is properly setup/released.

### DIFF
--- a/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
+++ b/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
@@ -480,8 +480,8 @@ public final class CastPlayer extends BasePlayer {
   @Override
   public void release() {
     SessionManager sessionManager = castContext.getSessionManager();
-    sessionManager.removeSessionManagerListener(statusListener, CastSession.class);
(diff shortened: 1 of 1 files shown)


AquilesCanta:
I think the change is reasonable, but the target branch needs to change to dev-v2, so we can't take this change as is. I can also push this change internally, if you wan't to save the work of creating a new pull request. Aside:

> However, I can't find the right time to start playing locally after session is ended at release().

Aren't you calling release? Can't you start playing locally right when you call release?

TakuSemba:
> Aren't you calling release? Can't you start playing locally right when you call release?

Oh, you're right. 😅 That would work too.

Anyways, I will fix the target branch and create a new PR soon.
If you've already fixed it internally, please close the PR.

PR #7936, opened 2020-09-17
remove statusListener after ended session

TakuSemba:
I recreated PR since I was targeting the wrong branch.

Detail: https://github.com/google/ExoPlayer/pull/7935

### DIFF
--- a/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
+++ b/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
@@ -496,8 +496,8 @@ public final class CastPlayer extends BasePlayer {
   @Override
   public void release() {
     SessionManager sessionManager = castContext.getSessionManager();
-    sessionManager.removeSessionManagerListener(statusListener, CastSession.class);
     sessionManager.endCurrentSession(false);
+    sessionManager.removeSessionManagerListener(statusListener, CastSession.class);
   }
 
   @Override


AquilesCanta:
I apologize for not raising this before but I have a few objections on this API change:

1. release() is the counterpart of the constructor. The constructor doesn't call onCastSessionAvailable when a session is available while constructing the CastPlayer instance. Calling onCastSessionUnavailable() in the release() method introduces asymmetry into the API.
2. Generally, when a user calls release(), they expect that no further callbacks are invoked. API-wise this could be counter-productive. The app must not interact with the instance after calling release().
3. The change doesn't really introduce new functionality. The app is in control of the release call. We are introducing an API change without concrete advantages to this (some could consider this API change to be bad for them).

Feedback is welcome. So provide any reasons why you think this change is important and we can revisit this change.

AquilesCanta:
Let me expand a bit more on the second point above:

Let's imagine that an existing app (like the demo app) transfers the playback state from the cast player to the local player when a cast session ends. If they perform that action in onCastSessionUnavailable, then this change will break them, because they'll query the state of a player on which release has been called.

Hope this clarifies point 2.

TakuSemba:
Hi, thank you for considering my change and raising concerns.
I think your points are pretty reasonable and I felt my change was a bit wrong way to go.

When I was filling this PR, I was calling `release` just to stop a current session.
However, as the point `2` says, since I was using a CastPlayer instance across multiple sessions, I should have not call `release` just to end the session.

I'm going to try calling `sessionManager.endCurrentSession()` in my case and I think it will work fine.
I'm closing this for now, and I will let you know if I have a problem later.

Anyways, thank you. 👍

AquilesCanta:
Thank you for the Pull Requests!

PR #7938, opened 2020-09-17
text: Fix handling of unbreakable-space in ssa subtitles

bennettpeter:
The character sequence \h is used for "unbreakable space".
Replace these sequences with space to avoid strings of \h
showing up on screen.

HandBrake inserts many \h characters in captions transcoded
from mpeg-ts to mkv. I see phrases like this on screen:
"I, um, can't go anywhere\h\h\h\h\h\h\h\hby myself"
This is difficult to read.

References:
https://fileformats.fandom.com/wiki/SubStation_Alpha
https://www.md-subs.com/line-spacing-in-ssa

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -264,7 +264,8 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
     String text =
         SsaStyle.Overrides.stripStyleOverrides(rawText)
             .replaceAll("\\\\N", "\n")
-            .replaceAll("\\\\n", "\n");
+            .replaceAll("\\\\n", "\n")
+            .replaceAll("\\\\h", " ");
     Cue cue = createCue(text, style, styleOverrides, screenWidth, screenHeight);
 
     int startTimeIndex = addCuePlacerholderByTime(startTimeUs, cueTimesUs, cues);


icbaker:
Thanks for the contribution! `\h` isn't mentioned in the 'spec' I've seen before: http://moodub.free.fr/video/ass-specs.doc (linked from https://wiki.multimedia.cx/index.php/SubStation_Alpha). But it seems reasonable to support it since it is mentioned elsewhere (as you've linked, thanks!) and SSA/ASS is very much "consensus" based :)

I have a question though - in the example you've given, do you expect there to be lots of spaces rendered on the screen? i.e. your code transforms that example: 
```
I, um, can't go anywhere        by myself
```

And I'd expect those spaces to show up on the screen.

I think this is expected - since the `\h` is 'non-breaking space' (i.e. `&nbsp;` in HTML), so it's expected to be rendered, but it just seems a bit surprising in the context of your example.

bennettpeter:
Actually there is also a carriage return in the source, so the subtitle contains "I, um, can't go anywhere\n\h\h\h\h\h\h\h\hby myself". What I see on screen is 
```
I, um, can't go anywhere
\h\h\h\h\h\h\h\hby myself
```
WIth the fix, this becomes
```
I, um, can't go anywhere
        by myself
```
It seems that HandBrake is using non-breaking spaces for indenting lines.

icbaker:
Oh right, sorry I missed the `\n` because it blended in with all the `\h` - thanks for clarifying!

PR #7968, opened 2020-09-23
Update MPEG-DASH AudioChannelConfiguration Parsing

ybai001:
Add support for schemeIdUri="urn:mpeg:mpegB:cicp:ChannelConfiguration".
See: https://dashif.org/identifiers/audio_source_metadata/ and clause 8.2 in ISO/IEC 23001-8.


ybai001:
Current ExoPlayer doesn't handle this schemeIdUri so that this kind of DASH streams channel configuration can't be parsed.
After aligned between DASH IF and Dolby, Dolby AC-4 DASH stream uses this schemeIdUri value. 

See G.2.5 The AudioChannelConfiguration DASH descriptor in ETSI TS 103 190-2 V1.2.1 (2018-02).

If you need test stream, please write email to ybai@dolby.com. Thanks.

Issue #5097, opened 2018-11-12
Improve MPEG-TS seeking support


ebr11:
### Issue description
Not sure if this is a bug or incompatibility with a particular stream or other issue but seeking in a .TS file does not work in the new version.

### Reproduction steps
Play the following test content in the Demo app and the seek buttons will not be enabled.  When played in our app using Exo 2.9.1 any attempt to seek appears to try to do it but ends up just going back to the start of the video.  This is what has always happened when attempting to seek .TS files.

The Release Notes don't appear to state that it is required but we did try setting DefaultExtractorsFactory.setConstantBitrateSeekingEnabled(true) with no effect.

### Link to test content
https://www.dropbox.com/s/3270h1gbnebgchy/WBTV%203%20News%20at%204pm%202018_11_12_16_00_00.ts?dl=0

### Version of ExoPlayer being used
2.9.1

### Device(s) and version(s) of Android being used
Tested on Xiaomi Mi 3 running Android 8.0.

### A full bug report captured from the device
Unable to capture at this time.

ojw28:
Increasing `TsDurationReader.DURATION_READ_PACKETS` fixes this. I tried 300 and that was enough. It's unclear how that value was chosen, and whether we should just increase it or do something more complicated. If we increase it, we probably need to increase `TsBinarySearchSeeker.TIMESTAMP_SEARCH_PACKETS` as well.

ojw28:
@botaydotcom - Was that value chosen to be an arbitrary value we thought was big enough? Or was there more logic to it than that :)?

ebr11:
Confirmed!

Thanks very much!

majidebrahimii:
@ojw28 @ebr11 How use TsDurationReader.DURATION_READ_PACKETS and TsBinarySearchSeeker.TIMESTAMP_SEARCH_PACKETS for fix bug? Can you give me an example code?

ebr11:
Hi.  As indicated above, just set those values to 300 instead of their current 200.

majidebrahimii:
@ebr11 Hi. I use version 2.9.1 but unknown this line code:
` TsDurationReader.DURATION_READ_PACKETS = 300 
`
 How did you do it?

ojw28:
@sarabellaa You need to have the ExoPlayer code checked out as source. There is no API that you can use to modify the values.

ojw28:
Things we should look at to fix this properly:
[...]

PR #7988, opened 2020-09-25
TS Extractor: Fix problem with locating first PCR in HD-PVR recordings

bennettpeter:
TV shows recorded with Hauppauge HD-PVR will play, but do not allow
seeking and do not show the length. This is because the TsDurationReader
cannot locate the first PCR in the first 112800 bytes. Increasing that
buffer to 122200 bytes solves the problem. To be on the safe side I am
setting it to double the original, or 225600 bytes. The search for last
PCR does not have this problem so I am leaving that at 122200 bytes.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsDurationReader.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsDurationReader.java
@@ -38,7 +38,8 @@ import java.io.IOException;
  */
 /* package */ final class TsDurationReader {
 
-  private static final int TIMESTAMP_SEARCH_BYTES = 600 * TsExtractor.TS_PACKET_SIZE;
+  private static final int TIMESTAMP_SEARCH_BYTES_START = 1200 * TsExtractor.TS_PACKET_SIZE;
+  private static final int TIMESTAMP_SEARCH_BYTES_END = 600 * TsExtractor.TS_PACKET_SIZE;
 
   private final TimestampAdjuster pcrTimestampAdjuster;
   private final ParsableByteArray packetBuffer;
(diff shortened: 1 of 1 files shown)


ojw28:
Note: This is related to https://github.com/google/ExoPlayer/issues/5097.

troyhough:
https://drive.google.com/file/d/1G1dOY5eu67kRu7h9IgNJglgqjsVgBB3M/view?usp=sharing

Here's a sample file that this change does not help with. Can the scope of this fix be expanded to help with a broader range of mpegts files?

Thanks in Advance!

kim-vde:
The current method to compute the duration is already requiring a non-trivial amount of memory so we should avoid increasing it too much. We should come up with a method that requires less memory in the future.

@bennettpeter If increasing the buffer to 122200 bytes solves the issue, then I suggest we do that, to limit the memory consumption. I'll change the PR code myself.

@troyhough With the current way of computing the duration, supporting the stream you provided would require increasing the buffer from 112 800 bytes to more than 488 800 bytes. I will therefore not make the change as it would significantly increase the memory consumption for all TS files. Hopefully your stream will be better supported once we have optimized the duration computation.

AquilesCanta:
@kim-vde, should we consider letting the app choose a value for this through a TsExtractor parameter? Seems like that'd be a bit more scalable than updating on a per-demand basis.

I guess that if we were building a heuristic for this, that'd be based on the Transport Stream's bitrate, but that would require a considerably larger effort.

kim-vde:
@AquilesCanta I was thinking of fetching the data by chunks instead of having one large buffer. But the parameter is a good idea. I'll do that instead.

bennettpeter:
@kim-vde A parameter would be good. I thought of that but it seems it would require changes in several classes to pass the parameter through, since the client code does not have access to the TSDurationReader class.

troyhough:
@kim-vde Just so you know the test file I sent you was just a standard 30 minute HDHomerun recording from OTA, untouched. Probably the most common tuner on the market. Having said that I think it's pretty important to get seeking to work properly on something so common. Thanks for the consideration.

bennettpeter:
@kim-vde Can I update the pull request to allow the specification of the buffer size to be set in DefaultExtractorsFactory and TsExtractor? I will do that if it is likely to be accepted. Are there any special standards to be followed in this?

kim-vde:
@bennettpeter I am actually busy doing that myself. It should be available in a few days.

bennettpeter:
I see that the sample from @troyhough requires SEARCH_BYTES_START at 1200 packets and SEARCH_BYTES_END at 2600 packets, where my original pullrequest was using 1200 and 600. So there will need to be the ability to modify both of those, or use one modifiable value for both.

kim-vde:
Yes, I was going to use one variable for both.

PR #8025, opened 2020-10-01
Adjust initial selection of HlsChunckSource.InitializationTrackSelection

matamegger:
With the changes done in 0670782de80560a42874dbe6f94b070d94d8390d, there exists the possibility that `indexOf(group.getFormat(0))` returns `-1` (i.e. the track is not part of the selection).
This will earlier or later result in an `ArrayIndexOutOfBoundsException`.*

After having a look at the `BaseTrackSelection` I assumed that the `selectedIndex` may not be `-1` or in general any invalid index for the selection.
To always have a valid selection we just select the first one. (Which is basically equivalent to the previous behaviour if the above mentioned change was not implemented.)

---
_* I observed the exception in the `updateSelectedTrack -> isBlacklisted` call_

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -592,7 +592,7 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 
     public InitializationTrackSelection(TrackGroup group, int[] tracks) {
       super(group, tracks);
-      selectedIndex = indexOf(group.getFormat(0));
+      selectedIndex = 0;
     }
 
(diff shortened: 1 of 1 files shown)


AquilesCanta:
I think the described behavior may happen when the master playlist's first stream-inf tag is an i-frame-only tag.

The fix looks good, but it introduces a (I assume involuntary) functional change: It will make preparation start with the lowest bitrate variant, which changes the historical ExoPlayer bevahior of choosing the first stream-inf in the master playlist. I think it could make sense to choose the lowest bitrate one, as a way to speed up preparation. But there are better ways of achieving this, so it's a change that should be analysed independently of this bugfix.

Thanks for reporting this. I will send a fix for this internally, since it's faster than us having a back and forth to get this merged.

Issue #7873, opened 2020-09-09
Support PMT updates with additional tracks in TS extractor


lsgfly:
#7455 # [REQUIRED] Issue description
I use the Exoplayer to play "https://d4licq9vet44p.cloudfront.net/hls-vod/dvr/ch1.20200909_0200.m3u8".
It is no sound.
I use media player to play the link. It is ok.

### [REQUIRED] Reproduction steps
using the exoplayer to play  play "https://d4licq9vet44p.cloudfront.net/hls-vod/dvr/ch1.20200909_0200.m3u8".

### [REQUIRED] Link to test content
It is the link:  https://d4licq9vet44p.cloudfront.net/hls-vod/dvr/ch1.20200909_0200.m3u8

### [REQUIRED] Version of ExoPlayer being used
2.8.2 .  I also test in the lastet version, It can be reproduced.

### [REQUIRED] Device(s) and version(s) of Android being used
Vivo X50 ,Android 10.
I also test other device.It also can be reproduced.

kim-vde:
The audio is not played because the first PMT table does not contain any audio track. I don't know if it is allowed though.

I'll assign to @AquilesCanta who probably knows the answer.

AquilesCanta:
As @kimvde points out, the media is malformed. I will check whether MediaPlayer successfully plays the audio and I'll add a workaround for this if MediaPlayer effectively plays the media.

AquilesCanta:
I can confirm the media plays with audio on MediaPlayer. I'll add a workaround to fix this in ExoPlayer.

lsgfly:
When will this be fixed?

AquilesCanta:
Thanks for your patience. I cannot provide any timelines, as there are other high priority tasks that need attention. I'm sorry for the delay.

PR #8063, opened 2020-10-11
Progressive Media: Handle empty audio track

bennettpeter:
If there is a second audio track with no data (bitrate 0), playback never starts and throws an exception
ExoPlayerImplInternal:   Caused by: java.lang.IllegalStateException: Playback stuck buffering and not loading.
This seems to be typical of recordings from UK DVB-T2 stations. It has also been seen on channel 21.2
Rochester, NY.
This fix:
- Checks that there is at least one valid video track and one valid audio track
- If so, sets the mime type of the uninitialized track to a dummy value audio/null
  so that playback can proceed.

Once playing, if you select the second audio track it continues playing with silence,
which is a reasonable way to handle an empty audio stream.

Sample video file
https://www.filezzz.com/d/5138183d7312459eac2d0025a7545c6b/Exoplayer_null_audio_failure.ts/preview

Analysis of tracks in the sample:
https://www.filezzz.com/d/1dfa0be10d5a4b4b965e2906555817d0/Exoplayer_null_audio_analysis.txt/preview

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/ProgressiveMediaPeriod.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/ProgressiveMediaPeriod.java
@@ -744,11 +744,37 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
     if (released || prepared || !sampleQueuesBuilt || seekMap == null) {
       return;
     }
-    for (SampleQueue sampleQueue : sampleQueues) {
-      if (sampleQueue.getUpstreamFormat() == null) {
-        return;
+
+    // If all streams have an upstream format we can finish preparing.
+    // If there is a video and an audio stream with upstream format
+    // and there is one single stream without upstream format, it is likely
+    // an audio stream with no data. Set it to a dummy mime type of
+    // audio/null so that playback can continue.
+
+    int nullStream = -1;
+    int nullStreamCount = 0;
+    boolean videoFound = false;
+    boolean audioFound = false;
+    for (int i = 0; i < sampleQueues.length; i++) {
+      Format upstreamFormat = sampleQueues[i].getUpstreamFormat();
+      if (upstreamFormat == null) {
+        nullStream = i;
+        ++nullStreamCount;
       }
+      else if (MimeTypes.isVideo(upstreamFormat.sampleMimeType))
+        videoFound = true;
+      else if (MimeTypes.isAudio(upstreamFormat.sampleMimeType))
+        audioFound = true;
     }
+    if (nullStreamCount == 1 && videoFound && audioFound) {
+      Format.Builder builder = new Format.Builder();
+      builder.setSampleMimeType("audio/null");
+      sampleQueues[nullStream].format(builder.build());
+      nullStreamCount = 0;
+    }
+    if (nullStreamCount > 0)
+      return;
+
     loadCondition.close();
     int trackCount = sampleQueues.length;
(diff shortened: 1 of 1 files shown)


andrewlewis:
I think `ProgressiveMediaPeriod` is probably not the right place to work around this issue, as it sounds specific to MPEG-TS. Perhaps `TsExtractor` should avoid exposing the audio track at all if its bitrate is zero? (I'm not familiar with MPEG-TS in detail so I'm not sure if that's signaled explicitly.) If you already have an issue open for this please reference it here.

andrewlewis:
@AquilesCanta @kim-vde Please could you take a look and advise on what would be the best way to handle this?

AquilesCanta:
To add to what @andrewlewis said, I don't see how the proposed change addresses the issue. The audio/null workaround should only be applied when audioFound is true, which in turn will only happen when an audio format is found upstream.

Moreover, the provided file analysis doesn't show an empty audio track:

```
|=============================================================================|
|  PID: 0x00CA (202)                                      HE-AAC Audio (eng)  |
|  PES stream id: 0xC0 (Audio 0)                                              |
|  Service: 0x0001 (1) (unknown)                                              |
|-----------------------------------------------------------------------------|
|  Single Service PID        Transport:                Discontinuities:       |
|  Bitrate: ... 356,872 b/s  Packets: ........ 18,617  Expected: ......... 0  |
|  Access: .......... Clear  Adapt.F.: ........ 7,345  Unexpect: ......... 0  |
|                            Duplicated: .......... 0  PES:                   |
|                            PCR: ................. 0  Packets: ...... 3,677  |
|                                                      Inv.Start: ........ 0  |
|=============================================================================|
```

So I would expect the file to play normally on ExoPlayer.

Confirm the following for me please: Is the issue that some of your streams have a PMT that declares a track which is not present in the file? If so, this is being tracked by #7873. The root cause of #7873 is not the same, but the fix is. I'm currently working on a fix for this, but haven't managed to get it done due to other priorities.

bennettpeter:
@AquilesCanta there are two audio tracks in the file, but by default it tries to play the one that does not work.
The empty audio track I inferred from PID 0x00CC, see below extract from the analysis.
"a PMT that declares a track which is not present in the file" sounds like it could be the same problem.
The fix I propose is a kludge and I agree that there is likely a better way to do it.
My fix is working, but I will wait for the fix for #7873 and try it on my setup.
Thanks.

```
|     PID  Usage                                     Access          Bitrate  |
|   Total  Undefined .................................... C    4,078,242 b/s  |
|  0x00C9  AVC video (1920x1080, high profile, level 4.0  C    3,713,855 b/s  |
|  0x00CA  HE-AAC Audio (eng) ........................... C      356,872 b/s  |
|  0x00CB  Subtitles (eng, DVB subtitles, high definitio  C            0 b/s  |
|  0x00CC  HE-AAC Audio (eng, visual impaired commentary  C            0 b/s  |
|  0x03EC  PMT .......................................... C        7,514 b/s  |
|          (C=Clear, S=Scrambled, +=Shared)                                   |
```

AquilesCanta:
Ok, I misunderstood the issue. However, this change will break some TS files where the audio track is actually there, I'm afraid. For example:
```
track(1) <-- video
track(2) <-- audio 1
track(3) <-- audio 2
track(1).format(video)
track(2).format(audio 1) <-- With this format call preparation will finish, by assigning audio/null to the last track. Even though its format is coming after this call.
track(3).format(audio 2)
```

Issue #8073, opened 2020-10-13
Mov file with pcm_u8 audio track couldn't play correctly


xufuji456:
Hello, guys of developer, nice to contact you again. I Found that Mov file with pcm_u8 audio track couldn't play correctly. It uses Mp4Extractor, when running on the method of readSample, the sampleSize of audioTrack is always 1. As a result, audio frame couldn't be decoded correctly. Looking forward to your reply.

Here is the testing video: [https://drive.google.com/file/d/1QMX--yv2jloMbANXWQbAepF1n71Pm7RJ/view](url)

marcbaechinger:
Mov is the Apple Quicktime format. This is similar to MP4 but not exactly. Some files may work with ExoPlayer while some do not. So in your case it seems that the MP4 extractor does not like it, I'm afraid:

```
Caused by: com.google.android.exoplayer2.source.UnrecognizedInputFormatException: None of the available extractors (FlvExtractor, FlacExtractor, WavExtractor, FragmentedMp4Extractor, Mp4Extractor, AmrExtractor, PsExtractor, OggExtractor, TsExtractor, MatroskaExtractor, AdtsExtractor, Ac3Extractor, Ac4Extractor, Mp3Extractor) could read the stream.
```

Why are you thinking it has something to do with the pcm_u8 audio track? Have you been able to play a mov file with another audio track?

> , the sampleSize of audioTrack is always 1

Can you give me a bit more details on how you see the error message you are reporting above? For me the file can not be extracted at all, so I don't see tracks at all like you say.

marcbaechinger:
I assign to @kim-vde so you have someone looking into it who actually knows all the extractor details.

xufuji456:
> Mov is the Apple Quicktime format. This is similar to MP4 but not exactly. Some files may work with ExoPlayer while some do not. So in your case it seems that the MP4 extractor does not like it, I'm afraid:
> 
> ```
> Caused by: com.google.android.exoplayer2.source.UnrecognizedInputFormatException: None of the available extractors (FlvExtractor, FlacExtractor, WavExtractor, FragmentedMp4Extractor, Mp4Extractor, AmrExtractor, PsExtractor, OggExtractor, TsExtractor, MatroskaExtractor, AdtsExtractor, Ac3Extractor, Ac4Extractor, Mp3Extractor) could read the stream.
> ```
> 
> Why are you thinking it has something to do with the pcm_u8 audio track? Have you been able to play a mov file with another audio track?
[...]

PR #8083, opened 2020-10-17
Fix MOV with pcm_alaw which is fixedSampleSize

xufuji456:
Hello, I found that MOV with pcm_alaw or pcm_mulaw couldn't play, which use Mp4Extractor. The pcm_alaw is also one of raw audio, though we don't know how much bits it uses. However, pcm_alaw and pcm_mulaw also use fixedSampleSize, we can see the sample_size which in the box of "moov/trak/mdia/minf/stbl/stsz". As a result, we should treat pcm_alaw and pcm_mulaw as fixedSampleSize, and recheck it to calculate the real frame size.

Here is the testing video:
pcm_alaw: [https://drive.google.com/file/d/1J5LcRO49HHWqr-hfoSB8Y9M4PVlmgDDy/view](url)
pcm_mulaw: [https://drive.google.com/file/d/1ulKeHKvqK6bppmFct1ZLd2SnaOlrZtZs/view](url)

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/AtomParsers.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/AtomParsers.java
@@ -232,7 +232,9 @@ import org.checkerframework.checker.nullness.compatqual.NullableType;
     // Fixed sample size raw audio may need to be rechunked.
     boolean isFixedSampleSizeRawAudio =
         sampleSizeBox.isFixedSampleSize()
-            && MimeTypes.AUDIO_RAW.equals(track.format.sampleMimeType)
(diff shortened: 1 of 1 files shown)


andrewlewis:
Looks like this is related to #8073.

xufuji456:
> Looks like this is related to #8073.

Yeah, This uses to fix mov with pcm_alaw/pcm_mulaw/pcm_u8 which is fixedSampleSize

ojw28:
@xufuji456 - The change looks good, but please could you resolve the CLA issue mentioned above? Thanks!

xufuji456:
> @xufuji456 - The change looks good, but please could you resolve the CLA issue mentioned above? Thanks!

OK, I will resolve the CLA, and try to pull request again. Thank you for your appreciation.

PR #8088, opened 2020-10-19
Fix MOV with pcm_alaw/pcm_mulaw which is fixedSampleSize

xufuji456:
Hello, I found that MOV with pcm_alaw or pcm_mulaw couldn't play, which use Mp4Extractor. The pcm_alaw is also one of raw audio, though we don't know how much bits it uses. However, pcm_alaw and pcm_mulaw also use fixedSampleSize, we can see the sample_size which in the box of "moov/trak/mdia/minf/stbl/stsz". As a result, we should treat pcm_alaw and pcm_mulaw as fixedSampleSize, and rechunk it to calculate the real frame size.

This is from #8083 , try to resolve CLA, and pull request again.

Here is the testing video:
pcm_alaw: https://drive.google.com/file/d/1J5LcRO49HHWqr-hfoSB8Y9M4PVlmgDDy/view
pcm_mulaw: https://drive.google.com/file/d/1ulKeHKvqK6bppmFct1ZLd2SnaOlrZtZs/view


ojw28:
Related to #8073.

PR #8133, opened 2020-10-29
fix the end point of span exceeds the length of SpannableStringBuilder

xufuji456:
Hello, guys of developer. I found that the end point of span exceeds the length of SpannableStringBuilder in Tx3gDecoder sometime. In this case, it will throw exception when setting span. Therefore, we may add a judgment whether the end point will exceed the length.

Here is the test url: https://drive.google.com/file/d/1ETVLaO5Y3hiZ8UlP1-EZn-lt3sU5VYKD/view?usp=sharing

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/tx3g/Tx3gDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/tx3g/Tx3gDecoder.java
@@ -185,6 +185,9 @@ public final class Tx3gDecoder extends SimpleSubtitleDecoder {
     int fontFace = parsableByteArray.readUnsignedByte();
     parsableByteArray.skipBytes(1); // font size
     int colorRgba = parsableByteArray.readInt();
+    if (end > cueText.length()) {
+      end = cueText.length();
+    }
     attachFontFace(cueText, fontFace, defaultFontFace, start, end, SPAN_PRIORITY_HIGH);
     attachColor(cueText, colorRgba, defaultColorRgba, start, end, SPAN_PRIORITY_HIGH);
   }


ojw28:
@icbaker - Please take a look!

icbaker:
Thanks for the PR - and thanks for providing the test media to demonstrate the problem.

Having looked the media, I think we're tripping up because it contains multi-byte UTF-8 code-points and it seems to use byte offsets (rather than code-point offsets) for the `end` field in the `styl` box.

Concretely:
```
# Extract the tx3g track:
$ MP4Box ~/Downloads/Birds.mp4 -raw 3
# Inspect it:
$ hexdump -C ~/Downloads/Birds_track3.tx3g
```

And then note the problematic part (with reference to the [tx3g spec](https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-SW81)):
```
// text length (bytes)
00 00 00 37 
// subtitle text:
C3 A2 E2 84 A2 C2 AA 20 54 68 65 20 62 69 72 64 73 20 69 6E 20 74 68 65 20 73 6B 79
20 67 6F 20 74 68 72 6F 75 67 68 20 74 68 65 20 61 69 72 20 C3 A2 E2 84 A2 C2 AA 

// Followed by this styl atom:
00 00 00 16   // styl atom length
73 74 79 6C   // "styl"
00 01  // entry count
00 00  // start character
00 37  // end character
00 01  // font identifier
02  // font face
12  // font size
FF FF FF FF  // foreground color
```

That text decodes (as UTF-8) to "â™ª The birds in the sky go through the air â™ª"

(aside: this seems slightly unexpected - is that the subtitle string you'd expect to see?)

When stored as a UTF-16 string in Java that string only has `length=47`, while the end offset in the `styl` box is `0x37 == 55`.

And so when we try to build a span that ends at index 55, we get the exception you're seeing.

The spec I linked above is a little bit muddy here to be honest - it's clear that the length of the string is in bytes. But when talking about the styl offsets it says it's measured in "characters" - which isn't a precise term in unicode:
>Start character
>  A 16-bit value that is the offset of the first character that is to use the style specified in this record. Zero (0) is the first character in the subtitle.
>
>End character
>  A 16-bit value that is the offset of the character that follows the last character to use this style.

It's clear that the string should be interpreted as UTF-8:
>The subtitle text is Unicode text, encoded either as UTF-8 text or UTF-16 text beginning with a UTF-16 BYTE ORDER MARK ('\uFEFF') in big or little endian order.

Nevertheless, it seems harmless & safe to limit the start & end indices to be within the bounds of the string in Java. The worst that happens if something goes wrong is that the styling looks a bit strange, rather than no subtitles being shown at all (or playback failing on older versions). So this PR makes sense - I'll work on getting it merged.

xufuji456:
@icbaker  It's my pleasure to commit the PR. You are right, the real reason is  character encoding. 
This boundary detection is used to avoid exception. ( â™ª The birds in the sky go through the air â™ª) This subtitle looks complete, and I see that MX Player displays like that.

PR #8154, opened 2020-11-03
Correctly handling Exception: java.nio.file.FileSystemException:... N…

samoylenkodmitry:
Correctly handling Exception:
java.nio.file.FileSystemException:  No space left on device.

By default methods File.makeDir() and File.makeDirs() can return 'false' if file aleady exists or can not be created. Such silent ignore of the situation propagates misbehavior to the caller: CacheDataSink#173 : new FileOutputStream(file). And then it throws not correct exception type 'FileNotFoundException'. While correct exception should be 'no space left on the device'.

This can be fixed only with 'Files.createDirectories()' method that throws correct exception type.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
@@ -15,6 +15,7 @@
  */
 package com.google.android.exoplayer2.upstream.cache;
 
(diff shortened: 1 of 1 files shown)


ojw28:
We merged a slightly different version of this, which gives a less specific failure message (i.e., that the directory could not be created), but works across all API levels. It does fix the problem of the failure being deferred and then the caller throwing a more obscure exception. Thanks for the contribution!

Issue #7788, opened 2020-08-20
Dolby(EC3) not playing on androidTV when connected to bluetooth speaker


PratimMallick:
###  Issue description
Trying to play ec3 content on Android TV os 9.0 via exoplayer demo app while connected to **Bluetooth speaker/headphone.** Getting **Playback failed** error

From the logs:
`AudioTrack-JNI: Error -38 initializing AudioTrack`
`android.media.AudioTrack: Error code -20 when initializing AudioTrack` 

###  Reproduction steps
100% reproducible
Add the following to `media.exolist.json` to exoplayer demo app
Connect the device(I have tested it only dolby enabled TV) to bluetooth speaker or headphone and play this content

###  Link to test content
 {
        "name": "DASH DD+ 5.1 clip",
        "uri": "https://nsesh-s3.s3.ap-south-1.amazonaws.com/VOD/5ch_DASH/ChID_voices_51_256_ddp.mpd"
      }

### A full bug report captured from the device
[bugreport](https://drive.google.com/file/d/1KTRbFNsTPZ_apypF4n_UXjeWP55qDkMM/view?usp=sharing)

### Version of ExoPlayer being used
2.11.0

###  Device(s) and version(s) of Android being used
TCL android TV, OS - 9.0

andrewlewis:
From the bug report:

```
APM_AudioPolicyManager: getOutputForDevice() could not find output for stream 3, sampling rate 48000, format 0xa000000, channels 0x3f, flags 0x1
```

... means that the `AudioTrack` is failing to initialize for encoded audio output (`ENCODING_E_AC3`).

I suspect that the device is still reporting to the app that DD+ is supported via passthrough even though the actual output path is now via headphones. Can you detect the headphone connection event and restart the player with a new `AudioCapabilities` instance that doesn't include encoded formats?

PratimMallick:
@andrewlewis Yes the audio output settings are set to `passthrough`. So there is no other way apart from detecting a bluetooth headset connection?

Also is it possible to decode ec3 in the device(if it supports) and pass the decoded stream to the speakers via BT?

PratimMallick:
@andrewlewis After detecting headphone connection and sending new capabilities that do not include Encoded format plays fine

andrewlewis:
> So there is no other way apart from detecting a bluetooth headset connection?

I'm not aware of another way to do this, I'm afraid.

[...]

PR #8178, opened 2020-11-06
Fix issues with pass-thru surround audio and external audio devices

sneelavara:
Broadcom based AndroidTV devices (and likely others) present the raw (pass-thru) audio CODEC when an external PCM
only (Bluetooth or wired headphones) device is connected. his occurs as well if the connection is
dynamic (Bluetooth device powers up and pairs with the AndroidTV during pass-thru playback).

Basically, the AndroidTV device reports the pass-thru CODEC via
([MediaCodecInfo](https://developer.android.com/reference/android/media/MediaCodecInfo)) if an surround
capable HDMI device is connected regardless of the current audio route.

This is the cause of issue [#7788](https://github.com/google/ExoPlayer/issues/7788)

The solution in this commit watches the
[MediaRouter](https://developer.android.com/guide/topics/media/mediarouter#router-callback) and modifies the returned AudioCapabilities based on changes.

### DIFF
--- a/library/core/build.gradle
+++ b/library/core/build.gradle
@@ -37,6 +37,8 @@ dependencies {
     api project(modulePrefix + 'library-common')
     api project(modulePrefix + 'library-extractor')
     implementation 'androidx.annotation:annotation:' + androidxAnnotationVersion
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.mediarouter:mediarouter:1.1.0'
     compileOnly 'com.google.code.findbugs:jsr305:' + jsr305Version
     compileOnly 'org.checkerframework:checker-qual:' + checkerframeworkVersion
     compileOnly 'org.checkerframework:checker-compat-qual:' + checkerframeworkCompatVersion
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilities.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilities.java
@@ -48,6 +48,9 @@ public final class AudioCapabilities {
   /** Global settings key for devices that can specify external surround sound. */
   private static final String EXTERNAL_SURROUND_SOUND_KEY = "external_surround_sound_enabled";
 
+  private static boolean externalPcmOnlySound = false;
+  public static final int[] EXTERNAL_PCM_ONLY = {AudioFormat.ENCODING_PCM_16BIT,
+          AudioFormat.ENCODING_PCM_8BIT, AudioFormat.ENCODING_PCM_FLOAT, AudioFormat.ENCODING_MP3};
   /**
    * Returns the current audio capabilities for the device.
    *
@@ -113,6 +116,9 @@ public final class AudioCapabilities {
     this.maxChannelCount = maxChannelCount;
   }
 
+  public static void setExternalPcmOnlySoundset(boolean flag) {
+    externalPcmOnlySound = flag;
+  }
   /**
    * Returns whether this device supports playback of the specified audio {@code encoding}.
    *
@@ -120,6 +126,9 @@ public final class AudioCapabilities {
    * @return Whether this device supports playback the specified audio {@code encoding}.
    */
   public boolean supportsEncoding(@C.Encoding int encoding) {
+    if (externalPcmOnlySound) {
+      return Arrays.binarySearch(EXTERNAL_PCM_ONLY, encoding) >= 0;
+    }
     return Arrays.binarySearch(supportedEncodings, encoding) >= 0;
   }
 
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
@@ -24,7 +24,15 @@ import android.database.ContentObserver;
 import android.media.AudioManager;
 import android.net.Uri;
 import android.os.Handler;
+import android.util.Log;
+
 import androidx.annotation.Nullable;
+import androidx.mediarouter.media.MediaControlIntent;
+import androidx.mediarouter.media.MediaRouteSelector;
+import androidx.mediarouter.media.MediaRouter;
+import androidx.mediarouter.media.MediaRouter.Callback;
+import androidx.mediarouter.media.MediaRouter.ProviderInfo;
+import androidx.mediarouter.media.MediaRouter.RouteInfo;
 import com.google.android.exoplayer2.util.Assertions;
 import com.google.android.exoplayer2.util.Util;
 
@@ -57,6 +65,10 @@ public final class AudioCapabilitiesReceiver {
   @Nullable /* package */ AudioCapabilities audioCapabilities;
   private boolean registered;
 
+  Intent stickyIntent = null;
+  private MediaRouter mediaRouter;
+  private MediaRouteSelector mSelector;
+
   /**
    * @param context A context for registering the receiver.
    * @param listener The listener to notify when audio capabilities change.
@@ -73,6 +85,11 @@ public final class AudioCapabilitiesReceiver {
             ? new ExternalSurroundSoundSettingObserver(
                 handler, context.getContentResolver(), externalSurroundSoundUri)
             : null;
+
+    mediaRouter = MediaRouter.getInstance(context);
+    mSelector = new MediaRouteSelector.Builder()
+            .addControlCategory(MediaControlIntent.CATEGORY_LIVE_AUDIO)
+            .build();
   }
 
   /**
@@ -91,7 +108,6 @@ public final class AudioCapabilitiesReceiver {
     if (externalSurroundSoundSettingObserver != null) {
       externalSurroundSoundSettingObserver.register();
     }
-    Intent stickyIntent = null;
     if (receiver != null) {
       IntentFilter intentFilter = new IntentFilter(AudioManager.ACTION_HDMI_AUDIO_PLUG);
       stickyIntent =
@@ -99,6 +115,10 @@ public final class AudioCapabilitiesReceiver {
               receiver, intentFilter, /* broadcastPermission= */ null, handler);
     }
     audioCapabilities = AudioCapabilities.getCapabilities(context, stickyIntent);
+
+    mediaRouter.addCallback(mSelector, mediaRouterCallback,
+            MediaRouter.CALLBACK_FLAG_REQUEST_DISCOVERY);
+
     return audioCapabilities;
   }
 
@@ -117,6 +137,9 @@ public final class AudioCapabilitiesReceiver {
     if (externalSurroundSoundSettingObserver != null) {
       externalSurroundSoundSettingObserver.unregister();
     }
+
+    mediaRouter.removeCallback(mediaRouterCallback);
+
     registered = false;
   }
 
@@ -163,4 +186,27 @@ public final class AudioCapabilitiesReceiver {
     }
   }
 
+  // Define the Callback object and its method to detect the Bluetooth speaker connenction
+  private final MediaRouter.Callback mediaRouterCallback =
+          new MediaRouter.Callback() {
+            @Override
+            public void onRouteChanged(MediaRouter router, RouteInfo route) {
+            }
+
+            @Override
+            public void onRouteSelected(MediaRouter router, RouteInfo route) {
+              if (route.getDeviceType() == 3) {
+                AudioCapabilities.setExternalPcmOnlySoundset(true);
+                onNewAudioCapabilities(AudioCapabilities.getCapabilities(context, null));
+              }
+            }
+
+            @Override
+            public void onRouteUnselected(MediaRouter router, RouteInfo route, int reason) {
+              if (route.getDeviceType() == 3) {
+                AudioCapabilities.setExternalPcmOnlySoundset(false);
+                onNewAudioCapabilities(AudioCapabilities.getCapabilities(context, stickyIntent));
+              }
+            }
+          };
 }
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
@@ -195,7 +195,7 @@ public final class AudioCapabilitiesReceiver {
 
             @Override
             public void onRouteSelected(MediaRouter router, RouteInfo route) {
-              if (route.getDeviceType() == 3) {
+              if (route.getDeviceType() == RouteInfo.DEVICE_TYPE_BLUETOOTH) {
                 AudioCapabilities.setExternalPcmOnlySoundset(true);
                 onNewAudioCapabilities(AudioCapabilities.getCapabilities(context, null));
               }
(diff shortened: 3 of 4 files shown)


icbaker:
@krocard please take a look

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
```diff
@@ -163,4 +186,27 @@ public void onChange(boolean selfChange) {
...
+  // Define the Callback object and its method to detect the Bluetooth speaker connenction
+  private final MediaRouter.Callback mediaRouterCallback =
+          new MediaRouter.Callback() {
+            @Override
+            public void onRouteChanged(MediaRouter router, RouteInfo route) {
+            }
+
+            @Override
+            public void onRouteSelected(MediaRouter router, RouteInfo route) {
+              if (route.getDeviceType() == 3) {
```
krocard: Please avoid the magic, use `== DEVICE_TYPE_BLUETOOTH`

library/core/src/main/java/com/google/android/exoplayer2/trackselection/TrackSelector.java
```diff
@@ -143,7 +143,7 @@ public abstract TrackSelectorResult selectTracks(
    * Calls {@link InvalidationListener#onTrackSelectionsInvalidated()} to invalidate all previously
    * generated track selections.
    */
-  protected final void invalidate() {
+  public final void invalidate() {
```
krocard, line 146: Why make public?

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
```diff
@@ -163,4 +186,27 @@ public void onChange(boolean selfChange) {
...
+                AudioCapabilities.setExternalPcmOnlySoundset(true);
+                onNewAudioCapabilities(AudioCapabilities.getCapabilities(context, null));
+              }
+            }
+
+            @Override
+            public void onRouteUnselected(MediaRouter router, RouteInfo route, int reason) {
+              if (route.getDeviceType() == 3) {
```
krocard: Please avoid the magic, use `== DEVICE_TYPE_BLUETOOTH`

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
```diff
@@ -91,14 +108,17 @@ public AudioCapabilities register() {
...
     if (receiver != null) {
       IntentFilter intentFilter = new IntentFilter(AudioManager.ACTION_HDMI_AUDIO_PLUG);
       stickyIntent =
           context.registerReceiver(
               receiver, intentFilter, /* broadcastPermission= */ null, handler);
     }
     audioCapabilities = AudioCapabilities.getCapabilities(context, stickyIntent);
+
+    mediaRouter.addCallback(mSelector, mediaRouterCallback,
+            MediaRouter.CALLBACK_FLAG_REQUEST_DISCOVERY);
```
krocard, line 120: Would using `CALLBACK_FLAG_REQUEST_DISCOVERY` not miss some state change, leading to erroneous capabilities?

> The CALLBACK_FLAG_REQUEST_DISCOVERY flag should be used while the application is in the foreground to request that passive discovery be performed if there are sufficient resources to allow continuous passive discovery. On low-RAM devices this flag will be ignored to conserve resources.
https://developer.android.com/reference/androidx/mediarouter/media/MediaRouter#addCallback(androidx.mediarouter.media.MediaRouteSelector,%20androidx.mediarouter.media.MediaRouter.Callback,%20int)

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
krocard: I'm concern about hardcoding the capabilities of `DEVICE_TYPE_BLUETOOTH`. Why would bluetooth only support PCM and all the other types the capabilities from the sticky intent.
I'm wondering if https://developer.android.com/reference/android/media/AudioManager#getDevices(int) should not be used to retrieve the capabilities of the live route type.
Though I'm not sure if it is possible to find the live device from the  live `RouteInfo`.

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilities.java
```diff
@@ -48,6 +48,9 @@
   /** Global settings key for devices that can specify external surround sound. */
   private static final String EXTERNAL_SURROUND_SOUND_KEY = "external_surround_sound_enabled";
 
+  private static boolean externalPcmOnlySound = false;
+  public static final int[] EXTERNAL_PCM_ONLY = {AudioFormat.ENCODING_PCM_16BIT,
+          AudioFormat.ENCODING_PCM_8BIT, AudioFormat.ENCODING_PCM_FLOAT, AudioFormat.ENCODING_MP3};
```
krocard, line 53: I think `DEFAULT_AUDIO_CAPABILITIES` should be used instead.

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
```diff
@@ -163,4 +186,27 @@ public void onChange(boolean selfChange) {
...
 
+  // Define the Callback object and its method to detect the Bluetooth speaker connenction
+  private final MediaRouter.Callback mediaRouterCallback =
+          new MediaRouter.Callback() {
+            @Override
+            public void onRouteChanged(MediaRouter router, RouteInfo route) {
+            }
+
+            @Override
+            public void onRouteSelected(MediaRouter router, RouteInfo route) {
```
krocard, line 197: It this called on registration with the current route on registration? If not, then `externalPcmOnlySoundset` state will be incorrect until the route change.

krocard (commented):
This seems like a sensible way to get the capabilities of the audio route currently selected.
I have a couple of reservation regarding the state management and capabilities assumption.
Though this could be an improvement, until AudioDeviceInfo query is implemented.
@andrewlewis what is your opinion?

library/core/src/main/java/com/google/android/exoplayer2/trackselection/TrackSelector.java
stevemayhew: If you want to deal with 'hot-plug' (which we do, for example if the bluetooth device connects after you start playback) you need to listen to the route changes and trigger Track Selection when they occur.   There are no changes to the constraint based parameters ( `TrackSelectionParameters`) but the audio route will change will force a renderer/codec change.

The code we have in our application does this:

```java
    audioCapabilitiesReceiver = new AudioCapabilitiesReceiver(context, audioCapabilities -> {
      if (trackSelector != null) {
        trackSelector.invalidate();
      }
    });
    audioCapabilitiesReceiver.register();
```

To make this part of the change, we see there are a few options here:

1. Open this method up and document why
1. Add a new parameter to `TrackSelectionParameters` that follows the API desired for when AudioDeviceInfo is available (for now, all it has to do is produce a change to the parameters to force an invalidate()
1. Incorporate this listener code down in `SimpleExoPlayer`, still this requires wider access to invalidate(), but with non documentation
1. Incorporate this listener code in the `trackselection` package, note this would pollute this package with retaining a reference to `Context`, which it currently wisely does not

We are ok with any of these choices, my personal feeling is option 3 (in SimpleExoPlayer) is probably best, depends on where you are with the API for using `AudioDeviceInfo`.  We wanted to allow the team to provide input before we added code to this pull request.  So, @ojw28 and @krocard your thoughts?

library/core/build.gradle
```diff
@@ -37,6 +37,8 @@ dependencies {
     api project(modulePrefix + 'library-common')
     api project(modulePrefix + 'library-extractor')
     implementation 'androidx.annotation:annotation:' + androidxAnnotationVersion
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
ojw28, line 40: I'm not convinced we want to start taking extra dependencies in core for this. Core currently has exactly one dependency, on a module that provides only annotations. See also the first rule at https://jlbp.dev/ :).

Is there a reason why you can't maintain your own `AudioCapabilitiesReceiver` that contains the workaround in this pull request? I don't think it's actually used inside the library anywhere, so it's possibly only used from your own application code anyway?

stevemayhew: Sure @ojw28 my [comment](https://github.com/google/ExoPlayer/pull/8178#discussion_r527150635) above speaks to the question of where the code should land up that uses this class. 

The class itself needs to listen to route changes from `MediaRouter` and that means including the AndroidX MediaRouter AAR.   We only see the results of the refactoring of core, so whatever you think is best for this `com.google.android.exoplayer2.audio` package.   class does depend (obviously) on `AudioCapabilities` but that pulls a pretty long thread all the way to the `MediaCodecAudioRenderer`

andrewlewis: Related to Kevin's earlier comment, if we implemented querying devices using `AudioDeviceInfo` from API 23 would this change still be needed? If not that seems like a better long term solution, and keeping the workaround from this change in the app for now seems preferable to adding this then removing it.

ojw28: Right. In which case, what's the minimum change that _needs_ to go into the library, if anything, to support a workaround in the application?

* I don't think `AudioCapabilitiesReceiver` needs to change at all
* I don't think `TrackSelector.invalidate` needs making public because it's trivial to extend `DefaultTrackSelector` and add a method like `public void forceInvalidate() { invalidate(); }`. That seems preferable, because otherwise we'll be exposing a method in the library API that never does anything useful (unless the app is making other customizations, in which case it's reasonable to expect the app to customize this as well).
* The ability to change the `AudioCapabilities` that the `DefaultAudioSink` has seems like the one piece that's missing for an easy application workaround. Is that the case? If so then IMO we should focus on a smaller change that supports replacing those, if it's a feasible thing to do. It would be preferable to do this without removing the immutability property of `AudioCapabilities`.

stevemayhew: To answer Andrew (@andrewlewis ):  The change to `AudioCapabilitiesReceiver` allows it to detect the change in audio routing, that is essential to the fix.   I've not seen the design for using an `AudioDeviceInfo` query, but it seems to me in any case we will need:

1. some set of `BroadcastReceiver`'s that are notified of the audio route change (HDMI hotplug, Bluetooth device attaches, etc)
2. mechanism to reconfigure the CODEC / Renderer / AudioTrack to work with the new route.

To answer Oliver (@ojw28)  I like the idea of extending `DefaultTrackSelector` and put it in the package with `SimpleExoPlayer` so we can implement option 3 and still keep everything package private at highest access.  I do think it is better to include the fix in the library, otherwise the code the application would need is rather obscure (see the code snippet in my comment above.)  We are putting the listener in `SimpleExoPlayer` ourselves so our applications do not need to include it.

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
sneelavara: Thanks @krocard for the input. Related the magic number. 
Bluetooth A2DP profile can stream audio in stereo only. It appears the limitation of A2DP. We tested with different speaker models including Apple AirPods. All of them fails the similar way when 5.1 channel AC-3 audio is passed through.

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
sneelavara: Changed to use "== DEVICE_TYPE_BLUETOOTH"

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
sneelavara: Thanks @krocard for the feedback. We decided to use CALLBACK_FLAG_REQUEST_DISCOVERY based on https://developer.android.com/reference/androidx/mediarouter/media/MediaRouter#addCallback(androidx.mediarouter.media.MediaRouteSelector,%20androidx.mediarouter.media.MediaRouter.Callback,%20int)

CALLBACK_FLAG_FORCE_DISCOVERY seem to have significant performance impact.

In our experiment we are being notified when BlueTooth speaker connects without having to scan. i.e. even if the BT speaker connected earlier, when App comes up, the notification arrives here and we are able to set the PCM_ONLY Audioapabilities.

krocard: If the capabilities are eventually consistent with `CALLBACK_FLAG_REQUEST_DISCOVERY`, it should indeed be enough. I do wonder though when ExoPlayer is playing in a foreground service how long the callback can be delayed.
If it is by more than a few seconds users are going to consider that a bug.

library/core/build.gradle
krocard: I believe the MediaRouter Androidx library is currently too big to be added as an ExoPlayer dependency. If we want to provide the proposed dynamic routing mechanism, we should do so in a MediaRouter ExoPlayer extension that clients can chose to use or not depending on their need and APK size requirements.

ojw28: > Otherwise the code the application would need is rather obscure

This is true, but the problem is also somewhat obscure. It certainly doesn't seem to be widely reported, and it sounds a bit like there's a better long term fix anyway. So I don't really see the justification for these changes needing to go into the library, particularly given they'll either require a (likely unacceptable) additional dependency, or an entirely separate library module.

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilities.java
stevemayhew: I think that should work Kevin.  Most everything supports 16 bit PCM ("WAV format").

library/core/src/main/java/com/google/android/exoplayer2/audio/AudioCapabilitiesReceiver.java
stevemayhew: Yes, it is called as a `handleMessage` callback on the `Looper` that calls register if the bluetooth device is connected.  It looks like it may not for HDMI connection, So this looks like a bug.  Good catch, thanks!

library/core/build.gradle
stevemayhew: @ojw28 FWIW Issue #7788 occurs on the new Google TV as well when it's connected to any media center (HDMI sound-bar or receiver that supports DD sound), so you may get more reports down the line.   I can't imagine a use case on phones that would run into this, might be one explanation for why you have not seen it reported.

I still think the core library (or some module) will need to listen for and respond to audio routing changes as part of any long term fix.  We can certainly move all the code accept for the change to `AudioCapabilities` to our own code, would be good to have `AudioCapabilitiesReceiver` so we would not need to copy all of that code.

MartB: Seems like #8318 is related to this?
But its not about bluetooth, just format changes, could your PR help in this case too?

library/core/src/main/java/com/google/android/exoplayer2/trackselection/TrackSelector.java
krocard: I agree that 3 is better, but I would not trigger this directly.
Triggering a track reselection is not enough. The `AudioCapabilities` of the `DefaultAudioSink` are currently immutable and need to be added a setter. That setter could send a `onCapabilityChange` notification to the renderer that could propagate a `onSupportedFormatChanged` or similar that would force a rerouting at the player level rather than been handled directly by the track selector.
We might want to use this logic for offload too, rather than the current behaviour of throwing an exception and recreating the renderer from scratch.

tonihei:
I'll close this PR for now. Based on the discussion above and the linked issue (https://github.com/google/ExoPlayer/issues/7788), it seems clear that we won't fix the issue directly for now, and we also won't make `invalidate` public. The rest of the PR also adds dependencies we shouldn't take. So if we decide to provide a way for apps to workaround this issue and/or more closely integrate this into ExoPlayer itself, the code likely looks very different anyway.

PR #8214, opened 2020-11-12
Fix for Apple's I-Frame-only stream playback.

mac-dobrzynski-vizio:


mac-dobrzynski-vizio:
Need rework. Closing.

Issue #7512, opened 2020-06-16
Fix playback of I-FRAME only variants in Apple's advanced FMP4 stream


AquilesCanta:
To reproduce, simply play the "Apple master playlist advanced (FMP4)" sample from the demo app and select (using the track selection UI) an I-FRAME only track.

Note: I-FRAME only variants are only accessible through track selection overrides.

stevemayhew:
@AquilesCanta what is happening is the `FragmentedMp4Extractor` is returning exiting the read() loop (returning `RESULT_END_OF_INPUT`), but `ExtractorInput.getPosition()` returns the length of the stream (like it is not fully read).

This makes the subrange [computed on HlsMediaChunk#L365](https://github.com/google/ExoPlayer/blob/1a5b304b1faf244ea91e2b2aca8ea241aa237dd1/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java#L365) invalid (as the left over value in `nextLoadPosition` is incorrect.

I have sample mp4 files that seem to  re-create the issue, if you are ok with it I will add this to the  `FragmentedMp4Extractor` test cases and submit a pull request if it fixes the issue. 

Thanks for catching it!.

AquilesCanta:
I'm still not clear how the nextLoadPosition is involved. I thought nextLoadPosition was relevant when there was an error while reading the media. The error I ran into when I tried the media was an end of input exception. But when I added a catch to ignore that exception, and treat the chunk as completed (In which case nextLoadPosition should not matter), playback still failed. Any PRs are welcome though. If just for reference you can explain the phenomenon, it would be really useful. To be honest I didn't really dig that much. It's possible it's really simple.

stevemayhew:
Well, nothing is ever really simple ;-).   I'm not clear on the history (looks very long back) on the comments on [line 355](https://github.com/google/ExoPlayer/blob/1a5b304b1faf244ea91e2b2aca8ea241aa237dd1/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java#L355) but it seems the range in the `dataSpec` is (was) being used to partially fetch segments this would seem to overlap/conflict with the fact that all the iframe only segments are already byte-ranged to a single sample.

At the completion of parsing of one of these ranged segments `input.getPosition() - dataSpec.absoluteStreamPosition` is  non-zero so that messes up the next byte-range fetch.

[...]

PR #8215, opened 2020-11-13
Fix for Apple's I-Frame-only stream playback.

mac-dobrzynski-vizio:
Fixes issue #7512.  Apple's i-Frame playlist is published as byte-range indexes into the main fMP4 segments.  

These byte ranges have the IDR sample required, but this results in a partial MDAT which leaves the `FragmentedMp4Extractor` in an unusable state for the next segment load.

The fix simply resets the Extractor by performing a vacuous `Extractor.seek()` operation.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/BundledHlsMediaChunkExtractor.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/BundledHlsMediaChunkExtractor.java
@@ -68,6 +68,11 @@ public final class BundledHlsMediaChunkExtractor implements HlsMediaChunkExtract
   }
 
   @Override
+  public void seek(long position, long timeUs) {
+    extractor.seek( position, timeUs );
+  }
+
+  @Override
   public boolean isPackedAudioExtractor() {
     return extractor instanceof AdtsExtractor
         || extractor instanceof Ac3Extractor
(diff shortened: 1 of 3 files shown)


mac-dobrzynski-vizio:
@googlebot I signed it!

mac-dobrzynski-vizio:
@googlebot I signed it!

mac-dobrzynski-vizio:
@googlebot I signed it!

mac-dobrzynski-vizio:
Hi @AquilesCanta, in my comment [link](https://github.com/google/ExoPlayer/issues/7512#issuecomment-704535978) I mentioned making a use of `FragmentedMp4Extractor.seek()` method, though it looked there was no direct access to it from the `HlsMediaChunk` class. Well, it turned out it wasn't that hard to make it available after all.

AquilesCanta:
Hi, the change looks good, in general. I can't avoid feeling that we are leaking Extractor implementation details by using seek to reset the state of the fMp4 extractor. 

However, I cannot think of a clearly better way of doing this (fixing in the extractor is the only other option I'm considering right now). I'll discuss with the team and merge if there are no objections.

mac-dobrzynski-vizio:
Thanks @AquilesCanta for taking a look at the fix. It was also my first thought to implement it entirely in the fMp4 extractor. The problem is that the extractor has no knowledge about the byte ranges. Object `ExtractorInput`, which the extractor holds, does not keep the opening byte position (keeps only the current one) so we can't calculate how many bytes is left in the actual range. We can't rely strictly on EOF exceptions as they can happen in other unrelated scenarios too. We could instrument the `ExtractorInput` to keep that info but this seems a bit harder to implement and test.

AquilesCanta:
This has been merged internally and will be available publicly soon. Thanks for your contribution.

PR #8294, opened 2020-11-29
IndexOutOfBoundsException fixed when representation is empty

emircanpekyurek:


emircanpekyurek:
i signed it.

emircanpekyurek:
@googlebot I signed it!

ojw28:
@emircanpekyurek - You need to send this pull request to the `dev-v2` branch, as per our [contribution guidelines](https://github.com/google/ExoPlayer/blob/release-v2/CONTRIBUTING.md).

It's also somewhat unclear whether this is the right thing to do. An `AdaptationSet` with no `Representation` elements is a violation of the DASH spec. If this is your own content, shouldn't you just fix this? If we were to change something, I think throwing a `ParserException` would be more appropriate than silently dropping the `AdaptationSet` from the parsed object.

Closing this, since you'll need to send a new pull request to the `dev-v2` branch anyway.

PR #8303, opened 2020-12-01
IndexOutOfBoundsException fixed when representation is empty

emircanpekyurek:


emircanpekyurek:
@googlebot I fixed it.

ojw28:
As per my previous comment, do you not think it would be preferable to throw `ParserException` in this case, given an empty adaptation set is a violation of the DASH spec? Else we're silently dropping something from the parsed data structure, which seems confusing.

Is it your own media that contains empty adaptation sets, and if so, why not fix the media rather than add a workaround on the client?

Thanks!

emircanpekyurek:
representation empty in the media I play. It is not in my power to fix it.
I wish I could fix it. but I can play this content with the code I sent.

Oliver Woodman <notifications@github.com>, 1 Ara 2020 Sal, 16:56 tarihinde
şunu yazdı:

ojw28:
Who does own the content, in that case? Can you provide a link to it? We should report the issue to the content producer, in addition to any fix that we might want to merge into the library.

ojw28:
@emircanpekyurek - Please address the comment above if you wish us to proceed any further with this pull request. Thanks!

ojw28:
I'm going to close this for now due to lack of response to my comments above, and because the pull request is being made to the wrong branch. 

If you still want to see this change in ExoPlayer, please address the comment above, and open a new pull request for the change to the correct branch, which is the `dev-v2` branch as per our [contribution guidelines](https://github.com/google/ExoPlayer/blob/release-v2/CONTRIBUTING.md). Thanks!

PR #8340, opened 2020-12-10
Manually set up the Widevine L3 Workaround

emircanpekyurek:
some content (possibly DRM based) gives an error. When securityLevel is manually set to L3, the player continues to play. I haven't a sample video

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
@@ -91,8 +91,12 @@ public final class FrameworkMediaDrm implements ExoMediaDrm {
    * @throws UnsupportedDrmException If the DRM scheme is unsupported or cannot be instantiated.
    */
   public static FrameworkMediaDrm newInstance(UUID uuid) throws UnsupportedDrmException {
+    return newInstance(uuid, false);
+  }
+
+  public static FrameworkMediaDrm newInstance(UUID uuid, Boolean forceWidevineL3) throws UnsupportedDrmException {
     try {
-      return new FrameworkMediaDrm(uuid);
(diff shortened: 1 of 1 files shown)


icbaker:
I'm not sure I understand how this change is better than just calling `FrameworkMediaDrm#setPropertyString()` after you've constructed it.

Would you be able to offer a bit of context for how you intend to use this change?

emircanpekyurek:
drm init:
```
val mediaDrm = FrameworkMediaDrm.newInstance(DRM_SCHEME).also {
            if (forceWideVineL3) it.setPropertyString("securityLevel", "L3")
        }
```

This code works fine but is deprecated:
```
        return DefaultDrmSessionManager(DRM_SCHEME, mediaDrm!!, drmCallback, null, true)
```

The current code is as follows. It gives an crash when I use it :

```
DefaultDrmSessionManager.Builder()
                .setUuidAndExoMediaDrmProvider(DRM_SCHEME) { mediaDrm!! }
                .setKeyRequestParameters(null)
                .setMultiSession(true)
                .build(drmCallback)
```

Contains private mediaDrm in FrameworkMediaDrm. setPropertyString is called from this variable. this may be the reason

icbaker:
Are you saying that you're able to avoid the crash when using the `DefaultDrmSessionManager.Builder` by using the change you've proposed in this PR? Can you explain exactly what that looks like?

Regarding the crash: This is only a guess, since without the stacktrace & clear repro steps it's impossible to be sure, but I wonder if it would be fixed if you use `ExoMediaDrm.AppManagedProvider` - I think with your existing code there's a risk that the `mediaDrm` instance gets completely released.

Debugging crashes is a better fit for an issue rather than a PR - so if you want us to look deeper into the crash you're seeing can you open a new issue?

emircanpekyurek:
erasable

PR #8351, opened 2020-12-14
fix second seek sync position for live

zenxu88:
For the case the first segment number is not 0 based, getSegmentCount() could be smaller than segmentNum.
This will make seek always pick the first keyframe. To fix this, we use getLastSegmentNumber() instead.


ojw28:
@tonihei - Does it make sense for you to take a look at this, given you were adjusting some related logic in `DefaultDashChunkSource` for DASH-LL recently? Thanks!

tonihei:
Please sign the CLA before we can have a look at your PR. Thanks!

tonihei:
Closing because we didn't receive a CLA. We'll submit a fix for the problem ourselves instead.

zenxu88:
@tonihei I finally got CLA through my company email. However it seems you have applied a fix already?

tonihei:
Yes, sorry. It seemed worth a fix, so we went ahead and applies our own fix to the problem. Thanks for looking into the CLA anyway :) Maybe it becomes useful in the future.

Issue #1807, opened 2016-09-08
EAI-708 support


zsmatyas:
Previous version of the v1 and v2-dev branches had the entire SEI data put into a buffer and then read by the Eia608Parser. The parser had a check that the type and validity bits are set to channel 1 of the 608 content, and everything else was dropped. See: [V1](https://github.com/google/ExoPlayer/blob/release-v1/library/src/main/java/com/google/android/exoplayer/text/eia608/Eia608Parser.java#L141)
`if (ccType != 0) {
        seiBuffer.skipBits(16);
        continue;
}`
Here the ccType values mean:
- 0 = channel 1 and 2 of EIA-608
- 1 = channel 3 and 4 of EIA-608
- 2 = Start of EIA-708 content (channel ID is defined inside the stream)
- 3 = Chunk of EIA-708 content

As only value 0 was handled, and no additional checks were used to separate channel 1 and 2: a single 608 channel was supported. I added my additional channel and EIA-708 support by updating the Eia608Parser to collect the data instead of dropping it, so channel 3-4 and EIA-708 bits were collected and parsed appropriately. 

In the latest v2 code, all these additional data are dropped at SeiReader level:
[V2](https://github.com/google/ExoPlayer/blob/dev-v2/library/src/main/java/com/google/android/exoplayer2/extractor/ts/SeiReader.java#L65)
like:
`
if (ccValidityAndType != 0x04) {
      seiBuffer.skipBytes(2);
  }
`
The change was: [Link](https://github.com/google/ExoPlayer/commit/a0c15957259384f3fd4a8c893324c50b2c3f1ab2)

SeiReader seems to be the correct location to filter 608/708 and channel information, but needs serious changes in the architecture I had. Are you guys planning to ever support EIA-708 (as it is an FCC requirement from 2017)? Did you consider supporting the other channels of EIA-608? Any similar changes like the one you made in the v2-dev branch requires lots of changes to support all subtitles correctly, that is why I am interested in the plans about supported formats. 

Are you willing to accept pull requests related to these features or are you maybe already working on these issues?

Big Thanks!

ojw28:
@cdrolle is looking at 708 support. Could you comment on the above? Thanks!

zsmatyas:
Any news? :chicken:

zsmatyas:
:honey_pot:

zsmatyas:
[...]

PR #8356, opened 2020-12-16
CEA-708 Decoder fixes for issue #1807.

sneelavara:
This pull request is for issue#1807. In this commit -

1) Handling the sequence number discontinuity in caption channel packet header.

2) Handle the multiple Service Blocks within a single caption channel packet. An outer while loop is added to function processCurrentPacket() to achieve that.

3) The processCurrentPacket returns if the packet length does not match with the currentIndex. That assumption is wrong.
As per spec the the packet can end on reception of next cc_type = 0x3.

3) Added a logic to handle multi byte commands. These commands can be within the service block or spread across service blocks.

4) Processing of service block needs to wait until all the command bytes arrive. An ArrayList is used to accumulate the bytes belonging to same service block.

In this change the current flow is not altered, i.e. the parsing of service block is not started until the complete caption channel packet arrives.

These changes are verified through Sarnoff Test streams. Many tests are still failing after this change. Subsequent changes are planned to address them.


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
```diff
@@ -270,81 +300,113 @@ private void finalizeCurrentPacket() {
...
-          handleC3Command(command);
-        } else if (command <= GROUP_G3_END) {
-          handleG3Character(command);
-          cuesNeedUpdate = true;
+        // Convert the ArrayList to byte[].
+        byte[] blk = new  byte[serviceBlockList.size()];
+        for (int i = 0; i < serviceBlockList.size(); i++) {
+          blk[i] = serviceBlockList.get(i);
+        }
+        serviceBlockPacket.reset(blk, serviceBlockList.size());
```
icbaker, line 366: I might not be following exactly what's happening here (forgive my relative inexperience with CEA-708), but it seems like each time you go round the outer while loop (that starts on L314), if you find a service block with the selected number you add its bytes to serviceBlockList and then allow that to be parsed/consumed by the (mostly existing) inner while loop that starts on L363.

But if you find multiple (N) service blocks with a matching service number (maybe that's impossible...?) then you'll parse the first one N times, the second N-1 etc.

Something like:
```
selectedServiceNum=2
currentDtvCcData = {
  blockA { num=1 },
  blockB { num=2 },
  blockC { num=3 },
  blockD { num=2 },
}
```

Will cause the while loop on L363 to first be executed with:
```
serviceBlockPacket = {
  blockB { num=2 },
}
```

And then again with:
```
serviceBlockPacket = {
  blockB { num=2 },
  blockD { num=2 },
}
```

a) Have I understood what the code is doing correctly
b) Is that intended/necessary?

I have a copy of the CEA-708-B spec (dated Dec 1999) - so if there's sections in there that you can point to to help me understand what's meant to be going on that might be helpful!

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
```diff
@@ -270,81 +300,113 @@ private void finalizeCurrentPacket() {
...
-      serviceBlockPacket.skipBits(2);
-      serviceNumber = serviceBlockPacket.readBits(6);
-      if (serviceNumber < 7) {
-        Log.w(TAG, "Invalid extended service number: " + serviceNumber);
+    int currOffset = 0;
+    // Caption channel packet can contain more than one service blocks. Handle all the service
+    // blocks in the current packet. If the previous command spread across the paket, that is
+    // accumulated and handled after all the bytes are received
+    while (currOffset < currentDtvCcPacket.currentIndex) {
+      int serviceNumber = (currentDtvCcPacket.packetData[currOffset] & 0xE0) >> 5; //3 bits
```
icbaker, line 330: It seems far preferable to leverage `ParsableBitArray` rather than manually bit shifting.

I wonder if it's possible to re-use the `serviceBlockPacket` one everywhere by some cunning position checking (and rename it `allServiceBlocks`?), or failing that to introduce a second one called `allServiceBlocks` to separate the two.

I had a go locally with the first approach (use `serviceBlockPacket` everywhere) by making edits on top of this change - but I ran into a lack of understanding of exactly how the code is meant to behave (see my next comment) which I think affects how feasible that approach will be.

icbaker (commented):
Points 1 and 2 in the description seem simple & uncontroversial - is it worth splitting them out into a separate PR?

3 is the one that seems to introduce the most complexity - and (I think?) 4 and 5 both depend on 3.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
sneelavara: Thank you @icbaker for looking at this. I think I can split 1 and 3 into a separate PR. Though 3 aligns more with 4 and 5 but coding wise it looks independent. But for 2, it depends on restructuring of serviceBlockPacket and I think its better if we handle 2, 4 and 5 together. I am open for suggestions.  

For the scenario -
```
selectedServiceNum=2
currentDtvCcData = {
  blockA { num=1 },
  blockB { num=2 },
  blockC { num=3 },
  blockD { num=2 },
}
```
Will cause the while loop on L363 to first be executed with:
```
serviceBlockPacket = {
  blockB { num=2 },
}
```
Before we proceed to next iteration of outer loop, the L407 clears the  serviceBlockList.
```
if (ccDataState == CC_DATA_STATE_COMPLETE_PARAM) {
  serviceBlockList.clear();
}
```
And then, the blockD is added.  Then the inner while loop is called with -  
```
serviceBlockPacket = {
  blockD { num=2 },
}
```
With that, blockB and blockD are decoded as expected.
 
The outer loop adds all the service blocks from the selected service number to serviceBlockList. The purpose of serviceBlockList is to handle the commands with multiple parameters. For example DefineWindow command (CEA-708-D  section 8.10.5.2 DEFINE WINDOW - (DF0 … DF7)) has 6 parameters each of 1 byte. It makes Define Window coding is of total 7 bytes. All these 7 bytes can either be part of single service block or can be spread across multiple service blocks.  ParsableBitArray does not provide a way to add/extended an existing object.  ParsableBitArray.reset() is going to clear the previous data and leads to disassociation of command and its associated parameters. 

In this fix, if the serviceBlock ends before it provides all the parameters required for the command, the ccDataState is set to CC_DATA_STATE_WAITING_PARAM and the inner while loop breaks at LN376 and the serviceBlockList is not cleared at LN408. In the next iteration of the outer while loop, the next service block is appended to serviceBlockList. The serviceBlockList now contains 2 service blocks which together describes a command and its parameters. The inner while loop now has all the information to decode a command.

For example
```
selectedServiceNum=2
currentDtvCcData = {
  blockA { num=1 },
  blockB { num=2,  has partial DF command  like “0x9A, 0x38, 0x4A, 0xD1” },
  blockC { num=3 },
  blockD { num=2 has the next parameters for command 0x9A which are like “0x8B, 0x0F, 0x11”},
}
```
Will cause the while loop on L363 to first executed with:
```
serviceBlockPacket = {
  blockB { num=2  with “0x9A, 0x38, 0x4A, 0xD1”},
}
```
The condition at LN 590 becomes true
```
if (serviceBlockPacket.bitsLeft() < DFS_COMMAND_LEN) {
  ccDataState = CC_DATA_STATE_WAITING_PARAM;
  break;
}
```
That makes cuesNeedUpdate to remain false and skips serviceBlockList.clear() at LN408. The serviceBlockList now holds blockB.
Then the blockD (without the service block header) is appended to serviceBlockList on the next iteration of outer while loop. Now the inner loop is called with 
```
serviceBlockPacket = {
 blockB { num=2 }
 blockD { num = 2 }
}
```
This now describes command DF (0x9A) with all 6 parameters and ccDataState becomes CC_DATA_STATE_COMPLETE_PARAM. This time the new cue is added.

I am referring to  CEA-708-D from August 2008. The https://en.wikipedia.org/wiki/CEA-708 also provides a good overview.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
sneelavara: Thanks @icbaker. My understanding of `ParsableBitArray` is that it can not be resized after creation. As per CEA-708 spec, the `cc_data()` can be split across multiple caption channel packets (`section 4.4.1.1` - `A DTVCC Caption Channel Packet may continue from one cc_data() structure to the next`). In such case, we need to retain the previous data to interpret the commands correctly. But  ParsableBitArray.reset() clears everything. If we could extend ParsableBitArray with append feature, I think we can still use ParsableBitArray here.   As a simple workaround, I used ArrayList which can be resized easily.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
sneelavara: @icbaker points 1 and 3 are separated into PR #8415.

sneelavara:
@icbaker any suggestion how can we add unit tests for these files (Cea708Decoder and Cea608Decoder)?
 Since I have shared the test streams which would effectively execute almost all the aspects of the spec, would that be sufficient?

icbaker:

PR #8371, opened 2020-12-18
Update media.exolist.json Reslove playready test stream can't play's issue

Sunshipeng:
Add Super speed(Playready) Test Stream's KeyUrl and
force use the key url to sure exoplayer can play this
est stream.


tonihei:
Please sign the CLA if you want us to look at the PR.

ojw28:
Note - This should not be necessary in general, but appears to be in this case because the media specifies a cleartext (http) license URL, rather than a secure (https) one. We can submit an equivalent fix into the `dev-v2` branch, which is where such changes should be made. Thanks!

Issue #8393, opened 2020-12-23
ExoPlayer doesn't check if AVC/HEVC profile/level are supported when playing media file directly


equeim:
It seems that ExoPlayer can only check if video profile/level is supported by decoder (as reported by MediaCodecInfo.getCapabilitiesForType) only when profile/level information is provided in DASH/HLS manifest.

The only exception is MPEG-TS extractor that can get profile/level information from AVC bitstream.

If that is not an intentional design decision, I would like to make a pull request that will allow ExoPlayer to extract profile/level information from AVC and HEVC bitstreams for other extractors too.

ojw28:
I don't think there was ever an intentional decision _not_ to implement this. That said, the value of doing so seems smaller than for DASH/HLS. For DASH/HLS, it might be the difference between playback succeeding and failing, since variant streams may use different profiles/levels, and a subset may be supported. For progressive streams I would assume it's the difference between playback failing with a clear error and playback failing with a less clear error.

Does that align with your understanding, and is your goal here just to make the failure mode clearer? If so, that seems fine, although given the relatively limited usefulness, I think we'd want to keep the implementation fairly straightforward. If it's fundamentally extremely complicated then it might be preferable not to implement it at all. Hopefully that's not the case though!

So, I think the next steps here would be:

* If your goal is to make the failure mode clearer and if you think it's fairly straightforward, feel free to go ahead and send a pull request to the `dev-v2` branch.
* If you have other goals, let's discuss what those are :).

As a final note, it's possible this might also fix https://github.com/google/ExoPlayer/issues/5882.

Thanks!

equeim:
>For progressive streams I would assume it's the difference between playback failing with a clear error and playback failing with a less clear error.

The problem is that not all decoders fail with error in these cases. Some will "decode" frames, but show black screen or distorted image, for example. This makes it hard for application to appropriate action in these cases (which is especially important for things like ads when you would want to skip unsupported ad video and show actual content).

I suspect that it's decoders fault to not indicate error when unsupported profile is passed for decode, but there is an API for check if is supported beforehand and it makes for player to use it.

Regarding implementation complexity: for AVC it is very easy, since the code is already there. All that is needed is a few lines to actually hook it up in extractors. For HEVC it would require a bit of additional SPS parsing, but it's quite straightforward too.

ojw28:
Thanks for the clarification! Perhaps send a pull request for AVC to get started, in that case, and follow up with the HEVC one separately :)?

[...]

PR #8401, opened 2020-12-27
HEVC: initialize Format.codecs (#8393)

equeim:
This will allow ExoPlayer to check if video track's profile and level are supported by decoder when playing progressive media sources.

### DIFF
--- a/library/common/src/main/java/com/google/android/exoplayer2/video/HevcConfig.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/video/HevcConfig.java
@@ -17,7 +17,9 @@ package com.google.android.exoplayer2.video;
 
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.ParserException;
+import com.google.android.exoplayer2.util.CodecSpecificDataUtil;
 import com.google.android.exoplayer2.util.NalUnitUtil;
+import com.google.android.exoplayer2.util.ParsableBitArray;
 import com.google.android.exoplayer2.util.ParsableByteArray;
 import java.util.Collections;
 import java.util.List;
@@ -26,9 +28,9 @@ import java.util.List;
  * HEVC configuration data.
  */
 public final class HevcConfig {
-
   @Nullable public final List<byte[]> initializationData;
   public final int nalUnitLengthFieldLength;
+  public final String codecs;
 
   /**
    * Parses HEVC configuration data.
@@ -40,7 +42,9 @@ public final class HevcConfig {
    */
   public static HevcConfig parse(ParsableByteArray data) throws ParserException {
     try {
-      data.skipBytes(21); // Skip to the NAL unit length size field.
+      data.skipBytes(1); // Skip to the general_profile_space
+      final String codecs = buildCodecString(data);
+      data.skipBytes(20); // Skip to the NAL unit length size field.
       int lengthSizeMinusOne = data.readUnsignedByte() & 0x03;
 
       // Calculate the combined size of all VPS/SPS/PPS bitstreams.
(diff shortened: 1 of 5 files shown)


library/common/src/main/java/com/google/android/exoplayer2/video/HevcConfig.java
```diff
@@ -72,20 +79,47 @@ public static HevcConfig parse(ParsableByteArray data) throws ParserException {
...
   }
 
-  private HevcConfig(@Nullable List<byte[]> initializationData, int nalUnitLengthFieldLength) {
+  private static String buildCodecString(ParsableByteArray data, int nalUnitLength) {
+    final ParsableNalUnitBitArray bitArray = new ParsableNalUnitBitArray(data.getData(), data.getPosition(), data.getPosition() + nalUnitLength);
+    bitArray.skipBits(24);
+    final int generalProfileSpace = bitArray.readBits(2);
+    final boolean generalTierFlag = bitArray.readBit();
+    final int generalProfileIdc = bitArray.readBits(5);
+    final int generalProfileCompatibilityFlags = bitArray.readBits(32);
```
andrewlewis, line 96: @equeim I think the order of bits in the resulting int is reversed compared to what's needed to build the codecs string. Please could you confirm whether that's consistent with your understanding or if I'm missing something?

I will take care of making the change if it's needed when merging this, so please just let me know whether it is correct. Thanks!

equeim: I'm not sure, actually. MPEG-DASH for DVB specification says that bits must be reversed, but MPEG-4 Part 15 (where codecs string for HEVC is defined officially) doesn't mention it. Although I don't have access to the latest version of MPEG-4 Part 15 yet so I can't check whether it was added later.
It seems that almost all HLS/DASH manifests out there have it reversed though, and that was google/shaka-packager does too (https://github.com/google/shaka-packager/blob/d90cf9a0fd87833f7d9210ffe65d175dcf994c42/packager/media/codecs/hevc_decoder_configuration_record.cc#L133).

equeim:
Updated PR to build codec string from HEVCDecoderConfigurationRecord instead of SPS in case of MP4/Matroska.

andrewlewis:
> Updated PR to build codec string from HEVCDecoderConfigurationRecord instead of SPS in case of MP4/Matroska.

Are there any particular benefits to doing it this way round? I was going to parse in one place for both `HevcConfig` and `H265Reader` but they can't share code if they don't both use `ParsableNalUnitBitArray`.

equeim:
Not really, aside from the fact that specification mentions HEVCDecoderConfigurationRecord specifically. I just though that code in HevcConfig would be clearer that way.

More interesting, in Apple sample HLS stream with HEVC and MP4 segments (https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_adv_example_hevc/master.m3u8), I discovered that information in HEVCDecoderConfigurationRecord and in SPS are not the same. In particular, generalProfileCompatibilityFlags in HEVCDecoderConfigurationRecord seems to have its bit order inverted compared to SPS. I'm not sure how it conforms to MP4 specification (probably doesn't), and I haven't seen other MP4 files like this (and ffmpeg generates MP4 with SPS and HEVCDecoderConfigurationRecord equivalent).

andrewlewis:
Thanks for the info. I'll test with that stream to make sure the compatibility flags look right in the output string.

ojw28:
This was merged in https://github.com/google/ExoPlayer/commit/ff8c8645abb1490fe130a9cd5a4bf4576734d140, although for some reason our tooling didn't update the pull request to reflect this. Closing manually since it's now merged. Thanks!

andrewlewis:
Just in case you are trying out ff8c8645abb1490fe130a9cd5a4bf4576734d140 on the development branch: I introduced a bug in how the buffer limit is set. I plan to submit a fix for that soon. Sorry for the inconvenience.

PR #8414, opened 2020-12-30
Improved generic exception "Error reading non-2xx response body" 

tidoemanuele:
This will allow the **onPlayerError** method to get the response http status code, message and headers also when there is an error getting the **errorResponseBody**

### DIFF
--- a/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
+++ b/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
@@ -304,13 +304,13 @@ public class OkHttpDataSource extends BaseDataSource implements HttpDataSource {
 
     // Check for a valid response code.
     if (!response.isSuccessful()) {
-      Map<String, List<String>> headers = response.headers().toMultimap();
       byte[] errorResponseBody;
       try {
         errorResponseBody = Util.toByteArray(Assertions.checkNotNull(responseByteStream));
       } catch (IOException e) {
         errorResponseBody = Util.EMPTY_BYTE_ARRAY;
       }
+      Map<String, List<String>> headers = response.headers().toMultimap();
       closeConnectionQuietly();
       InvalidResponseCodeException exception =
           new InvalidResponseCodeException(
--- a/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
+++ b/extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
@@ -304,14 +304,14 @@ public class OkHttpDataSource extends BaseDataSource implements HttpDataSource {
 
     // Check for a valid response code.
     if (!response.isSuccessful()) {
+      Map<String, List<String>> headers = response.headers().toMultimap();
       byte[] errorResponseBody;
       try {
         errorResponseBody = Util.toByteArray(Assertions.checkNotNull(responseByteStream));
       } catch (IOException e) {
-        throw new HttpDataSourceException(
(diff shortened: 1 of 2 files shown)


extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
```diff
@@ -304,14 +304,14 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
...
     // Check for a valid response code.
     if (!response.isSuccessful()) {
+      Map<String, List<String>> headers = response.headers().toMultimap();
       byte[] errorResponseBody;
       try {
         errorResponseBody = Util.toByteArray(Assertions.checkNotNull(responseByteStream));
       } catch (IOException e) {
-        throw new HttpDataSourceException(
-            "Error reading non-2xx response body", e, dataSpec, HttpDataSourceException.TYPE_OPEN);
+        throw new InvalidResponseCodeException(
```
ojw28: Isn't it better just to replace this with a `errorResponseBody = Util.EMPTY_BYTE_ARRAY` and then fall through, so that the lines below take care of throwing what you want? I think that should just work. Note that you need to do it this way for the special handling of `416` to work as well.

Note that it will not be possible to disambiguate between "no response body" and "failed to read response body" in this case. If you think it's important to be able to disambiguate then you can use `errorResponseBody = null` instead. I'm not sure it's worth it though?

library/core/src/main/java/com/google/android/exoplayer2/upstream/DefaultHttpDataSource.java
```diff
@@ -377,8 +377,8 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
         errorResponseBody =
             errorStream != null ? Util.toByteArray(errorStream) : Util.EMPTY_BYTE_ARRAY;
       } catch (IOException e) {
-        throw new HttpDataSourceException(
-            "Error reading non-2xx response body", e, dataSpec, HttpDataSourceException.TYPE_OPEN);
+        throw new InvalidResponseCodeException(
```
ojw28: Same comment as in `OkHttpDataSource`. I think it's better just to set `errorResponseBody` to Util.EMPTY_BYTE_ARRAY` and fall through.

Note that `CronetDataSource` should also be modified as part of this change. Probably by enclosing [this while loop](https://github.com/google/ExoPlayer/blob/1347d572ef9ce79aacd667cfffa7d7468c8408a4/extensions/cronet/src/main/java/com/google/android/exoplayer2/ext/cronet/CronetDataSource.java#L555) in a `try/catch` block and then in the `catch` block setting `responseBody` back to `Util.EMPTY_BYTE_ARRAY` before falling through.

extensions/okhttp/src/main/java/com/google/android/exoplayer2/ext/okhttp/OkHttpDataSource.java
```diff
@@ -304,14 +304,14 @@ public long open(DataSpec dataSpec) throws HttpDataSourceException {
...
+            responseCode, response.message(), headers, dataSpec, null);
```
ojw28: The `responseBody` argument isn't `@Nullable`, so you should use `Util.EMPTY_BYTE_ARRAY` as suggested above. Alternatively, you'll need to add `@Nullable` annotations through `InvalidResponseCodeException`. I think it's probably best to use `Util.EMPTY_BYTE_ARRAY` unless you can think of a case where it's important to be able to disambiguate between the cases described above.

ojw28 (changes requested):

ojw28:
Please see comments inline. Thanks!

tidoemanuele (commented):
you are right, better to set the responseBody to empty there

PR #8415, opened 2020-12-31
CEA-708 Decoder fixes for issue #1807 (Refactoring the PR #8356)

sneelavara:
This pull request is for issue #1807. Refactoring the PR #8356

In this change -

1. Handling the sequence number discontinuity in caption channel packet header.

2. The processCurrentPacket returns if the packet length does not match with the currentIndex. That assumption is wrong. As per spec the the packet can end on reception of next cc_type = 0x3.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
@@ -145,6 +145,7 @@ public final class Cea708Decoder extends CeaDecoder {
 
   private final ParsableByteArray ccData;
   private final ParsableBitArray serviceBlockPacket;
+  private int lastSequenceNo = -1;
   // TODO: Use isWideAspectRatio in decoding.
   @SuppressWarnings({"unused", "FieldCanBeLocal"})
   private final boolean isWideAspectRatio;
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
```diff
@@ -270,10 +278,11 @@ private void finalizeCurrentPacket() {
...
   private void processCurrentPacket() {
     if (currentDtvCcPacket.currentIndex != (currentDtvCcPacket.packetSize * 2 - 1)) {
-      Log.w(TAG, "DtvCcPacket ended prematurely; size is " + (currentDtvCcPacket.packetSize * 2 - 1)
+      Log.d(TAG, "DtvCcPacket ended prematurely; size is " + (currentDtvCcPacket.packetSize * 2 - 1)
           + ", but current index is " + currentDtvCcPacket.currentIndex + " (sequence number "
-          + currentDtvCcPacket.sequenceNumber + "); ignoring packet");
-      return;
+          + currentDtvCcPacket.sequenceNumber + ");");
+      // This is not invalid packet. As per CEA-708 section 4.4.1.1, the Packect end can happen
+      // when the reception of cc_type = 0x03 (binary 11).
```
icbaker, line 285: I'm going to update this comment to the following - does that sound right to you?

```
// We've received cc_type=0x03 (packet start) before receiving packetSize byte pairs of data.
// This might indicate a byte pair has been lost, but we'll still attempt to process the data
// we have received.
```

icbaker:
Thanks for splitting this out of the larger PR - this is much easier to review (and thanks also for the references to the spec, very helpful).

You sent through a number of test streams - are there any specific ones that I can use to verify this change in particular? i.e. something that clearly fails before this change and succeeds after it?

I tried the first one but it seems unaffected by this change.

library/core/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
sneelavara: @icbaker I agree with the above comment.

sneelavara:
@icbaker thank you for looking at this PR. I have sent an e-mail to `dev.exoplayer@gmail.com` with subject `Issue #1807` at `Jan 8, 2021, 6:58 PM `
There are two streams provided in the mail to validate the two items from this PR. 

The stream which has the packet length set to 128 bytes but actual packet length shorter than that, unfortunately requires to be tested with the changes from PR #8356 to get the right caption displayed. If you are testing the changes from this PR alone, that simply displays few character with only one row, where as the actual caption has 4 rows.

Issue #8253, opened 2020-11-19
How to handle CacheException?


vhow:
I’m using v2.11.7, and it throws CacheException sometimes.

> E/ExoPlayerImplInternal: Source error
> **E/ExoPlayerImplInternal:   com.google.android.exoplayer2.upstream.cache.Cache$CacheException: Failed to create cache directory: /data/user/0/com.x.x/cache/**
> E/ExoPlayerImplInternal:       at com.google.android.exoplayer2.upstream.cache.p.d(SimpleCache.java:522)
> E/ExoPlayerImplInternal:       at com.google.android.exoplayer2.upstream.cache.p.a(SimpleCache.java:50)
> E/ExoPlayerImplInternal:       at com.google.android.exoplayer2.upstream.cache.p$1.run(SimpleCache.java:268)
> ...
> **E/DownloadManager: Task failed: progressive:xxx, false**
> **E/DownloadManager:   com.google.android.exoplayer2.upstream.cache.Cache$CacheException: Failed to create cache directory: /data/user/0/com.xxx/cache/**
> E/DownloadManager:       at com.google.android.exoplayer2.upstream.cache.p.d(SimpleCache.java:522)
> E/DownloadManager:       at com.google.android.exoplayer2.upstream.cache.p.a(SimpleCache.java:50)
> E/DownloadManager:       at com.google.android.exoplayer2.upstream.cache.p$1.run(SimpleCache.java:268)

marcbaechinger:
The exception is thrown on L522 which with 2.11.7 is when the cache is initialized and the directory for storing the cached data needs to be created because it does not exist.  The path ` /data/user/0/com.x.x/cache/` you are using is the cache directory you get from `context.getCacheDir()`. That actually seems about right to me and I would expect that it exists when starting up and it does not need to be created. `context.getExternalCacheDir()` would be an option, but I don't think it should fail with the path you are using, because it should actually exist.

You are saying it throws the exception only sometimes. Does this happen when you deploy from the IDE with adb  to a emulator or a test device? When it does not work does it fail repeatedly? In general, do you have some more information about in what situation it fails?

AChep:
This actually does happen for me every time on the first installation. Looking through the code
```
  private static void createCacheDirectories(File cacheDir) throws CacheException {
    if (!cacheDir.mkdirs()) {
      String message = "Failed to create cache directory: " + cacheDir;
      Log.e(TAG, message);
      throw new CacheException(message);
    }
  }
```
[...]

PR #8458, opened 2021-01-14
Double-check that the cache directory does not exist

AChep:
Check https://github.com/google/ExoPlayer/issues/8253#issuecomment-760218637 for more info

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
@@ -837,7 +837,7 @@ public final class SimpleCache implements Cache {
   }
 
   private static void createCacheDirectories(File cacheDir) throws CacheException {
-    if (!cacheDir.mkdirs()) {
+    if (!cacheDir.mkdirs() && !cacheDir.exists()) {
       String message = "Failed to create cache directory: " + cacheDir;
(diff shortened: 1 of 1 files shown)


lcf87:
Please move the pull request to dev-v2.

AChep:
Will do it this evening.

library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
```diff
@@ -837,7 +837,7 @@ private static long parseUid(String fileName) {
   }
 
   private static void createCacheDirectories(File cacheDir) throws CacheException {
-    if (!cacheDir.mkdirs()) {
+    if (!cacheDir.mkdirs() && !cacheDir.exists()) {
```
ojw28, line 840: Would it be a bit better to use:

`if (!cacheDir.mkdirs() && (!cacheDir.exists() || !cacheDir.isDirectory())) {`

ojw28:
@AChep - Thanks for this. I'll close this PR because I'm pretty certain you'll need to use a fresh one to target `dev-v2` as required. Please send the one for `dev-v2` when you have a chance. I also put one comment inline (see above). Thanks!

AChep:
@ojw28 do you mind if I'll add _more messages and stuff_ if the file is not a directory?

PR #8459, opened 2021-01-14
Before throwing, double check the cache dir does not exist

AChep:
Check https://github.com/google/ExoPlayer/issues/8253#issuecomment-760218637 for more info

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
@@ -837,7 +837,7 @@ public final class SimpleCache implements Cache {
   }
 
   private static void createCacheDirectories(File cacheDir) throws CacheException {
-    if (!cacheDir.mkdirs()) {
+    if (!cacheDir.mkdirs() && !cacheDir.isDirectory()) {
       String message = "Failed to create cache directory: " + cacheDir;
       Log.e(TAG, message);
       throw new CacheException(message);


AChep:
We can probably delete the existing file if it's not a directory and create a dir instead, what do you guys think?

library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/SimpleCache.java
```diff
@@ -408,18 +408,15 @@ public synchronized File startFile(String key, long position, long length) throw
     CachedContent cachedContent = contentIndex.get(key);
     Assertions.checkNotNull(cachedContent);
     Assertions.checkState(cachedContent.isFullyLocked(position, length));
-    if (!cacheDir.exists()) {
-      // The cache directory has been deleted from underneath us. Recreate it, and remove in-memory
+    if (createCacheDirectories(cacheDir)) {
```
ojw28: Checking the directory is really a directory every time a file is started seems like overkill. What is the problem this is actually trying to solve? The user replacing the cache directory with a file, whilst the app is simultaneously running, seems like something that would never happen in practice.

I think I preferred the original, more minimal version of this change. Could we consider restoring the pull request to that version?

I see now my suggestion to check `isDirectory` may not have been a good suggestion, since it kind of implied that it would make sense to do this. But on reflection I'm not sure what it's really helping with. Apologies for that!

AChep:
@ojw28 yeah, that was a little bit overkill :) Reverted to a simpler solution

PR #8462, opened 2021-01-15
Fix VP9 on Android M and below by advertising profile level

zeninsta:
On Android M and below, VP9 codecCapabilities do not advertise profile level support. In this case, estimate the level from MediaCodecInfo.VideoCapabilities instead. https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel.html

This change is ported from chromium media: https://chromium.googlesource.com/chromium/src/media/+/master/base/android/java/src/org/chromium/media/MediaCodecUtil.java#296

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
@@ -47,7 +47,6 @@ import com.google.android.exoplayer2.util.Assertions;
 import com.google.android.exoplayer2.util.Log;
 import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.Util;
-import java.util.ArrayList;
 
 /** Information about a {@link MediaCodec} for a given mime type. */
 @SuppressWarnings("InlinedApi")
@@ -301,10 +300,10 @@ public final class MediaCodecInfo {
     }
 
     CodecProfileLevel[] codecProfileLevels = getProfileLevels();
-    if (MimeTypes.VIDEO_VP9.equals(mimeType) &&
-        Util.SDK_INT <= 23 &&
-        codecProfileLevels.length == 0 &&
-        capabilities != null) {
+    if (MimeTypes.VIDEO_VP9.equals(mimeType)
+        && Util.SDK_INT <= 23
+        && codecProfileLevels.length == 0
+        && capabilities != null) {
       codecProfileLevels = getVp9CodecProfileLevelsV23(capabilities);
     }
 
--- a/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
@@ -47,6 +47,7 @@ import com.google.android.exoplayer2.util.Assertions;
 import com.google.android.exoplayer2.util.Log;
 import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.Util;
+import java.util.ArrayList;
 
 /** Information about a {@link MediaCodec} for a given mime type. */
 @SuppressWarnings("InlinedApi")
@@ -298,7 +299,15 @@ public final class MediaCodecInfo {
       // which may not be widely supported. See https://github.com/google/ExoPlayer/issues/5145.
       return true;
     }
-    for (CodecProfileLevel capabilities : getProfileLevels()) {
+
+    final CodecProfileLevel[] codecProfileLevels;
+    if (MimeTypes.VIDEO_VP9.equals(mimeType) && Util.SDK_INT <= 23 && capabilities != null) {
+      codecProfileLevels = getVp9CodecProfileLevelsV23(capabilities);
+    } else {
+      codecProfileLevels = getProfileLevels();
+    }
+
+    for (CodecProfileLevel capabilities : codecProfileLevels) {
       if (capabilities.profile == profile && capabilities.level >= level) {
         return true;
       }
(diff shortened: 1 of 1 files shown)


library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -572,6 +581,41 @@ public boolean isAudioChannelCountSupportedV21(int channelCount) {
...
+  /**
+   * On versions L and M, VP9 codecCapabilities do not advertise profile level
+   * support. In this case, estimate the level from MediaCodecInfo.VideoCapabilities
+   * instead. Assume VP9 is not supported before L. For more information, consult
+   * https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel.html
+   */
+  private static CodecProfileLevel[] getVp9CodecProfileLevelsV23(CodecCapabilities capabilities) {
+    // https://www.webmproject.org/vp9/levels
+    final int[][] bitrateMapping = {
+        {200, 10}, {800, 11}, {1800, 20}, {3600, 21}, {7200, 30}, {12000, 31}, {18000, 40},
```
ojw28: I don't think the level values are correct here. For example, `180000bits/s` does indeed map to level `5.2` according to https://www.webmproject.org/vp9/levels. But the level value that's supposed to be used in a `CodecProfileLevel` in this case is [CodecProfileLevel.VP9Level52](https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel#VP9Level52), which has a constant value `1024`. I think the right thing to do is to replace the second value in each sub-array with the correct `CodecProfileLevel.VP9Level[X]` constant.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -298,7 +299,15 @@ public boolean isCodecSupported(Format format) {
       // which may not be widely supported. See https://github.com/google/ExoPlayer/issues/5145.
       return true;
     }
-    for (CodecProfileLevel capabilities : getProfileLevels()) {
+
+    final CodecProfileLevel[] codecProfileLevels;
+    if (MimeTypes.VIDEO_VP9.equals(mimeType) && Util.SDK_INT <= 23 && capabilities != null) {
+      codecProfileLevels = getVp9CodecProfileLevelsV23(capabilities);
+    } else {
+      codecProfileLevels = getProfileLevels();
```
ojw28: What does `getProfileLevels()` actually return on affected devices? If an empty array is returned, it seems better to use that as a condition for using the workaround path, rather than hardcoding which path to use based only an API level check. This way, if a device with `SDK_INT <= 23` does actually return something from `getProfileLevels()`, we would avoid overriding it with estimated information.

Also, it seems that this code is going to override for J and K as well as L and M. This is probably the correct thing to do (`getProfileLevels()` seems to return the empty array on a K device I have), but please update the pull request title to accurately reflect what the pull request does. Also the comment block further down in this file that starts with "On versions L and M". Thanks!

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -572,6 +581,41 @@ public boolean isAudioChannelCountSupportedV21(int channelCount) {
     return true;
   }
 
+  /**
+   * On versions L and M, VP9 codecCapabilities do not advertise profile level
+   * support. In this case, estimate the level from MediaCodecInfo.VideoCapabilities
+   * instead. Assume VP9 is not supported before L. For more information, consult
+   * https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel.html
+   */
+  private static CodecProfileLevel[] getVp9CodecProfileLevelsV23(CodecCapabilities capabilities) {
```
ojw28, line 590: I don't think there's any need to be returning multiple `CodecProfileLevel` entries from this method, given they all have the same `profile` value. You just need to return a single entry with `level` set to the highest level that's supported. Support for that level implies support for lower levels.

ojw28 (commented):
Please see the inline comments. If these can be addressed then we will get this merged. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
zeninsta: Good catch. Changed to use CodecProfileLevel constant.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
zeninsta: Yes getProfileLevels() returns empty array. Changed to check the returned list from getProfileLevels() as well. Updated the title and comment as well.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
zeninsta: Changed to return the highest level only.

ojw28:
Thanks! We will get this merged in.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -572,6 +581,49 @@ public boolean isAudioChannelCountSupportedV21(int channelCount) {
...
+  /**
+   * On versions M and below, VP9 codecCapabilities do not advertise profile level
+   * support. In this case, estimate the level from MediaCodecInfo.VideoCapabilities
+   * instead. For more information, consult
+   * https://developer.android.com/reference/android/media/MediaCodecInfo.CodecProfileLevel.html
+   */
+  private static CodecProfileLevel[] getVp9CodecProfileLevelsV23(CodecCapabilities capabilities) {
+    // https://www.webmproject.org/vp9/levels
+    final int[][] bitrateMapping = {
+        {180000, CodecProfileLevel.VP9Level52},
```
ojw28, line 593: I think the bitrate values here need to be 1000x larger, since the table at `https://www.webmproject.org/vp9/levels` is in `x1000 bit/s`.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -572,6 +581,49 @@ public boolean isAudioChannelCountSupportedV21(int channelCount) {
...
+      if (videoCapabilities.getBitrateRange().contains(bitrate)) {
+        CodecProfileLevel profileLevel = new CodecProfileLevel();
+        // Assume all platforms before N only support VP9 profile 0.
+        profileLevel.profile = CodecProfileLevel.VP9Profile0;
+        profileLevel.level = level;
+        return new CodecProfileLevel[] { profileLevel };
+      }
+    }
+
+    return new CodecProfileLevel[0];
```
ojw28, line 624: I think it would make sense to always assume `VP9Level1` is supported.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
```diff
@@ -572,6 +581,49 @@ public boolean isAudioChannelCountSupportedV21(int channelCount) {
...
+        {3600, CodecProfileLevel.VP9Level21},
+        {1800, CodecProfileLevel.VP9Level2},
+        {800, CodecProfileLevel.VP9Level11},
+        {200, CodecProfileLevel.VP9Level1},
+    };
+
+    VideoCapabilities videoCapabilities = capabilities.getVideoCapabilities();
+
+    if (videoCapabilities == null) {
+      return new CodecProfileLevel[0];
```
ojw28, line 609: I think it would make sense to always assume `VP9Level1` is supported.

ojw28 (commented):
See some more inline comments. In particular, I think the bitrate values are incorrect.

ojw28:
Hi. I merged this with the bitrate adjustments and a few other tweaks. You can view and check the final version [here](https://github.com/google/ExoPlayer/commit/c40d1c6620afa33134dc15195fe97ac800c95437). Please let us know if there are any issues, and thanks for your contribution!

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecInfo.java
zeninsta: Ah, thanks for catching this! So this is not correct for chromium media either...

ojw28: It looks to me like both the bitrate constants _and_ the profile level constants are wrong in the Chromium code you reference. I could be wrong, but it looks like it just doesn't do the right thing, as far as I can see!

PR #8490, opened 2021-01-21
Add support for SSA (V4+) PrimaryColour style

szaboa:
@icbaker This is an initial PR for SSA (V4+) PrimaryColour style.

A question in general, just to make sure we start this style support in the right way:

- Parsing of the style attributes should be done in the `SsaStyle` so then this class will have all the styles prepared to be used directly from the `SsaDecoder`, or the `SsaStyle` should be just a holder and the parsing needs to be done on the decoder side?

Open point:
- Not sure what is the code of the the primary colour style **override**, as the spec on this from [Matroska](http://moodub.free.fr/video/ass-specs.doc) is pretty confusing, and it is not aligned with [this](https://fileformats.fandom.com/wiki/SubStation_Alpha) other spec. I'll try to figure it out by taking VLC for reference.

### DIFF
--- a/demos/main/src/main/assets/media.exolist.json
+++ b/demos/main/src/main/assets/media.exolist.json
@@ -507,6 +507,13 @@
         "subtitle_language": "en"
       },
       {
+        "name": "SubStation Alpha colors",
+        "uri": "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-avc-baseline-480.mp4",
+        "subtitle_uri": "https://drive.google.com/uc?export=download&id=13EdW4Qru-vQerUlwS_Ht5Cely_Tn0tQe",
+        "subtitle_mime_type": "text/x-ssa",
+        "subtitle_language": "en"
+      },
+      {
         "name": "MPEG-4 Timed Text",
         "uri": "https://storage.googleapis.com/exoplayer-test-media-1/mp4/dizzy-with-tx3g.mp4"
       }
--- a/demos/main/src/main/assets/media.exolist.json
+++ b/demos/main/src/main/assets/media.exolist.json
@@ -507,13 +507,6 @@
         "subtitle_language": "en"
       },
       {
-        "name": "SubStation Alpha colors",
-        "uri": "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-avc-baseline-480.mp4",
-        "subtitle_uri": "https://drive.google.com/uc?export=download&id=13EdW4Qru-vQerUlwS_Ht5Cely_Tn0tQe",
-        "subtitle_mime_type": "text/x-ssa",
-        "subtitle_language": "en"
-      },
-      {
         "name": "MPEG-4 Timed Text",
         "uri": "https://storage.googleapis.com/exoplayer-test-media-1/mp4/dizzy-with-tx3g.mp4"
       }
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
@@ -21,10 +21,12 @@ import static java.lang.annotation.RetentionPolicy.SOURCE;
 
 import android.graphics.PointF;
 import android.text.TextUtils;
+import androidx.annotation.ColorInt;
 import androidx.annotation.IntDef;
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.util.Assertions;
+import com.google.android.exoplayer2.util.ColorParser;
 import com.google.android.exoplayer2.util.Log;
 import com.google.android.exoplayer2.util.Util;
 import java.lang.annotation.Documented;
@@ -83,12 +85,16 @@ import java.util.regex.Pattern;
   public static final int SSA_ALIGNMENT_TOP_CENTER = 8;
   public static final int SSA_ALIGNMENT_TOP_RIGHT = 9;
 
+  public static final int SSA_COLOR_UNKNOWN = -1;
+
   public final String name;
   @SsaAlignment public final int alignment;
+  @ColorInt public int primaryColor;
 
-  private SsaStyle(String name, @SsaAlignment int alignment) {
+  private SsaStyle(String name, @SsaAlignment int alignment, @ColorInt int primaryColor) {
     this.name = name;
     this.alignment = alignment;
+    this.primaryColor = primaryColor;
   }
 
   @Nullable
@@ -105,7 +111,9 @@ import java.util.regex.Pattern;
     }
     try {
       return new SsaStyle(
-          styleValues[format.nameIndex].trim(), parseAlignment(styleValues[format.alignmentIndex]));
+          styleValues[format.nameIndex].trim(),
+          parseAlignment(styleValues[format.alignmentIndex]),
+          parsePrimaryColor(styleValues[format.primaryColorIndex]));
     } catch (RuntimeException e) {
       Log.w(TAG, "Skipping malformed 'Style:' line: '" + styleLine + "'", e);
       return null;
@@ -144,6 +152,16 @@ import java.util.regex.Pattern;
     }
   }
 
+  @ColorInt
+  private static int parsePrimaryColor(String primaryColorStr) {
+    try {
+      return ColorParser.parseSsaColor(primaryColorStr);
+    } catch (IllegalArgumentException ex) {
+      Log.w(TAG, "Failed parsing color value: " + primaryColorStr);
+    }
+    return SSA_COLOR_UNKNOWN;
+  }
+
   /**
    * Represents a {@code Format:} line from the {@code [V4+ Styles]} section
    *
@@ -154,11 +172,13 @@ import java.util.regex.Pattern;
 
     public final int nameIndex;
     public final int alignmentIndex;
+    public final int primaryColorIndex;
     public final int length;
 
-    private Format(int nameIndex, int alignmentIndex, int length) {
+    private Format(int nameIndex, int alignmentIndex, int primaryColorIndex, int length) {
       this.nameIndex = nameIndex;
       this.alignmentIndex = alignmentIndex;
+      this.primaryColorIndex = primaryColorIndex;
       this.length = length;
     }
 
@@ -171,6 +191,7 @@ import java.util.regex.Pattern;
     public static Format fromFormatLine(String styleFormatLine) {
       int nameIndex = C.INDEX_UNSET;
       int alignmentIndex = C.INDEX_UNSET;
+      int primaryColorIndex = C.INDEX_UNSET;
       String[] keys =
           TextUtils.split(styleFormatLine.substring(SsaDecoder.FORMAT_LINE_PREFIX.length()), ",");
       for (int i = 0; i < keys.length; i++) {
@@ -181,9 +202,14 @@ import java.util.regex.Pattern;
           case "alignment":
             alignmentIndex = i;
             break;
+          case "primarycolour":
+            primaryColorIndex = i;
+            break;
         }
       }
-      return nameIndex != C.INDEX_UNSET ? new Format(nameIndex, alignmentIndex, keys.length) : null;
+      return nameIndex != C.INDEX_UNSET
+          ? new Format(nameIndex, alignmentIndex, primaryColorIndex, keys.length)
+          : null;
     }
   }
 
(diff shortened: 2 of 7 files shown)


library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
```diff
@@ -83,12 +85,16 @@
   public static final int SSA_ALIGNMENT_TOP_CENTER = 8;
   public static final int SSA_ALIGNMENT_TOP_RIGHT = 9;
 
+  public static final int SSA_COLOR_UNKNOWN = -1;
```
icbaker: Unfortunately I don't think you can do this, because the Android ColorInt representation uses all 32 bits to represent the 4 ARGB channels (8 bits each), so there are no 'reserved' invalid values.

Specifically -1 is 0xFFFFFFFF, i.e. fully transparent black - which means if someone writes a SSA file with a format line with primaryColour="&HFFFFFFFF" then your code in SsaDecoder will ignore this (thinking it's unset).

Arguably 100% transparent colors are silly, but i think it's still confusing to arbitrarily declare a valid part of the color-space to be our token 'invalid' value.

I think you probably need to track "primary color set" as a separate boolean - see `Cue.windowColor` and `windowColorSet` for somewhere we've had to do a similar thing.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
```diff
@@ -181,9 +202,14 @@ public static Format fromFormatLine(String styleFormatLine) {
           case "alignment":
             alignmentIndex = i;
             break;
+          case "primarycolour":
+            primaryColorIndex = i;
```
icbaker, line 235: [non-actionable comment]

Speaking as a British programmer who has learnt to misspell `colour` as `color` whenever coding, I enjoy we end up with both versions here.

I think what you've done makes perfect sense (name the field color).

library/core/src/main/java/com/google/android/exoplayer2/util/ColorParser.java
```diff
@@ -67,6 +67,27 @@ public static int parseCssColor(String colorExpression) {
...
+  /**
+   * Parses a SSA V4+ color expression.
+   *
+   * @param colorExpression The color expression.
+   * @return The parsed ARGB color.
+   */
+  @ColorInt
+  public static int parseSsaColor(String colorExpression) {
+    // SSA V4+ color format is &HAABBGGRR.
+    if (colorExpression.length() != 10 || !"&H".equals(colorExpression.substring(0, 2))) {
```
icbaker: nit: It's probably cheaper to do colorExpression.startsWith("&H") because it saves allocating the returned substring each time.

imo it's also more concise and more readable

icbaker (commented):
Thanks for the PR - looks like a great start! I added some specific comments in the code, but only after that did I spot this part of the moodub spec (emphasis mine):
> Field 4:	PrimaryColour. A **long integer BGR (blue-green-red)** value. ie. the byte order in the hexadecimal equivelent of this number is BBGGRR

That sounds like these values will actually be (64-bit?) numbers. Combined with this line:

> Field 4-7:	The color format contains the alpha channel, too. (AABBGGRR)

This seems weird. Does this mean that in v4 there are 3 colors (B G R) and each one has 64/3=21 bits (with 3 unused bits)?

It makes a bit more sense in v4+ where it's clear that there are 4 colors, so each one can have 64/4 = 16 bits.

And then the fileformats.fandom spec you linked to has **no** mention of this numeric representation at all and just says this:
> Color values are expressed in hexadecimal BGR format as &HBBGGRR& or ABGR (with alpha channel) as &HAABBGGRR&. Transparency (alpha) can be expressed as &HAA&. Note that in the alpha channel, 00 is opaque and FF is transparent.

I wonder if to be conservative we need to handle both? And presumably also be robust against missing leading zeroes (same as for the override format below).

So I think we should handle **both** files like this:
```
[V4+ Styles]
Format: Name, PrimaryColour
Style: Default,&H00FFFFFF
```

And like this:
```
[V4+ Styles]
Format: Name, PrimaryColour
Style: Default,16777215
```

-----

> * Parsing of the style attributes should be done in the `SsaStyle` so then this class will have all the styles prepared to be used directly from the `SsaDecoder`, or the `SsaStyle` should be just a holder and the parsing needs to be done on the decoder side?

Since the alignment parsing logic is already inside SsaStyle, let's keep following that pattern - so what you've done (add new static methods) is perfect.

> * Not sure what is the code of the the primary colour style **override**, as the spec on this from [Matroska](http://moodub.free.fr/video/ass-specs.doc) is pretty confusing, and it is not aligned with [this](https://fileformats.fandom.com/wiki/SubStation_Alpha) other spec. I'll try to figure it out by taking VLC for reference.

I think the two specs you linked are aligned. It seems the number before the `c` is optional - although neither spec seems to make it clear what it means if it's missing - presumably it means the foreground color? That would seem like a sensible default to me. The moodub doc says "\alpha defaults to \1a" so I think we can extrapolate that "\c defaults to \1c"

So you can either use `{\c...}` or `{\1c...}`

There's some trickiness in the moodub doc which I think we should handle:
>**Leading zeroes are not required.** e.g.
>* `{\c&amp;HFF&amp;}` This is pure, full intensity red
>* `{\c&amp;HFF00&amp;}` This is pure, full intensity Green
>* `{\c&amp;HFF0000&amp;}` This is pure, full intensity Blue
>* `{\c&amp;HFFFFFF&amp;}` This is White
>* `{\c&amp;HA0A0A&amp;}` This is dark grey
 

It seems reasonable to split the Format line support away from the override support, so each PR can stay small and focussed. Do you want to add a new column to your table indicating whether ExoPlayer supports the property in Format lines vs in overrides? Then imo let's keep this PR focussed on the Format line, and once we've got that hammered out you can then raise a second one to add override support.

szaboa:
> I wonder if to be conservative we need to handle both? And presumably also be robust against missing leading zeroes (same as for the override format below).

Yep, let's handle both of them.

> Do you want to add a new column to your table indicating whether ExoPlayer supports the property in Format lines vs in overrides? Then imo let's keep this PR focussed on the Format line, and once we've got that hammered out you can then raise a second one to add override support.

Yep, I'll add a new column to the table for the overrides, and agree with you that we should first focus on the Format line.

I'll update the PR when everything is ready :)

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
szaboa: Yeah, I missed it.. big time :) I'll fix it similarly how the `windowColorSet` is solved.

szaboa:
Handling both the hex- and decimal format of the colors has been added. Note that the original alpha value needs to be inverted, because 0xFF indicates full transparency in case of SSA (V4+) - this is based on the tests **performed on VLC** and also fileformats.fandom specs states this. And because we decided to handle the leading zeros, the V4 hex colors are also supported out of the box.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
szaboa: In the end I've decided to go with a dedicated `SsaColor` class which holds this state, because we'll have multiple color attributes in the style which would lead to having a lot of `xyColorSet` flags. What do you think?

demos/main/src/main/assets/media.exolist.json
```diff
@@ -506,6 +506,13 @@
         "subtitle_mime_type": "text/x-ssa",
         "subtitle_language": "en"
       },
+      {
+        "name": "SubStation Alpha colors",
+        "uri": "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-avc-baseline-480.mp4",
+        "subtitle_uri": "https://drive.google.com/uc?export=download&id=13EdW4Qru-vQerUlwS_Ht5Cely_Tn0tQe",
+        "subtitle_mime_type": "text/x-ssa",
+        "subtitle_language": "en"
+      },
```
icbaker: I think we probably won't add this to the demo app, and instead update the test-subs-position.ass file above to include a range of SSA/ASS features.

If you'd like to have a go at crafting that file (would probably make sense to include features we don't support yet too) that would be cool - you can send a standalone PR for it and I'll upload it to the storage bucket.

This colors-only file you've crafted is still handy - I think we should turn it into automated test data in this PR. You can put it here: 
https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/ssa

And then add a test using it here:
https://github.com/google/ExoPlayer/blob/dev-v2/library/core/src/test/java/com/google/android/exoplayer2/text/ssa/SsaDecoderTest.java

That will help make sure future editors don't mess up the careful bit twiddling :)

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
```diff
@@ -144,6 +150,35 @@ private static boolean isValidAlignment(@SsaAlignment int alignment) {
...
+
+  /**
+   * Represents an SSA V4+ style color in ARGB format.
+   */
+  /* package */ static final class SsaColor {
+
+    public static SsaColor UNSET = new SsaColor(0, false);
+
+    public final @ColorInt int value;
+    public final boolean isSet;
```
icbaker: (I'm not sure this is a good idea, just a thought)

What about making these private and having an `int getColor()` method that throws an exception if `isSet == false`? There would also be a `boolean isSet()` method. This is similar to `Optional.get()`.

(in fact this class probably wouldn't be needed if we could use Java 8's `OptionalInt`, but I think our minSdk level is still too low for that)

library/core/src/main/java/com/google/android/exoplayer2/util/ColorParser.java
```diff
@@ -67,6 +67,45 @@ public static int parseCssColor(String colorExpression) {
...
+      // Parse color from hex format (&HAABBGGRR).
+      if (colorExpression.startsWith("&H")) {
+        StringBuilder rgbaStringBuilder = new StringBuilder(colorExpression);
+        if (colorExpression.length() < 10) {
+          // Add leading zeros if necessary.
+          while (rgbaStringBuilder.length() != 10) {
+            rgbaStringBuilder.insert(2, "0");
+          }
+        }
+        abgr = (int) Long.parseLong(colorExpression.substring(2), 16);
```
icbaker: should this use rgbaStringBuilder?

icbaker (approved):
Looks good! If you can add the testdata + test then I'll work on getting this merged.

Thanks for this, it's great :)

library/core/src/main/java/com/google/android/exoplayer2/util/ColorParser.java
szaboa: of course it should use that, sorry it was too late yesterday :) I'll fix it.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
szaboa: I like this idea, I've changed the code the way you described.

demos/main/src/main/assets/media.exolist.json
szaboa: Removed the entry from media.exolist.json. And of course, I can try to craft a file with more features - but I would wait a bit until we add 1-2 more style support just to get more familiar with it :)

szaboa:
Testdata + test added, ready for review.

icbaker:
> > Field 4:	PrimaryColour. A **long integer BGR (blue-green-red)** value. ie. the byte order in the hexadecimal equivelent of this number is BBGGRR
> 
> That sounds like these values will actually be (64-bit?) numbers. Combined with this line:
> 
> > Field 4-7:	The color format contains the alpha channel, too. (AABBGGRR)
> 
> This seems weird. Does this mean that in v4 there are 3 colors (B G R) and each one has 64/3=21 bits (with 3 unused bits)?
> 
> It makes a bit more sense in v4+ where it's clear that there are 4 colors, so each one can have 64/4 = 16 bits.

I've spent a bit more time thinking about this, and playing with your sample file in VLC (thanks!). I think the "long integer" doesn't mean 64 bits, it means an unsigned 32-bit integer. And (obviously?) in the 3-color case only the lowest 24 bits are used (top 8 bits are just always zero).

Your parsing code seems correct but I realised a couple of things:
1. You don't need to manually add zeroes to the string before parsing, they're implicit when you parse to int (i.e. `Long.parseLong("123") == Long.parseLong("000123")`, and equally `Long.parseLong("ABC", 16) == Long.parseLong("000ABC", 16)`
2. It's probably safer to just use a long for storing the bits, to avoid hitting any weird java behaviour around automatic propagation of negative values if the 32nd bit of the input is high.

So I'm going to make those tweaks as I merge it.

szaboa:
> I've spent a bit more time thinking about this, and playing with your sample file in VLC (thanks!). I think the "long integer" doesn't mean 64 bits, it means an unsigned 32-bit integer. And (obviously?) in the 3-color case only the lowest 24 bits are used (top 8 bits are just always zero).

Yes, I came to the same conclusion when I was testing out different color representations in VLC. 

 > You don't need to manually add zeroes to the string before parsing

You are right, good catch :)

PR #8546, opened 2021-02-04
Add option to add playback speed button and audio track selection button via xml

TadejZupancic:
So far only settings button existed that opened a submenu with the playback speed button and audio track selection button. If you need to control e.g. only playback speed, then it makes sense to provide a direct way of accessing it instead of going inside a submenu. You can use the following xml to add the buttons:
 ```
<ImageButton android:id="@+id/exo_playback_speed" style="@style/ExoStyledControls.Button.Bottom.PlaybackSpeed" />
<ImageButton android:id="@+id/exo_audio_track" style="@style/ExoStyledControls.Button.Bottom.AudioTrack" />
```


ojw28:
Thanks for sending this. What you're trying to do seems fairly reasonable, although I think it's problematic that the menu always pops up at the far bottom left. If the custom button is somewhere else in the UI then this is fairly confusing, and in the worst case the user might not even notice it. How hard would it be to make the popup appear positioned correctly relative to the button?

TadejZupancic:
Thanks for noticing that. I tested the buttons on the same position as the settings button and also the settings button opened the popup always on the bottom right side regardless of its position. Now it should open above the button or below it if the button is somewhere at the top.

ojw28:
> Thanks for noticing that. I tested the buttons on the same position as the settings button and also the settings button opened the popup always on the bottom right side regardless of its position. Now it should open above the button or below it if the button is somewhere at the top.

Sorry for the back-and-forth, but we merged the pull request without this additional fix in the end. It was subtly but negatively changing where the menus were being displayed in the regular case. It didn't seem worth investing more time trying to improve it, given you don't need it. It can be left for future work should anyone ever need the menus to pop-up somewhere significantly different. You can see the end result here: https://github.com/google/ExoPlayer/commit/6be3a59354635a633d597576f011934cb1c70425.

I also submitted some simplification first, which you can see here: https://github.com/google/ExoPlayer/commit/70f831a6470be10b25ea6ee7110c7e6b23dfc9fe.

Thanks for the contribution!

TadejZupancic:
What was the problem with the position of the menus? Vertical position? I moved it to not cover the button since it was set that the second click on the button should close it. And if it was over the button, that could not happen. I can fix it if you tell me, what was wrong.

ojw28:
If I remember correctly, it moved the menus slightly to the right and down compared to where they were before.

More generally, I had some second thoughts about whether positioning relative to the button would be optimal. It seems fairly clear that if the button is all the way on the left hand side, having the menu pop-up on the right is pretty confusing, and the change you made would be clearly preferable for this case. However, for the more common case (including the default case) where buttons with pop-up menus are all toward the bottom right, there was something nice about having the menu pop-up in the same place for all of the buttons, at a position chosen by our UX team, and so the change may have been marginally negative. Given you didn't need it to pop-up elsewhere, that we'd likely want UX input if we were to change it, and that we don't have a lot of spare capacity time-wise to do that, it seemed better to leave it as-is for the time-being.

PR #8589, opened 2021-02-15
let evictor change cache max bytes

halaei:
This PR lets the app's users change cache size at runtime. Also it lets applications to decrease cache size when low storage space is detected.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/LeastRecentlyUsedCacheEvictor.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/LeastRecentlyUsedCacheEvictor.java
@@ -21,7 +21,7 @@ import java.util.TreeSet;
 /** Evicts least recently used cache files first. */
 public final class LeastRecentlyUsedCacheEvictor implements CacheEvictor {
 
-  private final long maxBytes;
+  private long maxBytes;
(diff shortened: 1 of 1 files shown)


halaei:
@googlebot I signed it!

ojw28:
It doesn't seem important that we support this as a core feature in the library, since applications can easily inject their own `CacheEvictor` if they need this functionality. I'm also somewhat unclear on the thread safety of the change being made.

Closing, since application developers already have a straightforward way to implement this if they need it.

halaei:
Thanks for @ojw28 for expressing your doubts about thread safety. It really makes me wonder how I should implement such a feature without causing threading issues. I guess my code so far is OK, but somehow I should choose the right thread to call `changeMaxBytes()`. Do you know on what thread should I run my code and how to access that thread?

ojw28:
Thanks for prompting me to take a bit of a closer look at this; it's actually a little tricky :). What you've done is currently safe if you make sure to synchronize on your `SimpleCache` when calling the custom method. Whilst this works at the moment, it's pretty fragile and liable to break. I can't think of a way of achieving thread safety that's both safe and robust, given what we currently provide. I've filed https://github.com/google/ExoPlayer/issues/8628 to add a mechanism for updating custom `CacheEvictor` components in a safe way.

PR #8653, opened 2021-03-01
Support tts:textEmphasis in TTML parser and WebView output

dlafayet:


library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
```diff
@@ -0,0 +1,201 @@
...
+
+    // Handle "auto" or unknown value
+    // If an implementation does not recognize or otherwise distinguish an emphasis style value,
+    // then it must be interpreted as if a style of auto were specified; as such, an implementation
+    // that supports text emphasis marks must minimally support the auto value.
+    return new TextEmphasis(MARK_AUTO, POSITION_UNKNOWN);
+  }
+
+  private static @Nullable
+  TextEmphasis handleTwoNodes(@NonNull String mark, @NonNull String position) {
```
icbaker: I don't think you can assume this ordering (mark always coming before position), based on my reading of the TTML spec syntax:

> a double vertical bar (||) separates two or more options, of which one or more must appear in any order without duplication, separated by <lwsp>;

https://www.w3.org/TR/2018/REC-ttml2-20181108/#conventions

Based on the `text-emphasis`, `emphasis-style` and `emphasis-position` sections I think these orders are possible:
```
"before circle filled"
"before filled circle"
"circle filled before"
```
In fact I think the only permutations that aren't possible are ones that split the style, e.g.:

```
"circle before filled"
"filled before circle"
```

Does that match your understanding of the spec or have I misread it?

If you agree then I think we probably need to slightly re-think the approach here - maybe by doing `.contains()` checks on the whole list of words? In fact, if we had Sets of possible words  for each 'type' then we could perhaps use Guava's `Sets.intersection()` to find matches - not sure!

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
```diff
@@ -106,6 +107,19 @@
...
+  public static final String TEXT_EMPHASIS_AUTO = "auto";
+  public static final String TEXT_EMPHASIS_MARK_DOT = "dot";
+  public static final String TEXT_EMPHASIS_MARK_SESAME = "sesame";
+  public static final String TEXT_EMPHASIS_MARK_CIRCLE = "circle";
+  public static final String TEXT_EMPHASIS_MARK_FILLED = "filled";
+  public static final String TEXT_EMPHASIS_MARK_OPEN = "open";
+
+  public static final String TEXT_EMPHASIS_POSITION_AFTER = "after";
+  public static final String TEXT_EMPHASIS_POSITION_BEFORE = "before";
+  public static final String TEXT_EMPHASIS_POSITION_OUTSIDE = "outside";
```
icbaker: WDYT about merging these with RUBY_BEFORE and RUBY_AFTER?

Since they're defined in the same place in the spec:
https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-value-annotation-position

library/core/src/main/java/com/google/android/exoplayer2/text/span/TextEmphasisSpan.java
```diff
@@ -0,0 +1,134 @@
...
+  public @interface Position {
+
+  }
+
+  /**
+   * The text emphasis position is unknown.
+   */
+  public static final int MARK_UNKNOWN = -1;
+
+  public static final int MARK_AUTO = 1;
```
icbaker: In my mental model of ExoPlayer's subtitle parsing logic the `Cue` representation (and styling Spans) are independent of the many different format/specs.

This `auto` concept (and the way it should map depending on writing direction) seems to be TTML-specific - so I don't think it should have an entry in this IntDef defined on the Span.

I think the `auto` concept should only exist inside the TTML parsing code - and by the time it spits out a Cue with styling spans, the necessary resolution has already happened.

Otherwise a `TextOutput` receiving a Cue with `TextEmphasisSpan` with `MARK_AUTO` won't really know what to do with it (not knowing if it comes from TTML or elsewhere...).

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
```diff
@@ -674,6 +676,105 @@ public void rubies() throws IOException, SubtitleDecoderException {
     assertThat(sixthCue).hasNoRubySpanBetween(0, sixthCue.length());
   }
 
+  @Test
+  public void textEmphasis() throws IOException, SubtitleDecoderException {
+    TtmlSubtitle subtitle = getSubtitle(TEXT_EMPHASIS_FILE);
+
+    Spanned firstCue = getOnlyCueTextAtTimeUs(subtitle, 10_000_000);
+    assertThat(firstCue)
+        .hasNoTextEmphasisSpanBetween("None ".length(), "None おはよ".length());
```
icbaker: nit: It's probably more robust to just use `hasNoSpans` here?

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
```diff
@@ -674,6 +676,105 @@ public void rubies() throws IOException, SubtitleDecoderException {
...
+
+    Spanned secondCue = getOnlyCueTextAtTimeUs(subtitle, 20_000_000);
+    assertThat(secondCue)
+        .hasTextEmphasisSpanBetween("Auto ".length(), "Auto ございます".length())
+        .withMarkAndPosition(TextEmphasisSpan.MARK_FILLED_CIRCLE,
+            TextEmphasisSpan.POSITION_UNKNOWN);
```
icbaker: Shouldn't this be `POSITION_OUTSIDE`?

>If no emphasis position is specified, then the emphasis position must be interpreted as if a position of outside were specified.

https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-textEmphasis

testdata/src/test/assets/media/ttml/text_emphasis.xml
```diff
@@ -0,0 +1,65 @@
+<tt xmlns:ttm="http://www.w3.org/2006/10/ttaf1#metadata"
+  xmlns:ttp="http://www.w3.org/2006/10/ttaf1#parameter"
+  xmlns:tts="http://www.w3.org/2006/10/ttaf1#style"
+  xmlns="http://www.w3.org/ns/ttml">
+  <head>
+    <region xml:id="region1" tts:extent="80.000% 80.000%" tts:origin="10.000% 10.000%" tts:writingMode="tbrl"/>
+    <region xml:id="region2" tts:extent="80.000% 80.000%" tts:origin="10.000% 10.000%" tts:writingMode="tblr"/>
+    <region xml:id="region3" tts:extent="80.000% 80.000%" tts:origin="10.000% 10.000%" tts:writingMode="tb"/>
+    <region xml:id="region4" tts:extent="80.000% 80.000%" tts:origin="10.000% 10.000%" tts:writingMode="lr"/>
```
icbaker: Since these regions are really only used to change the writing direction, do you think it makes sense to name them to reflect that?

e.g. `region_tbrl`, `region_tblr` etc.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SpannedToHtmlConverter.java
```diff
@@ -197,6 +198,14 @@ private static String getOpeningTag(Object span, float displayDensity) {
       }
     } else if (span instanceof UnderlineSpan) {
       return "<u>";
+    } else if (span instanceof TextEmphasisSpan) {
+      TextEmphasisSpan textEmphasisSpan = (TextEmphasisSpan) span;
+      String style = getTextEmphasisStyle(textEmphasisSpan.mark);
+      String position = getTextEmphasisPosition(textEmphasisSpan.position);
+      return Util
+          .formatInvariant("<span style='-webkit-text-emphasis-style: %s; text-emphasis-style: %s; "
+                  + "-webkit-text-emphasis-position: %s; text-emphasis-position: %s;'>",
```
icbaker: (I'm not a web developer, apologies if this is a silly question)

If you know that this is always going to be used in a WebView, do you need to use the `-webkit` prefixed styles as well?

This table suggests `text-emphasis-style` (I assume without the webkit prefix) is supported from 4.4 (i.e. KitKat):
https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style#browser_compatibility

Is this to make things also work on Jelly Bean (ExoPlayer's min supported version)?

I'm mostly asking in case we should add the same prefix for other CSS in this file...

testutils/src/main/java/com/google/android/exoplayer2/testutil/truth/SpannedSubject.java
```diff
@@ -1110,4 +1152,95 @@ public String toString() {
...
+        if (o == null || getClass() != o.getClass()) {
+          return false;
+        }
+
+        TextEmphasisSubject.MarkAndPosition that = (TextEmphasisSubject.MarkAndPosition) o;
+        return (position == that.position) && (mark == that.mark);
+      }
+
+      @Override
+      public int hashCode() {
```
icbaker, line 1243: I think we can remove this and equals() - not sure it really adds much for a Truth subject implementation

icbaker (commented):
Thanks for sending this - the parsing in particular looks really tricky! I've added some comments

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: Agreed with your interpretation. Reworking the implementation using Guava's `Sets.intersection()`. Great suggestion!

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
dlafayet: Fixing.

library/core/src/main/java/com/google/android/exoplayer2/text/span/TextEmphasisSpan.java
dlafayet: Agreed with the mental model. I'll refactor this part.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SpannedToHtmlConverter.java
dlafayet: The way they wrote it in the documentation is a little confusing but in the legend section, you'll see that it says
'-' or 'x-' - requires a vendor prefix or different name to use. 

If you try this code on the chrome browser, you'll see that the text emphasis is not rendered correctly on Chrome without the webkit prefix, but renders correctly on Firefox or Safari.

```
<body>
  <span style='text-emphasis: filled double-circle blue;'>Text emphasis</span>
</body>
```

testutils/src/main/java/com/google/android/exoplayer2/testutil/truth/SpannedSubject.java
dlafayet: Actually, these are required to avoid checking for instance equality using Truth.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SpannedToHtmlConverter.java
icbaker: Got it, thanks!

library/core/src/main/java/com/google/android/exoplayer2/text/span/TextEmphasisSpan.java
```diff
@@ -0,0 +1,84 @@
...
+  public static final int MARK_FLAG_FILLED = 0;
+  public static final int MARK_FLAG_OPEN = 4;
+
+  // Below are the mark style constants
+  public static final int MARK_FILLED_CIRCLE = MARK_FLAG_CIRCLE | MARK_FLAG_FILLED;
+  public static final int MARK_FILLED_DOT = MARK_FLAG_DOT | MARK_FLAG_FILLED;
+  public static final int MARK_FILLED_SESAME = MARK_FLAG_SESAME | MARK_FLAG_FILLED;
+  public static final int MARK_OPEN_CIRCLE = MARK_FLAG_CIRCLE | MARK_FLAG_OPEN;
+  public static final int MARK_OPEN_DOT = MARK_FLAG_DOT | MARK_FLAG_OPEN;
+  public static final int MARK_OPEN_SESAME = MARK_FLAG_SESAME | MARK_FLAG_OPEN;
```
icbaker: I appreciate the bit twiddling cleverness :) But I think it's best to be able to use @IntDefs without relying on the underlying values at all, which suggests to me this should maybe be two separate defs, one for the shape and one for the fill-style?

Maybe `@MarkShape` and `@MarkFill`.

-----

This is slightly awkward in the TTML-specific code because the syntax in the spec allows `auto` **or** `( "filled" | "open" ) || ( "circle" | "dot" | "sesame" )`, so it's nice to have a single value that covers both 'filled' and 'circle', but imo we shouldn't let that awkwardness influence the API decisions here.

I think you can get away with adding a custom "auto" value to a TTML-specific `@MarkShape` int-def, and in that case ignoring whatever `@MarkFill` ends up being set to. It's not ideal - but imo it's better to hide the slight messiness inside the TTML-parsing logic than in this API surface.

----

On a similar subject: WDYT about representing "none" as a span with `mark=none` instead of no span? It feels slightly nicer and less magic to me to try and match the TTML representation as closely as possible (instead of mapping none -> null). It might make a difference if considering how nested `<span tts:textEmphasis="...">` nodes will interact, if the outer one sets a value but the inner one sets `none` then mapping `none -> null` might end up rendering differently.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
```diff
@@ -0,0 +1,189 @@
...
+          break;
+        case TtmlNode.TEXT_EMPHASIS_MARK_SESAME:
+          mark |= TextEmphasisSpan.MARK_FLAG_SESAME;
+          break;
+        case TtmlNode.TEXT_EMPHASIS_MARK_CIRCLE:
+        default:
+          mark |= TextEmphasisSpan.MARK_FLAG_CIRCLE;
+      }
+    } else {
+      mark |= TextEmphasisSpan.MARK_FLAG_CIRCLE;
```
icbaker: This should be `AUTO`, no?

> If an implementation does not recognize or otherwise distinguish an emphasis style value, then it must be interpreted as if a style of auto were specified; as such, an implementation that supports text emphasis marks must minimally support the auto value.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
```diff
@@ -0,0 +1,189 @@
...
+     *  <a href="https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-textEmphasis">tts:textEmphasis</a>
+     */
+    @Position int position = POSITION_OUTSIDE;
+    if (positionSet.size() == 1) {
+      switch ((String) positionSet.iterator().next()) {
+        case TtmlNode.ANNOTATION_POSITION_AFTER:
+          position = TextAnnotation.POSITION_AFTER;
+          break;
+        case TtmlNode.ANNOTATION_POSITION_OUTSIDE:
+          position = POSITION_OUTSIDE;
```
icbaker, line 203: wdyt about just using BEFORE here and getting rid of the `POSITION_OUTSIDE` constant and custom IntDef?

(i.e. move the OUTSIDE -> BEFORE mapping logic from TtmlRenderUtil to here)

library/ui/src/main/java/com/google/android/exoplayer2/ui/SpannedToHtmlConverter.java
```diff
@@ -186,17 +188,25 @@ private static String getOpeningTag(Object span, float displayDensity) {
...
+              style, style, position, position);
```
icbaker: optional: you can use indexed substitution parameters to avoid needing to repeat style & position here. Something like (untested):

```
formatInvariant(
    "<span style='-webkit-text-emphasis-style: %1$s; text-emphasis-style: %1$s; "
        + "-webkit-text-emphasis-position: %2$s; text-emphasis-position: %2$s;'>",
    style,
    position);
```

testutils/src/main/java/com/google/android/exoplayer2/testutil/truth/SpannedSubject.java
icbaker: Ah misread this and thought it was on the subject not the inner MarkAndPosition - my bad! 🙈

library/core/src/main/java/com/google/android/exoplayer2/text/span/TextEmphasisSpan.java
dlafayet: I also went back and forth myself whether to split into two IntDefs or not. The other consideration I had was if in the future, someone wants to add support for custom marks, they would also have to ignore the `MarkFill` value. But I agree with your assessment that although it's not ideal, it's probably better to have straightforward API's where we don't rely on the enum values themselves. I will split them into two IntDefs.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: Good catch! I think depends on the number of valid arguments. i.e. "filled", vs. "invalid". Both would result in an empty `markSet`. The former should be set to `circle`, while the latter should be set to `auto`

> If only filled or open is specified, then it is equivalent to filled circle and open circle, respectively.

dlafayet: This gets a little tricky. Consider the following if the user makes a typo:
`filed circle` -> equivalent to "auto" or "circle"?
`open sesamee` -> equivalent to "auto" or "open"?
Based on the wording of the spec, one could conclude that this is an unrecognized emphasis style value, and should be treated as "auto". What do you think?

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: I think it's better to preserve the node specification here. Logic can be added to `TtmlRenderUtil` to determine the line positioning of the span and evaluate `OUTSIDE` -> BEFORE` | `AFTER`.

library/core/src/main/java/com/google/android/exoplayer2/text/span/TextEmphasisSpan.java
dlafayet: Agreed on the 3rd point regarding `mark=none`. Will update.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: On further thought, there is no way to know whether the invalid attribute was intended to be a style value or position value or color (not supported) value. This is my revised position on this:
"filed circle" -> equivalent to "circle"
"open sesamee" -> equivalent to "open"
wdyt?

icbaker: I don't think 'filed' can match a color, because they [either start with `#` or `rgb`](https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-value-color) or are one of the [19 predefined named colors](https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-value-named-color).

I assume the `<quoted-string>` part of `emphasis-style` would have **extra** quotes around it (as well as not being supported by your impl here)? i.e. something like `'foo' blue` - in which case `filed` can't be mistaken for a quoted string either.

So I think a rigid interpretation of the spec probably rejects `filed blue` completely as malformed (i.e. doesn't match anything). Not sure what the TTML spec expects in that case, imo it seems reasonable to reject the entire node and behave as if it never existed (probably logging a warning).

Implementing that with the existing set-based logic might be a bit tricky to do though (and arguably user-hostile - it's a tricky line between being over-precise and over-tolerant) - so it's also probably fine to treat it like `circle` as you propose. It doesn't seem worth spending ages handling a misspelt edge-case...

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private TextEmphasis(@MarkShape int shape, @TextEmphasisSpan.MarkFill int fill,
+      @TextAnnotation.Position int position) {
+    this.markShape = shape;
+    this.markFill = fill;
+    this.position = position;
+  }
+
+  @Override
+  public String toString() {
```
icbaker: Is this used? Wondering if I can remove it while merging...

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: I understand what you're saying that user-intent can be inferred even in misspelled attributes. What I am suggesting is that a typo on a non-quoted string will still be ambiguous. i.e.`b` could be intended for color, position or styling. 
In the example you gave, i.e. `filed blue`, my interpretation of the spec is that `auto` styling should apply. If we had an implementation that supported the color attribute, then the equivalent of `auto blue` should apply.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: It's not used. Will remove it.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
icbaker: > In the example you gave, i.e. `filed blue`, my interpretation of the spec is that `auto` styling should apply. If we had an implementation that supported the color attribute, then the equivalent of `auto blue` should apply.

Interesting - are you approaching this from the perspective of "the code should be written to allow the spec to evolve later"? Because with the **current** spec I don't see how you go from `filed blue` to `auto blue`. `filed` isn't an `emphasis-style`, an `emphasis-color` or an `emphasis-position` - right? That's why it seems to me the whole line should strictly be considered invalid.

If the code needs to assume there might be un-quoted `emphasis-style` names that were added to the spec since the code was written (or predefined color names I guess, bleh) - then I get your point about ambiguity. I think that's a reasonable & realistic stance, because of course specs evolve. And resolving unknown things to the `auto` behaviour seems like a pretty safe default, so I agree we should go with that.

Do you think it's worth adding some additional cues to `text_emphasis.xml` with some 'invalid' lines like this to show we thought about how they should resolve?

dlafayet:
One other thing I would like to point out is that the background color doesn't seem to be applied for text emphasis marks on the WebView output. I confirmed the same behavior on Firefox, Chrome and Safari.

Screenshot_1615227433

css-demo-text-emphasis

icbaker:
Ah, well spotted! I had the same problem with rubies because I didn't realise `background-color` doesn't [inherit](https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance) automatically, unlike most other CSS properties (did I mention I'm not a web developer?).

I fixed that by 'forcing' the inheritance using CSS: https://github.com/google/ExoPlayer/commit/cdfee89181b4d52a99c4375bac7b26f76b51da98

But that doesn't help here because there's no DOM element representing the emphasis marks apart from the span around the text.

So is it a rendering bug? It's clear the line height changes when adding a `text-emphasis` style property to a span (as in, the lines end up further apart) but when selecting the element in the chrome inspector only the 'base text' seems to be considered part of the span element. So what element are you supposed to use to cover the background of the emphasis marks?

Maybe we need to revisit the way I fixed the ruby problem and instead of forcing inheritance using CSS, we could try and draw a bounding box and set the background on that. Not immediately sure how to do that without JS, which I was hoping to avoid at all costs...

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: I wasn't approaching it from the perspective that spec might evolve later, at least not consciously. Let me see if I can explain it better.

> <text-emphasis>   : <emphasis-style> \|\| <emphasis-color> \|\| <emphasis-position>
--

The question I am debating is whether we consider the validity of each attribute (style, color, position) value independently? For example, "filed blue' has a valid color value, but an unspecified or invalid style or position value. Do we 1) disregard the entire description and either revert to `auto` or `none` or 2) apply the valid values but revert to defaults for the invalid or unspecified values?

For approach 1) we would be looking for values outside the set of valid values. 
`filed blue` would resolve to `auto` | `none`
`filled dot blue after foo` would resolve to `auto` | `none`
This is a more rigid interpretation of the spec, and less future-friendly if the specs evolve, requires full implementation of the spec (parsing color and quoted string styles) but more deterministic in that any invalid descriptions would always resolve to the same result.

For approach 2) we would be looking for values in the set of valid values and applying the ones we detect, and discarding the rest.
`filed blue` would resolve to `auto blue`
`filled dot blue after foo` would resolve to `filled dot blue after`
This is a looser interpretation of the spec, and is more future-friendly if the specs evolve, but less deterministic as partial features may be applied (i.e. color only but not style or position).

The current implementation takes approach 2), but after more consideration, I believe the correct interpretation of the spec should be 1), which I believe is your position already.

For reference, the IMSC JS parser http://sandflow.com/imsc1_1/index.html takes approach 1).
All three browsers (Firefox, Chrome, Safari) takes approach 1) based on the CSS attributes for `text-emphasis`.
All of the above implementations fallback to `none` if the description contains invalid values.

I'd love to hear your thoughts before I make any changes.

icbaker: Thanks for explaining further - I understand now.

Thanks for pointing out that my 'strict' proposal implicitly requires a **full** spec implementation, including quoted strings and colors - I hadn't fully appreciated that (though seems obvious in retrospect). Or at least the parser has to be aware of the full spec, even if it then throws away bits it can't implement and reverts them to `auto`.

The fact that the existing IMSC parser, and browsers for the CSS property, all take the strict approach does suggest we should ideally do the same. On a pragmatic level though, it seems pretty unlikely that approach 2 will be truly problematic...

I think there are three options here:
a. Merge this PR as-is with approach 2.
b. Change this PR to approach 1 and add full parse support for colors & quoted strings (even if they're not supported by TextEmphasisSpan or the HTML/CSS rendering yet).
c. Do (a) and then do (b) in a follow-up PR.

I'll leave it up to you - I'm happy to take it as-is with no plan to make it stricter in future, but also happy to work with you on either (b) or (c). Let me know what you prefer!

icbaker:
> One other thing I would like to point out is that the background color doesn't seem to be applied for text emphasis marks on the WebView output. I confirmed the same behavior on Firefox, Chrome and Safari.

I played around a bit and found that by adding `display: inline-block` to the `<span>` style it seemed to extend the background behind the emphasis marks - do you see any problems with that approach?

dlafayet:
> I played around a bit and found that by adding `display: inline-block` to the `<span>` style it seemed to extend the background behind the emphasis marks - do you see any problems with that approach?

Great find! I am also not a web developer 😆 but I think it's pretty safe to add it. 

Here are some screenshots using our test file. Just note that the inline-block area can be different based on the text emphasis marking used. In this file, each character has its own style span. 

In practice, most creators would probably use the same emphasis mark within a caption, so this might not be something that we would encounter typically.

inline-block-vert

inline-block

Design aesthetics aside, I think having the background color behind the mark is better for readability than not having it, so I'm all for adding it.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: I propose we do a) as it addresses the typical use cases for Japanese subtitles. Additionally, I'll add comments to explain the approach as well as the edge case examples in the tests as you suggested.

This doesn't preclude b) from happening in the future, but we can just make it low priority for now.

icbaker: Yep sounds good - I'll work on getting this merged. Thanks!

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
```diff
@@ -0,0 +1,216 @@
...
+    this.markFill = fill;
+    this.position = position;
+  }
+
+  @Nullable public static TextEmphasis createTextEmphasis(@Nullable String value) {
+    if (value == null) {
+      return null;
+    }
+
+    String parsingValue = value.toLowerCase().trim();
```
icbaker: While merging this some internal static analysis flagged that the behaviour of this `toLowerCase()` call depends on the locale of the device.

If we can assume this only contains ASCII then it recommends using Guava's [`Ascii.toLowerCase()`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Ascii.html#toLowerCase-java.lang.String-) instead - but that will look wrong if we support the 'quoted string' behaviour of `emphasis-style` in future, since I expect they generally won't be ASCII chars!

I think it will still work (non-ascii chars are copied without modification  by that function), but I'm inclined to remove the call and make this parsing case sensitive, so a future reader doesn't have to stop and check it's safe.

Reading the TTML spec I only see 2 explicit references to parsing things case-insensitively:
* when parsing hex strings: https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-value-hex-digit
* and when parsing color names: https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-value-named-color

So from that I infer it's reasonable to assume the rest of the spec is defined with case-sensitive parsing in mind.

That means removing/changing some of the tests you added exercising the case-insensitivity.

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
```diff
@@ -674,6 +677,120 @@ public void rubies() throws IOException, SubtitleDecoderException {
...
+    assertThat(firstCue).hasTextEmphasisSpanBetween("None ".length(), "None おはよ".length())
+        .withMarkAndPosition(TextEmphasisSpan.MARK_SHAPE_NONE,
+            TextEmphasisSpan.MARK_FILL_UNSPECIFIED,
+            TextAnnotation.POSITION_BEFORE);
+
+    Spanned secondCue = getOnlyCueTextAtTimeUs(subtitle, 20_000_000);
+    assertThat(secondCue)
+        .hasTextEmphasisSpanBetween("Auto ".length(), "Auto ございます".length())
+        .withMarkAndPosition(TextEmphasisSpan.MARK_SHAPE_CIRCLE,
+            TextEmphasisSpan.MARK_FILL_UNSPECIFIED,
```
icbaker: I think this should be `MARK_FILL_FILLED` right?

Since TTML should resolve it from `auto` -> `filled circle`. 

That means there's also a bit of logic missing in the auto-handling of `TtmlRenderUtil`. I'll change this test & fix the logic while merging - just flagging in case I've got something wrong.

Same in a few tests below.

dlafayet:
Thank you! 😁 🙏

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
dlafayet: Good catch! This was something I missed in the refactoring from one style enum to two enums for shape and fill. I have a fix.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TextEmphasis.java
dlafayet: Actually in `TtmlDecoder.parseStyleAttributes()`, it already converts the attribute value to lower case, so the `toLowerCase`  here is redundant.

```
      case TtmlNode.ATTR_TTS_TEXT_EMPHASIS:
          style = createIfNull(style).setTextEmphasis(
              TextEmphasis.createTextEmphasis(Util.toLowerInvariant(attributeValue)));
          break;
```

icbaker:
This is merged into dev-v2 - but GitHub didn't detect it for some reason - closing.

PR #8654, opened 2021-03-02
Add support for SSA Bold Italic style

abeljim:
@icbaker @szaboa
This is initial PR for Bold and Italic SSA style. I created a SSA file with most of the styles but it seems like VLC has a limit to the number of styles so I wasn't able to have them all. I created a new entry for it in media.exolist.json. Tell me if missed something. Thanks.

### DIFF
--- a/demos/main/src/main/assets/media.exolist.json
+++ b/demos/main/src/main/assets/media.exolist.json
@@ -507,6 +507,13 @@
         "subtitle_language": "en"
       },
       {
+        "name": "SubStation Alpha Style",
+        "uri": "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-avc-baseline-480.mp4",
+        "subtitle_uri": "https://drive.google.com/uc?export=download&id=16IrvtynQ6-ANRpRX7hU6xEQeFU91LmXl",
+        "subtitle_mime_type": "text/x-ssa",
+        "subtitle_language": "en"
+      },
+      {
         "name": "MPEG-4 Timed Text",
         "uri": "https://storage.googleapis.com/exoplayer-test-media-1/mp4/dizzy-with-tx3g.mp4"
       }
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -18,9 +18,11 @@ package com.google.android.exoplayer2.text.ssa;
 import static com.google.android.exoplayer2.text.Cue.LINE_TYPE_FRACTION;
 import static com.google.android.exoplayer2.util.Util.castNonNull;
 
+import android.graphics.Typeface;
 import android.text.Layout;
 import android.text.SpannableString;
 import android.text.style.ForegroundColorSpan;
+import android.text.style.StyleSpan;
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.text.Cue;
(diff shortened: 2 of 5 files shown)


abeljim:
@googlebot I signed it!

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
```diff
@@ -207,6 +215,36 @@ private static float parseFontSize(String fontSize) {
     }
   }
 
+  private static boolean parseBold(String bold) {
+    try {
+      int boldFlag = Integer.parseInt(bold);
+      if(boldFlag == 1 || boldFlag == -1){
```
szaboa: Based on the specs, only -1 means true, or am I missing something? Also you could simplify it by inlining the return like `return Integer.parseInt(bold) == -1;`. Same for `parseItalic`.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -319,6 +321,27 @@ private static Cue createCue(
...
             Cue.TEXT_SIZE_TYPE_FRACTIONAL_IGNORE_PADDING);
       }
+      if (style.bold && style.italic) {
+        spannableText.setSpan(
+            new StyleSpan(Typeface.BOLD_ITALIC),
+            0,
+            spannableText.length(),
+            SpannableString.SPAN_EXCLUSIVE_EXCLUSIVE);
+      }
+      else if(style.bold){
```
szaboa: I wouldn't put the `else` in new lines, to be consistent with the rest of the code. Also if we want to make it really consistent I'd add these leading comments `/* start= */` and `/* end= */` at the start of the 2nd and 3rd params of `setSpan` - similarly how it is done for the primary color span.

And lastly, you have a couple of missing whitespaces, e.g. `if(style.bold){` -> `if (style.bold) {`

szaboa (commented):
Thanks for this, did a quick review, but it all depends on @icbaker ;)

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
abeljim: Yeah based on the specs it should only be -1 but when I was testing with vlc they took 1 as well. Do you think we should go for a clean implementation or a more forgiving one?

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
abeljim: Thank you. I will fix this.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
szaboa: I've also checked the libass implementation and it seems they are doing the same (-1, 1): https://github.com/libass/libass/blob/82b225b3d6653091d028b39d561d185ed76a7be5/libass/ass_parse.c#L118

So IMO we can go with the more forgiving implementation, but lets wait for @icbaker 's confirmation.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
icbaker: I agree with @szaboa - but don't worry too much about the styling/whitespace. When we merge PRs we run them through https://github.com/google/google-java-format (same for all the code we author ourselves), which sorts all this stuff out without needing to think too hard about it.

library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
icbaker: +1 to more forgiving

though weirdly for italic it seems libass **only** consider 1 to mean `true` (just below the line @szaboa linked), while the spec only allows `-1`...? P

Maybe it would be easier just to do `!= 0`?

demos/main/src/main/assets/media.exolist.json
```diff
@@ -506,6 +506,13 @@
         "subtitle_mime_type": "text/x-ssa",
         "subtitle_language": "en"
       },
+      {
+        "name": "SubStation Alpha Style",
+        "uri": "https://storage.googleapis.com/exoplayer-test-media-1/gen-3/screens/dash-vod-single-segment/video-avc-baseline-480.mp4",
+        "subtitle_uri": "https://drive.google.com/uc?export=download&id=16IrvtynQ6-ANRpRX7hU6xEQeFU91LmXl",
+        "subtitle_mime_type": "text/x-ssa",
+        "subtitle_language": "en"
+      },
```
icbaker, line 515: Thanks for including this to make testing easier, it's really helpful - FYI I'll drop it before actually merging the PR though.

icbaker (approved):
Looks good, thanks for the contribution! I'll work on getting this merged

icbaker:
> Looks good, thanks for the contribution! I'll work on getting this merged

Correction: I'll wait to see if you decide to use `!= 0` and then merge it :)

abeljim:
I think we should stick with 1 and -1 for now. This way defaults to false when input is unclear. This also follows the libass way. I will later go through and make a file with actually styles from content to have another reference.

demos/main/src/main/assets/media.exolist.json
icbaker: Ah I didn't realise this exercises more styles than just italic/bold - this does actually look like a useful file to add to the demo app (but probably not as part of this PR).

If you're happy to contribute this file, can you open a separate PR containing just the SSA file and I'll take it from there an upload it to our cloud bucket and add an entry to the demo app. Thanks!

PR #8658, opened 2021-03-02
Detect external surround sound on Xiaomi devices

joakimzhang:
exoplayer can enable dolby passthrough on xiaomi device when get
EXTERNAL_SURROUND_SOUND_KEY

Signed-off-by: zhangq <zhangqi16@xiaomi.com>

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -215,6 +215,8 @@
     manipulation API.
 *   Demo app: Retain previous position in list of samples.
 *   Add Guava dependency.
+*   IMA extension: Fix the way 'content complete' is handled to avoid repeatedly
+    refreshing the timeline after playback ends.
 
 ### 2.11.6 (2020-06-19) ###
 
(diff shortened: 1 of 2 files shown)


joakimzhang:
Play Dolby stream by "play movie" on XiaomiTV,  when set output mode to Spdif and passthrough, Exoplayer just out raw uncompressed PCM to audio hal, but AVR can support decode Dolby audio.
So,we need encoded audio be output in this case, exoplayer can enable dolby passthrough on xiaomi device when  Global.getInt(context.getContentResolver(), EXTERNAL_SURROUND_SOUND_KEY, 0) == 1)

@ojw28 Also, could you confirm that if external_surround_sound_enabled is set to 1 on any Xiaomi device, then both AC3 and EAC3 are supported? Thanks.
Reply: I only tested some of the Xiaomi TV，we can try to test all xiaomi TV  after commit this patch,and this patch will take effect when we set EXTERNAL_SURROUND_SOUND_KEY=1,so I think it have no bad effect for all Xiaomi TV.

krocard:
@joakimzhang, it seems your `zhangqi16@xiaomi.com` address email is not registered as having accepted the Contributor License Agreement. See https://github.com/google/ExoPlayer/pull/8658#issuecomment-788769174.

Until you sign such agreement, we cannot look at your pull request.

Thanks for your understanding.

PR #8720, opened 2021-03-15
Support tts:shear in TTML parser and WebView output

dlafayet:
@icbaker

### DIFF
--- a/library/common/src/main/java/com/google/android/exoplayer2/text/Cue.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/text/Cue.java
@@ -270,6 +270,12 @@ public final class Cue {
   public final @VerticalType int verticalType;
 
   /**
+   * The shear angle in degrees expressed in graphics coordinates to be applied to this block. This
+   * results in a skew transform for the block along the inline progression axis.
+   */
+  public final float shearDegrees;
+
+  /**
    * Creates a text cue whose {@link #textAlignment} is null, whose type parameters are set to
    * {@link #TYPE_UNSET} and whose dimension parameters are set to {@link #DIMEN_UNSET}.
    *
@@ -370,7 +376,8 @@ public final class Cue {
         /* bitmapHeight= */ DIMEN_UNSET,
         /* windowColorSet= */ false,
         /* windowColor= */ Color.BLACK,
-        /* verticalType= */ TYPE_UNSET);
+        /* verticalType= */ TYPE_UNSET,
+        0f);
   }
 
   /**
@@ -415,7 +422,8 @@ public final class Cue {
         /* bitmapHeight= */ DIMEN_UNSET,
         windowColorSet,
         windowColor,
-        /* verticalType= */ TYPE_UNSET);
+        /* verticalType= */ TYPE_UNSET,
+        0f);
   }
 
   private Cue(
@@ -433,7 +441,8 @@ public final class Cue {
       float bitmapHeight,
       boolean windowColorSet,
       int windowColor,
-      @VerticalType int verticalType) {
+      @VerticalType int verticalType,
+      float shearDegrees) {
     // Exactly one of text or bitmap should be set.
     if (text == null) {
       Assertions.checkNotNull(bitmap);
@@ -455,6 +464,7 @@ public final class Cue {
     this.textSizeType = textSizeType;
     this.textSize = textSize;
     this.verticalType = verticalType;
+    this.shearDegrees = shearDegrees;
   }
 
   /** Returns a new {@link Cue.Builder} initialized with the same values as this Cue. */
@@ -479,6 +489,7 @@ public final class Cue {
     private boolean windowColorSet;
     @ColorInt private int windowColor;
     @VerticalType private int verticalType;
+    private float shearDegrees;
 
     public Builder() {
       text = null;
@@ -514,6 +525,7 @@ public final class Cue {
       windowColorSet = cue.windowColorSet;
       windowColor = cue.windowColor;
       verticalType = cue.verticalType;
+      shearDegrees = cue.shearDegrees;
     }
 
     /**
@@ -795,6 +807,14 @@ public final class Cue {
     }
 
     /**
+     * Sets the shear angle for this Cue
+     */
+    public Builder setShearDegrees(float shearDegrees) {
+      this.shearDegrees = shearDegrees;
+      return this;
+    }
+
+    /**
      * Gets the vertical formatting for this Cue.
      *
      * @see Cue#verticalType
@@ -821,7 +841,8 @@ public final class Cue {
           bitmapHeight,
           windowColorSet,
           windowColor,
-          verticalType);
+          verticalType,
+          shearDegrees);
     }
   }
 }
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlDecoder.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlDecoder.java
@@ -615,13 +615,9 @@ public final class TtmlDecoder extends SimpleSubtitleDecoder {
           style =
               createIfNull(style)
                   .setTextEmphasis(TextEmphasis.parse(Util.toLowerInvariant(attributeValue)));
+          break;
         case TtmlNode.ATTR_TTS_SHEAR:
-          style = createIfNull(style);
-          try {
-            parseShear(attributeValue, style);
-          } catch (SubtitleDecoderException e) {
-            Log.w(TAG, "Failed parsing shear value: " + attributeValue);
-          }
+          style = createIfNull(style).setShearPercentage(parseShear(attributeValue));
           break;
         default:
           // ignore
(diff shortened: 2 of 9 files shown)


dlafayet:
italics-horz

italics-vert

italics-ruby

library/common/src/main/java/com/google/android/exoplayer2/text/Cue.java
```diff
@@ -269,6 +269,12 @@
    */
   public final @VerticalType int verticalType;
 
+  /**
+   * The shear angle expressed in graphics coordinates to be applied to this block. This results
+   * in a skew transform for the block along the inline progression axis.
+   */
+  public final float shear;
```
icbaker: would it make sense to call this `shearDegrees` and the one on TtmlStyle `shearFraction`?

Just to be clear about the different 'units' being used - which hopefully helps explain some of the conversions too.

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
```diff
@@ -382,6 +383,16 @@ private void applyStyleToOutput(
...
     if (resolvedStyle != null) {
       TtmlRenderUtil.applyStylesToSpan(text, start, end, resolvedStyle, parent, globalStyles);
+      if (resolvedStyle.getShear() != 0.0f && TAG_P.equals(tag)) {
+        // Shear style should only be applied to P nodes
+        // https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-shear
+        // The spec doesn't specify the coordinate system to use for block shear
+        // however the spec shows examples of how different values are expected to be rendered.
+        // See: https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-shear
+        // https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-fontShear
+        // This maps the shear percentage to shear angle in graphics coordinates
```
icbaker, line 419: nit: resolvedStyle.getShear() isn't a percentage (because you've already divided by 100 to convert it to a fraction between -1.0 and +1.0 above in TtmlDecoder#parseShear)

I guess this comment should refer to 'shear fraction' instead?

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlDecoderTest.java
```diff
@@ -674,6 +676,44 @@ public void rubies() throws IOException, SubtitleDecoderException {
...
+
+  @Test
+  public void shear() throws IOException, SubtitleDecoderException {
+    TtmlSubtitle subtitle = getSubtitle(SHEAR_FILE);
+
+    Cue firstCue = getOnlyCueAtTimeUs(subtitle, 10_000_000);
+    assertThat(firstCue.shear).isEqualTo(0f);
+
+    Cue secondCue = getOnlyCueAtTimeUs(subtitle, 20_000_000);
+    assertThat(secondCue.shear).isEqualTo(shearPercentageToSkewAngle(0.1667f));
```
icbaker: I think it's better to remove the helper function & use test values that can be easily converted by eye (which 16.67 is since it's ~15 degrees)

You can use Truth to do fuzzy assertions on floats to avoid needing masses of decimal places:
https://truth.dev/api/1.0.1/com/google/common/truth/FloatSubject.html#isWithin-float-

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlStyleTest.java
```diff
@@ -46,6 +46,7 @@
   private static final int RUBY_POSITION = RubySpan.POSITION_UNDER;
   private static final Layout.Alignment TEXT_ALIGN = Layout.Alignment.ALIGN_CENTER;
   private static final boolean TEXT_COMBINE = true;
+  public static final float SHEAR_PERCENTAGE = 16f;
```
icbaker, line 52: nit: this is an unrealistic value right? Should be between -1.0 and +1.0?

(and it's also not a percentage)

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
```diff
@@ -285,7 +284,7 @@ private void updateWebView() {
                       + "writing-mode:%s;"
                       + "font-size:%s;"
                       + "background-color:%s;"
-                      + "transform:translate(%s%%,%s%%);"
+                      + "transform:translate(%s%%,%s%%)%s;"
```
icbaker: maybe we should split the different transform parameters onto separate lines? Something like:

```
+ "transform:
+ "translate(%s%%,%s%%) "
+ "%s;"
```

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
```diff
@@ -320,6 +320,16 @@ private void updateWebView() {
         "base64");
   }
 
+  private String getBlockShear(Cue cue) {
+    if (cue.shear != 0.0f) {
+      String direction =
+          (cue.verticalType == Cue.VERTICAL_TYPE_LR || cue.verticalType == Cue.VERTICAL_TYPE_RL) ?
+              "skewY" : "skewX";
+      return Util.formatInvariant(" %s(%.2fdeg)'", direction, cue.shear);
```
icbaker: nit: remove the space from the string here and put it on L287 instead.

icbaker (commented):
I've just merged #8653 with some changes made during internal review - would it be possible to rebase this on top of those changes?

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
dlafayet: I will change TtmlStyle's shear value to be a percentage to keep it consistent with the spec, i.e. no more shear fraction.

library/core/src/test/java/com/google/android/exoplayer2/text/ttml/TtmlStyleTest.java
dlafayet: I will change to use shear percentage units in all the Ttml classes.

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
dlafayet: This will insert an extra space `translate(0%, 0%) ;` in the case where there is no skew transform.

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
dlafayet: This will insert an extra space `translate(0%, 0%) ;` in the case where there is no skew transform.

dlafayet:
Rebased and address code review comments.

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
icbaker: Ah I see, good point.

Some quick testing suggests the space isn't actually needed at all, Chrome seems happy with this: `transform:translate(-1%,-50%)skewX(25deg);`

WDYT about removing the space completely?

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlDecoder.java
```diff
@@ -755,6 +764,27 @@ private static void parseFontSize(String expression, TtmlStyle out) throws
     }
   }
 
+  private static void parseShear(String expression, TtmlStyle out) throws
```
icbaker: Is it possible to switch this to return `float` instead of taking an output parameter?

library/ui/src/main/java/com/google/android/exoplayer2/ui/WebViewSubtitleOutput.java
dlafayet: Sure thing. Removed.

icbaker:
Thanks for the fixes - I'll work on getting this merged.

icbaker:
During an internal review of this change, @AquilesCanta raised the following question:

> I don't know how faithfully we are trying to imitate the w3 example, but I'm under the impression that the provided example:
> 
> https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-shear
> 
> does not only apply a skew, but also a small scaling factor so that the bounding rectangle remains of the same size. I'm referring to this image:
> 
> https://www.w3.org/TR/2018/REC-ttml2-20181108/images/shear_1.png
> 
> I think the text on the left is downscaled a bit, it's not only skewed. If this is acceptable, please ignore.

@dlafayet - wdyt, would it make sense to apply the scaling factor too?

It looks like this is the relevant part of the spec guidance:

> its maximum inline progression dimension length prior to shear transformation, is reduced by `B*sin(θ)`, where `B` is the block progression dimension of the outer block area and `θ` is the shear angle

dlafayet:
My understanding of the spec, in consultation with two TTML spec contributors, is that a padding should be applied to reduce the height of the content area. To determine how much padding, we need to measure the block width after layout, which is the value of `B` that is mentioned in the spec in the padding calculation. Applying the padding, in turn, could cause a reflow as some text might wrap to the next line, which would again change the value of `B`.

In the current WebView output implementation, I do not see a way to compute the value of `B` from the Java side. One idea is to introduce some javascript to do post-processing on the DOM but that probably requires a bigger discussion. 

For the current patch, we do have the possibility of text being rendered out of the region bounds. 

As a side note, we have the same limitation for other layout-dependent properties, for example, implementing `rubyPosition` or `textEmphasis` `position` with value `outside`. We might be able to do an approximation on the Java side, but an exact implementation would require post-layout processing on the DOM. Would love to hear your thoughts on this.

icbaker:
Thanks for the extra details!

> In the current WebView output implementation, I do not see a way to compute the value of B from the Java side. One idea is to introduce some javascript to do post-processing on the DOM but that probably requires a bigger discussion.

Yeah, I'm **really** trying to avoid enabling JavaScript in this WebView - it raises so many questions around security I'd rather keep it switched off - I don't think this is a convincing enough use-case to confront all those questions unfortunately.

So we'll stick the code as-is.

PR #8764, opened 2021-03-28
Calculate HLS target live offset from #EXT-START-TIME if available

uvjustin:
From r2.12.3 to r2.13.0 there seems to be a regression where #EXT-START-TIME is not honored for live playlists. This PR tries to get the target live offset from the parsed playlist start time before falling back to the other methods.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaSource.java
@@ -616,7 +616,9 @@ public final class HlsMediaSource extends BaseMediaSource
     HlsMediaPlaylist.ServerControl serverControl = playlist.serverControl;
     // Select part hold back only if the playlist has a part target duration.
     long offsetToEndOfPlaylistUs;
-    if (serverControl.partHoldBackUs != C.TIME_UNSET
+    if (playlist.startOffsetUs != C.TIME_UNSET) {
+      offsetToEndOfPlaylistUs = playlist.durationUs - playlist.startOffsetUs;
+    } else if (serverControl.partHoldBackUs != C.TIME_UNSET
         && playlist.partTargetDurationUs != C.TIME_UNSET) {
       offsetToEndOfPlaylistUs = serverControl.partHoldBackUs;
     } else if (serverControl.holdBackUs != C.TIME_UNSET) {


uvjustin:
@googlebot I signed it!

christosts:
Hi,

Thank you for sending this pull-request. Before merging it, may I ask for two things:

1. Reading the [HLS spec](https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-08), I think that `EXT-X-SERVER-CONTROL` should have higher priority (if defined) over `EXT-X-START`. Therefore, the check for `playlist.startOffsetUs` should move down, just before applying the fallback case. I will make that change - no need to update the pull-request.

2. Can you share a link to a live playlist with EXT-X-START, so that we can test? You can email it to dev.exoplayer@gmail.com with "Issue #8764", if you can't share links publicly. If you send an email, please add comment here too.

Thanks

uvjustin:
Hi Chris,
Thanks for taking the time to look at this!
Regarding 1), Your suggested approach <strike>works for my use case</strike>, but my reading of the RFC is different than yours. The difference centers on the word "minimum" - since HOLD-BACK and PART-HOLD-BACK say that this should be the minimum distance to the live edge, this suggests that this is just a (inequality) constraint and not a (equality) proposed start time. Am I missing something here?
For 2), I actually don't have a server that is serving HLS. We actually use HLS in the [Home Assistant](https://github.com/home-assistant) project for distributing camera feeds to various browsers and client devices. One of the main drawbacks to this approach is the latency introduced by the HLS pipeline, so I've been playing with LL-HLS to see if this is a possible solution. Since the HLS feed is actually served through Home Assistant, the HLS playlist and media are dynamically created and are not available statically.
Would it be ok to send a static copy of the playlists and sample media with the endpoints changed to a local server eg 127.0.0.1?

Edit: Actually I realized it doesn't work for my use case. The primary use for me is actually to use a `PRECISE` start time to override starting from the nearest segment boundary, as outlined in the other draft PR.
A secondary use is that upon fallback to non LL-HLS, there might be some risk of BLWE when using a short (3 segment) window and starting right at the window boundary. (I know this short of a window has been discouraged by @ojw28 and others, but it is in one of the Apple samples, and it hasn't really caused any problems for us so far.) In this case, setting a start time of 2 segments back from the live edge, although technically a violation of HOLD-BACK, seems to eliminate any potential problems (and obviously improve latency). For this secondary use the approach chosen in 1) doesn't matter.

ojw28:
I also don't see anything in the spec that suggests `EXT-X-SERVER-CONTROL` has higher priority. @christosts - Could you quote that part of the RFC that you think implies that, so it's clear which bit specifically you're looking at? Thanks!

christosts:
Here's my reading of the spec (I'm also copying the spec definitions at the end of this comment).

- For the three directives involved, the spec uses the term **should** for all of them. It's not clear to me which of the three *should*s has higher priority, so my reading is context specific. But it does use capital "SHOULD" for EXT-X-START.
- Indeed, the server directives indicate a `minimum` distance from the live-edge, as @uvjustin says. However, the spec also says that they "_indicate the server-recommended minimum distance from the end of the Playlist **at which clients should begin to play or to which they should seek**_". So, it's not only an inequality constraint but also a recommendation where exactly to start playing. The player then needs to make sure to stay that much behind the live-edge, and not less.
- The spec is very explicit on the priorities between `PART-HOLD-BACK` and `HOLD-BACK` and what to do in the absence of them. 

So my reading was that, if the content supports server directives, there's a very specific behavior expected from the client, hence the spec is so detailed. And if the content creator has made the effort to support server directives, why define a different start-time which might effectively lead to a larger latency than the server recommends in the server directives?

I can see though that it's a matter of reading the spec. One might argue that `EXT-X-START` is meant to override server directives anyway.

ojw28:
If `EXT-X-START` does not get priority, then inclusion of both `EXT-X-START` and `EXT-X-SERVER-CONTROL` wouldn't ever make sense. `EXT-X-START` could always be omitted from the manifest, since it would never have any effect.

If `EXT-X-START` does get priority, then inclusion of both would make sense. `EXT-X-START` would indicate the preferred start position, and `EXT-X-SERVER-CONTROL` would define behavior if the player is subsequently seeked to the live edge.

On this basis I think that `EXT-X-START` should get priority if both are included.

ojw28:
> However, the spec also says that they "indicate the server-recommended minimum distance from the end of the Playlist at which clients should begin to play or to which they should seek". So, it's not only an inequality constraint but also a recommendation where exactly to start playing.

I do not interpret this as a recommendation of where to start playing. It's only defining a _minimum distance from the end of the playlist_ at which a client should begin to play. Assuming the `EXT-X-START` position is a greater distance from the end of the playlist, which is almost certainly the case if it's included, then using the start position that it specifies is compatible with this minimum distance constraint.

uvjustin:
> > However, the spec also says that they "indicate the server-recommended minimum distance from the end of the Playlist at which clients should begin to play or to which they should seek". So, it's not only an inequality constraint but also a recommendation where exactly to start playing.

I can see how it could be read this way, but I think if it was intended this way, the wording would be more clear about the dual purpose of the parameter (as a minimum distance constraint and as a start time) if it said something like "indicate the server-recommended minimum distance from the end of the Playlist _and the distance_ at which clients should begin to play or to which they should seek". Also I think it makes sense that there is some flexibility to allow the player to start from a "safer" start time as opposed to the most aggressive time permitted by the server.

> 
> I do not interpret this as a recommendation of where to start playing. It's only defining a _minimum distance from the end of the playlist_ at which a client should begin to play. Assuming the `EXT-X-START` position is a greater distance from the end of the playlist, which is almost certainly the case if it's included, then using the start position that it specifies is compatible with this minimum distance constraint.

I read it this way as well.

PR #8773, opened 2021-03-29
feat: developers can add devices to the widevine l3 forced ones

mzander-cbc:
Hi developers,

there are several devices which won't work with L1 but with L3 they do (e.g. Pixel C).
Why shouldn't we let the developers decide which devices they want to force to L3.

Kind regards,
Moritz

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/drm/FrameworkMediaDrm.java
@@ -40,8 +40,10 @@ import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 
 /** An {@link ExoMediaDrm} implementation that wraps the framework {@link MediaDrm}. */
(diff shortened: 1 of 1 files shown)


mzander-cbc:
@googlebot I signed it!

icbaker:
We'd like to avoid as much mutable static state in the library as possible as it can make code extremely hard to reason about, and to test - so if we're going to expose an option like this I'd like to see it as an instance method.

However I don't think we need to make any changes to `FrameworkMediaDrm` - developers can already customise this behaviour by providing a custom `ExoMediaDrm.Provider` that sets any property strings they like after instantiating the `ExoMediaDrm` instance.

Something like this that delegates to `FrameworkMediaDrm.DEFAULT_PROVIDER`:

```
class CustomMediaDrmProvider implements ExoMediaDrm.Provider {

  private final ImmutableSet<String> l3ForcedDevices;

  CustomMediaDrmProvider(Set<String> l3ForcedDevices) {
    this.l3ForcedDevices = ImmutableSet.copyOf(l3ForcedDevices);
  }
  
  ExoMediaDrm acquireExoMediaDrm(UUID uuid) {
    ExoMediaDrm mediaDrm = FrameworkMediaDrm.DEFAULT_PROVIDER.acquireExoMediaDrm(uuid);
    if (C.WIDEVINE_UUID.equals(uuid) && l3ForcedDevices.contains(Util.MODEL)) {
      mediaDrm.setPropertyString("securityLevel", "L3");
    }
    return mediaDrm;
  }
} 
```

mzander-cbc:
Sounds legit, thanks.
I was a bit confused about the static ASUS device property in the FrameworkMediaDrm class.

Thanks,
Moritz

PR #8782, opened 2021-03-30
Update Netflix Japanese TTML test file

dlafayet:
@icbaker A small update to the Netflix Japanese TTML file. Removed two lines from the original file which could cause confusion due to the description header not matching the subtitle.


icbaker:
Thanks for the fix - I've uploaded this new file to our storage bucket under the same name as the old one, it should be available immediately.

PR #8812, opened 2021-04-12
Add test cases for iframe only with EXT-X-MAP

stevemayhew:
The "implicit" init segment logic in the `HlsPlaylistParser` in the original pull request for i-Frame only
support had an off by one bug.  This commit adds unit test cases to cover that bug and for validating both
explicit and implicit initialization segments.

Thanks to @marcbaechinger for fixing the bug when the original pull request was merged, just returning the favor by increasing the `HlsPlaylistParser` UT coverage a bit.


stevemayhew:
I see the implicit init unit test was covered already, and pretty much matched my test.   So this commit just includes EXT-X-MAP with a BYTERANGE attribute (which a couple of our Origin partners use)

AquilesCanta:
The title of the commit seems correct to me, but the pull request's doesn't. Should with be **with EXT-X-MAP**?

stevemayhew:
Thanks @AquilesCanta updated title

stevemayhew:
We've definitely got some cycles and interest in increasing the test coverage of the parsers

PR #8814, opened 2021-04-13
Fix bottom-cropped text in WebView output

dlafayet:
Use unitless numbers for WebView line-height values

See
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values


dlafayet:
@icbaker 

Before fix:
line-height-em

After fix:
line-height-no-units

icbaker:
Thanks for the fix & reference link - will get this merged.

PR #8829, opened 2021-04-19
 Exoplayer LinearBlockMode support

bany2000:
, my app is broadcast tv. we use exoplayer to play tv ,Linearblock object create from android_media_tv_tuner.cpp not MediaCodec.obtain().we insert every video linearblock into SampleQueue.MediacodecRender read linearblock from sampleQueue and queue linearblock to mediacodec. Therefore i need MediaCodedAdapter interface to transmit linearBlock to mediacodec.


bany2000:
@christosts i summit my whole solution for linearBlock ,please help to review

bany2000:
@christosts  could you provide some suggestion for my all changes？

christosts:
Thank you for your pull-request. 

Can you please tell us a bit more about your application and its requirements? Let me give you some context why my question:

If I get it right, your PR tries to remove one memory copy between reading the data from the stream and passing them down to MediaCodec. 

However, your pull-request changes a few classes in a way that violates their abstraction. Specifically, the pull-request is making Format, SampleQueue and DecoderInputBuffer have knowledge about MediaCodec's LinearBlocks. That is something that breaks the abstraction model of these classes. Keep in mind that these 3 classes are also used for playbacks that do not use MediaCodec, therefore it is questionable why they should have a dependency on MediaCodec. 

Admittedly, to achieve a single-copy from the data source to the MediaCodec, you need to make changes to some components outside MediaCodecAdapter and MediaCodecRenderer. However, if these changes are necessary, they should be done in away that it doesn't create a dependency to MediaCodec and respects the abstraction model. This is not an easy task to achieve, therefore I ask the question: do you really need to avoid the extra copy between the data source and the SampleQueue?  Do you have data suggesting that this copy is a reason of slow performance? Are you seeing playback issues that are solved by dropping this memory copy?

bany2000:
@christosts
1.Can you please tell us a bit more about your application and its requirements
[bany2000] our project use exoplayer to playback broadcast tv,so  ts data come from tuner and is filtered to a/v linearblock buffer in hardware dmx or hardware extractor . the dmx callback a/v linearblock to exoplayer source .
2.If I get it right, your PR tries to remove one memory copy between reading the data from the stream and passing them down to MediaCodec.
[bany2000] yes
3.Do you have data suggesting that this copy is a reason of slow performance
[bany2000] yes
4.Are you seeing playback issues that are solved by dropping this memory copy?
[bany2000] no, a new project ,new archtecture for another company  run broadcast tv  with  exoplayer2.x
[bany2000]yes, linearblock can implement share memory to remove one memory copy between reading the data from the stream and passing them down to MediaCodec.

bany2000:
@christosts as you mentioned  to achieve a single-copy from the data source to the MediaCodec, you need to make changes to some components outside MediaCodecAdapter and MediaCodecRenderer.This is not an easy task to achieve
     i know it is not  an easy task to achieve, If i want to do that , could you help to  provide some idear how to design  the components outside MediaCodecAdapter and MediaCodecRenderer  and respects the abstraction model for linearblock mode

christosts:
We evaluated this feature and we don't think we can proceed with the pull-request at the moment.

Achieving a zero-copy from data source to platform requires wide changes to many components which introduce substantial risk, considering the many use cases ExoPlayer supports. For such an effort, we would require to spend more time in design before jumping to code, but we don't have capacity at the moment.

I will leave the pull-request open to re-evaluate when we can have the capacity to address the issue.

In the meantime: you should be able to implement partial support for LinearBlock in your app *without* modifying the ExoPlayer sources. 

- You can implement your own MediaCodecAdapter that operates MediaCodec in block mode *without* changes in the MediaCodecAdapter interface. At runtime, you can instantiate ExoPlayer with a MediaCodecVideoRenderer that uses your own MediaCodecAdapter.Factory. 
- For your MediaCodecAdapter implementation, you can copy the existing AsynchronousMediaCodecAdapter and effectively change getInputBuffer() so that returns a ByteBuffer that is backed by a LinearBlock, and queueInputBuffer() so that it enqueues a LinearBlock to MediaCodec.
- If you do that, you might save one copy (from sample queue to MediaCodec) and potentially can make better use of memory as follows:
- When the renderer reads from the sample queue, it does not know beforehand how many bytes will be needed, therefore getInputBuffer() needs to return ByteBuffers big enough to read the possible max frame size. This is in `MediaFormat.KEY_MAX_INPUT_SIZE` when MediaCodecAdapter is configured.
- For each call to MediaCodecAdapter.getInputBuffer(), you can allocate a LinearBlock of size `MediaFormat.KEY_MAX_INPUT_SIZE`  to make sure the sample fits in the buffer. However, this will use redundant memory. Alternatively, your adapter can allocate a single LinearBlock to fit a few samples and each MediaCodecAdapter.getInputBuffer() will allocate `MediaFormat.KEY_MAX_INPUT_SIZE` bytes from there. When queueInputBuffer() is called, you can find out exactly how many bytes were actually written and consider the remaining allocated space as free in order to fine tune memory usage in the single LinearBlock.

As I mentioned, this approach does not require changes to ExoPlayer sources, may improve the overall memory usage on your device, and might reduce memory copies, if the MediaCodec supports no-copy from the LinearBlock (this depends on the decoder used). However, the memory copies from the data source to the sample queue will still occur.

icbaker:

PR #8848, opened 2021-04-21
Add support for SSA Underline and Strikeout

abeljim:
@icbaker @szaboa
This is initial PR for Underline and Strikeout SSA style.


abeljim:
My bad did they wrong will recreate.

PR #8858, opened 2021-04-22
Support ebutts:multiRowAlign in TTML text renderer

dlafayet:
Fix bug in text alignment inheritance where child does not correctly inherit ancestor's setting

@icbaker

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
@@ -72,6 +72,7 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
   public static final String ATTR_TTS_TEXT_EMPHASIS = "textEmphasis";
   public static final String ATTR_TTS_WRITING_MODE = "writingMode";
   public static final String ATTR_TTS_SHEAR = "shear";
+  public static final String ATTR_EBUTTS_MULTI_ROW_ALIGN = "multiRowAlign";
 
   // Values for ruby
   public static final String RUBY_CONTAINER = "container";
@@ -376,7 +377,6 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
       return;
     }
     String resolvedRegionId = ANONYMOUS_REGION_ID.equals(regionId) ? inheritedRegion : regionId;
-
     for (Map.Entry<String, Integer> entry : nodeEndsByRegion.entrySet()) {
       String regionId = entry.getKey();
       int start = nodeStartsByRegion.containsKey(regionId) ? nodeStartsByRegion.get(regionId) : 0;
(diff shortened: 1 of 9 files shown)


dlafayet:
ebutts:multiRowAlign (before fix)
multi-row-align-bug
ebutts:multiRowAlign (after fix)
multi-row-align-fix

library/core/src/main/java/com/google/android/exoplayer2/text/ttml/TtmlNode.java
```diff
@@ -409,17 +409,29 @@ private void applyStyleToOutput(
...
+          // Multi-row align should only be applied to P nodes
+          // https://www.w3.org/TR/ttml-imsc1.1/#text-multiRowAlign
+          regionOutput.setMultiRowAlignment(resolvedStyle.getMultiRowAlign());
+        }
+      }
+
+      if (resolvedStyle.getTextAlign() != null) {
+        // Only set text alignment if defined in current node, otherwise could overwrite
+        // inherited text alignment
+        regionOutput.setTextAlignment(resolvedStyle.getTextAlign());
```
icbaker: This looks strange to me - could you explain a bit more why it's needed? (and why it's only needed for text alignment, and none of the other properties)

icbaker: (by 'this' I meant the `if`, even though I commented on the `set` line...)

I tried removing the `if` (i.e. reverting to the previous code that just always calls `setTextAlignment()`) and ran the demo file you provided and it seemed to work fine for the I.3 case you screenshotted but not the I.14 case - is that what this enables to work?

icbaker: Ah is this an existing bug in the TTML implementation?

I can reproduce with this test data:
```
<div>
   <p begin="70s" end="78s" tts:textAlign="start"><span tts:fontSize="50%">Start</span> alignment</p>
</div>
```

And this expectation fails because `textAlignment` is actually null:
```
assertThat(seventhCue.textAlignment).isEqualTo(Layout.Alignment.ALIGN_NORMAL);
```

I'm still a bit suspicious that only `textAlignment` needs special treatment like this - if you can help me understand that I'd appreciate it, I'll keep playing around on my end too.

icbaker: Ah OK I see it's completely the opposite way round: `textAlignment` is the only property in this bit of the code that isn't already null-checked - sorry for being a bit slow here.

Shouldn't it also only be applied to `<p>` nodes (like shear above, and now multiRowAlign)?
https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-textAlign

I think this probably deserves to be in a separate change as a 'bug fix' - would you like to pull it out into a separate PR or shall I just make it as a commit before merging this on top?

dlafayet: Created https://github.com/google/ExoPlayer/pull/8877 for the text alignment bug

icbaker:
This has been merged but copybara didn't integrate properly with github again :(

PR #8877, opened 2021-04-27
Fix bug in TTML inherited text alignment

dlafayet:
Fix bug where child does not correctly inherit ancestor's text
alignment setting.

Make it so that alignment is only applied on a P node.
https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-textAlign

@icbaker


dlafayet:
Before fix: (Vertical text is expected to be start-aligned) 
vertical-alignment-bug

After fix:
vertical_alignment_fix

Issue #8847, opened 2021-04-21
Live DASH video track selection broken since 2.13.0


cdongieux:
I'm currently updating my app from ExoPlayer 2.12.1 to 2.13.3, I notice the video track selection on Live DASH streams seems broken since 2.13.0.

Here are available video tracks:
```
MediaCodecVideoRenderer [
  Group:0, adaptive_supported=YES_NOT_SEAMLESS [
    [X] Track:0, id=379, mimeType=video/avc, bitrate=400000, codecs=avc1.64000d, res=384x216, fps=25.0, supported=YES
    [X] Track:1, id=380, mimeType=video/avc, bitrate=800000, codecs=avc1.64001e, res=640x360, fps=25.0, supported=YES
    [X] Track:2, id=381, mimeType=video/avc, bitrate=1600000, codecs=avc1.64001f, res=896x504, fps=25.0, supported=YES
    [X] Track:3, id=3517, mimeType=video/hevc, bitrate=1600000, codecs=hvc1.1.2.L93, res=896x504, fps=50.0, supported=YES
    [X] Track:4, id=382, mimeType=video/avc, bitrate=3000000, codecs=avc1.64001f, res=1280x720, fps=25.0, supported=YES
    [X] Track:5, id=3518, mimeType=video/hevc, bitrate=4800000, codecs=hvc1.1.2.L123, res=1920x1080, fps=50.0, supported=YES
    [X] Track:6, id=3519, mimeType=video/hevc, bitrate=14800000, codecs=hvc1.1.2.L123, res=1920x1080, fps=50.0, supported=YES
  ]
]
```

Here is what was ExoPlayer behavior in 2.12.1 and what it's now:

**2.12.1:**
1. track 896x504 (1600000bps, video/avc) is selected
2. track 1920x1080 (14800000bps, video/hevc) is selected
=> that's the expected behavior

**2.13.3:**
1. track 896x504 (1600000bps, video/avc) is selected
2. track 1920x1080 (14800000bps, video/hevc) is selected
3. track 384x216 (400000bps, video/avc) is selected
=> this behavior is not expected

In 2.13.3, few seconds after step 2. the bandwidth estimation drops and the lowest bitrate track is selected, a higher quality seems to never be selected.

As stream URL and DRM info are sensitive, I'll provide you by email a sample code hosted on a private GitHub repository to help you reproduce the issue. Please let me know if you need anything else.

tonihei:
[...]

PR #8894, opened 2021-05-03
Make DefaultDashChunkSource also take into account segments ending before/until manifest 'publishTime'

cdongieux:
DefaultDashChunkSource did not take into account segments ending after 'now' in the specific case where 'publishTime' value is after 'now'. This commit fixes this behavior described in #8847.

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
@@ -818,6 +818,12 @@ public class DefaultDashChunkSource implements DashChunkSource {
     }
 
     public boolean isSegmentAvailableAtFullNetworkSpeed(long segmentNum, long nowPeriodTimeUs) {
+      if (segmentIndex.isExplicit()) {
+        // We don't support segment availability for explicit indices. Hence, also assume all
+        // segments in explicit indices are always available at full network speed even if they
+        // end in the future.
+        return true;
+      }
       return nowPeriodTimeUs == C.TIME_UNSET || getSegmentEndTimeUs(segmentNum) <= nowPeriodTimeUs;
     }
   }


cdongieux:
@googlebot I signed it!

cdongieux:
@googlebot I fixed it.

tonihei:
Thanks for the PR. As explained in https://github.com/google/ExoPlayer/issues/8847#issuecomment-831913316 we can't merge the PR as it is right now, but I'll leave it open for now in case you want to update it as suggested.

iThuggin:
***@***.***> كتب :

 *unsubscribe*

iThuggin:
—
unsubscribe

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
```diff
@@ -818,6 +818,9 @@ public long getLastAvailableSegmentNum(long nowUnixTimeUs) {
     }
 
     public boolean isSegmentAvailableAtFullNetworkSpeed(long segmentNum, long nowPeriodTimeUs) {
+      if (segmentIndex.isExplicit()) {
+        return true;
```
tonihei, line 825: Could you add a comment here to explain why this is done? Because technically the code is still correct as it is and this new addition is just a workaround because we don't support detecting unavailable segments for explicit indices.

For example:
"// We don't support segment availability for explicit indices. Hence, also assume all segments in explicit indices are always available at full network speed even if they end in the future."

tonihei (commented):
Thanks for updating. It looks like some merge didn't work as expected and it now shows many existing ExoPlayer commits as part of your PR. Could you fix this? Otherwise looks good now.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DefaultDashChunkSource.java
cdongieux: Done :)

tonihei:
> It looks like some merge didn't work as expected and it now shows many existing ExoPlayer commits as part of your PR. Could you fix this? 

That is still an issue. Do you need to rebase your changes somehow?

cdongieux:
> > It looks like some merge didn't work as expected and it now shows many existing ExoPlayer commits as part of your PR. Could you fix this?
> 
> That is still an issue. Do you need to rebase your changes somehow?

It's OK now, sorry for the delay and the inconvenience :)

PR #8900, opened 2021-05-05
Peeked event from Queue

Akash-Ramjyothi:
Removed the peaked element explicitly, and only running the event if it is removed, some concurrency is added to the method, although a ConcurrentQueue would be even safer.


ojw28:
This change doesn't compile because the new `event` variable is never initialized. There is then a compilation failure at the second added line that tries to use it. I think you must have uploaded the wrong thing? It's also a bit unclear what you're trying to achieve with this commit. Please also note that:

1. You'll need to send your pull request to the dev-v2 branch, not the release-v2 branch, as per our contributing file.
1. You'll need to sign the CLA as described above.

If you think a change is needed here, and are able to sign the CLA, then please upload a corrected pull request to the `dev-v2` branch. Thanks!

PR #8943, opened 2021-05-14
Preserve Japanese language features when not applying embedded styling

dlafayet:
Fix bug where rubies and boutens are missing, tate-chu-yoko is rendered incorrectly when `SubtitleView.setApplyEmbeddedStyles(false)`. This method should only affect styling elements and not remove any language features.

@icbaker 

Bug is observed when using the `WebViewSubtitleOutput` and player sets `SubtitleView.setApplyEmbeddedStyles(false)`.

### DIFF
--- a/library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
+++ b/library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
@@ -36,6 +36,7 @@ import androidx.annotation.IntDef;
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.text.Cue;
 import com.google.android.exoplayer2.text.TextOutput;
+import com.google.android.exoplayer2.text.span.LanguageFeatureStyle;
 import com.google.android.exoplayer2.util.Util;
 import java.lang.annotation.Documented;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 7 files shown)


library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleViewUtils.java
```diff
@@ -48,5 +54,27 @@ public static float resolveTextSize(
     }
   }
 
+  public static void preserveJapaneseLanguageFeatures(Spannable copy, Spanned original) {
```
icbaker: Instead of needing to encode this list of Span types in SubtitleViewUtils, what do you think about using a marker interface like `LanguageFeature` that would be implemented by spans like `RubySpan`?

(See Effective Java 3rd edition Item 41: Use marker interfaces to define types)

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleView.java
```diff
@@ -386,7 +386,12 @@ private Cue removeEmbeddedStyling(Cue cue) {
           cue.buildUpon().setTextSize(Cue.DIMEN_UNSET, Cue.TYPE_UNSET).clearWindowColor();
       if (cueText != null) {
         // Remove all spans, regardless of type.
-        strippedCue.setText(cueText.toString());
+        strippedCue.setText(new SpannableString(cueText.toString()));
```
icbaker: Since we're no longer stripping all Spans, I think it's probably easier just to directly copy the `Spanned` to a `SpannableString` (including **all** the spans) by passing it into the constructor, and then iterate over them calling `removeSpan()` unless the type is recognised as a 'language feature'.

I suspect it will also reduce the number of `instanceof` and cast calls needed (especially if you store it in a `SpannableString` local before calling `Cue#setText` later).

It seems less complicated than being sure we definitely copied all the flags etc over.

library/ui/src/main/java/com/google/android/exoplayer2/ui/SubtitleViewUtils.java
dlafayet: Great idea! Will update the PR

icbaker:
Thanks! I'll work on getting this merged.

PR #9023, opened 2021-06-07
Bugfix: DD+JOC track can't be selected if there are multi tracks in manifest file

ybai001:
Bugfix: DD+JOC track can't be selected if there are multi tracks in manifest file

Test content:
DASH: https://ott.dolby.com/OnDelKits/AC-4/Dolby_AC-4_Online_Delivery_Kit_1.5/Test_Signals/example_streams/DASH/OnDemand/MPD/Holi_25fps_example_2.mpd
HLS: https://ott.dolby.com/OnDelKits/AC-4/Dolby_AC-4_Online_Delivery_Kit_1.5/Test_Signals/example_streams/HLS/Manifest_fMP4/Holi_25fps_example_2.m3u8

The reason is same for both DASH and HLS: There is a codec mismatch between DD+JOC content (it is "ec+3" in ExoPlayer internal logic) and manifest parsing (it is "ec-3") when doing track selection. As result, DD+JOC track can't be selected and it is grey in demo application.

I think you can test DASH case on your side if you can find a Dolby enabled device (e.g. Samsung S series.) For HLS, it is a little complex since it is also related to a Dolby DD+JOC decoder implementation issue. (HLS manifest file has property “CHANNELS” and value is “16” and it will result in the channel count capability check fail.) We have delivered patch to OEMs but I guess your test device has not been upgraded to that version. But this fix is needed so that I suggest you merge it.


ybai001:
More info:
It will result in this track (DD+JOC) mimetype is NOT supported in MediaCodecInfo::isCodecSupported(Format format)
 
String codecMimeType = MimeTypes.getMediaMimeType(format.codecs);
……
  if (!mimeType.equals(codecMimeType)) {
      logNoSupport("codec.mime " + format.codecs + ", " + codecMimeType);
      return false;
    }
 
Here, mimeType is “eac3-joc” while codecMimeType is “ec-3”.

ojw28:
@ybai001 - I don't follow your explanation. In particular:

* `Here, mimeType is “eac3-joc” while codecMimeType is “ec-3”.` - The method that determines `codecMimeType` doesn't appear to have any code paths that can return "ec-3", so I don't follow how this is possible.
* `The reason is same for both DASH and HLS: There is a codec mismatch between DD+JOC content (it is "ec+3" in ExoPlayer internal logic) and manifest parsing (it is "ec-3") when doing track selection. As result, DD+JOC track can't be selected and it is grey in demo application.` - This points to a mismatch between ExoPlayer logic for manifest parsing and track selection, but `We have delivered patch to OEMs but I guess your test device has not been upgraded to that version.` suggests there's some kind of underlying platform issue. It's unclear to me what the platform issue is, and how it's related.

Please could you clarify exactly what the problem is that this is fixing?

ybai001:
Hi, @ojw28 

Sorry that I didn't explain the issue well. You can reproduce issue in below two ways:

1. You can use ExoPlayer demo app to play the DASH content I provided. 
DASH: https://ott.dolby.com/OnDelKits/AC-4/Dolby_AC-4_Online_Delivery_Kit_1.5/Test_Signals/example_streams/DASH/OnDemand/MPD/Holi_25fps_example_2.mpd
When you click "SELECT TRACKS" button, you will find DD+JOC audio tracks can't be selected (It is grey.)

2. If you use below code snippet in application code, then its always picking “AAC”, though this URL above has EAC3 JOC signalling

trackSelector.setParameters(
    trackSelector
        .buildUponParameters()
        .setPreferredAudioMimeTypes(MimeTypes.AUDIO_E_AC3_JOC,
                                    MimeTypes.AUDIO_AAC));

ojw28:
What I'm really looking for is a clear explanation of the underlying problem. Specifically, what is the problem that you're fixing in the platform (via the patch that you've delivered to OEMs)? And how does this problem end up surfacing via ExoPlayer to cause the problem with track selection?

ybai001:
> What I'm really looking for is a clear explanation of the underlying problem. Specifically, what is the problem that you're fixing in the platform (via the patch that you've delivered to OEMs)? And how does this problem end up surfacing via ExoPlayer to cause the problem with track selection?

For HLS, the DD+JOC track can't be selected due to two issues. We need to fix both of them then DD+JOC track can be selected.
(1) The issue in ExoPlayer (Fixed in this pull request)
(2) The issue in Dolby decoder implementation. (ExoPlayer checks what's maximum channel count support in device DD+JOC decoder and the result is 8 while the content has channel-count 16 so that ExoPlayer treats this device has no capability to decode this stream. We have delivered the patch to OEMs and it modifies the maximum channel count from 8 to 16 in media_codecs.xml.)

For DASH, only the fix in ExoPlayer is enough.

I'm very sorry to make you confused. Hope this answer can make it clear. (BTW, I need to go to bed now. And I'll reply your futher questions tomorrow morning.) Good luck and have a nice day!

BR,
Yanning

ojw28:
I understand the ExoPlayer issue now. To determine codec support, `MediaCodecInfo.isCodecSupported` is called, with `format.codec = "ec-3"`, which comes from the source content. Within that method:

* `codecMimeType = MimeTypes.getMediaMimeType("ec-3") = "audio/eac3"`
* `mimeType = "audio/eac3-joc"`

Hence  `mimeType.equals(codecMimeType)` is `false`, and so that's what the method returns.

Looking [here](http://mp4ra.org/#/codecs), it looks like `"ec+3"` used to be the correct codec string value, but it was deprecated in favour of using `"ec-3"`. Do you know why that happened? It seems strictly worse, because `"ec+3"` allowed for disambiguation between JOC and non-JOC EAC3 from just the codec string, where-as the new approach does not.

I'm a bit undecided on what the best fix for this is. The pull request proposed here opts to keep using the deprecated type inside of ExoPlayer, which seems pretty strange given it's deprecated. That said, it does seem strictly better for disambiguation, as outlined above. An alternative would be to use `"ec-3"`, accept that disambiguation isn't possible, and add a special case into `isCodecSupported` to handle the mismatch the MIME types for this specific case.

What do you think? It would also be good to understand why an identical codec string is being used for JOC and non-JOC. It doesn't seem like an ideal choice to me?

ojw28:
I think the best approach is probably to return early from `MediaCodecInfo.isCodecSupported` if `mimeType == format.sampleMimeType == MimeTypes.AUDIO_E_AC3_JOC`. Once you've established that, you know that `isCodecSupported` isn't going to perform any useful checks anyway, because the `codec` string doesn't contain any useful additional information.

ybai001:
> I think the best approach is probably to return early from `MediaCodecInfo.isCodecSupported` if `mimeType == format.sampleMimeType == MimeTypes.AUDIO_E_AC3_JOC`. Once you've established that, you know that `isCodecSupported` isn't going to perform any useful checks anyway, because the `codec` string doesn't contain any useful additional information.

I changed the pull request as below (In MediaCodecInfo.java):
`    if (MimeTypes.AUDIO_E_AC3_JOC.equals(mimeType) && MimeTypes.AUDIO_E_AC3.equals(codecMimeType)) {
      return true;
    }`
It works.

In fact, there is a long internal discussion before this pull request is submitted. We discussed how to fix this bug is better. If this special handling can be accepted from your side, I think it is better fix than mine. (Because it avoids two more "ec+3" in code.) I'll explain "ec+3" in another comment. Should I update this pull request? or you will modify it directly?

ybai001:
> Looking [here](http://mp4ra.org/#/codecs), it looks like `"ec+3"` used to be the correct codec string value, but it was deprecated in favour of using `"ec-3"`. Do you know why that happened? It seems strictly worse, because `"ec+3"` allowed for disambiguation between JOC and non-JOC EAC3 from just the codec string, where-as the new approach does not.

This is very complex story actually. I need to find the decision maker and reply to you. My personal understanding is DD+JOC is a "RICH" DD+ format. Legacy DD+ decoder still can decode it as legacy 5.1 DD+ content (no Atmos effect.) The traditional DD+JOC signaling ("ec+3") in manifest will result in some players reject to play this DD+JOC content (ec+3/ec-3 mismatch), which is not what Dolby expects. Anyway, I'll try to find official answer. Stay tuned.

ojw28:
Thanks for the extra information.

> I changed the pull request as below (In MediaCodecInfo.java):
> ```
> if (MimeTypes.AUDIO_E_AC3_JOC.equals(mimeType) && MimeTypes.AUDIO_E_AC3.equals(codecMimeType)) {
>   return true;
> }
> ```
> It works.

This approach calls `MimeTypes.getMediaMimeType(format.codecs)`, which returns the wrong MIME type (because that method has insufficient information to return the correct one now that `ec-3` is used for both E-AC3 and E-AC3-JOC). Your addition is then to have a special case that detects this.

Given `MimeTypes.getMediaMimeType(format.codecs)` is returning the wrong thing, I'd suggest adding the special case before this even happens, so as to avoid the confusion. I think it would work to do:
```
if (MimeTypes.AUDIO_E_AC3_JOC.equals(mimeType)
    && MimeTypes.AUDIO_E_AC3_JOC.equals(format.sampleMimeType)) {
  return true;
}
```
Do you see any downsides of that approach? I think it's basically saying: "this is an E-AC3-JOC decoder, and the format is an E-AC-JOC format, and in this case I know that the `codecs` string isn't going to tell me any additional information for determining support, so I'm just going to return true".

The `isCodecSupported` method name is a bit confusing, but you can leave that for us to try and clean up. It's already using `sampleMimeType` in the call it makes to `MediaCodecUtil.getCodecProfileAndLevel(format)`, so there's no particularly strong reason to avoid doing this, that I can think of.

If you agree with this approach, please update the pull request and we'll get it merged. Thanks!

ojw28:
Other cases seem to assume that `isFormatSupported` is only called with formats whose sample MIME types are suitable for the decoder, so you can probably simplify the special case down to simply:

```
if (MimeTypes.AUDIO_E_AC3_JOC.equals(mimeType)) {
  return true;
}
```

ybai001:
> Other cases seem to assume that `isFormatSupported` is only called with formats whose sample MIME types are suitable for the decoder, so you can probably simplify the special case down to simply:
> 
> ```
> if (MimeTypes.AUDIO_E_AC3_JOC.equals(mimeType)) {
>   return true;
> }
> ```

I updated the pull request as this modification. It is OK for MPEG-DASH. But I found a side effect for HLS stream.
The media codec info on device is as below:
```
        <MediaCodec name="c2.dolby.eac3.decoder" >
            <Type name="audio/ac3">
                <Alias name="OMX.dolby.ac3.decoder" />
                <Limit name="channel-count" max="6" />
                <Limit name="sample-rate" ranges="32000,44100,48000" />
                <Limit name="bitrate" range="32000-640000" />
            </Type>
            <Type name="audio/eac3">
                <Alias name="OMX.dolby.eac3.decoder" />
                <Limit name="channel-count" max="8" />
                <Limit name="sample-rate" ranges="32000,44100,48000" />
                <Limit name="bitrate" range="32000-6144000" />
            </Type>
            <Type name="audio/eac3-joc">
                <Alias name="OMX.dolby.eac3-joc.decoder" />
                <Limit name="channel-count" max="16" />    // Note: we changed it from 8 to 16 and delivered it to OEM.
                <Limit name="sample-rate" ranges="48000" />
                <Limit name="bitrate" range="32000-6144000" />
            </Type>
            <Attribute name="software-codec" />
        </MediaCodec>
```

If the original patch is applied, the DD+JOC track can be selected. But for the new patch (modifying the logic in MediaCodecInfo.java), since codec is "ec-3" rather than "ec+3", the decoder with type name "audio/eac3" is selected and the maximum channel count support is 8. And the track selection fails in 
```
  public boolean isAudioChannelCountSupportedV21(int channelCount) {
    ......
    int maxInputChannelCount =
        adjustMaxInputChannelCount(name, mimeType, audioCapabilities.getMaxInputChannelCount());
    if (maxInputChannelCount < channelCount) {
      logNoSupport("channelCount.support, " + channelCount);
      return false;
    }
    ......
  }
```
As result, on track selection UI, DD+JOC tracks are grey.

If I modify device media codec info by changing channel count to 16 for type name "audio/eac3" too, the new patch can work for HLS. I'm thinking whether there is a better solution in ExoPlayer. I'll also discuss with team whether we should change media_codecs.xml (channel count 8 to 16 for "audio/eac3" too) Any suggestion?

ojw28:
I see the problem. There are a couple of different places in the HLS module that convert `codecs` back to MIME type again. Using `ec-3` means that we convert back to `audio/eac3` rather than `audio/eac3-joc`. This MIME type is then exposed in the track, and is also used for decoder selection.

Continuing along the path I've been proposing would involve adding special cases at places where this happens, to account for the fact that we cannot unambiguously map `ec-3` to the correct MIME type. But since it's happening in multiple places in the HLS module, this starts to get quite error prone. In light of this, let's go with your original fix where you were adjusting `codecs` to use `ec+3` in the parsers. It seems the ability to unambiguously map back to MIME type is worth using the deprecated codec string!

Apologies for the back and forth. Good to have the discussion though :).

ybai001:
> I see the problem. There are a couple of different places in the HLS module that convert `codecs` back to MIME type again. Using `ec-3` means that we convert back to `audio/eac3` rather than `audio/eac3-joc`. This MIME type is then exposed in the track, and is also used for decoder selection.
> 
> Continuing along the path I've been proposing would involve adding special cases at places where this happens, to account for the fact that we cannot unambiguously map `ec-3` to the correct MIME type. But since it's happening in multiple places in the HLS module, this starts to get quite error prone. In light of this, let's go with your original fix where you were adjusting `codecs` to use `ec+3` in the parsers. It seems the ability to unambiguously map back to MIME type is worth using the deprecated codec string!
> 
> Apologies for the back and forth. Good to have the discussion though :).

I have restored the patch to the original one. I truly learned a lot from this discussion, especially your idea about how to modify the code for maintainability. It helps me to think this issue more.

And eac3/eac3-joc is also a problem we met many times in our daily work. So that it is not strange for me to discuss it back and forth. :)

PR #9050, opened 2021-06-11
HLS: Allows playback through gap > maxBufferMs

### LINKED ISSUE #8959: Playback stops at EXT-X-GAP tag if GAP exceeds buffering time
If there is a gap across all renditions that lasts longer than the buffered time limit (50s default) the player stops indefinitely in the buffering state.   This is of course because buffering is calculated based on the playlist not the samples yet rendering depends on the samples.

This issue is related to #6087 (in that if the gap is smaller than 50s playback simply skips the duration of the audio gap)

### Reproduction Steps

1. Play a VOD sample with EXT-X-GAP duration longer than 50s in all renditions
2. Observe playback stalls in buffering forever

I have sent two sample stream URL's with a gap in all renditions starting at 1:18 seconds into the stream to the Exo developers email address.

#### Use Cases
Gaps in the stream happen in broadcast streams for a couple of reasons:
1.  **PID drops** &mdash;  the source transport stream drops an audio PID, this can happen for alternate language tracks
2. **Outages** &mdash; drops in the multicast feed, issues with satellites, transcoding broadcast stream.  All of these can result in gaps across all renditions.  Very possible this can exceed buffered time

ExoPlayer's current behavior (Skipping the media time with an instantaneous jump) is quite reasonable for the *Outages* use case (logging an analytics event would be nice to allow the UX of showing a whisper banner indicating an outage caused the jump).  The use cases for *PID drops* are adequately covered in issue #6087 

#### Suggested Fix

The `SequenceableLoader` interface needs to understand one or all of its constituent loaders are experiencing a GAP and continue loading across the gap until the discontinuity (first sample after the GAP) is loaded.  Here the decision to jump past the discontinuity or play through (if only one rendition has a GAP) can be intelligently made.    This may be specific to HLS as DASH appears to be looking at a different approach see issue #4616

stevemayhew:
This change fixes #8959

The issue is caused because the current logic assumes it is fully buffered after
fetching `minBufferMs` worth of GAP segments, which of course contain no
samples.  ExoPlayer normally jumps past the GAP but the infinite buffering
stall prevents this.

The fix changes the HLS implementation of `SequenceableLoader.getBufferedPositionUs()`
so to not report a GAP segment as potentially adding to the buffered time.
This allows deferring to the `SampleQueue`, a more complete source of truth.

Note the jump to end of the GAP occurs with no buffer, and loading has not triggered
so at least one handler loop is required to prevent false trigger of the
"Playback stuck buffering and not loading" error.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -2312,8 +2312,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
       return false;
     }
     MediaPeriodHolder loadingPeriodHolder = queue.getLoadingPeriod();
+    long bufferedPositionUs = loadingPeriodHolder.getBufferedPositionUs();
     long bufferedDurationUs =
-        getTotalBufferedDurationUs(loadingPeriodHolder.getNextLoadPositionUs());
+        getTotalBufferedDurationUs(bufferedPositionUs);
     long playbackPositionUs =
         loadingPeriodHolder == queue.getPlayingPeriod()
             ? loadingPeriodHolder.toPeriodTime(rendererPositionUs)
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java
@@ -413,6 +413,15 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     isPublished = true;
   }
 
+  /**
+   * If their are any samples loaded in this chunk.
+   *
+   * @return true if samples were loaded
+   */
+  boolean hasSamples() {
+    return ! hasGapTag;   // TOOD - could expand upon this to be more accurate, but for sure GAP has none
+  }
+
   // Internal methods.
 
   @RequiresNonNull("output")
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
@@ -715,7 +715,7 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
           lastMediaChunk.isLoadCompleted()
               ? lastMediaChunk
               : mediaChunks.size() > 1 ? mediaChunks.get(mediaChunks.size() - 2) : null;
-      if (lastCompletedMediaChunk != null) {
+      if (lastCompletedMediaChunk != null && lastCompletedMediaChunk.hasSamples()) {
         bufferedPositionUs = max(bufferedPositionUs, lastCompletedMediaChunk.endTimeUs);
       }
       if (sampleQueuesBuilt) {


stevemayhew:
Feel free to simply take this, change it around a bit or suggest a completely different approach.

I've done developer testing on the change with the sample URL I sent with the bug, it skips the gap now fine.  

There is one additional bug I'm looking at where the gap jumps cause the check for AudioTrack timestamps to report an error (again probably a false error because the actual time is exactly the end of the GAP):

````
com.google.android.exoplayer2.audio.AudioSink$UnexpectedDiscontinuityException: Unexpected audio track timestamp discontinuity: expected 78250666, got 130129999
````

File: library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsMediaChunk.java [hunk 1]
```diff
@@ -400,6 +400,15 @@ public void publish() {
...
 
+  /**
+   * If their are any samples loaded in this chunk.
+   *
+   * @return true if samples were loaded
+   */
+  boolean hasSamples() {
+    return ! hasGapTag;   // TOOD - could expand upon this to be more accurate, but for sure GAP has none
+  }
+
```
stevemayhew, line 424: In theory, we could have the parsing state with committed samples as well maybe, thread safety, as this is called from the playback thread.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -671,7 +671,7 @@ public long getBufferedPositionUs() {
           lastMediaChunk.isLoadCompleted()
               ? lastMediaChunk
               : mediaChunks.size() > 1 ? mediaChunks.get(mediaChunks.size() - 2) : null;
-      if (lastCompletedMediaChunk != null) {
+      if (lastCompletedMediaChunk != null && lastCompletedMediaChunk.hasSamples()) {
         bufferedPositionUs = max(bufferedPositionUs, lastCompletedMediaChunk.endTimeUs);
```
stevemayhew, line 719: Idea is pretty basic, The sample stream wrapper should not proactively report the last loaded chunk (segment) as having buffered time if it is known to have no samples (e.g. an EXT-X-GAP tag has no media to load, so it cannot add to the buffered time).

This way, the largest queued timestamp (below on line 679) will take precedence for the buffered time.

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -1046,7 +1048,9 @@ private void doSomeWork() throws ExoPlaybackException, IOException {
       }
       if (!playbackInfo.isLoading
           && playbackInfo.totalBufferedDurationUs < 500_000
-          && isLoadingPossible()) {
+          && isLoadingPossible()
+          && stuckLoadingRetryCount++ >= 3) {
+        stuckLoadingRetryCount = 0;
```
stevemayhew: Once the player jumps past the gap it needs one more handler message (the load request one is pending here).  If we wanted to be more pedantic about it one could check the message queue was devoid of anything else (seek, etc.) that might clear the stuck-ness before triggering this error.

BTW, @ojw28 still working on the check in queuing samples for un-reported discontinuities.   In our experience that is the largest reason why this set of conditions occur in the render loop.  That code will require some re-work depending on how we fall on reverting back to only allowing video timestamp master.

stevemayhew: This stuckLoading issue is no longer possible with GAP.  See my last commit b759e28fc 
The problem is `shouldContinueLoading()` is not using the correct method to compute buffered duration.

@ojw28 Could this be more of the stuck while loading is possible issues?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: @christosts Why do we ever care about the `endTimeUs`, at least as long as `sampleQueuesBuilt`

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -2312,8 +2312,9 @@ private boolean shouldContinueLoading() {
       return false;
     }
     MediaPeriodHolder loadingPeriodHolder = queue.getLoadingPeriod();
+    long bufferedPositionUs = loadingPeriodHolder.getBufferedPositionUs();
     long bufferedDurationUs =
-        getTotalBufferedDurationUs(loadingPeriodHolder.getNextLoadPositionUs());
+        getTotalBufferedDurationUs(bufferedPositionUs);
     long playbackPositionUs =
```
stevemayhew, line 2318: One line change to use `getBufferedPositionUs()`.  Local variables is for logging added to commit 78cf217c8f

christosts:
Apologies for the silence @stevemayhew, we're pretty busy focused getting [media3](https://github.com/androidx/media) stable out. Can I get back this PR once we start working towards ExoPlayer 2.19/media3 1.1.0, where this PR will be included anyway?

stevemayhew:
Absolutely, I can close it and move it to media3.  I posted a question over there if there are any tools or best practices for doing this and preserving history.

I'm thinking to do the packaging move on a branch here in our ExoPlayer fork then moving it somehow to media3.  If you guys have any tips on how you did that we'd appreciate it and just move our pull requests there 

Thanks @christosts, and no worries on quick replies we're both pressed for schedules so totally understand 

We're shipping with these changes in our current 2.15.1 release so will watch for issues and keep this pull up to date

christosts:
> I posted a question over there if there are any tools or best practices for doing this and preserving history.

@marcbaechinger is the best person to answer https://github.com/androidx/media/issues/217 and I see he's already on it :) 

> We're shipping with these changes in our current 2.15.1 release so will watch for issues and keep this pull up to date

Thank you very much for your contributions

sebastiangansca:
@stevemayhew Thank you for this changes and PR. We are using it with Exo 2.18.1 and it works perfectly for our use cases in our project. 

However we've identified one use case that is not covered and I want to ask if you can help on that also. 

We have also VAST for ads played with Exo. 

Problem is: 
If there is a stream with gap in the first segment and there is ad played before stream is started then the player remains stuck in buffering. Manual skip to any position will fix the issue with this PR changes as player will skip the gap also and resumes from first available segment. 

Thanks!

icbaker:

stevemayhew:
@icbaker thanks for the heads up, we definitely need to re-open this under Androidx as we are shipping with this change for over a year.  Hopefully we can get more traction towards reviewing and merging it there.

@sebastiangansca VAST playback is not a use case we have tested with pre-roll, we use VAST to play trailers with no content  (a `SilenceMediaSource`) with the extension-ima plugin.  I would open a bug with AndroidX/Media3 with your use case (not a big migration from 2.18.1) once I have the pull request migrated then I can comment on it.  Kind of pointless here as the ExoPlayer project is deprecated.

Issue #7954, opened 2020-09-21
Crash on Android 5.1 when try to release player


WagesOfSin:
### [REQUIRED] Issue description
Updated on the latest version 2.12.0 and I have crash on Android 5, Android 6 when I try to release the player. Don't have this crash on version 2.11.8. I can't reproduce this issue on your demo app. 

### [REQUIRED] Reproduction steps
1) Open the app with player.
2) Release player.

### [REQUIRED] Link to test content
It doesn't mater 

### [REQUIRED] A full bug report captured from the device
    java.lang.NoClassDefFoundError: com.google.android.exoplayer2.-$$Lambda$ExoPlayerImplInternal$vdH6YQpqyxP6KbK0sRxQC-SGllM
        at com.google.android.exoplayer2.ExoPlayerImplInternal.release(ExoPlayerImplInternal.java:402)
        at com.google.android.exoplayer2.ExoPlayerImpl.release(ExoPlayerImpl.java:708)
        at com.google.android.exoplayer2.SimpleExoPlayer.release(SimpleExoPlayer.java:1710)
        at com.everydoggy.android.presentation.view.adapter.viewholders.VideoTypeViewHolder.releasePlayer(VideoTypeViewHolder.kt:164)
        at com.everydoggy.android.presentation.view.adapter.viewholders.VideoTypeViewHolder.access$releasePlayer(VideoTypeViewHolder.kt:47)
        at com.everydoggy.android.presentation.view.adapter.viewholders.VideoTypeViewHolder$observer$1.onStateChanged(VideoTypeViewHolder.kt:254)
        at androidx.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:361)
        at androidx.lifecycle.LifecycleRegistry.backwardPass(LifecycleRegistry.java:316)
        at androidx.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:334)
        at androidx.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:145)
        at androidx.lifecycle.LifecycleRegistry.handleLifecycleEvent(LifecycleRegistry.java:131)
        at androidx.fragment.app.FragmentViewLifecycleOwner.handleLifecycleEvent(FragmentViewLifecycleOwner.java:51)
        at androidx.fragment.app.Fragment.performPause(Fragment.java:2874)
        at androidx.fragment.app.FragmentStateManager.pause(FragmentStateManager.java:383)
[...]

Issue #8821, opened 2021-04-16
updating guava


umerov1999:
Can i update guava from 27.1-android to 30.1.1-android? Does Exoplayer work correctly, if i update guava?

icbaker:
ExoPlayer doesn't use any Guava APIs marked `@Beta`. Guava's commitment to binary compatibility from 21.0 onwards means that ExoPlayer should run fine against our current declared dependency (27.1-android) and any future versions: https://github.com/google/guava/wiki/PhilosophyExplained#non-beta-apis

However we deploy this library into a context where 27.1 is the newest version available - so for now we're not going to change the version we compile against (in order to prevent accidentally depending on any APIs added after 27.1):
https://github.com/google/ExoPlayer/blob/0ba317b1337eaa789f05dd6c5241246478a3d1e5/constants.gradle#L24

Hopefully that answers your question - let me know if not.

Issue #8182, opened 2020-11-07
NoClassDefFoundError Mp4Extractor in Android 5.X and 6.X versions


idish:
### [REQUIRED] Issue description
Trying to prepare a media file causes a NoClassDefFoundError fatal crash on (seemingly all) Android 5.X and Android 6.X devices.
I'm able to reproduce that crash with an Android 6.0 emulator, but not on 7.0 + as expected.
Min api is 21, target API level is 29.

### [REQUIRED] A full bug report captured from the device
`E/LoadTask: Unexpected error loading stream
      java.lang.NoClassDefFoundError: com.google.android.exoplayer2.extractor.mp4.-$$Lambda$Mp4Extractor$IP-_bKEryH3151zsYMHu9R1OlAQ
        at com.google.android.exoplayer2.extractor.mp4.Mp4Extractor.processMoovAtom(Mp4Extractor.java:418)
        at com.google.android.exoplayer2.extractor.mp4.Mp4Extractor.processAtomEnded(Mp4Extractor.java:381)
        at com.google.android.exoplayer2.extractor.mp4.Mp4Extractor.readAtomPayload(Mp4Extractor.java:372)
        at com.google.android.exoplayer2.extractor.mp4.Mp4Extractor.read(Mp4Extractor.java:195)
        at com.google.android.exoplayer2.source.BundledExtractorsAdapter.read(BundledExtractorsAdapter.java:127)
        at com.google.android.exoplayer2.source.ProgressiveMediaPeriod$ExtractingLoadable.load(ProgressiveMediaPeriod.java:1046)
        at com.google.android.exoplayer2.upstream.Loader$LoadTask.run(Loader.java:415)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
        at java.lang.Thread.run(Thread.java:818)`
### [REQUIRED] Version of ExoPlayer being used
version: 2.12.1

### [REQUIRED] Device(s) and version(s) of Android being used
Android 5.X, Android 6.X devices

kim-vde:
Did you enable [Java 8 support](https://exoplayer.dev/hello-world.html#turn-on-java-8-support)? Are you importing the extractors library with the same ExoPlayer version as the core library? Does the crash also occur in our [demo app](https://github.com/google/ExoPlayer/tree/release-v2/demos/main)?

[...]

PR #9073, opened 2021-06-17
Update guava version to latest

Merlinkoss:
Fix issue https://github.com/google/ExoPlayer/issues/7954. Update guava to latest to project, which do not prefer to use resolution conflict on top of Gradle lvl


icbaker:
It looks like #7954 didn't conclusively identify the Guava version as the cause of that crash. https://github.com/google/ExoPlayer/issues/8182#issuecomment-726429302 suggests it's caused by the `missing-metadata-guava` plugin, which was a suggested fix for a Guava-related problem present in 2.12.0 but fixed/worked-around in 2.12.1 (https://github.com/google/ExoPlayer/commit/a635bdd4ba61075971432108c341ce0c5d2a6f24) - and not something ever used directly by ExoPlayer.

Please see the reasoning here for why we'd like to stick to Guava 27.1 if possible: https://github.com/google/ExoPlayer/issues/8821#issuecomment-821197003

If you can provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) that demonstrates the problem in a way that we can build locally and clearly shows the version of Guava causes the problem then we will reconsider this PR, but based on the current evidence I'm afraid we're unlikely to take it.

This could be an Android Studio project on GitHub, or zipped up and sent to dev.exoplayer@gmail.com using a subject in the format "Issue #1234", where "#1234" should be replaced with your issue number. Please also update this issue to indicate you’ve done this.

Issue #9026, opened 2021-06-08
[Cast] CastTimeline stays empty on reconnect


inv3rse:
The `CastTimeline` initially stays empty when reconnecting to a cast device even though it has an active `MediaQueueItem`. 
Only after a state change (play / pause) will the timeline be populated.

How to reproduce:
1. Launch the sender app and cast something to a device.
2. Pause the content (not required but helps to avoid extra state changes from buffering etc.).
3. Completely restart the sender app (re-launch through Android Studio).
4. Wait for the automatic re-connect to the cast or do so manually.
5. The `currentTimeline` of the  `CastPlayer` stays empty (as long as there is no other interaction with the cast).

Expected behavior:
- The timeline contains the currently playing item.

The issue is also observable in the demo-cast project if one adds logging and changes the timing a bit: 
https://github.com/inv3rse/ExoPlayer/commit/ec6e303de1185095559afd85e3a7947c9e2de915

The issue appears to be the usage of `remoteMediaClient.getMediaQueue().getItemIds()` in the `CastTimelineTracker`. 
After a reconnect it will always return an empty array until there are further state changes on the cast.

I am not sure about the internals of the MediaQueue, but simply getting the ids from the MediaStatus instead appears to be working fine: https://github.com/google/ExoPlayer/pull/9078

ExoPlayer version: 2.14.0
Android version: 10
Android device: Galaxy S9

AquilesCanta:
This is something we've been looking into for some time.

The problem we have is that we cannot reliably get the entire timeline from a receiver. We might be able to get the number of items from it, but we might be missing other stuff like MediaItem info, duration, etc. We need support from the Cast SDK to do this properly [Internal tracker: b/182255537], and we don't currently have the resources to implement workarounds ourselves. 

For the time being, we've opted to clear the receiver queue when connecting to an existing session, so the cast extension currently supports a single sender device. However, if you think you could contribute a reasonably high quality solution for this problem, you're very welcome.

AquilesCanta:
@marcbaechinger Assigning to you in the context of [Internal ref: b/182255537]. Hope that's ok.

inv3rse:
I have looked a bit more into it, but it seems like with the workaround all the other info is also set appropriately.  
The position and duration is correct, and the MediaItem is also there. 

I can not see the internal tracker, so I might be missing something. 
But since it looked fine to me I created a pull request.

PR #9078, opened 2021-06-18
Fix empty CastTimeline on reconnect

inv3rse:
After connecting to a cast device that is already playing, the `CastTimeline` would initially stay empty.  
And in turn all methods that rely on the timeline would also report incorrect data.  

The reason for this is that the `MediaQueue` would always return an empty array as active ids until there are further state changes on the cast. This does appear to be an issue in the cast framework, but it can easily be worked around by checking the `MediaStatus` instead.

Fixes https://github.com/google/ExoPlayer/issues/9026


inv3rse:
I resolved the merge conflicts and checked the behavior again. 
Without this change I could still reproduce the issue as described in https://github.com/google/ExoPlayer/issues/9026#issue-915027097. 
With this change the timeline was populated as expected.

inv3rse:
@marcbaechinger can you have a look?

icbaker:

PR #9115, opened 2021-06-25
Support cast live seekable ranges.

inv3rse:
Currently the seekable range of live streams is not exposed in the `CastPlayer`.
The duration is always reported as `C.TIME_UNSET`.

This PR will check the media status of the active item for a `LiveSeekableRange`.
The range is used to create the timeline with a corresponding seekable window at the start of the range.
Because the range might not start at 0 the seek method is now adding the optional window offset.


icbaker:

PR #9119, opened 2021-06-26
Use better builtin ControlDispatcher methods in media2 extension

### LINKED ISSUE #9084: Use `ControlDispatcher` for dispatching next/previous in media2 extension
[The current implementation](https://github.com/google/ExoPlayer/blob/release-v2/extensions/media2/src/main/java/com/google/android/exoplayer2/ext/media2/PlayerWrapper.java#L238) calculates the previous position itself and uses `ControlDispatcher.seekTo` to seek to it. This is different from [the default behaviour in the mediasession extension](https://github.com/google/ExoPlayer/blob/bd54394391b0527893f382c9d641b8a55ffca765/extensions/mediasession/src/main/java/com/google/android/exoplayer2/ext/mediasession/TimelineQueueNavigator.java#L148). 

I was wondering why this decision was taken, since this makes it impossible (or at least a **lot** harder than it should be) to override the behaviour of skip actions when using the media2 extension. The same question can be asked for more than just skip to previous, the behaviour of skip to next and other related methods should also be overridable, IMO (or at least, it should be clear why this was not made possible).

To avoid an [XY problem](https://en.wikipedia.org/wiki/XY_problem), what I really want to do is make it so that when the skip to previous action happens if a track has been playing for a few seconds, that it skips back to the start of the track instead of skipping to the previous track. I've found [SkipCallback](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/ext/media2/SessionCallbackBuilder.SkipCallback.html), but the problem there is that I didn't find any way to call the [skipBackward](https://developer.android.com/reference/androidx/media2/session/MediaController#skipBackward()) method from a notfication (or wearables, or OS-level controls, ...). Those all use the `skipToPreviousItem` action. Note that this worked out of the box when using the mediasession extension, so I assume that if `ControlDispatcher.dispatchPrevious(Player)` would be used there, it would remove the need for me to start messing around with it.

chvp:
Fixes #9084.

CLA should be signed.

### DIFF
--- a/extensions/media2/src/main/java/com/google/android/exoplayer2/ext/media2/PlayerWrapper.java
+++ b/extensions/media2/src/main/java/com/google/android/exoplayer2/ext/media2/PlayerWrapper.java
@@ -233,21 +233,13 @@ import java.util.List;
   public boolean skipToPreviousPlaylistItem() {
     Timeline timeline = player.getCurrentTimeline();
     Assertions.checkState(!timeline.isEmpty());
-    int previousWindowIndex = player.getPreviousWindowIndex();
-    if (previousWindowIndex != C.INDEX_UNSET) {
-      return controlDispatcher.dispatchSeekTo(player, previousWindowIndex, C.TIME_UNSET);
-    }
-    return false;
+    return controlDispatcher.dispatchPrevious(player);
   }
 
   public boolean skipToNextPlaylistItem() {
     Timeline timeline = player.getCurrentTimeline();
     Assertions.checkState(!timeline.isEmpty());
-    int nextWindowIndex = player.getNextWindowIndex();
-    if (nextWindowIndex != C.INDEX_UNSET) {
-      return controlDispatcher.dispatchSeekTo(player, nextWindowIndex, C.TIME_UNSET);
-    }
-    return false;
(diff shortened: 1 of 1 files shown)


extensions/media2/src/main/java/com/google/android/exoplayer2/ext/media2/PlayerWrapper.java
```diff
@@ -233,21 +233,13 @@ public boolean movePlaylistItem(
   public boolean skipToPreviousPlaylistItem() {
     Timeline timeline = player.getCurrentTimeline();
     Assertions.checkState(!timeline.isEmpty());
```
ojw28: Is there any value in this assertion? It doesn't look like other components (e.g., `PlayerControlView`) bother. If you can't see any value either, I'd suggest removing them (ditto for line `skipToNextPlaylistItem` below).

chvp: I don't think they have any value either. Removed.

ojw28: Thanks. We will get this merged in.

PR #9144, opened 2021-07-02
Optimize jvmargs in gradle.properties

Goooler:
[AGP 4.2 Optimizing Gradle builds for JDK 11](https://developer.android.com/studio/releases/gradle-plugin#kts)

image

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,4 +2,4 @@
 android.useAndroidX=true
 android.enableJetifier=true
 buildDir=buildout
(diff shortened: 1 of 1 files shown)


ojw28:
This commit appears to be changing three args, but the commit message only (partially) explains one of them. Could you provide a more detailed explanation of what you're doing? Also, did you actually profile the `+UseParallelGC` option and find it to be faster? If so, could you share the results?

PR #9162, opened 2021-07-08
Added DTS:X Audio Mime Type

ProtoScott:
For dtsx mimetype to be IANA compliant using vnd.dts.uhd instead of vnd.dts.x


krocard:
Thanks for the pull request. Nevertheless we don't usually list formats that are not handled by ExoPlayer.
Why do you need this value if ExoPlayer doesn't support it?

krocard:
Additionally `vnd.dts.uhd` has a required parameter `profile`. I believe only `profile=p2` corresponds to `dtsx` so I don't think the pull request is correct.
Source: https://www.iana.org/assignments/media-types/audio/vnd.dts.uhd

ProtoScott:
Hi @krocard 
We have written an extension library to support dtsx and are looking to use the MimeType class from the exoplayer core library.
The specific MimeType in the PR was specified by the Xperi (DTS:X) team

ProtoScott:
Closing for alternative PR: https://github.com/google/ExoPlayer/pull/9163

Issue #7975, opened 2020-09-23
Start position of 0 always resolves to default position for live playbacks


DavidMihola:
### [REQUIRED] Searched documentation and issues
Since the problem only appeared after updating from 2.11.8 to 2.12.0 I searched the recent issues for "seek" and "startPosition" but didn't find anything that seemed related.

### [REQUIRED] Question
In our TV app we offer users the possibility to watch the current program from the beginning, i. e. instead of simple live stream with a short live window we have a stream with a fixed start but which grows in length every time a new segment is available on the streaming server.

Since the default position in those streams seems to be right at the live edge we have always passed the start position (`simpleExoPlayer.seekTo(0)`) explicitly before `prepare()`ing the player. Up until and including 2.11.8 this has worked without any problems.

Since updating to 2.12.0 a few days ago this "initial seek position" seems to be ignored and the player starts at the default position, at the live edge. I saw that the `prepare(mediaSource, false, true)` which we were using so far is now deprecated. But replacing it with

```
simpleExoPlayer.setMediaSource(buildMediaSource(), startPositionInMs)
simpleExoPlayer.prepare()
```

does not resolve the problem. Even here, it seems that the `startPositionInMs` is ignored.

Unfortunately, I cannot provide a public test stream to demonstrate the problem, but this is the HLS master playlist:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:BANDWIDTH=1963735,PROGRAM-ID=1,CODECS="avc1.4d4028,mp4a.40.2",RESOLUTION=960x540
04(start=1600850100_stop=1600864500).m3u8
#EXT-X-STREAM-INF:BANDWIDTH=4636146,PROGRAM-ID=1,CODECS="avc1.4d4028,mp4a.40.2",RESOLUTION=1280x720
01(start=1600850100_stop=1600864500).m3u8
#EXT-X-STREAM-INF:BANDWIDTH=3055720,PROGRAM-ID=1,CODECS="avc1.4d4028,mp4a.40.2",RESOLUTION=1280x720
02(start=1600850100_stop=1600864500).m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2340502,PROGRAM-ID=1,CODECS="avc1.4d4028,mp4a.40.2",RESOLUTION=1280x720
03(start=1600850100_stop=1600864500).m3u8
[...]

PR #9174, opened 2021-07-09
Test case for fix for issue #7975

stevemayhew:
First steps for fixing issue #7975 a unit test case[s] that show the bug.   

The first commit, https://github.com/google/ExoPlayer/pull/9174/commits/b9d959cdcdd367d2dabc09bd74f3492c3918c5bf is a functional test, if the fix goes outside of `MaskingMediaSource` itself then we'll need this test case

The next commit, https://github.com/google/ExoPlayer/pull/9174/commits/ac044ae60e is a unit test case for`MaskingMediaSource` itself, if the code change to fix involves how the position override works in this class then this test case should be all we need.

Looking at the logic in [line 2575](https://github.com/google/ExoPlayer/blob/3f5dbf2ef3fe19f392d46027b7ffce0783d28427/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java#L2575)

````java
        } else {
          newPeriodUid = periodPosition.first;
          newContentPositionUs = periodPosition.second;
          // Use explicit initial seek as new target live offset.
          setTargetLiveOffset = true;
        }
````

The initial seek is resolved here, against the `PlaylistTimeline` this happens when the initial `MedaiItem`[s] are set so we have no idea even the duration of the window for the live playlist.   

Also, I believe using this as the target live offset may be a mistake, the usual use case for setting a position in a live playlist with an initial seek is a SOCU (start-over-catchup) where live is buffered to a rolling buffer, in this case you want to treat it more like VOD (the rolling buffer playlist is typically labeled "EVENT" type until the program ends at which point an end tag is added.  

So, there are these basic use cases for live:

1. No initial seek/position set -- default to the playlist requested live offset (default or EXT-X-START)
1. Initial position (usually 0) requested -- start a SOCU (here the playlist duration will be the entire event up to present time, the playlist live point is irrelevant)
1. Saved position ( 0 - duration ) -- return to a saved position in a SOCU playlist.  Here the user has navigated away from playback but returns via a "My Shows" list to the in-progress SOCU recording

The most strait forward fix I see is to make the initial `preparePositionUs` in the `MaskingMediaPeriod` / `MaskingMediaSource` have an unset value unless an initial seek or other explicit user action set it.  This frees up 0 from serving as the default value as well as a possible user choice.

@tonihei and @marcbaechinger you both have way more invested in this code and no doubt a bunch of refactoring plans I'm unaware of, so I don't want to add more test cases to code you may be refactoring away.   As long as the first commit (the functional test) passes with 0 as the seek value the requirements are met and bug #7975 can likely be closed.

### DIFF
--- a/library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
+++ b/library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
@@ -9374,6 +9374,54 @@ public final class ExoPlayerTest {
   }
 
   @Test
+  public void seekTo_toPositionZero_overridesLiveStartPosition()
+          throws Exception {
+    ExoPlayer player =
+        new TestExoPlayerBuilder(context)
+            .setClock(
+                new FakeClock(/* initialTimeMs= */ 20_000, /* isAutoAdvancing= */ true))
+            .setUseLazyPreparation(true)  // This is default for ExoPlayer.Builder()
+            .build();
+
+    // TODO - if you set expectedInitialPositionUs to 0, the test will fail... And it shouldn't
+    int expectedInitialPositionUs = 10_000;
+    player.seekTo(expectedInitialPositionUs);
+    FakeMediaSource mediaSource = new FakeMediaSource();
+    player.setMediaSource(mediaSource, false);
+    player.prepare();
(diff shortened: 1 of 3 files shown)


cyberwick011:
Please release apk version for this player, I don't know anything when it comes to zip files, please release an apk file for the next update 🙏

icbaker:

PR #9176, opened 2021-07-12
Updating WorkManager from 2.5.0 to 2.7.0-alpha03

seventhmoon:
There is a behavior change in Android 12 which requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent if the app is targeting Android 12. Current stable versions of WorkManager don’t have these flags, and so it affects all apps directly, or indirectly using WorkManager. ExoPlayer's extension has a dependency on an older version of WorkManager and so is affected.

Updating the version 2.7.0-alpha03 to solve this issue.


seventhmoon:
Closing this PR.
Submitted #9180 to dev branch.

PR #9196, opened 2021-07-15
Ensure EventSampleStream is updated if EventStreams disappeared from period

jruesga:
In case an `EventStream` with Events was present in a period and in a later update the EventStream is removed from the manifest, the `EventSampleStream` will never be updated with a false `eventStreamAppendable` flag, and it will EventSampleStream will never marked as ended and will continue reporting `RESULT_NOTHING_READ` indefinitelly (`BUFFER_FLAG_END_OF_STREAM` never is set) and the playback will be stuck because of still waiting this buffer to end.

xe: 

```
Initial manifest
=====
Period 1
  EventStream 2
Period 2
  EventStream 2

Fist manifest update
=====
Period 1
Period 2
  EventStream 2

Second manifest update
=====
Period 1
Period 2
  EventStream 2

... and so on
```

with this sequence, `eventStreamAppendable` in `EventSampleStream` is never called and player will be stucked when Period 2 is reached.

This commit ensures that if there is an `EventSampleStream` read previous a `EventStream` and that one is no present anymore in the new manifest update, then readd the old reported one to ensure `EventSampleStream` consistency when a new period is reached (if no more new EventStream came into the stream).

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/DashMediaPeriod.java
@@ -61,6 +61,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


jruesga:
@googlebot I signed it!

ojw28:
The DASH specification doesn't appear to indicate either way whether removal of an `EventStream` from a `Period` is permitted, but this feels like a strange thing to do. I suspect that it should be forbidden, and that if this is not specified anywhere, it's an omission.

Why would it make sense to completely remove the `EventStream`, rather than leaving it there without any `Event` children? Note that the former retains signalling that there were events in the period, where-as the latter does not. The former also seems more consistent with use of `InbandEventStream`, which would not generally be removed after than final event (because the component constructing the manifest would probably not have a way of determining that there are no more in-band events). And more consistent with not removing other elements from `Period`, such as adaptation sets (and their contained representations).

jruesga:
Thanks Oliver, I'll check as well with our encoder provider.

ojw28:
@jruesga - Did you manage to resolve this on the manifest generation side? I'm hoping we can close this pull request.

jruesga:
yes, we managed to workaround the issue.

PR #9207, opened 2021-07-20
Fix bug where rubyPosition in text node is not applied

dlafayet:
Also fix bug where rubyPosition in ruby container style is not applied
@icbaker


icbaker:
Looks good, thanks! I'll get this merged.

PR #9281, opened 2021-08-10
8906-Prevented negative seek position

### LINKED ISSUE #8906: Seeking to a negative value makes the ExoPlayer to pause over HLS
Passing a negative value to the `player.seekTo()` function makes the ExoPlayer puts itself in a "pause" state and it never re-positions the `playhead` to 0 seconds. Still, we can see that the seek command is executed in the logcat.

To reproduce this we can use the Exo demo app. In the `PlayerActivity` set a delay of, could be 10 seconds, and then execute a `player.seetTo(-3000)`. Then the video will pause, the current position label will be at -0:00 but the Play/Pause button will stay as playing, and nothing else will happen. 
There is no error message returned/thrown by the ExoPlayer nor printed in the logcat. 

This is only with HLS videos. DASH and SmoothStreaming videos work fine.

Could you take a look, please?

dino-saurabh:
0. According to the discussion I saw on #8906, tried to resolve this Issue
1. Added Assertion for preventing negative seek position. closes #8906

### DIFF
--- a/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
+++ b/extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
@@ -431,6 +431,7 @@ public final class CastPlayer extends BasePlayer {
   @SuppressWarnings("deprecation")
   @Override
   public void seekTo(int windowIndex, long positionMs) {
+    Assertions.checkArgument(positionMs >= 0 || positionMs == C.TIME_UNSET);
     MediaStatus mediaStatus = getMediaStatus();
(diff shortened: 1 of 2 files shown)


nick-tkachov:
You can have negative seek position in Live playback if you have content buffered outside of the sliding window.

dino-saurabh:
> You can have negative seek position in Live playback if you have content buffered outside of the sliding window.

Oops Did not know about it, will read about it more .

ojw28:
Adding an assertion that'll crash applications if they attempt to pass negative seek positions seems pretty risky to me, and I'm not sure it's the correct solution. Let's use the issue (https://github.com/google/ExoPlayer/issues/8906) for any discussion about the correct solution.

Issue #9307, opened 2021-08-17
Broken recognize of the initial type of network


brozikcz:
It seems to be broken with https://github.com/google/ExoPlayer/commit/f8727b5e875e4e8e7abe3bfebd2c101dd41541aa commit.

The initial network type is broken in `DefaultBandwidthMeter` because the value from `getNetworkType` is `C.NETWORK_TYPE_UNKNOWN` after the first call.

```
  private DefaultBandwidthMeter(
      @Nullable Context context,
      Map<Integer, Long> initialBitrateEstimates,
      int maxWeight,
      Clock clock,
      boolean resetOnNetworkTypeChange) {
    this.initialBitrateEstimates = ImmutableMap.copyOf(initialBitrateEstimates);
    this.eventDispatcher = new EventDispatcher();
    this.slidingPercentile = new SlidingPercentile(maxWeight);
    this.clock = clock;
    this.resetOnNetworkTypeChange = resetOnNetworkTypeChange;
    if (context != null) {
      NetworkTypeObserver networkTypeObserver = NetworkTypeObserver.getInstance(context);
      networkType = networkTypeObserver.getNetworkType();
      bitrateEstimate = getInitialBitrateEstimateForNetworkType(networkType);
      networkTypeObserver.register(/* listener= */ this::onNetworkTypeChanged);
    } else {
      networkType = C.NETWORK_TYPE_UNKNOWN;
      bitrateEstimate = getInitialBitrateEstimateForNetworkType(C.NETWORK_TYPE_UNKNOWN);
    }
  }```

ojw28:
It's a little bit ambiguous whether this is broken or working as intended. In the new version the network type is initially `NETWORK_TYPE_UNKNOWN` because it's determined asynchronously. Once it's determined, `onNetworkTypeChanged` is called on any registered listeners. The reason for the change is to allow for disambiguation between 4G and 5G NSA, which I think either has to be asynchronous, or is more conveniently asynchronous, due to the underlying platform APIs used to disambiguate between these two cases.

That said, we did also see some problems internally as a result of the change, and we have [Internal ref: b/190021699] tracking whether we want to do something different.

brozikcz:
I see, I'll rewrite our bandwidth provider impl.

ojw28:
[...]

PR #9308, opened 2021-08-17
Fix broken recognize of the initial type of network

brozikcz:
#9307

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/util/NetworkTypeObserver.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/util/NetworkTypeObserver.java
@@ -109,7 +109,7 @@ public final class NetworkTypeObserver {
     mainHandler = new Handler(Looper.getMainLooper());
     listeners = new CopyOnWriteArrayList<>();
     networkTypeLock = new Object();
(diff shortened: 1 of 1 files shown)


brozikcz:
@googlebot I signed it!

library/core/src/main/java/com/google/android/exoplayer2/util/NetworkTypeObserver.java
```diff
@@ -109,7 +109,7 @@ private NetworkTypeObserver(Context context) {
     mainHandler = new Handler(Looper.getMainLooper());
     listeners = new CopyOnWriteArrayList<>();
     networkTypeLock = new Object();
-    networkType = C.NETWORK_TYPE_UNKNOWN;
+    networkType = getNetworkTypeFromConnectivityManager(context);
```
ojw28, line 112: This will cause the initial network type to be misreported as 4G for cases where a device is actually connected to 5G NSA. It might be worth making the change anyway, since it gets all other cases right and the network type will be corrected to 5G NSA quickly in the broken case. That said, it is technically less correct to how it is now. We'll sync about this internally.

PR #9318, opened 2021-08-19
Parse mid-stream SPS NAL units in MP4 streams and update the aspect ratio accordingly

MarcusWichelmann:
This PR resolves an issue with HLS/DASH fMP4 streams with changing aspect ratios.

These can occur e.g. in live TV (DVB) streams where the program, for example an old 4:3 movie, is interrupted by a modern 16:9 commercial with a different aspect ratio.  In this scenario, the H264/HEVC elementary stream that was packaged into the fMP4 segments contains SPS NAL units which inform the player about the aspect ratio change.

Currently ExoPlayer ignores them completely which causes the picture to be stretched after an aspect ratio change. 
Also, in case of HEVC, the initial SPS from the `hvcC` atom/box was not parsed for its aspect ratio which resulted in all content being played back with aspect ratio 1:1 ([see here](https://github.com/google/ExoPlayer/compare/dev-v2...MarcusWichelmann:mp4-parse-mid-stream-sps?expand=1#diff-c4f8a7953ee8a7eb6b5c11b5d978b31fe3ab2008951f581840da58f33aaeb33fR1143)).

This PR ...
- introduces a `parseH265SpsNalUnitPayload` method to parse the SPS NAL unit, just like it's done for H.264, too. The parsing code is derived from `H265Reader.parseMediaFormat`.
- extends the `HevcConfig` class to parse the initial SPS and sets the aspect ratio correctly when reading the `hvcC` atom.
- makes sure that the received mid-stream SPS NAL units are parsed instead of being ignored.
- refactors the `buildHevcCodecStringFromSps` method to re-use the parsed SPS data - just like it's done for H.264.

These changes should work just as well for regular MP4 files with changing aspect ratios.

As mentioned above, this PR duplicates the HEVC parsing code from `H265Reader.java`, just like `NalUnitUtil.java` already contained duplicated H.264 parsing code from `H264Reader.java`. This is not optimal and could probably be improved, but I thought that removing redundancies in existing code should not be in the scope of this PR. And also would be a bit more work refactoring work. Instead I tried to keep my new code as close and consistent as possible to how things are currently implemented for H.264.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
@@ -170,6 +170,7 @@ public class FragmentedMp4Extractor implements Extractor {
   private int sampleBytesWritten;
   private int sampleCurrentNalBytesRemaining;
   private boolean processSeiNalUnitPayload;
+  private boolean processSpsNalUnitPayload;
 
   // Outputs.
   private ExtractorOutput extractorOutput;
@@ -1407,23 +1408,45 @@ public class FragmentedMp4Extractor implements Extractor {
           processSeiNalUnitPayload =
               ceaTrackOutputs.length > 0
                   && NalUnitUtil.isNalUnitSei(track.format.sampleMimeType, nalPrefixData[4]);
+          processSpsNalUnitPayload =
+              NalUnitUtil.isNalUnitSps(track.format.sampleMimeType, nalPrefixData[4]);
           sampleBytesWritten += 5;
           sampleSize += nalUnitLengthFieldLengthDiff;
         } else {
           int writtenBytes;
-          if (processSeiNalUnitPayload) {
-            // Read and write the payload of the SEI NAL unit.
+          if (processSeiNalUnitPayload || processSpsNalUnitPayload) {
+            // We need to process the payload further, so read it into the temporary buffer
+            // before writing.
             nalBuffer.reset(sampleCurrentNalBytesRemaining);
             input.readFully(nalBuffer.getData(), 0, sampleCurrentNalBytesRemaining);
             output.sampleData(nalBuffer, sampleCurrentNalBytesRemaining);
             writtenBytes = sampleCurrentNalBytesRemaining;
-            // Unescape and process the SEI NAL unit.
-            int unescapedLength =
-                NalUnitUtil.unescapeStream(nalBuffer.getData(), nalBuffer.limit());
-            // If the format is H.265/HEVC the NAL unit header has two bytes so skip one more byte.
-            nalBuffer.setPosition(MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType) ? 1 : 0);
-            nalBuffer.setLimit(unescapedLength);
-            CeaUtil.consume(sampleTimeUs, nalBuffer, ceaTrackOutputs);
+            if (processSeiNalUnitPayload) {
+              // Unescape and process the SEI NAL unit.
+              int unescapedLength =
+                  NalUnitUtil.unescapeStream(nalBuffer.getData(), nalBuffer.limit());
+              // If the format is H.265/HEVC the NAL unit header has two bytes so skip one more byte.
+              nalBuffer.setPosition(
+                  MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType) ? 1 : 0);
+              nalBuffer.setLimit(unescapedLength);
+              CeaUtil.consume(sampleTimeUs, nalBuffer, ceaTrackOutputs);
+            } else if (processSpsNalUnitPayload) {
+              // Parse SPS and update the aspect ratio
+              float pixelWidthHeightRatio = track.format.pixelWidthHeightRatio;
+              if (MimeTypes.VIDEO_H264.equals(track.format.sampleMimeType)) {
+                NalUnitUtil.SpsData spsData = NalUnitUtil.parseSpsNalUnitPayload(
+                    nalBuffer.getData(), 0, nalBuffer.limit());
+                pixelWidthHeightRatio = spsData.pixelWidthHeightRatio;
+              } else if (MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType)) {
+                // TODO: Why is offset 1 necessary?
+                NalUnitUtil.H265SpsData spsData = NalUnitUtil.parseH265SpsNalUnitPayload(
+                    nalBuffer.getData(), 1, nalBuffer.limit());
+                pixelWidthHeightRatio = spsData.pixelWidthHeightRatio;
+              }
+              Format format =
+                track.format.buildUpon().setPixelWidthHeightRatio(pixelWidthHeightRatio).build();
+              output.format(format);
+            }
           } else {
             // Write the payload of the NAL unit.
             writtenBytes = output.sampleData(input, sampleCurrentNalBytesRemaining, false);
--- a/.jitpack.yml
+++ b/.jitpack.yml
@@ -1,0 +1,1 @@
+jdk: openjdk11
--- a/library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
@@ -158,6 +158,7 @@ public final class NalUnitUtil {
   private static final int H264_NAL_UNIT_TYPE_SEI = 6; // Supplemental enhancement information
   private static final int H264_NAL_UNIT_TYPE_SPS = 7; // Sequence parameter set
   private static final int H265_NAL_UNIT_TYPE_PREFIX_SEI = 39;
+  private static final int H265_NAL_UNIT_TYPE_SPS = 33;
 
   private static final Object scratchEscapePositionsLock = new Object();
 
@@ -265,6 +266,21 @@ public final class NalUnitUtil {
             && (nalUnitHeaderFirstByte & 0x1F) == H264_NAL_UNIT_TYPE_SEI)
         || (MimeTypes.VIDEO_H265.equals(mimeType)
             && ((nalUnitHeaderFirstByte & 0x7E) >> 1) == H265_NAL_UNIT_TYPE_PREFIX_SEI);
+  }
+
+  /**
+   * Returns whether the NAL unit with the specified header contains a sequence parameter set.
+   *
+   * @param mimeType The sample MIME type, or {@code null} if unknown.
+   * @param nalUnitHeaderFirstByte The first byte of nal_unit().
+   * @return Whether the NAL unit with the specified header is a SPS NAL unit. False is returned if
+   *     the {@code MimeType} is {@code null}.
+   */
+  public static boolean isNalUnitSps(@Nullable String mimeType, byte nalUnitHeaderFirstByte) {
+    return (MimeTypes.VIDEO_H264.equals(mimeType)
+            && (nalUnitHeaderFirstByte & 0x1F) == H264_NAL_UNIT_TYPE_SPS)
+        || (MimeTypes.VIDEO_H265.equals(mimeType)
+            && ((nalUnitHeaderFirstByte & 0x7E) >> 1) == H265_NAL_UNIT_TYPE_SPS);
   }
 
   /**


ojw28:
Could we split this pull request into two pieces:

1. Extend `HevcConfig` to include the pixel aspect ratio, and add the missing code in `AtomParsers` to propagate this into the `Format` in the absence of a `pasp` atom. The purpose of this pull request would be to fix:

> Also, in case of HEVC, the initial SPS from the hvcC atom/box was not parsed for its aspect ratio which resulted in all content being played back with aspect ratio 1:1

2. The remainder of the change that handles in-stream changes to the aspect ratio. This change feels a bit more controversial to me, because I think this kind of stream property change should trigger the start of a new `Period` in a DASH manifest, and the inclusion of an `EXT-X-DISCONTINUITY` with a fresh `EXT-X-MAP` in an HLS media playlist. I suspect (but did not confirm for HLS) that ExoPlayer handles the pixel aspect ratio change correctly if the manifest level signalling is present.

One reason to split the PR is that we can get the first (non-controversial) part merged, independently of any discussion that we need to have about the second part.

MarcusWichelmann:
Thanks for you answer.

I've noticed an error in the H.265 SPS parsing code that makes some streams fail to play and am currently analyzing it further. Will split up the PR when I've resolved it.

MarcusWichelmann:
Ooups, wrong commit email address. :/ Please just ignore the cla message above.

MarcusWichelmann:
@ojw28 As you've probably seen already, I've now split up the PR. This PR is now dependant on #9421.

> This change feels a bit more controversial to me, because I think this kind of stream property change should trigger the start of a new Period in a DASH manifest, and the inclusion of an EXT-X-DISCONTINUITY with a fresh EXT-X-MAP in an HLS media playlist. I suspect (but did not confirm for HLS) that ExoPlayer handles the pixel aspect ratio change correctly if the manifest level signalling is present.

I did some tests before working on this PR and `EXT-X-DISCONTINUITY` does currently not suffice to make ExoPlayer notice the aspect ratio change. For Apple AVKit and VLC, a discontinuity is sufficient, but required to make the aspect ratio change get noticed. And FFmpeg/FFplay notices the aspect ratio changes even without discontinuities.

RFC8216 [says about this](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.3):
>  The EXT-X-DISCONTINUITY tag SHOULD be present if there is a change in
   any of the following characteristics:
> - encoding parameters
> - encoding sequence

So at least to my understanding, a discontinuity is not strictly required in this case. And my opinion is that, if it doesn't cause any issues on the player side, why not make PAR changes just work out of the box, even without discontinuities?

This would at least make the stream creation easier. e.g. in my case, I had to write a patch (https://github.com/google/shaka-packager/pull/976) for shaka-packager to make it insert these discontinuities because Apples player requires them. But even then the switches are still not as smooth as with FFmpeg/FFplay (and thanks to this PR with ExoPlayer, too) because the segment cuts might not happen exactly where the aspect ratio change occurs. And fixing this will require even more effort and complexity on the stream-creation side.

So as you can see, I highly prefer how FFmpeg does it because it's way less of a burden to get it to work and I don't see a reason not to support this.

ojw28:
> I did some tests before working on this PR and EXT-X-DISCONTINUITY does currently not suffice to make ExoPlayer notice the aspect ratio change.

Ok. We should definitely make sure the aspect ratio change is handled in this case.

> This would at least make the stream creation easier. e.g. in my case, I had to write a patch (google/shaka-packager#976) for shaka-packager to make it insert these discontinuities because Apples player requires them. But even then the switches are still not as smooth as with FFmpeg/FFplay (and thanks to this PR with ExoPlayer, too) because the segment cuts might not happen exactly where the aspect ratio change occurs. And fixing this will require even more effort and complexity on the stream-creation side.

If Apple's implementation requires a discontinuity then I don't see a huge amount of benefit to having ExoPlayer not require them. In practice, HLS packagers need to produce streams that work well on Apple devices for them to be useful, which seems to imply inserting a discontinuity and cutting the segment at the point where the aspect ratio change occurs. One of the main goals of manifest level signalling is that it's less complicated and more efficient for a client to parse. So one reason for not supporting the case where discontinuities are not inserted is that it adds extra complexity into the client. I can take a more detailed look at this PR once we get the first one committed, to see exactly how much complexity is involved.

MarcusWichelmann:
I've rebased this PR onto latest `dev-v2` now.

Your fix of making `parseH265SpsNalUnitPayload` skip 1 byte less caused the changed `readSample` in this PR to break, so I had to increase the parsing offset to 1 to compensate that. See [line 1443](https://github.com/google/ExoPlayer/pull/9318/commits/68a15769d026eb1057e9ca5bf78871063cdd4dcd#diff-793f10aa618621cccb28e4b688e95913b4027a13a39616c136c20b4695b41f9aR1443).
I'm not yet sure why that's necessary, maybe you have an idea what that remaining byte is for so I can add a comment?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1407,23 +1408,45 @@ private boolean readSample(ExtractorInput input) throws IOException {
...
+              CeaUtil.consume(sampleTimeUs, nalBuffer, ceaTrackOutputs);
+            } else if (processSpsNalUnitPayload) {
+              // Parse SPS and update the aspect ratio
+              float pixelWidthHeightRatio = track.format.pixelWidthHeightRatio;
+              if (MimeTypes.VIDEO_H264.equals(track.format.sampleMimeType)) {
+                NalUnitUtil.SpsData spsData = NalUnitUtil.parseSpsNalUnitPayload(
+                    nalBuffer.getData(), 0, nalBuffer.limit());
+                pixelWidthHeightRatio = spsData.pixelWidthHeightRatio;
+              } else if (MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType)) {
+                // TODO: Why is offset 1 necessary?
```
ojw28, line 1441: This is necessary because the payload has been read assuming a 1 byte NAL unit header, but in H265 the header is actually 2 bytes, so the first byte in `nalBuffer` is actually the second byte of the header.

What you're doing here is identical to what's already happening on L1430 above for SEI NAL units, where `nalBuffer.setPosition` is passed `1` for HEVC and `0` otherwise. There is probably a clearer way that this could be written in code, but you can gain confidence from the pre-existing case that this is the right thing to do!

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -1407,23 +1408,45 @@ private boolean readSample(ExtractorInput input) throws IOException {
...
-            int unescapedLength =
-                NalUnitUtil.unescapeStream(nalBuffer.getData(), nalBuffer.limit());
-            // If the format is H.265/HEVC the NAL unit header has two bytes so skip one more byte.
-            nalBuffer.setPosition(MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType) ? 1 : 0);
-            nalBuffer.setLimit(unescapedLength);
-            CeaUtil.consume(sampleTimeUs, nalBuffer, ceaTrackOutputs);
+            if (processSeiNalUnitPayload) {
+              // Unescape and process the SEI NAL unit.
+              int unescapedLength =
+                  NalUnitUtil.unescapeStream(nalBuffer.getData(), nalBuffer.limit());
```
ojw28, line 1427: For SEI NAL unit handling, the data is unescaped before being processed. I'm not sure, but might this be necessary for the SPS case as well?

MarcusWichelmann: I just tested it and unescaping the SPS NALUs too does at least not break things. But I had no luck yet finding any documentation regarding what this unescaping is for and when it's needed. Could you point me to a term or chapter in the ISO standard(s)?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
MarcusWichelmann: Ah, thank you. I didn't notice that the code in line 1430 is already inside the if for `processSeiNalUnitPayload` and therefore doesn't get executed when processing the SPS. Now it totally makes sense that the offset is needed.

Theoretically, the whole
```java
// Unescape and process the SEI NAL unit.
int unescapedLength =
    NalUnitUtil.unescapeStream(nalBuffer.getData(), nalBuffer.limit());
// If the format is H.265/HEVC the NAL unit header has two bytes so skip one more byte.
nalBuffer.setPosition(
    MimeTypes.VIDEO_H265.equals(track.format.sampleMimeType) ? 1 : 0);
nalBuffer.setLimit(unescapedLength);
```
could be moved before line 1424 to deduplicate things. Assuming that the unescaping is needed/acceptable for SPS NAL units.

ojw28:
> I did some tests before working on this PR and EXT-X-DISCONTINUITY does currently not suffice to make ExoPlayer notice the aspect ratio change. For Apple AVKit and VLC, a discontinuity is sufficient, but required to make the aspect ratio change get noticed. And FFmpeg/FFplay notices the aspect ratio changes even without discontinuities.

Returning to this, and more to the fundamental question about whether we want to be parsing SPS/PPS directly from the bitstream ... Could you provide a test stream that has an `EXT-X-DISCONTINUITY` tag in it at the point of pixel aspect ratio change? I'd like to understand why it's not sufficient, now that the change to fix the extractor has been submitted.

MarcusWichelmann:
@ojw28 Of course. Here is a test stream that's packaged using my [fork](https://github.com/MarcusWichelmann/shaka-packager) of shaka-packager: http://185.223.147.231/master.m3u8

This stream is an infinite loop of a short clip that contains two discontinuities. My modified chunking handler will split the segments and add an `EXT-X-DISCONTINUITY` tag at each detected discontinuity. The master playlist contains two variant streams, H.264 and H.265.

I verified that the stream plays correctly using Apple AVKit and the patched ExoPlayer from this PR. When playing without this patch, the commercial is stitched.

I hope you can work with that.

In case the above stream crashes, I also tried to create a VOD stream of the same content ([Download](https://drive.wichelmann.cloud/s/R4enAeeyyEDkYqy)). But somehow this doesn't play correctly in VLC, not sure what's wrong there.

ojw28:
I'm not convinced that the playlist is valid. It contains a single `#EXT-X-MAP` tag, which points to an `init.mp4` that includes SPS/PPS. If the SPS/PPS are changing at the point where `#EXT-X-DISCONTINUITY` tags are being inserted, then there should be new `#EXT-X-MAP` tags at these points, pointing to init segments containing the correct SPS/PPS data. If these were in place then I think playback would work correctly.

MarcusWichelmann:
Ah, I get what you mean now, that makes sense. Although this will get a bit tricky to implement into the packager. 😄 

What I don't understand yet is why AVKit doesn't require it. They seem to parse the parameter set from the following segment after the discontinuity and use the SPS from that instead. At least when no new `EXT-X-MAP` is present. But you don't want this behaviour, right?

I think this PR can be closed then and I'll play around with generating new media initialization sections and see if that works as expected.

Thanks for your clarifications! 👍🏼

ojw28:
> What I don't understand yet is why AVKit doesn't require it. They seem to parse the parameter set from the following segment after the discontinuity and use the SPS from that instead. At least when no new EXT-X-MAP is present. But you don't want this behaviour, right?
>
> I think this PR can be closed then and I'll play around with generating new media initialization sections and see if that works as expected.

The spec mandating a new `#EXT-X-MAP` doesn't preclude the player from also reading the parameter set from the following segment. It does mean that it would be valid for Apple to update their implementation to _not_ do this in the future, so in general making sure that your media is compliant to the spec (rather than the particular player implementation) is a good idea.

The main concern with this PR is whether the additional cost/complexity is worth it, given it shouldn't be required if the media conforms to the spec. It's worth leaving the PR open; we might take a look at it sometime and figure out whether it's cheap enough to be worth taking anyway. Additional providers sharing streams with us that have the same setup might also provide more weight to taking it. For now, we consider it non-urgent because it's not a bug as such, so looking at it more carefully is pushed down the stack of things that we need to do a little bit :).

icbaker:

Issue #9264, opened 2021-08-06
RTSP: stream does not play (blackscreen)


holoc285:
implementation 'com.google.android.exoplayer:exoplayer:2.14.2'
implementation 'com.google.android.exoplayer:exoplayer-rtsp:2.14.2'

play: rtsp://wowzaec2demo.streamlock.net/vod/mp4:BigBuckBunny_115k.mov
cannot play:  rtsp://113.163.157.230:9717/rtsph2641080p

Debug Log
2021-08-06 10:13:32.544 8410-8410/com.example.demo D/EventLogger: surfaceSize [eventTime=0.09, mediaPos=0.00, window=0, period=0, 900, 1440]
2021-08-06 10:13:32.547 8410-8410/com.example.demo D/EventLogger: timeline [eventTime=0.09, mediaPos=0.00, window=0, period=0, periodCount=1, windowCount=1, reason=SOURCE_UPDATE
2021-08-06 10:13:32.547 8410-8410/com.example.demo D/EventLogger:   period [?]
2021-08-06 10:13:32.547 8410-8410/com.example.demo D/EventLogger:   window [?, seekable=false, dynamic=false]
2021-08-06 10:13:32.547 8410-8410/com.example.demo D/EventLogger: ]
2021-08-06 10:13:32.548 8410-8410/com.example.demo D/EventLogger: loading [eventTime=0.09, mediaPos=0.00, window=0, period=0, true]

Thanks

claincly:
FYI `rtsp://wowzaec2demo.streamlock.net/vod/mp4:BigBuckBunny_115k.mov` plays on my end.

Please provide a log of RTSP messages.

Add logs at

https://github.com/google/ExoPlayer/blob/b2333c86c1eac9a1f95992960a8495f1e5b79200/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspMessageChannel.java#L182

and 

https://github.com/google/ExoPlayer/blob/b2333c86c1eac9a1f95992960a8495f1e5b79200/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspClient.java#L423

The easiest way is to use

`Log.d("RTSP", message)`

holoc285:
How can I override this methods to add logs. Thanks

claincly:
You don't need to override it, adding one line of code `Log.d("RTSP", message)` to the pointed locations will be sufficient.

holoc285:
[...]

PR #9353, opened 2021-08-29
[BUGFIX] RTSP: RTP-Info response validation

mmalisz:
This commit adds a regex test to validate quirky camera server
reponses, that do not return the full URL, but rather a single
parameter, i.e. 'track1' is returned instead of
'rtsp://113.163.157.230:9717/rtsph2641080p/track1'.

The regex test is currently valid for responses like
track1, track222, trackID=1, trackid=222, etc.

The test itself is held as a static constant placed in RtspResponse
class.

Related: #9264

### DIFF
--- a/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspMediaPeriod.java
+++ b/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspMediaPeriod.java
@@ -512,13 +512,15 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
           trackUrisWithTiming.add(checkNotNull(uriPath));
         }
       }
-      for (int i = 0; i < selectedLoadInfos.size(); i++) {
-        RtpLoadInfo loadInfo = selectedLoadInfos.get(i);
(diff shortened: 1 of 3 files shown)


mmalisz:
@googlebot I signed it!

ojw28:
@claincly - PTAL. Thanks!

claincly:
Thanks for sending us the PR!

Since it's the first case we see this quirky behaviour, and the code in this PR fixes your use case. I will take a notice in case the same issue comes up in the future; but until then, I'd be hesitant merging this.

mmalisz:
> Thanks for sending us the PR!
> 
> Since it's the first case we see this quirky behaviour, and the code in this PR fixes your use case. I will take a notice in case the same issue comes up in the future; but until then, I'd be hesitant merging this.

No problem. I've tried to take a generic approach, so it can be easily extended (hence the regex).

ojw28:
What model of camera is sending these quirky responses?

mmalisz:
@ojw28,\
DAHUA IPC-HDBW1320E-W-0280B.

However, I could see the same behaviour reported in #9264, though the faulty response was different (included in the regex test).

claincly:
We'll have a generic patch for this soon, so I'm closing it for now.

mmalisz:
Great! Thanks for the info.

PR #9380, opened 2021-09-02
Handle trailing space on RTSP RtpMap attribute

### LINKED ISSUE #9379: IllegalArgumentException on RSTP stream from ip-cam because of trailing space in RtpMap
Exoplayer can't show stream from ip camera because of RtpMap header with trailing space. I don't know if it is a camera's firmware bug or it was made intentionally, but no change in camera settings removes that trailing space. Some players (e.g VLC) plays this steam correctly. I'm still using an ExoPlayer fork with RTSP support (before is was added to official repo), and nothing of them has such blocker.

As described in #9014 I have RTSP response:

```
RTSP/1.0 200 OK
Content-Type: application/sdp
Server: H264DVR 1.0
Cseq: 1
Content-Base: rtsp://192.168.1.11:554/user=admin&password=&channel=1&stream=0.sdp/
Cache-Control: private
x-Accept-Retransmit: our-retransmit
x-Accept-Dynamic-Rate: 1
Content-Length: 412

v=0
o=- 38990265062388 38990265062388 IN IP4 192.168.1.11
s=RTSP Session
c=IN IP4 192.168.1.11
t=0 0
a=control:*
a=range:npt=0-
m=video 0 RTP/AVP 96
a=rtpmap:96 H264/90000 
a=range:npt=0-
a=framerate:0S
a=fmtp:96 profile-level-id=4d0029; packetization-mode=1; sprop-parameter-sets=J00AKY1uBQBboQAAAwABAAADADKE,KO4Fcg==
a=framerate:25
a=control:trackID=3
m=audio 0 RTP/AVP 8
a=control:trackID=4
```

Here is additional trailing space that causes error:
`a=rtpmap:96 H264/90000␣`

### Proposed solution
As I see, header looks normal and can be used as normal with skipping trailing spacer after splitting parameter. Moreover, the library uses only first two values, but checks that split result counts to exactly two in [MediaDescriptor.java:47](https://github.com/google/ExoPlayer/blob/dev-v2/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/MediaDescription.java#L47):
```
      String[] rtpmapInfo = Util.split(rtpmapString, " ");
      checkArgument(rtpmapInfo.length == 2);
```

Also, [Util.split()](https://github.com/google/ExoPlayer/blob/373db56a52e9943df79ff3540da061264d6dc4b9/library/common/src/main/java/com/google/android/exoplayer2/util/Util.java#L731) just calls  `value.split(regex, -1);`. Calling `rtpmapString.split(" ")` or more specific call `rtpmapString.split(" ", 0)` will be enough to enhance the situation.

Similar headers processing issues: #9247, #9114, #9182

artroman:
Fixes #9379

### DIFF
--- a/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/MediaDescription.java
+++ b/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/MediaDescription.java
@@ -43,7 +43,7 @@ import java.util.HashMap;
 
     /** Parses the RTPMAP attribute value (with the part "a=rtpmap:" removed). */
     public static RtpMapAttribute parse(String rtpmapString) throws ParserException {
-      String[] rtpmapInfo = Util.split(rtpmapString, " ");
+      String[] rtpmapInfo = rtpmapString.split(" ");
       checkArgument(rtpmapInfo.length == 2);
       int payloadType = parseInt(rtpmapInfo[0]);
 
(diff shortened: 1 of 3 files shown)


artroman:
@googlebot I signed it!

claincly:
Thanks for sending this in. We already have a fix internally and will be pushed soon.

Issue #9347, opened 2021-08-27
seekTo(0) followed by a TrackSelection jumps back to live point.


stevemayhew:
## Steps To Reproduce

Add this code to the demo `PlayerActivity`:

````java

  private static boolean bug_triggered = false;

  private class PlayerEventListener implements Player.Listener {

    @Override
    public void onPlaybackStateChanged(@Player.State int playbackState) {
      Log.d(TAG, "onPlaybackStateChanged() - " + playbackState);

      if (getIntent().getBooleanExtra("trigger_bug", false) && ! bug_triggered) {
        playerView.setControllerAutoShow(false);
        playerView.setControllerShowTimeoutMs(-1);
        playerView.showController();
        if (playbackState == Player.STATE_READY) {
          bug_triggered = true;
          DefaultTrackSelector.Parameters savedParameters = trackSelector.getParameters();
          DefaultTrackSelector.Parameters disableTracksParameters =
              savedParameters
              .buildUpon()
              .setMaxVideoBitrate(3000000 - 1)
              .build();
          trackSelector.setParameters(disableTracksParameters);
          long pos = player.getCurrentPosition();
          while (pos > 0) {
            Log.d(TAG, "seekTo - " + pos);
            player.seekTo(pos);
            pos -= 600_000L;
            android.os.SystemClock.sleep(500);
          }
          Log.d(TAG, "seekTo final to 0");
          player.seekTo(0);
          Log.d(TAG, "run track selection");
          trackSelector.setParameters(savedParameters);
        }
      }
````

Run it sending the intent to the player:

````shell
adb shell am start -n com.google.android.exoplayer2.demo/.PlayerActivity -a com.google.android.exoplayer.demo.action.VIEW --ez trigger_bug true -d <live-url>
````

## Issue Analysis
The `MaskingMediaPeriod` forces position back to live start when TrackSelection is triggered following a seek to 0.
I believe this issue is related to issue #7975, specifically the code that handles changing the position, source [here](https://github.com/google/ExoPlayer/blob/1d36083e2543dd4be9e9e38a8f08bee2a3a3d3da/library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaPeriod.java#L181).

[...]

PR #9386, opened 2021-09-03
TrackSelection does not jump to live with HlsMediaPeriod

stevemayhew:
This is a fix for issue #9347

Once the MaskingMediaPeriod completes `prepare()` (when the masked HlsMediaPeriod reports the
first real Timeline with a duration and window start position) and the `preparePositionUs` is used
by the player in the `onPrepared()` callback to set the render position it should not use it again.

The bug is a track selection causes a jump to live if you are at position 0
in the timeline, even after playback was started from the live point (the prepare position override).

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaPeriod.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaPeriod.java
@@ -178,10 +178,6 @@ public final class MaskingMediaPeriod implements MediaPeriod, MediaPeriod.Callba
       @NullableType SampleStream[] streams,
       boolean[] streamResetFlags,
       long positionUs) {
-    if (preparePositionOverrideUs != C.TIME_UNSET && positionUs == preparePositionUs) {
-      positionUs = preparePositionOverrideUs;
-      preparePositionOverrideUs = C.TIME_UNSET;
-    }
     return castNonNull(mediaPeriod)
         .selectTracks(selections, mayRetainStreamFlags, streams, streamResetFlags, positionUs);
   }
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaSource.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaSource.java
@@ -140,11 +140,6 @@ public final class MaskingMediaSource extends CompositeMediaSource<Void> {
     @Nullable MediaPeriodId idForMaskingPeriodPreparation = null;
     if (isPrepared) {
       timeline = timeline.cloneWithUpdatedTimeline(newTimeline);
-      if (unpreparedMaskingMediaPeriod != null) {
-        // Reset override in case the duration changed and we need to update our override.
-        setPreparePositionOverrideToUnpreparedMaskingPeriod(
-            unpreparedMaskingMediaPeriod.getPreparePositionOverrideUs());
-      }
     } else if (newTimeline.isEmpty()) {
       timeline =
           hasRealTimeline
--- a/library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
+++ b/library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
@@ -3250,6 +3250,77 @@ public final class ExoPlayerTest {
   }
 
   @Test
+  public void trackSelectionDoesNotResetToDefaultPeriodPosition()
+      throws Exception {
+    FakeMediaSource mediaSource =
+        new FakeMediaSource(new FakeTimeline(), ExoPlayerTestRunner.VIDEO_FORMAT, ExoPlayerTestRunner.AUDIO_FORMAT);
+    FakeTrackSelector trackSelector = new FakeTrackSelector();
+    FakeRenderer videoRenderer = new FakeRenderer(C.TRACK_TYPE_VIDEO);
+    FakeRenderer audioRenderer = new FakeRenderer(C.TRACK_TYPE_AUDIO);
+    MediaItem mediaItem =
+        new MediaItem.Builder().setUri(Uri.EMPTY).build();
+    final long windowDefaultStartPositionUs = 5 * C.MICROS_PER_SECOND;
+    final AtomicLong expectedPosition = new AtomicLong(C.TIME_UNSET);
+    final int lastSeekPosition = 0;
+
+    Timeline liveTimeline =
+        new SinglePeriodTimeline(
+            /* presentationStartTimeMs= */ C.TIME_UNSET,
+            /* windowStartTimeMs= */ C.TIME_UNSET,
+            /* elapsedRealtimeEpochOffsetMs= */ C.TIME_UNSET,
+            /* periodDurationUs= */ 10 * C.MICROS_PER_SECOND,
+            /* windowDurationUs= */ 10 * C.MICROS_PER_SECOND,
+            /* windowPositionInPeriodUs= */ 0,
+            /* windowDefaultStartPositionUs= */ windowDefaultStartPositionUs,
+            /* isSeekable= */ true,
+            /* isDynamic= */ true,
+            /* manifest= */ null,
+            mediaItem,
+            mediaItem.liveConfiguration);
+
+    ActionSchedule actionSchedule =
+        new ActionSchedule.Builder(TAG)
+            .pause()
+            .waitForPlaybackState(Player.STATE_BUFFERING)
+            // Finish preparation.
+            .executeRunnable(() -> mediaSource.setNewSourceInfo(liveTimeline))
+            .waitForTimelineChanged()
+            .waitForPlaybackState(Player.STATE_READY)
+            .play()
+            .playUntilPosition(0, windowDefaultStartPositionUs)
+            .pause()
+            .seekAndWait(lastSeekPosition) // paused at this position, should not change from here
+            .executeRunnable(() -> {
+              DefaultTrackSelector.Parameters parameters = trackSelector.getParameters();
+              trackSelector.setParameters(parameters.buildUpon()
+                  .setMaxAudioBitrate(0)    // Cause track selection change
+                  .build()
+              );
+            })
+            .waitForPendingPlayerCommands()
+            .executeRunnable(new PlayerRunnable() {
+              @Override
+              public void run(SimpleExoPlayer player) {
+                expectedPosition.set(player.getCurrentPosition());
+              }
+            })
+            .stop()   // End live playback
+            .build();
+
+    new ExoPlayerTestRunner.Builder(context)
+        .setMediaSources(mediaSource)
+        .setTrackSelector(trackSelector)
+        .setRenderers(audioRenderer, videoRenderer)
+        .setActionSchedule(actionSchedule)
+        .setUseLazyPreparation(true)    // defer prepare until first source update (this is default)
+        .build()
+        .start()
+        .blockUntilEnded(TIMEOUT_MS);
+
+    assertThat(expectedPosition.get()).isEqualTo(lastSeekPosition);
+  }
+
+  @Test
   public void seekToUnpreparedWindowWithMultiplePeriodsInConcatenationStartsAtCorrectPeriod()
       throws Exception {
     long periodDurationMs = 5000;


stevemayhew:
## Use Cases

Most origin vendors support live playback with a rolling-buffer in support of Cloud DVR.  In these cases the live window represents only the part of what is available for playback with a simple `seekTo().`  

For example, a SOCU (Start Over / Catch UP) operation converts the sliding live playlist into a live [EVENT](https://developer.apple.com/documentation/http_live_streaming/example_playlists_for_http_live_streaming/event_playlist_construction) playlist, assuming the program is in progress.

The operations and how they are implemented are:

#### Initiate SOCU for in-progress Live
Here live playback is restarted playback (prepare) with URL indicating start-time / end-time (end time in this case is in the future).  The steps in the player are simple:

````java
	ExoPlayer player = ...
	liveEventSource = // code converts current live URL to SOCU
	player.setMediaSource(liveEventSource, 0);
	player.prepare();
````

#### Restart on BLWE
This (obviously) can happen prior to the user switching to SOCU they scrub or trick-play to the start of the live window, position is very possibly negative at the time of the error report. 

It is disruptive in this case to jump to live edge.  Instead, in the error handler, logically like to do:

````java
	player.seekTo(0);
	player.prepare();
````

During SOCU this is obviously never possible, even if the `Timeline` is still dynamic.

## Current Flow

The [SOCU Prepare Flow Sequence Diagram](https://sequencediagram.org/index.html?presentationMode=readOnly#initialData=C4S2BsFMAIGUHkDCBVaAFATpADgQy9AGLgD2A7gFB4agDGIeAdsNAKJoCSHFAsrgM4BrEIwDmPSABMQuWCQCuGWpAC0APgBUfISNEAVEAFtI4EZABcAHhFg1FALz3oiLLmAwAMiVzSx6SBggJJLQjhTsXOoRHOby2JJukGgBQZL8ABQAlBS4tKAAbolsnNzRUSXmhrgAngBGkMjxiV4+usmBwVk5eSCF7sVcFOEl6trCYhLSsgpKFrSu7u2pXbkFRWO6kzJyisq8AuPiUtszyupaB5vHuEvBVjbAaukiAGYkAHT8wPjAaCT8YCCjGyJGwLBA-EwOHwUkcL1w4H4MAAZMjoAALASwb40KRQ7BwhFI-Y6CbXHazdQACURW2muws2Cw1EgFOUXRp-DpbNUak53NOjOZMJ5HGYAUYCJWPT6MH55MFFEYJH6JHyAWg8qmPPMXx+-Gg2HANVMX2gpB8FC1JwZKnUag2ZO1gvMWGAikYFEkkFWvXWlydNtmXp9MqK1vpweVqvVGGgjqOzoZ5mw-xYtBIjGkoEz-HMAB1GNAIfiYZJCYjIIXoPJGEzoVhJAm6bcQhDoEjgBRIFnoMNIg6A4mZK2U8KsM9GG9Pjjfv9AZnkPxskqVTA1Rrm9dR6mzRmswvGHnq-WWX8ATnGEvQk4AAwn8dJeeX+CxwLe6+OZzvPQcHisAB9ZAADlYFYPQQ19WUBm4b0oKKaIHCcQgQAwM0eWgMU3mgAAlSAXiwfh0RvK1aQVW0+TIpNZnMTNMCMfBqjQY1qlNYB3g46U1n6CMeVIrlyMpQdSWHSNlFoxhEHREBwEkUUpw+Djum4mAt2ovZQXBSFH0kAAaGs6x0tSRxSYJoHUaAqhEklDgFCjjJuUzJHMeYfUWJyuL9foHNbGyrimVtzgjUdrEYWw-MDRyOkkakqJM6KxwbSB0mAaT+H0jcMHfJ9l2Ury5TiqLUgoUxGB9OMItE1tLEsO1gqc11IHdDBPQcnkartHyGrdD0SrMfBoBIF4XkgsMeMK3y4LGgqBPUyBRpU+MhxbJyFvypaRLs4NSvKwbPWjddYw22ztwa3d00zbMgWPItTxhc9D1fAJsqXRwyBESRyA7Wdq1rO7Gy66LiwNTtVxjTdlsE8TzugfcrtzAsixLHTHGADB5CrItMX4PCEQMYwdtR9HMb6sqBqGkaexCIYv1LAhEBIQwjSayADTCHbydayGAqc2r6oSuidK6aJasB1IJLpqQug5uMKf7Dhyi4cxMSzKBW0lyRPOgxDNPNbxfFEG9DRY3R5cVmJCKavCswCAIHsvLoqZppxYHkWokQARwx5g4EFTCFNw-DCOIsJeMFWLZqDcTBYYjAmJYtiOPeLXw0Kviw-s7mo4sTMpJkuTBSwxT3jy6C2sVXXkaSvSDP+qQxbMiyrMOSqtrOYSTrm3Umsl+2gSerKQG9JPS-9Taofm8vM-HnmBbfIekkfPvFwybIpsWjPg3X9ap+DKnoCAA) shows how the player flow works for the SOCU use case above.

The "Create Loading Period" works perfectly, picking up the requested prepare position into the `MaskingMediaPeriod`

Looking at the "First Source Info Refresh" issue #7975  comes into play, the masked media period (`HlsMediaPeriod`) in-correctly picks and overrides the requested prepare position.   For other use cases the position is correctly set to start at the live edge requested by the playlist.

Next, HLS library completes the "prepare" operation and reports `onPrepared()` event.  One option is to reset the `preparePositionOverrideUs` to `C.TIME_UNSET` at this point, as we have already "used" the prepare position.

The "Subsequent Source Info Refresh" flow may occur before or after the "Prepare Completes" flow (definitely possible for non-chunkless prepare).  Updating the 'prepare' position might be warranted in these cases.  But, of `onPrepared()` has been signaled there seems no point.

The block of code in `MaskingMediaPeriod` is what most certainly causes this issue #9347 

````java
  @Override
  public long selectTracks(
      ...
    if (preparePositionOverrideUs != C.TIME_UNSET && positionUs == preparePositionUs) {
      positionUs = preparePositionOverrideUs;
      preparePositionOverrideUs = C.TIME_UNSET;
    }
````

Unless there is a real use case for this (no test case fail with the code removed) then removing it is the most strait forward solution.  Otherwise resetting `preparePositionOverrideUs` after `onPrepared()` is signaled may also cover it.

I owe unit test cases for this request once we decide on the correct path forward @tonihei and @ojw28  what do you think.

**Comment by stevemayhew** [comment]
Test cases added.  See inline comments

File: library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaPeriod.java [hunk 1]
```diff
@@ -178,10 +178,6 @@ public long selectTracks(
       @NullableType SampleStream[] streams,
       boolean[] streamResetFlags,
       long positionUs) {
-    if (preparePositionOverrideUs != C.TIME_UNSET && positionUs == preparePositionUs) {
-      positionUs = preparePositionOverrideUs;
-      preparePositionOverrideUs = C.TIME_UNSET;
-    }
     return castNonNull(mediaPeriod)
```
stevemayhew, line 181: The position change to start at the default position has already happened when the `onPrepare()` callback is processed by `ExoPlayerImplInternal`.   Removing this causes no existing test cases to fail.

library/core/src/main/java/com/google/android/exoplayer2/source/MaskingMediaSource.java
```diff
@@ -140,11 +140,6 @@ protected void onChildSourceInfoRefreshed(
     @Nullable MediaPeriodId idForMaskingPeriodPreparation = null;
     if (isPrepared) {
       timeline = timeline.cloneWithUpdatedTimeline(newTimeline);
-      if (unpreparedMaskingMediaPeriod != null) {
-        // Reset override in case the duration changed and we need to update our override.
-        setPreparePositionOverrideToUnpreparedMaskingPeriod(
-            unpreparedMaskingMediaPeriod.getPreparePositionOverrideUs());
-      }
```
stevemayhew, line 147: This seems remotely possible, perhaps if the timeline is growing when it is a newly started live playback where it has not reached its target window size.  @marcbaechinger what do you think?  My new test case still passes if we remove this change, however I'd suggest if we leave it that it is covered by a unit test case (right now it is not).

library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
```diff
@@ -3249,6 +3249,77 @@ public void run(SimpleExoPlayer player) {
...
   }
 
+  @Test
+  public void trackSelectionDoesNotResetToDefaultPeriodPosition()
+      throws Exception {
+    FakeMediaSource mediaSource =
+        new FakeMediaSource(new FakeTimeline(), ExoPlayerTestRunner.VIDEO_FORMAT, ExoPlayerTestRunner.AUDIO_FORMAT);
+    FakeTrackSelector trackSelector = new FakeTrackSelector();
+    FakeRenderer videoRenderer = new FakeRenderer(C.TRACK_TYPE_VIDEO);
+    FakeRenderer audioRenderer = new FakeRenderer(C.TRACK_TYPE_AUDIO);
```
stevemayhew, line 3259: Start with a MediaSource with both audio and video

library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
```diff
@@ -3249,6 +3249,77 @@ public void run(SimpleExoPlayer player) {
...
+    MediaItem mediaItem =
+        new MediaItem.Builder().setUri(Uri.EMPTY).build();
+    final long windowDefaultStartPositionUs = 5 * C.MICROS_PER_SECOND;
+    final AtomicLong expectedPosition = new AtomicLong(C.TIME_UNSET);
+    final int lastSeekPosition = 0;
```
stevemayhew, line 3264: Window start override is 5 seconds, the seekTo 0 is what does not get honored.

library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
```diff
@@ -3249,6 +3249,77 @@ public void run(SimpleExoPlayer player) {
...
+    ActionSchedule actionSchedule =
+        new ActionSchedule.Builder(TAG)
+            .pause()
+            .waitForPlaybackState(Player.STATE_BUFFERING)
+            // Finish preparation.
+            .executeRunnable(() -> mediaSource.setNewSourceInfo(liveTimeline))
+            .waitForTimelineChanged()
+            .waitForPlaybackState(Player.STATE_READY)
+            .play()
+            .playUntilPosition(0, windowDefaultStartPositionUs)
```
stevemayhew, line 3290: Timeline update is completed and playback reaches the default window start.

library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
```diff
@@ -3249,6 +3249,77 @@ public void run(SimpleExoPlayer player) {
...
+            .pause()
+            .seekAndWait(lastSeekPosition) // paused at this position, should not change from here
+            .executeRunnable(() -> {
+              DefaultTrackSelector.Parameters parameters = trackSelector.getParameters();
+              trackSelector.setParameters(parameters.buildUpon()
+                  .setMaxAudioBitrate(0)    // Cause track selection change
+                  .build()
+              );
+            })
```
stevemayhew, line 3299: Playback starts, reaches the default window start and is paused there.  Issue a track selection that changes parameters so as to invalidate the current selection.

library/core/src/test/java/com/google/android/exoplayer2/ExoPlayerTest.java
```diff
@@ -3249,6 +3249,77 @@ public void run(SimpleExoPlayer player) {
...
+            .waitForPendingPlayerCommands()
+            .executeRunnable(new PlayerRunnable() {
+              @Override
+              public void run(SimpleExoPlayer player) {
+                expectedPosition.set(player.getCurrentPosition());
+              }
+            })
```
stevemayhew, line 3306: At this point, after track selection position should still be 0, but the removed code would have reset it to the default position (causing this test to fail)

stevemayhew:
@ojw28  this code is in our source tree now, please have someone take a look at the change.  It fixes the bug and includes a test case that demonstrates the bug.

stevemayhew:
GitHub is not allowing a convert to Draft for this pull request, but it has an issue that an error recovery on playback that attempts to reset to the default playback position for live will not work (as it looks like only the `MediaPeriod.selectTracks()` commits the pending seek position).

stevemayhew:
The test cases are good, but the fix is not.  It breaks other things.  Note this bug is related to #7975 (part of the same sequence of logic) but not the same.   See pull request #11285 for a much simpler change that does not break existing use cases.

PR #9413, opened 2021-09-09
Added resolution options to AdaptiveTrackSelection to retain low quality media on discard

jaeho-lee104:
Hello 🖐🏻
[The current implementation](https://github.com/google/ExoPlayer/blob/02f7aafe67b4893134a31ea55a3ba0f0535df145/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java#L425)  treats as low quality media when the resolution is less than height=720 and width=1280.
Users may want to discard 720P quality, or they may want to keep 480P quality.
Added resolution options to AdaptiveTrackSelection to retain specific low quality media on discard.
Thanks.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
@@ -53,6 +53,8 @@ public class AdaptiveTrackSelection extends BaseTrackSelection {
     private final int minDurationForQualityIncreaseMs;
     private final int maxDurationForQualityDecreaseMs;
     private final int minDurationToRetainAfterDiscardMs;
+    private final int maxWidthToDiscard;
+    private final int maxHeightToDiscard;
     private final float bandwidthFraction;
     private final float bufferedFractionToLiveEdgeForQualityIncrease;
     private final Clock clock;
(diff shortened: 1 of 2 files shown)


library/core/src/main/java/com/google/android/exoplayer2/trackselection/AdaptiveTrackSelection.java
```diff
@@ -257,6 +285,14 @@ public AdaptiveTrackSelection(TrackGroup group, int[] tracks, BandwidthMeter ban
    *     quality, the selection may indicate that media already buffered at the lower quality can be
    *     discarded to speed up the switch. This is the minimum duration of media that must be
    *     retained at the lower quality. It must be at least {@code minDurationForQualityIncreaseMs}.
+   * @param minHeightOfResolutionToRetainOnDiscard When switching to a track of significantly higher
```
ojw28: The change itself looks fine to me, but the variable naming and documentation are pretty confusing. The variable name, in particular, suggests that anything below this minimum will always be discarded when switching up, which is not the case.

Would it be clearer as `maxHeightToDiscard` and `maxWidthToDiscard`, documenting that these are the maximum values for which the selector may choose to discard when switching up to a higher quality (aside: I don't think you need `OfResolution`; it can't really be of anything else)? You'll need to use `<=` and reduce the default values by `1` to keep the behavior unchanged.

jaeho-lee104: @ojw28 
Thanks for pointing out what could be confusing.
I agree with your opinion.
I applied it.

PR #9415, opened 2021-09-09
Dev v2 truehd

g-lass:
Hi 

I would like to enable the extraction of Dolby TrueHD streams from ISOBMF files.
For this extraction, I have reused the bundling of 16 access units (referred as samples in MP4 files) introduced with the extraction of Dolby TrueHD from MKV files.

The second commit provides a corresponding MP4 test vector and extraction dumps.

Fragmented MP4 has not been addressed in these commits.

Regards
gael


library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.android.exoplayer2.util.MimeTypes;
+import com.google.android.exoplayer2.util.ParsableBitArray;
+import com.google.android.exoplayer2.util.ParsableByteArray;
+import java.nio.ByteBuffer;
+
+/** Utility methods for parsing MLP frames, which are access units in MLP bitstreams. */
+public final class MlpUtil {
+
+  /** The channel count of MLP stream. */
+  // TODO: Parse MLP stream channel count.
```
kim-vde, line 31: Is this intended that this TODO has not been implemented?

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
```diff
@@ -101,5 +101,10 @@ public void mp4SampleWithMhm1Track() throws Exception {
   public void mp4SampleWithColorInfo() throws Exception {
     ExtractorAsserts.assertBehavior(
         Mp4Extractor::new, "media/mp4/sample_with_color_info.mp4", simulationConfig);
+
+  @Test
+  public void mp4SampleWithDolbyTrueHDTrack() throws Exception {
```
kim-vde, line 107: This test doesn't pass. The changes also seem to have broken test mp4SampleWithAc4Track.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/Mp4Extractor.java
```diff
@@ -646,6 +649,64 @@ private int readSample(ExtractorInput input, PositionHolder positionHolder) thro
...
+   * <p>Returns {@link #RESULT_END_OF_INPUT} if no samples are left. Otherwise, returns {@link
+   * #RESULT_CONTINUE}.
+   *
+   * @param input The {@link ExtractorInput} from which to read data.
+   * @param positionHolder If {@link #RESULT_SEEK} is returned, this holder is updated to hold the
+   *     position of the required data.
+   * @return One of the {@code RESULT_*} flags in {@link Extractor}.
+   * @throws IOException If an error occurs reading from the input.
+   */
+  private int readSample(ExtractorInput input, PositionHolder positionHolder) throws IOException {
```
kim-vde, line 670: The method readSample should return as often as possible so it should ideally read only one sample before returning. To avoid that, you can use a rechunker as in the MatroskaExtractor that would check whether sampleMetadata should be called or not.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
g-lass: I've messed up while rebasing. First a closing bracket } is  missing on line 104.

As of the regression on AC4, I have indeed broken it while moving most of the readSample method code into a readSample1 method.
I have now fixed both issues in an updated set of commits.

g-lass: The mp4SampleWithDolbyTrueHDTrack() test still fails when simulating i/o errors but I am not quite sure how to debug this because don't quite understand how i/o errors get injected.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/Mp4Extractor.java
g-lass: I think that I get it... you mean "don't change the readSample to read 16 samples a time but to issue the metadata only once every 16 samples for TrueHD" ...

I'll try to implement this.

g-lass:
I have inadvertently deleted the branch while pushing my latest changes upstream....
I have in between "pushed" the branch up again....  but I apparently can't reopen the pull request , can I ?

PR #9421, opened 2021-09-13
MP4: Parse initial PAR from H.265 SPS and include it in HevcConfig

MarcusWichelmann:
Currently, the initial H.265 SPS from the `hvcC` atom/box is not parsed for its pixel aspect ratio which resulted in all content being played back with aspect ratio 1:1 if no `pasp` box is present.

This PR ...
- introduces a `parseH265SpsNalUnitPayload` method to parse a SPS NAL unit, just like it's done for H.264, too. The parsing code is derived from `H265Reader.parseMediaFormat`.
- extends the `HevcConfig` class to parse the initial SPS and propagate the PAR into the Format if no `pasp` is present.
- refactors the `buildHevcCodecStringFromSps` method to re-use the parsed SPS data - just like it's done for H.264.

This PR was split up from #9318.

### DIFF
--- a/library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
@@ -15,6 +15,8 @@
  */
 package com.google.android.exoplayer2.util;
 
+import static java.lang.Math.min;
+
 import androidx.annotation.Nullable;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
@@ -24,7 +26,7 @@ public final class NalUnitUtil {
 
   private static final String TAG = "NalUnitUtil";
 
-  /** Holds data parsed from a sequence parameter set NAL unit. */
+  /** Holds data parsed from a H.264 sequence parameter set NAL unit. */
   public static final class SpsData {
 
     public final int profileIdc;
@@ -68,6 +70,44 @@ public final class NalUnitUtil {
       this.picOrderCountType = picOrderCountType;
       this.picOrderCntLsbLength = picOrderCntLsbLength;
       this.deltaPicOrderAlwaysZeroFlag = deltaPicOrderAlwaysZeroFlag;
+    }
+  }
+
+  /** Holds data parsed from a H.265 sequence parameter set NAL unit. */
+  public static final class H265SpsData {
+
+    public final int generalProfileSpace;
+    public final boolean generalTierFlag;
+    public final int generalProfileIdc;
+    public final int generalProfileCompatibilityFlags;
+    public final int[] constraintBytes;
+    public final int generalLevelIdc;
+    public final int seqParameterSetId;
+    public final int picWidthInLumaSamples;
+    public final int picHeightInLumaSamples;
+    public final float pixelWidthHeightRatio;
+
+    public H265SpsData(
+        int generalProfileSpace,
+        boolean generalTierFlag,
+        int generalProfileIdc,
+        int generalProfileCompatibilityFlags,
+        int[] constraintBytes,
+        int generalLevelIdc,
+        int seqParameterSetId,
+        int picWidthInLumaSamples,
+        int picHeightInLumaSamples,
+        float pixelWidthHeightRatio) {
+      this.generalProfileSpace = generalProfileSpace;
+      this.generalTierFlag = generalTierFlag;
+      this.generalProfileIdc = generalProfileIdc;
+      this.generalProfileCompatibilityFlags = generalProfileCompatibilityFlags;
+      this.constraintBytes = constraintBytes;
+      this.generalLevelIdc = generalLevelIdc;
+      this.seqParameterSetId = seqParameterSetId;
+      this.picWidthInLumaSamples = picWidthInLumaSamples;
+      this.picHeightInLumaSamples = picHeightInLumaSamples;
+      this.pixelWidthHeightRatio = pixelWidthHeightRatio;
     }
   }
 
@@ -252,17 +292,16 @@ public final class NalUnitUtil {
   }
 
   /**
-   * Parses an SPS NAL unit using the syntax defined in ITU-T Recommendation H.264 (2013) subsection
-   * 7.3.2.1.1.
+   * Parses a SPS NAL unit payload using the syntax defined in ITU-T Recommendation H.264 (2013)
+   * subsection 7.3.2.1.1.
    *
    * @param nalData A buffer containing escaped SPS data.
-   * @param nalOffset The offset of the NAL unit header in {@code nalData}.
+   * @param nalOffset The offset of the NAL unit in {@code nalData}.
    * @param nalLimit The limit of the NAL unit in {@code nalData}.
    * @return A parsed representation of the SPS data.
    */
-  public static SpsData parseSpsNalUnit(byte[] nalData, int nalOffset, int nalLimit) {
+  public static SpsData parseSpsNalUnitPayload(byte[] nalData, int nalOffset, int nalLimit) {
     ParsableNalUnitBitArray data = new ParsableNalUnitBitArray(nalData, nalOffset, nalLimit);
-    data.skipBits(8); // nal_unit
     int profileIdc = data.readBits(8);
     int constraintsFlagsAndReservedZero2Bits = data.readBits(8);
     int levelIdc = data.readBits(8);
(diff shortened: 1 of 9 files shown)


library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
```diff
@@ -252,17 +292,16 @@ public static int getH265NalUnitType(byte[] data, int offset) {
...
+   * subsection 7.3.2.1.1.
    *
    * @param nalData A buffer containing escaped SPS data.
-   * @param nalOffset The offset of the NAL unit header in {@code nalData}.
+   * @param nalOffset The offset of the NAL unit in {@code nalData}.
    * @param nalLimit The limit of the NAL unit in {@code nalData}.
    * @return A parsed representation of the SPS data.
    */
-  public static SpsData parseSpsNalUnit(byte[] nalData, int nalOffset, int nalLimit) {
+  public static SpsData parseSpsNalUnitPayload(byte[] nalData, int nalOffset, int nalLimit) {
```
ojw28, line 316: Is there a strong reason for renaming this, given it's a breaking API change (which we can do if we need to, but it doesn't really feel like it's worth the potential disruption in this case)?

If not, let's remove "Payload" from the method below as well, so they're consistent.

library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
```diff
@@ -387,17 +426,166 @@ public static SpsData parseSpsNalUnit(byte[] nalData, int nalOffset, int nalLimi
...
+   * subsection 7.3.2.2.
    *
    * @param nalData A buffer containing escaped PPS data.
-   * @param nalOffset The offset of the NAL unit header in {@code nalData}.
+   * @param nalOffset The offset of the NAL unit in {@code nalData}.
    * @param nalLimit The limit of the NAL unit in {@code nalData}.
    * @return A parsed representation of the PPS data.
    */
-  public static PpsData parsePpsNalUnit(byte[] nalData, int nalOffset, int nalLimit) {
+  public static PpsData parsePpsNalUnitPayload(byte[] nalData, int nalOffset, int nalLimit) {
```
ojw28, line 626: Ditto.

library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
MarcusWichelmann: I renamed the method because I changed its behaviour. It now does no longer skip the 8 header bits because this would prevent the re-use of this method in `FragmentedMp4Extractor` in the other PR.

Yes, that's a breaking change. But if you want, I could re-add a `parseSpsNalUnit` method which skips 8 bits and then calls `parseSpsNalUnitPayload`. Would you prefer that?

Same goes for `parsePpsNalUnitPayload` and `parseH265SpsNalUnitPayload`.

library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
```diff
@@ -252,17 +292,16 @@ public static int getH265NalUnitType(byte[] data, int offset) {
...
 
   /**
-   * Parses an SPS NAL unit using the syntax defined in ITU-T Recommendation H.264 (2013) subsection
-   * 7.3.2.1.1.
+   * Parses a SPS NAL unit payload using the syntax defined in ITU-T Recommendation H.264 (2013)
+   * subsection 7.3.2.1.1.
    *
    * @param nalData A buffer containing escaped SPS data.
-   * @param nalOffset The offset of the NAL unit header in {@code nalData}.
+   * @param nalOffset The offset of the NAL unit in {@code nalData}.
```
ojw28: Can we add "payload" into this sentence? (maybe even explicitly clarify what it does not include).

library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
ojw28: Ah, I didn't spot that. It might be worth keeping
```
parseSpsNalUnit(...) {
  parseSpsNalUnitPayload(nalData, nalOffset + 1, nalLimit);
}
```

MarcusWichelmann:
Re-added the methods now and reverted some changes that were required to adopt to the breaking change. Way cleaner now. 👍🏼

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/H265Reader.java
```diff
@@ -243,10 +243,20 @@ private static Format parseMediaFormat(
...
     bitArray.skipBit(); // sps_temporal_id_nesting_flag
-
-    // profile_tier_level(1, sps_max_sub_layers_minus1)
-    bitArray.skipBits(88); // if (profilePresentFlag) {...}
-    bitArray.skipBits(8); // general_level_idc
+    int generalProfileSpace = bitArray.readBits(2);
+    boolean generalTierFlag = bitArray.readBit();
+    int generalProfileIdc = bitArray.readBits(5);
+    int generalProfileCompatibilityFlags = 0;
+    for (int i = 0; i < 32; i++) {
```
ojw28, line 250: Is this for loop the same as `int generalProfileCompatibilityFlags = bitArray.readBits(32)`?

[I don't think there's any reversing going on, but if there is you could probably use `Integer.reverse` to get them the right way round]

library/extractor/src/main/java/com/google/android/exoplayer2/video/HevcConfig.java
```diff
@@ -105,6 +118,9 @@ public static HevcConfig parse(ParsableByteArray data) throws ParserException {
   @Nullable public final List<byte[]> initializationData;
   /** The length of the NAL unit length field in the bitstream's container, in bytes. */
   public final int nalUnitLengthFieldLength;
+  public final int width;
```
ojw28, line 124: Please could you Javadoc these three fields. Thanks!

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/H265Reader.java
MarcusWichelmann: This code is just moved from [`buildHevcCodecStringFromSps`](https://github.com/google/ExoPlayer/pull/9421/files/45db77dc6ddfb0ca9de191faad184aaf90821012#diff-bfa623b40c5dce28bfd03d9cc30f13aca7fdbbbe035158284d12e2e61a2b31b3L101) I didn't take the risk of changing any logic here.

But it should be not equal to `bitArray.readBits(32)` because Java is big endian and the for loop reverses the bit order. `Integer.reverse` would probably work.

But as mentioned, I didn't want to cleanup/deduplicate existing code here. And when changing it in this place, I should problably also change it in `NalUnitUtil.java`. Do you want that?

ojw28: It seem fine to leave it as-is, if you're just moving it from one place to another. Thanks!

library/extractor/src/main/java/com/google/android/exoplayer2/video/HevcConfig.java
MarcusWichelmann: Fixed.

library/common/src/main/java/com/google/android/exoplayer2/util/NalUnitUtil.java
```diff
@@ -386,6 +438,169 @@ public static SpsData parseSpsNalUnit(byte[] nalData, int nalOffset, int nalLimi
...
+   *
+   * @param nalData A buffer containing escaped SPS data.
+   * @param nalOffset The offset of the NAL unit payload in {@code nalData}.
+   * @param nalLimit The limit of the NAL unit in {@code nalData}.
+   * @return A parsed representation of the SPS data.
+   */
+  public static H265SpsData parseH265SpsNalUnitPayload(byte[] nalData, int nalOffset, int nalLimit) {
+    ParsableNalUnitBitArray data = new ParsableNalUnitBitArray(nalData, nalOffset, nalLimit);
+    // Skip sps_video_parameter_set_id.
+    data.skipBits(8 + 4);
```
ojw28, line 466: It was pointed out during internal review that `sps_video_parameter_set_id` is only 4 bits long. Furthermore, it appears that the NAL unit header for H265 is 16 bits long, so L451 should be adding 2 to the offset rather than 1. We've fixed this as part of merging the change.

Please see the merge [here](https://github.com/google/ExoPlayer/commit/9108dc5b0ee399ed6dbce46664a83bea42a22637) and let us know if there are any problems! Thanks

MarcusWichelmann: @ojw28 Thanks for merging.

> Furthermore, it appears that the NAL unit header for H265 is 16 bits long ...

Than makes sense, thanks for fixing. Interesting that it worked anyway. 😄 

I just tested latest `dev-v2` and everything looks fine. 👍🏼

PR #9477, opened 2021-09-24
Fix metadata being reset when media is repeated

### LINKED ISSUE #9458: MediaMetaData returns null values if player is set to Player.REPEAT_MODE_ONE
- ExoPlayer version number: 2.15.0
- Android version: OS 10
- Android device: Samsung a35

I am playing a local mp3 and to get its information,I've override onMediaMetadataChanged to get the title information like below:

```
    override fun onMediaMetadataChanged(mediaMetadata: MediaMetadata) {
        title = mediaMetadata.albumTitle.toString()
    }
```
At first run, It's successfully showing the title of the currently playing item, but when I set the player.repeatmode to REPEAT_MODE_ONE and right after player goes back to 00:01 position, the `onMediaMetadataChanged` gets called again but right now all datas are null including the title. See image below:

Screen Shot 2021-09-21 at 4 01 08 PM

It should retain all the datas from before. Any idea why this is happening?

Thanks!

marcqtan:
This will fix #9458

When player is set to REPEAT Mode and goes back to position 00:01, all metadata are gone. 

This is working in r2.14.2 and commit `d47d1ebf19a15f59629b220ef9c82b24209e89a7` broke it.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java
@@ -1265,7 +1265,7 @@ import java.util.concurrent.CopyOnWriteArraySet;
                 .windowIndex;
         mediaItem = newPlaybackInfo.timeline.getWindow(windowIndex, window).mediaItem;
       }
-      newMediaMetadata = mediaItem != null ? mediaItem.mediaMetadata : MediaMetadata.EMPTY;
+      mediaMetadata = mediaItem != null ? mediaItem.mediaMetadata : MediaMetadata.EMPTY;
     }
     if (!previousPlaybackInfo.staticMetadata.equals(newPlaybackInfo.staticMetadata)) {
       newMediaMetadata =


library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java
```diff
@@ -1265,7 +1265,7 @@ private void updatePlaybackInfo(
                 .windowIndex;
         mediaItem = newPlaybackInfo.timeline.getWindow(windowIndex, window).mediaItem;
       }
-      newMediaMetadata = mediaItem != null ? mediaItem.mediaMetadata : MediaMetadata.EMPTY;
+      mediaMetadata = mediaItem != null ? mediaItem.mediaMetadata : MediaMetadata.EMPTY;
```
ojw28, line 1268: I think you're reintroducing another bug here. It happens to fix the bug you're seeing, but this line really _should_ be assigning to `newMediaMetadata`.

The real bug is that this line correctly updates `newMediaMetadata`, but then L1271-1272 below don't execute to inject in the metadata from the ID3 data.

This can be fixed by by changing the condition on L1270 to be:

```
if (mediaItemTransitioned
    || !previousPlaybackInfo.staticMetadata.equals(newPlaybackInfo.staticMetadata)) {
```

The fix is easy enough that it's probably quicker for us to do it internally, so I'll close this PR. Thanks!

PR #9495, opened 2021-09-27
Added support for RTSPS

### LINKED ISSUE #9494: Add support for RTSPS streaming
### Use case description
Currently RTSPS (RTSP over SSL) is not supported and URLs following a `rtsps://` scheme will result in an unknown protocol exception.

### Proposed solution
Detecting RTSPS streams by the URL scheme and using the `javax.net.ssl.SSLSocketFactory` instead of the standard `javax.net.SocketFactory` to handle the request.

manuelzomer:
Adds support for RTSPS (RTSP over SSL) streams by detecting `rtsps://` URL schemes and handling them by using `javax.net.ssl.SSLSocketFactory`. 
If no port is provided within a `rtsps://` URL, 322 is used as the default port, following IANA-registered default port for RTSPS.

Fixes #9494

### DIFF
--- a/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspClient.java
+++ b/library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspClient.java
@@ -289,13 +289,18 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 
   /** Returns a {@link Socket} that is connected to the {@code uri}. */
   private static Socket getSocket(Uri uri) throws IOException {
-    checkArgument(uri.getHost() != null);
-    if (uri.getScheme().equals("rtsps")) {
-      int rtspsPort = uri.getPort() > 0 ? uri.getPort() : DEFAULT_RTSPS_PORT;
-      return SSLSocketFactory.getDefault().createSocket(checkNotNull(uri.getHost()), rtspsPort);
-    }
-    int rtspPort = uri.getPort() > 0 ? uri.getPort() : DEFAULT_RTSP_PORT;
-    return SocketFactory.getDefault().createSocket(checkNotNull(uri.getHost()), rtspPort);
+    boolean isRtsps = uri.getScheme().equals("rtsps");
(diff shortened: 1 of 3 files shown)


library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspClient.java
```diff
@@ -288,6 +290,10 @@ private void maybeLogMessage(List<String> message) {
   /** Returns a {@link Socket} that is connected to the {@code uri}. */
   private static Socket getSocket(Uri uri) throws IOException {
     checkArgument(uri.getHost() != null);
+    if (uri.getScheme().equals("rtsps")) {
+      int rtspsPort = uri.getPort() > 0 ? uri.getPort() : DEFAULT_RTSPS_PORT;
+      return SSLSocketFactory.getDefault().createSocket(checkNotNull(uri.getHost()), rtspsPort);
+    }
```
claincly: Would it make sense to generalize it a bit to:

```
boolean isRtsps = uri.getScheme().equals("rtsps");
int port = uri.getPort() > 0 
   ? uri.getPort()
   : isRtsps
     ? DEFAULT_RTSPS_PORT
     : DEFAULT_RTSP_PORT;

SocketFactory factory = isRtsps 
  ? SSLSocketFactory.getDefault()
  : SocketFactory.getDefault();

return factory.createSocket(checkNotNull(uri.getHost()), port)
```

claincly (changes requested):
Thanks for sending the PR. I'm not an expert on TLS, etc, I'm a bit suspicious though that using SSLSocket alone will be enough to support RTSPS.

Could you at the meantime provide some test links that we can use?

library/rtsp/src/main/java/com/google/android/exoplayer2/source/rtsp/RtspClient.java
manuelzomer: Thanks for the feedback, I updated the function now accordingly!

PR #9496, opened 2021-09-27
Dev v2 truehd

g-lass:
reopening https://github.com/google/ExoPlayer/pull/9415 that was automatically closed after deleting the branch by mistake.

the latest commit do address the comments made in https://github.com/google/ExoPlayer/pull/9415 .
To summarize , it does :

- fix the MP4 Extractor test : after re-muxing the TrueHD test vector , all tests are now passing
- implement a TrueHD Rechunker for MP4 Extractor as done in the pre-existing MKV extractor


library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.android.exoplayer2.Format;
+import com.google.android.exoplayer2.drm.DrmInitData;
+import com.google.android.exoplayer2.util.MimeTypes;
+import com.google.android.exoplayer2.util.ParsableByteArray;
+
+/** Utility methods for parsing MLP frames, which are access units in MLP bitstreams. */
+public final class MlpUtil {
+
+  /** The channel count of MLP stream. */
+  private static final int CHANNEL_COUNT_2 = 2;
```
kim-vde, line 31: Could you explain why the channel count is set to a constant value of 2? Thanks.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
```diff
@@ -102,4 +102,10 @@ public void mp4SampleWithColorInfo() throws Exception {
     ExtractorAsserts.assertBehavior(
         Mp4Extractor::new, "media/mp4/sample_with_color_info.mp4", simulationConfig);
   }
+
+  @Test
+  public void mp4SampleWithDolbyTrueHDTrack() throws Exception {
+    ExtractorAsserts.assertBehavior(
+        Mp4Extractor::new, "media/mp4/sample_dthd.mp4", simulationConfig);
```
kim-vde, line 109: Would it be possible to reduce the size of the test file to less than 100KB (for example by reducing its duration)?

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: The channel count is "fairly" indicative.  A Dolby trueHD stream on a bluray generally embeds the necessary information to render multiple versions for the same stream (stereo, 5.1 or 7.1 and atmos).  A bluray player would select which version to render based on its own decoding capabilities as well as the end point's capability.

So even if the stream says 16 objects for a dolby atmos track , the mediacodec would most probably "render" these objects to stereo or bestcase 7.1 and expect a 2 or 8ch decoded buffer queue..

The same has been done for AC4 .. the extractor says 2ch as "indicative".. while the same stream can embed multiple representations. The decoder may decode a full 5.1 and ask for the exoplayer to resize the decoded buffer accordingly.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mp4/Mp4ExtractorTest.java
g-lass: The current test file represents 1s of audio.  For this test , a 0.4s teststream would certainly do..
I'll regenerate the stream and update the reference dumps.

g-lass:
the latest drop 
- cuts the testfile to less than 100kB (418ms total duration).
- fixes a crash observed after cutting the testfile.
- clarifies why I set the number of channels to 2 irrespective of the stream content.
- addresses a couple of warnings from the compiler.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
kim-vde: Is the channel count passed in the Format used in any way, or could we just pass anything there (e.g. the channel count from the stsd sample entry)?

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
```diff
@@ -0,0 +1,104 @@
...
+   * The number of samples to store in each output chunk when rechunking TrueHD streams. The number
+   * of samples extracted from the container corresponding to one syncframe must be an integer
+   * multiple of this value.
+   */
+  public static final int TRUEHD_RECHUNK_SAMPLE_COUNT = 16;
+
+  /**
+   * Rechunks TrueHD sample data into groups of {@link #TRUEHD_RECHUNK_SAMPLE_COUNT} samples.
+   */
+  public static class TrueHdSampleRechunker {
```
kim-vde, line 71: As samples are grouped by chunks, shouldn't you make sure that, when you seek, you sync to the start of a chunk? I think this is what is being done in the MatroskaExtractor.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private MlpUtil() {}
+
+  /**
+   * The number of samples to store in each output chunk when rechunking TrueHD streams. The number
+   * of samples extracted from the container corresponding to one syncframe must be an integer
+   * multiple of this value.
+   */
+  public static final int TRUEHD_RECHUNK_SAMPLE_COUNT = 16;
```
kim-vde, line 66: It would be great to re-use the constant "TRUEHD_RECHUNK_SAMPLE_COUNT" from Ac3Util here to avoid duplication. I am wondering, is there a link between TrueHD and AC3, or are they completely separate codecs? If they are separate, we should probably split Ac3Util into Ac3Util and TrueHdUtil. This can be done in a separate PR. What do you think?

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: A channel count is given in the TrueHD SampleEntry box. 
But the reliable info (including all possible renditions) is given in the elementary stream itself.
The decoder is expected to use the in-stream information as well as the knowledge of end-point characteristics to render appropriately.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: I have expected that the SEEK would systematically jump at the beginning of a sample chunk... i.e. when the user randomly jumps through the timeline, the extractor would be jumping to the closest chunck and start extracting samples from there.

Though I did not study it in details, the MatroskaExtractor seems to be looking for "major sync" and to my knowledge in an MP4 file the first sample of a chunk is a major sync (at least it is in the dolby-atmos-music files that I have played during testing).

If for some reason, this is not the case, then I'd be keen to better understand how this is working , maybe by trying to hit that situation in a test... Would the extractor "skip" some access units to "pre-align" A/V tracks ?  
If the skip happens above the extractor , I guess it would skip a set of 16 rechuncked samples so anyway the major-sync provided by the first rechuncked sample would be skipped and the decoder will drop every coming access units until it get the next major sync (in most TrueHD file that we encode, this is every 128  access units or 106.6 ms)

My understanding is that seeking at the beginning of the playback (after the reset of the rechunker) must definitely start with the major sync so that the decoder decodes the very first samples that it receives and so that the audio timestamps be accurate for the AV sync logic.... if one seek within the file while playing, I don't think this is so critical as the info given in the major sync should be fairly constant throughout a track.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: I had actually thought of refactoring the Ac3Util code moving the TrueHD related code into the MlpUtil where they actually belong but definitely as a separate PR.

AC3 and TrueHD have nothing in common.... 
I suppose that the TrueHD code was added to Ac3 Utils by convenience because the primary usecase was playback of BluRay content where TrueHD tracks are generally backed up by an AC3 track for compatibility with old systems .(I speculate that the ripping process from M2TS to MKV preserves the 2 tracks).

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: Note that I have tested seeking on the test files that I have using the NVidia Shield connected to a Dolby Atmos AVR and I did not see a problem.
Because my "test suite" primarily consists of music files, I'll generate new test content with some video and see how this behaves. But unless I stitch audio tracks from different albums  I don't expect a problem since the info conveyed by the major sync is constant. An TrueHD decoder receiving a first group of samples that is not a major sync would drop these access units until it get the major sync.

So does the defaultAudioSink as far as I understand.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
kim-vde: Thanks for your input on that. If you are making a new PR for that, can you add in the description that it is a follow up of this one so that it gets assigned to me?

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
kim-vde: Here is a summary of how the seek is working for MP4:

- When the user seeks to a specific time, method Mp4Extractor.getSeekPoints is called to determine 2 positions in the stream from where to start extracting: one before the requested time, and one after. The extractor then starts extracting from one of these positions, depending on the SeekParameters.
- The positions returned by getSeekPoints always correspond to key frames (flag C.BUFFER_FLAG_KEY_FRAME is set).
- For more details, see the code of getSeekPoints (code is clearer than words in this case).

I had a closer look at this specific case and here is what I found out:
- In the test file you provided, flag C.BUFFER_FLAG_KEY_FRAME is set every 128 samples, which is a multiple of TRUEHD_RECHUNK_SAMPLE_COUNT, so that the samples returned by getSeekPoints should always correspond to chunk starts. Assuming the flag is always set similarly, this means that we don't need to sync to a chunk start in the rechunker.
- When seeking, TrueHdSampleRechuncker.chunkSampleCount should be reset to 0. Otherwise the chunk borders computed could be wrong. I have done so in my own version of this CL so there is nothing to do there as well.

So I think you are right and that there is nothing left to do to make seek working.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
kim-vde: > The decoder is expected to use the in-stream information as well as the knowledge of end-point characteristics to render appropriately.

Can you clarify what you mean by end-point characteristics?

g-lass: I mean "whether the end point is stereo , 5.1 or designed for atmos" .
On most devices, the TrueHD bitstream will be passed-through  to an AVR or soundbar over HDMI / eARC and the decoder of that AVR/soundbar will take care of decoding and rendering to suit the speaker setup.
Some ATMOS anrdoid TVs will be able to decode it as well internally (most probably using an MSD HAL based decoder implementation so here as well the exoplayer would push the TrueHD access units to the HAL via a Direct Audio Track ).

in both cases (plain passthrough or MSD HAL based decoding) , the number of channels read by the extractor would not matter ... 

However I know that the number of channels is one of the criteria  for automatic track-selection but not the only one (some criteria like the codec would make sense as well) but if the exoplayer auto-selects the tracks based on the number of channels (following the motto the more , the better) then I guess I would tend to vote for declaring that the track is actually 8 channels by default.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: Thanks for the details.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: Sure, I'll refer to this PR.

library/common/src/main/java/com/google/android/exoplayer2/audio/MlpUtil.java
g-lass: Another note about this topic : the TrueHD sample entry box is specified as follow : 
`MLPSampleEntry().{`
`....BoxHeader.Size.......u(32)`
`....BoxHeader.Type.......u(32)...mlpa`
`....Reserved.............u(48)......0`
`....Data-reference-index.u(16)`
`....Reserved.............u(64)......0`
`....ChannelCount.........u(16)......2`
`....SampleSize...........u(16).....16`
`....Reserved.............u(32)......0`
`....SampleRate...........u(32)`
`....MLPSpecificBox`
`}`

so just reading this box would not give more info than hardcoding the channel count to 2.

kim-vde: Thanks for the details provided!

Channel count is indeed used for track selection but it might also be used to decide the size of the input and output buffers of the decoder (not sure though). I think it is safer to have it hard-coded to 2 (like for AC4) for now and we can still change it later on if it causes issues.

g-lass: Good.. let's do so.
Is there anything else I need to look at to make get this PR merged ?

kim-vde: No. I have refactored the code a bit and it is currently under review internally. I will merge as soon as I can.

Issue #2882, opened 2017-05-29
Support key-frame accurate seeking in HLS


mohammed-atif:
<h3>SimpleExoPlayer.seekTo(int millis) issue</h3>
<p>I have seen several issues requesting the same feature of smoother seeking in Exoplayer</p>
<p>At one thread I have seen your reply mentioning about the decoder but Android's VideoView default seeking is much more smoother when compared to ExoPlayer or MediaPlayer</p>

<p><b>Here is my code for reference</b></p>

```
@Override
    public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
        long seekTo = (long) (mPlayer.getCurrentPosition() - distanceX);
        Log.d(TAG, "current scroll position "+seekTo);
        mPlayer.seekTo(seekTo);
        return true;
    }

```
I am using locally stored mp4 file to play using Exoplayer with TextureView

current exoplayer version - r2.4.1
```com.google.android.exoplayer:exoplayer-core:r2.4.1```

Are there any configurations that I have to change for smoother seeking?

**Comment by ojw28** [comment]
The reason behavior is different is that ExoPlayer does frame-accurate seeking (in all cases). I think VideoView does key-frame accurate seeking, which is faster but less accurate. We can use this issue to track supporting key-frame accurate seeking as an option, but it's low priority.

**Comment by mohammed-atif** [comment]
@ojw28 , In other players, we have a callback which notifies about the successful seeking. If I make multiple calls for seek using onScroll method, I can ignore the inputs if seek is in progress, but in case of ExoPlayer there is no such callback and due to multiple seek requests, the performance of seeking further goes down. It would be really good if ExoPlayer too provides the callbacks for seek completion.

Apart from that, as you mentioned about the key-frame seeking, can you provide me details about the class which I can look into to do minor tweaks at my end for better seeking?

**Comment by goffioul** [comment]
@AtifMohammed Not sure this will help you, but I also needed some callback after seek completion. What I come up with was to drop a flag in my app when starting the seek operation, then listen to state change in onPlayerStateChanged. If my flag was set and the player state was STATE_READY, then I fired my callback. That seemed to work for my use case.

**Comment by mohammed-atif** [comment]
@goffioul Thanks for the response, I figured it out few days ago but couldn't update here. While seeking the state goes to Buffer and after seek it shifts to ready.

@ojw28 forward seek works good after adding the callbacks. But when I seek backwards, player's current time is not updated.

```
[...]

Issue #8592, opened 2021-02-16
Support renderer mode that only renders the first frame (for trick play)


davydovvladimir:
Hello dear developers!

First, let me explain  what I'm doing.
I'm trying to implement reverse trick mode by using I-Frames provided by #EXT-X-I-FRAME-STREAM-INF track.
The current algorithm is as follows:
1. Set playback to pause by player.setPlayWhenReady(false)
2. Select trick-play tracks by SelectionOverride.
3. Get current position by player.getCurrentPosition()
4. Set a new position reduced by N seconds relative to the current.
5. Wait for onRenderedFirstFrame() and perform step 3 again. So, steps 3,4,5 are in the loop. 

This algorithm is working for me but has some disadvantages.
The main problem is the time between seek() operation and onRenderedFirstFrame event.
It takes about 500ms and looks pretty big.
I have debuged the code and found that first frame is rendered only after 4-5 I-Frames pushed to the decoder.
I found it by additional trace before https://github.com/google/ExoPlayer/blob/4b1e0fa9fc643f96ab5bbf4cec9aa7a79e4bc789/library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java#L1349.
I've tried to control buffering  by overriding DefaultLoadControl.shouldContinueLoading() and I see if I load only 2 or 3 I-Frame chunks after seek(), then the onRenderedFirstFrame event does not come at all.

In theory, sending one i-frame to the decoder is enough to receive onRenderedFirstFrame.
But maybe MediaCodec or Renderer has some internal buffering and requires mulptiple frames to display the first.

Could you tell me why this is happening and how to avoid it?

Thanks.

andrewlewis:
> In theory, sending one i-frame to the decoder is enough to receive onRenderedFirstFrame.
> But maybe MediaCodec or Renderer has some internal buffering and requires mulptiple frames to display the first.

It is indeed the case that `MediaCodec`s generally require several frames to be queued before they output the first frame, even if (theoretically) there's enough information in the bitstream to output a frame immediately after it's queued. The way around this would be to queue end-of-stream after the I-frame to cause the decoder to output any pending data, but I don't think there's an easy way to get this behavior with the stock ExoPlayer video renderer (you would need to customize it a bit).

Side note: some information in https://github.com/google/ExoPlayer/issues/6794#issuecomment-573068211 may be interesting, if only to confirm your current solution is reasonable.

ojw28:
[...]

Issue #8952, opened 2021-05-17
Fix to allow Audio timestampMaster causes stream startup delay


stevemayhew:
The patch, allow audio master, commit [6f8a8f](https://github.com/google/ExoPlayer/commit/6f8a8fbc1cf974617056edac6ee0068b175a6454), fixes stall issues with ordering of updates of the DSN in segments across demuxed renditions however the cost is the seconds of delay on channel change or stream startup, a bad user experience. 

To reproduce this bug start playback of any HLS live stream where audio timestamps precede video in the sequence of segments.  

* Expected:
    1.  Initial load of video sets start time reference
    2. Audio render is not started (samples are discarded, until they match video time)
    3. Playback starts immediately after first frame shows (early audio samples are discarded)

* Actual:
   1. Audio playback starts immediately 
   2. First frame shows immediately (as this is a forced render, `MediaCodecVideoRenderer` forces first frame in `processOutputBuffer()`
   3. First frame is frozen until all audio preceding it plays out

This change to allow audio variants to reset timestamp adjustment fixes issues we have with live playback and marked discontinuities, so any fix must still address that issue.  The bug here is simple.

1. Demuxed audio rendition playlist is updated, changing the discontinuity sequence (DSN)
2. Next audio segment blocks for timestamps (as a freshly minted `TimestampAdjuster` is picked up)
3. Video fetches continue on a stale DSN / `TimestampAdjuster` (cached version, or just the time ordering of the origin writing the playlists) so playback freezes

Because audio and video are kept in step by the `CompositeSequenceableLoader` (not allowing timeline progress) then playback is frozen forever.

Note the video rendition playlist will update to the new DSN, but the player does not see it.

stevemayhew:
Very possible that issue #8952 is related to this.

stevemayhew:
@AquilesCanta 

Assuming the best fix is to return to using video must master timestamps.  This is the best fix I believe, and the timestamp master certainly should not change during playback, only on track selection if the track that was timestamp master is unselected.

This statement seems an interesting path: 
> The alternative that was considered was having the code that selects the next chunk to load be smart enough to not select a chunk which is not the timestampAdjuster master and has a new discontinuity sequence number

This could work, at least for our use case, the trick is we need to refresh the *timestamp master*'s playlist in order to have it pickup the correct updated DSN.   Here's why

[...]

PR #9536, opened 2021-10-06
Implements SeekParameters.*_SYNC variants for HLS

stevemayhew:
The HLS implementation of `getAdjustedSeekPositionUs()` now completely supports `SeekParameters.CLOSEST_SYNC` and it's brotheran, assuming the HLS stream indicates segments all start with an IDR (that is EXT-X-INDEPENDENT-SEGMENTS  is specified).

This fixes issue #2882 and improves (but does not completely solve #8592

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -257,8 +257,13 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
       mediaPlaylist = playlistTracker.getPlaylistSnapshot(playlistUrls[selectedIndex], /* isForPlayback= */ true);
     }
 
-    // Resolve to a segment boundary, current track is fine (all should be same).
-    // and, segments must start with sync (EXT-X-INDEPENDENT-SEGMENTS must be present)
+    // If segments must start with sync (EXT-X-INDEPENDENT-SEGMENTS is set) and the playlist is not empty
+    // resolve using the nearest segment start and the next segment (if any) start as the first and second
+    // sync points.
+    // Note, the position returned is normalized to the period, so it will work if a track selection changes
+    // variants before the seek is executed. It is possible it may not land exactly on a segment sync point in the rare
+    // case the segment boundaries do not align across variants.
+    //
     if (mediaPlaylist != null && mediaPlaylist.hasIndependentSegments && !mediaPlaylist.segments.isEmpty()) {
       long startOfPlaylistInPeriodUs = mediaPlaylist.startTimeUs - playlistTracker.getInitialStartTimeUs();
       long targetPositionInPlaylistUs = positionUs - startOfPlaylistInPeriodUs;
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
@@ -26,6 +26,7 @@ import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
+import com.google.android.exoplayer2.SeekParameters;
 import com.google.android.exoplayer2.source.BehindLiveWindowException;
 import com.google.android.exoplayer2.source.TrackGroup;
 import com.google.android.exoplayer2.source.chunk.BaseMediaChunkIterator;
@@ -235,6 +236,43 @@ import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
    */
   public void setIsTimestampMaster(boolean isTimestampMaster) {
     this.isTimestampMaster = isTimestampMaster;
+  }
+
+  /**
+   * Adjusts a seek position given the specified {@link SeekParameters}.  The HLS Segment start times
+   * are used as the sync points iff the playlist declares {@link HlsMediaPlaylist#hasIndependentSegments}
+   * indicating each segment starts with an IDR.
+   *
+   * @param positionUs The seek position in microseconds.
+   * @param seekParameters Parameters that control how the seek is performed.
+   * @return The adjusted seek position, in microseconds.
+   */
+  public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParameters) {
+    long adjustedPositionUs = positionUs;
+
+    int selectedIndex = trackSelection.getSelectedIndex();
+    boolean haveTrackSelection = selectedIndex < playlistUrls.length && selectedIndex != C.INDEX_UNSET;
+    @Nullable HlsMediaPlaylist mediaPlaylist = null;
+    if (haveTrackSelection) {
+      mediaPlaylist = playlistTracker.getPlaylistSnapshot(playlistUrls[selectedIndex], /* isForPlayback= */ true);
+    }
+
+    // Resolve to a segment boundary, current track is fine (all should be same).
+    // and, segments must start with sync (EXT-X-INDEPENDENT-SEGMENTS must be present)
+    if (mediaPlaylist != null && mediaPlaylist.hasIndependentSegments && !mediaPlaylist.segments.isEmpty()) {
+      long startOfPlaylistInPeriodUs = mediaPlaylist.startTimeUs - playlistTracker.getInitialStartTimeUs();
+      long targetPositionInPlaylistUs = positionUs - startOfPlaylistInPeriodUs;
+
+      int segIndex = Util.binarySearchFloor(mediaPlaylist.segments, targetPositionInPlaylistUs, true, true);
+      long firstSyncUs = mediaPlaylist.segments.get(segIndex).relativeStartTimeUs + startOfPlaylistInPeriodUs;
+      long secondSyncUs = firstSyncUs;
+      if (segIndex != mediaPlaylist.segments.size() - 1) {
+        secondSyncUs = mediaPlaylist.segments.get(segIndex + 1).relativeStartTimeUs + startOfPlaylistInPeriodUs;
+      }
+      adjustedPositionUs = seekParameters.resolveSeekPositionUs(positionUs, firstSyncUs, secondSyncUs);
+    }
+
+    return adjustedPositionUs;
   }
 
   /**
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
@@ -30,6 +30,7 @@ import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.Format;
 import com.google.android.exoplayer2.FormatHolder;
 import com.google.android.exoplayer2.ParserException;
+import com.google.android.exoplayer2.SeekParameters;
 import com.google.android.exoplayer2.decoder.DecoderInputBuffer;
 import com.google.android.exoplayer2.drm.DrmInitData;
 import com.google.android.exoplayer2.drm.DrmSession;
@@ -584,6 +585,31 @@ import org.checkerframework.checker.nullness.qual.RequiresNonNull;
     return chunkSource.onPlaylistError(playlistUrl, exclusionDurationMs)
         && exclusionDurationMs != C.TIME_UNSET;
   }
+
+  /**
+   * Check if the primary sample stream is video, {@link C#TRACK_TYPE_VIDEO}.  This
+   * HlsSampleStreamWrapper may managed audio and other streams muxed in the same
+   * container, but as long as it has a video stream this method returns true.
+   *
+   * @return true if there is a video SampleStream managed by this object.
+   */
+  public boolean isVideoSampleStream() {
+    return primarySampleQueueType == C.TRACK_TYPE_VIDEO;
+  }
+
+
+  /**
+   * Adjusts a seek position given the specified {@link SeekParameters}.  Method delegates to
+   * the associated {@link HlsChunkSource#getAdjustedSeekPositionUs(long, SeekParameters)}.
+   *
+   * @param positionUs The seek position in microseconds.
+   * @param seekParameters Parameters that control how the seek is performed.
+   * @return The adjusted seek position, in microseconds.
+   */
+  public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParameters) {
+    return chunkSource.getAdjustedSeekPositionUs(positionUs, seekParameters);
+  }
+
 
   // SampleStream implementation.
 
--- a/library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
+++ b/library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
@@ -1,0 +1,197 @@
+package com.google.android.exoplayer2.source.hls;
+import static com.google.common.truth.Truth.assertThat;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+import android.net.Uri;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+
+import com.google.android.exoplayer2.Format;
+import com.google.android.exoplayer2.SeekParameters;
+import com.google.android.exoplayer2.source.MediaPeriod;
+import com.google.android.exoplayer2.source.hls.playlist.HlsMediaPlaylist;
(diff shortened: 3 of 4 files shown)


stevemayhew:
@ojw28  Here is the first step to solving the faster first render requested in #8592.   In my testing on BRCM and AmLogic platforms it does not completely solve the issues, but we are looking at ways to intimidate the decoder into presenting a single IDR immediately.

At any rate, here's a fix for one bug on your plate :-).    I am happy to split out the test cases into a set that test the logic in `SeekParameters.resolveSeekPositionUs()` as a separate unit-test (seemed more proper way),  I did it this way to limit the change to a single module.   It could also be tech-debt I'm happy to take on.

Still working on a design for the fix for #8952.  Haven't forgotten about this one.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -237,6 +238,31 @@ public void setIsTimestampMaster(boolean isTimestampMaster) {
...
+    long adjustedPositionUs = positionUs;
+
+    int selectedIndex = trackSelection.getSelectedIndex();
+    boolean haveTrackSelection = selectedIndex < playlistUrls.length && selectedIndex != C.INDEX_UNSET;
+    @Nullable HlsMediaPlaylist mediaPlaylist = null;
+    if (haveTrackSelection) {
+      mediaPlaylist = playlistTracker.getPlaylistSnapshot(playlistUrls[selectedIndex], /* isForPlayback= */ true);
+    }
+
+    // Resolve to a segment boundary, current track is fine (all should be same).
```
ojw28: > current track is fine (all should be same)

I don't think this is required by the HLS spec, and we've definitely seen cases where streams deliberately use different segment durations for different playlists, in which case the segment boundaries are no longer aligned.

Presumably it's highly likely that the seek will be performed into the current track, and making this work for the common case definitely seems worth it for possibly getting things wrong in the edge case that there's a track change _and_ segments are not aligned. So this is still fine, but maybe delete the comment?

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -237,6 +238,31 @@ public void setIsTimestampMaster(boolean isTimestampMaster) {
...
+    // and, segments must start with sync (EXT-X-INDEPENDENT-SEGMENTS must be present)
+    if (mediaPlaylist != null && mediaPlaylist.hasIndependentSegments) {
```
ojw28: You probably need to check `!mediaPlaylist.segments.isEmpty()` as part of this condition, to avoid making any adjustment in the edge case where there aren't any segments in the playlist yet.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -237,6 +238,31 @@ public void setIsTimestampMaster(boolean isTimestampMaster) {
...
+      int segIndex = Util.binarySearchFloor(mediaPlaylist.segments, positionUs, true, true);
```
ojw28: This is using the absolute `positionUs`, but I think the search is expecting a relative position. So I think you need to do something like:

```
long startOfPlaylistInPeriodUs = playlist.startTimeUs - playlistTracker.getInitialStartTimeUs();
long targetPositionInPlaylistUs = positionUs - startOfPlaylistInPeriodUs;
```

and then use `targetPositionInPlaylistUs` to do the binary search. You'll then need to add `startOfPlaylistInPeriodUs` back into `firstSyncUs` and `secondSyncUs` so that they're absolute positions, before resolving the seek position.

Please double check you agree with this. As a point of reference, look at the other place in this class that calls `Util.binarySearchFloor`, and look at what value it's passing in. I think that my suggestion above is consistent with what's happening there.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -237,6 +238,31 @@ public void setIsTimestampMaster(boolean isTimestampMaster) {
     this.isTimestampMaster = isTimestampMaster;
   }
 
+  long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParameters) {
```
ojw28: Please make public, and add Javadoc.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
```diff
@@ -585,6 +586,16 @@ public boolean onPlaylistError(Uri playlistUrl, LoadErrorInfo loadErrorInfo, boo
         && exclusionDurationMs != C.TIME_UNSET;
   }
 
+  public boolean isVideoSampleStream() {
```
ojw28, line 596: Add Javadoc. Ditto for the method below.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
```diff
@@ -0,0 +1,171 @@
...
+        mockExtractorFactory,
+        mockPlaylistTracker,
+        new Uri[] {PLAYLIST_URI},
+        new Format[] { ExoPlayerTestRunner.VIDEO_FORMAT },
+        mockDataSourceFactory,
+        null,
+        new TimestampAdjusterProvider(),
+        null);
+
+    assertThat(testee).isNotNull();
```
ojw28: This assertion is guaranteed to be true from the line immediately above. Remove.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  @Test
+  public void getAdjustedSeekPositionUs_PreviousSync() {
+    when(mockPlaylistTracker.isSnapshotValid(eq(PLAYLIST_URI))).thenReturn(true);
```
ojw28: Any reason not to pull this up from every test method into `setup`?

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Mock
+  private HlsDataSourceFactory mockDataSourceFactory;
+  private HlsChunkSource testee;
+
+  @Before
+  public void setup() throws IOException {
+    // sadly, auto mock does not work, you get NoClassDefFoundError: com/android/dx/rop/type/Type
+//    MockitoAnnotations.initMocks(this);
+    mockExtractorFactory = Mockito.mock(HlsExtractorFactory.class);
```
ojw28: Why can't you just pass a real instance (i.e., `HlsExtractorFactory.DEFAULT`)? If this works then it's generally much better than mocking. In particular because this test doesn't randomly start failing due to some unrelated change that ends up meaning the mock starts to be called.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
```diff
@@ -0,0 +1,171 @@
...
+    mockPlaylistTracker = Mockito.mock(HlsPlaylistTracker.class);
+    mockDataSourceFactory = Mockito.mock(HlsDataSourceFactory.class);
```
ojw28, line 82: Ditto. I'm not sure about the `HlsPlaylistTracker`. It's possible that you actually need to mock that, but please take a look.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Yes, that sounds good.   We are actually working on another bug with how HlsChunkSource switches tracks using MSN to locate the matching segment.  This is explicitly forbidden in the spec when it is intra-variant:

From 6.3.2

Allowed within the same playlist:
>    A client MAY use the segment Media Sequence Number to track the
   location of a Media Segment within a Playlist when the Playlist is
   reloaded.

But not intra-variant:

>    A client MUST NOT assume that segments with the same Media Sequence
   Number in different Variant Streams or Renditions have the same
   position in the presentation; Playlists MAY have independent Media
   Sequence Numbers.  Instead, a client MUST use the relative position
   of each segment on the Playlist timeline and its Discontinuity
   Sequence Number to locate corresponding segments

Where this matters is we have an origin vendor that is using GAP to mark dropped GOP sections from the transcode stream for individual variants.   As such, they may have playlist that look like:

V1:  GAP(2s) 1a - Segment 1b (4s) - Segment 2(6s)
V2: Segment 1 (6s) - Segment 2 (6s)

So Segment 2 in the two variants has the same content, but different MSN.  I believe this will break the logic in `HlsChunkSource.getNextChunk()`

For this particular method, consider the case where the current index playlist is V1, if the time locates "Segment 1b" start as nearest sync.  If the track switch changes to V2 between this call to `getAdjustedSeekPositionUs()` and the load for the `seekTo()` operations then yes we would land in the middle of V2's Segment 1.

Bottom line, remove the comment is a good idea, it shouldn't matter if the segment counts/durations change because everything is playlist time based, which is mandated to be consistent intra-playlist:

> Instead, a client MUST use the relative position
   of each segment on the Playlist timeline and its Discontinuity
   Sequence Number to locate corresponding segments.

How about the test case?  Are you ok with this in HLS or should I move the logic to test `SeekParameters` itself (IMHO this is the better place if we are ok with changes to core library)

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Good call.  I'll add a test case for this too. `Util.binarySearchFloor()` will return 0 in this case, which will lead to an IndexOutOfBounds exception.   Thanks for finding that bug.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Yes, I'd like to ask about this...  So is the idea to normalize different staleness for playlist snapshots when you have to switch tracks?  Probably worth adding some unit test case logic to cover this case if it is possible.  If there is no playlist switch then the period relative time vs inter-playlist time shouldn't be different.  Maybe @tonihei or @AquilesCanta have  some insight here.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Sure, I'll follow what core did for DASH.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsSampleStreamWrapper.java
stevemayhew: sure.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
stevemayhew: I'll give it a try, yes sadly Mockito comes with a host of issues with Android mocks in Roboelectric.  Also, I'm fine with moving most of this to a test of `SeekParameters` directly.  Then it covers all users of that code not use HLS.

Only HLS specific stuff will be empty playlist and if we can test track switching (ABR)

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Ah, yes.. the `Player.seekTo()` is relative to the `Window` in the `Timeline` but looks like `ExoPlayerImplInternal` resolves this as period time (not clear in the `MediaPeriod` Javadoc, but thinking about it of course all `positionUs` values in that API are in period time).  Certainly hard to follow in `ExoPlayerImplInternal.seekToInternal()`

Will fix and add to unit tests to cover this.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: Fixed, see `getAdjustedSeekPositionUs_EmptyPlaylist()` test and isEmpty check

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: I deleted the comment,  we will put a sample stream together with tracks where segment media sequence is not aligned and enter bugs for this if there are playback issues as described above.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
stevemayhew: This is fixed in latest version of the commit to use `MediaPeriod` relative `positionUs` and covert it to/from playlist relative.  Test cases updated as well.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
stevemayhew: Yes, using the `DefaultHlsPlaylistTracker`, IMO, would add a lot of complexity (`Loader`, `DataSource`, etc) when all we need it to do is return the current playlist, report it valid and the period start time.

Always worth considering the cost/benefit of mocking an object vs using the defaults though.

stevemayhew:
@ojw28  I think all of the comments/issues have been addressed.   Do let me know, thanks so much for the review and catching the bug with `MediaPeriod` vs `Timeline` / playlist relative time.

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
stevemayhew: done

ojw28:
@stevemayhew - Are you certain you pushed all of your changes? Looking at your comment starting "I deleted the comment", it appears the comment is still there. That's not particularly important in itself, but it's very hard to tell whether there are also other changes that haven't been pushed?

I'm not sure if you got into a bad state and were forced into it, but force pushing a single commit on top of a previous review makes it incredibly hard to work out what changed. Please just push additional commits, if possible.

stevemayhew:
@ojw28  Yes I updated the commit and did a force-push.   Sure, in the future I'll make new commits each time for reviews you are on.   Some people don't like lots of commits on a pull request branch, but I can see addressing review comments it is nice to see the diffs.

I do like when people interactive rebase to a single change before the final merge, that makes a cleaner history.  But next time, I'll have you let me know when you are ready to merge then I can do that before the final merge.

Thanks for the review...  BTW, the change does speed up paused seeks on i-Frame only playlist a bit.  I've got some other things I'm working on for #8592 as well.

Once it is all a bit more baked I can add a track selector setting for iFrame only mode and add the scrub seek to the demo app.

ojw28:
So is the comment not being deleted just an omission? I guess I saw that along with the "done" comment, and was then unsure whether there are other accidental omissions too (and can't easily see a diff to check). If you could confirm that everything is as it should be, and that the comment not being deleted was just a singular omission, that would be great. Thanks!

ojw28:
In case it's not completely clear what I'm referring to, I'm referring to this (comment in the code is still there, but comment in the discussion thread says the comment in the code has been deleted):

https://github.com/google/ExoPlayer/pull/9536/files#diff-a500e9307d881d3b7b6a92999132c39808322d24b6f6a4a57a1c0d03cd6cf513R260

stevemayhew:
Let me double check it all over and diff against our version (which is very close, but based from release 2.12) just to be double sure.

The comment was not removed, it is misleading.

The resolved sync point position will be valid in any variant, however it may not match a segment boundary. If for example there were sub-segment duration `EXT-X-GAP` segments in only one playlist such as to cause an disparate list of segments inter-variant.

This actually will cause the MSN based algorithm later in this class to pick in incorrect segment on track switch, so probably better just to add a bug

stevemayhew:
@ojw28 I double checked all the diffs match up with what we have in our 2.12 version.   The unit test cases all run.

I updated the comment so it is correct now,  or I can remove it completely too if that is preferred.   Also, let me know if  you want me to squash the commits into one once we agree the review is complete and it is ready to merge.   Certainly makes it easier to merge then bunches of edit commits from review comments.  In the future I will not do this until it is ready to merge.   Sorry again for the confusion.

ojw28:
Thanks! We'll get this merged.

> Also, let me know if you want me to squash the commits into one once we agree the review is complete and it is ready to merge. Certainly makes it easier to merge then bunches of edit commits from review comments.

The tooling & internal review process we use gives us with a flat view anyway, so it doesn't really make any difference to us. The only benefit of squashing is that it makes the commit history look a little neater in the end. On the flip side, anyone wanting to revisit the review of this PR later and understand any of the discussion threads will find it much harder if there's a squash and force push, because some of those threads wont be tied to any accessible source code any more. I think being able to revisit the PR review is probably more important, so overall I think that not squashing is probably preferable for cases like this.

PR #9543, opened 2021-10-07
Add support for RF64 wave files

KasemJaffer:
This commit adds support for RF64 wave files without breaking the existing support for RIFF wave files


library/extractor/src/main/java/com/google/android/exoplayer2/extractor/wav/WavHeaderReader.java
```diff
@@ -117,14 +117,23 @@ public static WavHeader peek(ExtractorInput input) throws IOException {
     ParsableByteArray scratch = new ParsableByteArray(ChunkHeader.SIZE_IN_BYTES);
     // Skip all chunks until we find the data header.
     ChunkHeader chunkHeader = ChunkHeader.peek(input, scratch);
+    long dataSize = -1;
     while (chunkHeader.id != WavUtil.DATA_FOURCC) {
       if (chunkHeader.id != WavUtil.RIFF_FOURCC && chunkHeader.id != WavUtil.FMT_FOURCC) {
```
kim-vde: Can you change the condition here so that no warning is logged for RF64?

kim-vde (commented):
It would be great if you could add a unit test for RF64 in WavExtractorTest! Is it possible to generate a RF64 file that is fairly small in size (ideally less than 100KB)?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/wav/WavHeaderReader.java
KasemJaffer: Thanks for the review.. Can you tell me please what the warning say so i can fix it accordingly?

KasemJaffer:
I added a unit test and fixed a bug i found along the way.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/wav/WavExtractorTest.java
```diff
@@ -53,4 +53,10 @@ public void sample_imaAdpcm() throws Exception {
     ExtractorAsserts.assertBehavior(
         WavExtractor::new, "media/wav/sample_ima_adpcm.wav", simulationConfig);
   }
+
+  @Test
+  public void sample_RF64() throws Exception {
```
kim-vde, line 58: You also need to generate the dump files that go with the test. To do so, you need to,
- Set DUMP_FILE_ACTION to WRITE_TO_LOCAL in DumpFileAsserts.java
- Run the test. This will create dump files that are a summary of the data extracted by the WavExtractor.
- Set DUMP_FILE_ACTION back to COMPARE_WITH_EXISTING.
- Run the test. This will compare the extracted data with the files previously generated. This test need to pass.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/wav/WavHeaderReader.java
kim-vde: The warning says that the current chunk is being ignored. RF64 and DS64 chunks are not ignored anymore so the condition should be 

chunkHeader.id != WavUtil.RIFF_FOURCC && chunkHeader.id != WavUtil.FMT_FOURCC && chunkHeader.id != WavUtil. RF64_FOURCC && chunkHeader.id != WavUtil. DS64_FOURCC

KasemJaffer:
You are right!  I changed the condition.. it shouldn't log a warning now. 

Also, I followed the  steps you mentioned to generate dump files and run the test. The test is passing.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/wav/WavHeaderReader.java
KasemJaffer: Done :)

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/wav/WavExtractorTest.java
KasemJaffer: Just added the dump files and the test is passing :)

KasemJaffer:
I don't know why its saying there is a conflicting file. There shouldn't be because I have rebased my branch on the latest dev-v2 commit.

kim-vde:
Hey! I am currently working on merging this PR. I had to make some refactorings to make the code 100% correct. You don't need to do anything. The code PR should be merged in the following days/weeks.

PR #9549, opened 2021-10-08
Convert tag name to lowercase in webvtt

akirago:
This commit converts `getTagName()`'s value to lowercase to support `isSupportedTag()` which is case-sensitive function.

https://github.com/google/ExoPlayer/blob/02f7aafe67b4893134a31ea55a3ba0f0535df145/library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java#L513-L525

https://github.com/google/ExoPlayer/blob/02f7aafe67b4893134a31ea55a3ba0f0535df145/library/core/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java#L130-L137

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/webvtt/WebvttCueParser.java
@@ -743,7 +743,7 @@ public final class WebvttCueParser {
   private static String getTagName(String tagExpression) {
     tagExpression = tagExpression.trim();
     Assertions.checkArgument(!tagExpression.isEmpty());
-    return Util.splitAtFirst(tagExpression, "[ \\.]")[0];
+    return Util.splitAtFirst(tagExpression, "[ \\.]")[0].toLowerCase();
   }
 
   private static List<StyleMatch> getApplicableStyles(


akirago:
@googlebot I signed it!

icbaker:
My interpretation of the WebVTT spec is that the span 'names' are not intended to be case-insensitive. For example in the italics case it's written as an exact match for `i`: https://www.w3.org/TR/webvtt1/#webvtt-cue-italics-span

Do you have a link that suggests these ought to be interpreted case-insensitively?

akirago:
>Do you have a link that suggests these ought to be interpreted case-insensitively?

https://html.spec.whatwg.org/multipage/syntax.html#writing
> Tags contain a tag name, giving the element's name. HTML elements all have names that only use ASCII alphanumerics. In the HTML syntax, tag names, even those for foreign elements, may be written with any mix of lower- and uppercase letters that, when converted to all-lowercase, matches the element's tag name; tag names are case-insensitive.

In HTML, tag names are case-insensitive, so I think that tag names should be case-insensitive in WebVTT.

But there is [WebVTT spec](https://www.w3.org/TR/webvtt1), it must be followed.
So I close this PR.

Thanks for your reviews.

PR #9655, opened 2021-11-06
Fix getCurrentWindowIndex documentation

G00fY2:
I think there is a small mistake in the documentation that slipped through


tonihei:
Thanks for the fix. Will merge this change soon.

tonihei:
Actually, the PR needs to be committing into the dev-v2 branch. To simplify the fix, I'll just do the equivalent change internally.

Issue #9819, opened 2021-12-29
Downloads: Support per-download requirements


TheSikh:
I am trying to download dash content from a network storage device to client device using ExoPlayer's DownloadManager, However When I submit the downloadrequest and their is no internet connection. Downloadmanager simply put the download in Stop state and wait for internet connection requirement to be met.

I am downloading from local network so internet is not needed but current DownloadManager add Network as a mandatory requirement for process downloads.

Is there any way or api I can use to tell DownloadManager to skip Internet requirement because I am downloading content from local network.

here is the stack over flow question 
**https://stackoverflow.com/questions/70496661/downloading-content-from-local-network-storage-using-exoplayer**

ojw28:
You can tell `DownloadManager` not to require network simply by passing it a `Requirements` object that doesn't specify any requirements:
```
downloadManager.setRequirements(new Requirements(0));
```
See also [the documentation here](https://exoplayer.dev/downloading-media.html#setting-the-requirements-for-downloads-to-progress).

TheSikh:
Requirements Assume to be either Network_Unmetered or it adds Network to requirement flags by itself. 

https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/library/core/src/main/java/com/google/android/exoplayer2/scheduler/Requirements.java#L67-L75

ojw28:
> Requirements Assume to be either Network_Unmetered or it adds Network to requirement flags by itself.

That's not what that code does. It adds `NETWORK` if `NETWORK_UNMETERED` **is set**, not if it's unset. This is because a requirement for an unmetered network implies that there must also be a requirement for a network. If you pass 0 to the `Requirements` constructor, you should get a `Requirements` object that does not require anything.

TheSikh:
@ojw28  I understand that but Requirement imposes restriction on all downloads not on a single download. Suppose I am migrating X downloads from local and Y downloads from internet. the Y downloads will fail because there is no internet.

ojw28:
In that case, I think we'd need to add support with a generalized version of https://github.com/google/ExoPlayer/pull/9822, as per the comment I left there: https://github.com/google/ExoPlayer/pull/9822#issuecomment-1005587716.

This feels like a pretty niche feature, so marking as a low priority enhancement. It may be easier for you to fork the relevant pieces of ExoPlayer into your own codebase to get the functionality you need, but we will also consider a high quality and complete pull request adding support.

PR #9822, opened 2021-12-29
Skip Network check for local content

TheSikh:
Ability to continue downloading content from local storage while not connected to internet for example migrating existing downloaded content to ExoPlayer by skipping network requirement check.


ojw28:
Please see my response in https://github.com/google/ExoPlayer/issues/9819. If that approach is sufficient for your needs, then I don't think we need this.

If it's not sufficient because you need to indicate the flag on a per-download basis, then the kind of approach proposed in this pull request might be appropriate, however:

* The proposal seems very specific to your exact use case. If we were to support per-download requirements, I think we would want to support that generally somehow, rather than only supporting a specific boolean to bypass the network requirement.
* For both the specific and general approaches, it will be necessary to persist the per-download information in `DefaultDownloadIndex`. Else the extra information added to `DownloadRequest` will not be restored correctly when the downloads are read back from the index.

I'll go ahead and close this pull request for now, because it's either not required or requires substantial revisions. Thanks!

PR #9844, opened 2022-01-07
Fix issue 9782 - need to link program

alteredworlds:


alteredworlds:
This was only necessary against r2.16.1, not dev-v2

Issue #9861, opened 2022-01-13
PlayerView click handling if finger is lifted outside view


ivk1800:
If drag finger started on player view and end on outside view, triggered MotionEvent.ACTION_UP event and triggered toggle player controls. also if begin drag finger on playerview and end on it, player controls same triggered. 

Reproduction steps:
-------------------------

open demo app, modify player_activity.xml set some fixed hegiht for playerview, example 100dp
run on device or emulator. play any example from list.

- ExoPlayer version number: r2.16.1
- Android version: any supported version
- Android device: any supported device

icbaker:
I can reproduce the described behaviour (using `StyledPlayerView` for completeness) - over to Olly to decide whether both of these are bugs and how to handle them:
* drag finger starting on player view and end outside view: player controls are shown
* drag finger starting on player view and elsewhere inside the view: player controls are shown

I note that touching outside the player view and dragging a finger inside it before releasing **does not** trigger the controls.

ojw28:
Fixed in `dev-v2`. The fix will be included in 2.18.

Issue #9605, opened 2021-10-25
StyledPlayerView.setOnClickListener doesn't work when useController=false


eneim:
As the title suggests, if I set `useController` to false (either from XML, or code), the click listener just doesn't work.

In `onTouchEvent`, StyledPlayerView returns false in such a case, so I think it was implemented that way, so I guess the current design is `it should not accept a custom OnClickListener if the controller is disabled`. Is there any reason why? What should I do to have custom click listener without using the controller?

ExoPlayer version: 2.15.1.
Tested device: LG G7/Android 10.

eneim:
I found these 2 related issues that kinda the motivation to disable the touch:

https://github.com/google/ExoPlayer/issues/6340
https://github.com/google/ExoPlayer/issues/1923

IMO there are cases where the client doesn't enable the controller but still allows the user to click the video, for example, to switch between portrait and landscape. I'm curious if there is a way to enable such behavior.

I have an idea of using a custom Controller UI that doesn't show anything ... But want to know if there is another approach. Feel free to close this issue if the current decision is fixed.

ojw28:
This doesn't seem ideal. It looks like there are some other broken cases as well. For example, registering a long click listener doesn't work, regardless of whether there's a controller or not. I'm going to try and remove overriding of `onTouchEvent` to fix this.

ojw28:
Fixed in `dev-v2`. The fix will be included in 2.18.

Issue #8627, opened 2021-02-24
"player control" component behaves differently between TalkBack (Accessibility) on and off


ryxryx:
Steps to reproduce
- Enable "TalkBack" on your Android device
- Run exoplayer demo app
- Go to play any video
- Tap on player area -> TalkBack says "show player controls"
- Activate it (by double tap or however you set up accessibility on your device) -> Player controls are shown
- While player controls are still shown, activate again -> video is paused
- While player controls are still shown, activate again -> video is played

This is different behaviour compared with when Talkback is off:
- While video is play, tap player area -> Player controls are shown
- While player controls are still shown, tap player area -> Player controls are hidden

This is causing confusion to our users who use TalkBack. When player controls are shown and TalkBack focus is still on the player, and then users activate, what should happen is player controls get hidden. The key thing here is that TalkBack focus is still on the player, so the action should not apply onto the play/pause button. User with accessibility might not see why this is happening.

- ExoPlayer version number: using demo app
- Android version: Android 10
- Android device: Samsung Tab A
- Also tested on older version 2.11.3 on Android 7 on Nexus 9

Samrobbo:
This issue slipped through the cracks, sorry about that.

I can confirm this exact issue is happening on my device. 

@ojw28 is this something you can look into? I'm conscious there seems to be a few other Talkback related issues, such as it repeatedly 'saying' the current progress time? Is that expected behaviour?

It could be worth a full dive into the UI talkback situation?

ojw28:
I did take a look at this a while back, and was able to reproduce the issue. I wasn't able to figure out the root cause though. It felt like it was more likely to be a problem with talkback itself than a problem with ExoPlayer's UI components. It seems quite likely there exists a way we could work around whatever the underlying problem is. We'll need to carve out some more time to try and figure out how to do that.

halilozel1903:
Can't it be done by naming the buttons? 

Like this `android:contentDescription="volume button"`

ojw28:
This seems to be related to `StyledPlayerView` overriding `onTouchEvent`. It may be possible to fix it whilst fixing https://github.com/google/ExoPlayer/issues/9605 and https://github.com/google/ExoPlayer/issues/9861.

ojw28:
Fixed in `dev-v2`. The fix will be included in 2.18.

PR #9862, opened 2022-01-13
Handle properly player controls click

ivk1800:
This is a fix for issue https://github.com/google/ExoPlayer/issues/9861


ojw28:
Thanks for the contribution! That said, I think overriding of `onTouchEvent` is more generally problematic (it's also responsible for https://github.com/google/ExoPlayer/issues/9605 and https://github.com/google/ExoPlayer/issues/8627), so I plan to fix the problem in a different way. Please follow along on https://github.com/google/ExoPlayer/issues/9861 for updates. Thanks!

PR #9864, opened 2022-01-16
Add full vorbis comment support

OxygenCobalt:
This PR fully fleshes out the vorbis comment capabilities with the ExoPlayer metadata retrieval system.

# The problem
My music player app needs to extract audio metadata at multiple points, such as with high-quality cover art or ReplayGain tags. However, I began to notice that I could only extract metadata from ID3v2 and FLAC tags, which seemed strange as OGG Vorbis tags are nearly identical to FLAC tags. Looking into this, I discovered that:

- `OggReader` would parse vorbis comments and then never expose them.
- `OpusReader` would just skip vorbis comments entirely.

As a result, I decided to implement parsing for both of these data structures.

# What this does
This PR extends `VorbisComment` and `PictureFrame` to OGG Vorbis and OPUS streams, enabling full metadata functionality for those formats.

- Previously internal FLAC methods responsible for building `VorbisComment` and `PictureFrame` instances have been moved to generic implementations in `VorbisUtil`.
    - When parsing Vorbis Comments, if a tag has the key of `METADATA_BLOCK_PICTURE`, then the tag is actually decoded from Base64 and then parsed as if it is a `PictureFrame`. This is outlined by the specification [here.](https://wiki.xiph.org/index.php/VorbisComment#Cover_art)
    - What used to be the single FLAC method for `PictureFrame` parsing has been split into two blocks. `FlacReader` will first copy the contents of the picture frame into a `ParseableByteArray` on it's own, and then pass it to the utility afterwards. This is required so that vorbis comments can be transformed into picture frames as well.
- In `VorbisReader`, once `VorbisSetup` is returned, the `CommentsHeader` instance is transformed into a `Metadata` instance using the aforementioned utilities. This instance is then provided to `SetupData.format`.
- In `OpusReader`, after the identification header frame is parsed, the subsequent frame is checked to be a Vorbis Comment frame. If that is the case, it is parsed with the aforementioned utilities and immediately transformed into a `Metadata` instance that is then added to to `SetupData.format`.
- To reflect the new nature of the metadata parsing, all contents of the `metadata.flac` module have been moved to a new module called `metadata.vorbis`, with the previous module being deprecated. I can change this back if desired.
- No metadata parsing was added to the OPUS extension.

I updated the testing configuration to reflect these changes and confirmed that the test suite succeeds. I'm not fully sure what knock-on effects this might have, but hopefully this change only affects local files, like I intended to. If it's not, then I'm okay with this not being merged.

As for documentation, I don't fully know how to work with javadoc or the jekyll docs, so there's some directories that still correspond to the old `flac` module name. If you need me to update anything I will.

Feel free to raise any concerns with my PR.


OxygenCobalt:
By the way: I will be signing the CLA, give me a second.

**Edit:** I signed the CLA.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/flac/VorbisComment.java
```diff
@@ -18,12 +18,12 @@
 import static com.google.android.exoplayer2.util.Util.castNonNull;
 
 import android.os.Parcel;
-import android.os.Parcelable;
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.MediaMetadata;
 import com.google.android.exoplayer2.metadata.Metadata;
 
-/** A vorbis comment. */
+/** @deprecated Use {@link com.google.android.exoplayer2.metadata.vorbis.VorbisComment} instead. */
```
icbaker, line 26: We need to think about ideally maintaining backwards compatibility with existing consumers of `Metadata` objects. They will likely have written code something like:

```java
import com.google.android.exoplayer2.metadata.flac.VorbisComment;
...
for (int i = 0; i < metadata.length(); i++) {
  if (metadata.get(i) instanceof VorbisComment) {
    handleVorbisComment((VorbisComment) metadata.get(i));
  }
}
```

This PR keeps this old type (deprecated), but updates `FlacExtractor` to produce the new type. This means that the code above will 'silently' stop working (OK, there will be a deprecation warning, but it's likely to be missed by some developers).

IMO it would be better to either:
1. Completely remove the old type (so devs see a compilation error rather than a silent failure at run-time). We would add a release note entry pointing people at the new location for this type.
2. Or keep both types with this one deprecated, but keep the code above working by making the new type a (mostly empty) subclass of the old one. 
    * All the 'producer' code will be updated to roduce the new sub-class, but any existing 'consumer' code expecting the old type will still work.
    * This requires removing the `final` keyword, but I think that's OK - because we're simultaneously adding the deprecation marking - so anyone subclassing this outside this project is already asking for trouble.
     * When we come to clean-up the deprecation by removing the old type in a few releases time we will just move all the class content into the new type.

(2) is a preferable approach imo - if you can get it working. Can you give that a try?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
```diff
@@ -248,6 +263,89 @@ public static CommentHeader readVorbisCommentHeader(
...
   }
 
+  /**
+   * Builds a metadata instance from Vorbis comments.
+   *
+   * METADATA_BLOCK_PICTURE comments will be transformed into picture frames.
+   * All others will be transformed into vorbis comments.
+   *
+   * @param vorbisComments The raw input of comments, as a key-value pair KEY=VAL.
+   * @return A metadata instance containing the processed tags.
```
icbaker: nit: Can you document when this returns null (same below)? I think it's when no valid vorbis comments were found in `vorbisComments`?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
```diff
@@ -248,6 +263,89 @@ public static CommentHeader readVorbisCommentHeader(
...
+    return metadataEntries.isEmpty() ? null : new Metadata(metadataEntries);
+  }
+
+  /**
+   * Transforms a picture metadata block into a picture frame.
+   *
+   * @param scratch The bytes of the picture.
+   * @return A picture frame from the scratch data.
+   */
+  static PictureFrame buildPictureFrame(ParsableByteArray scratch) {
```
icbaker: nit: can this be `private`?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
```diff
@@ -248,6 +263,89 @@ public static CommentHeader readVorbisCommentHeader(
...
+    }
+
+    ArrayList<Metadata.Entry> metadataEntries = new ArrayList<>();
+    for (int i = 0; i < vorbisComments.size(); i++) {
+      String vorbisComment = vorbisComments.get(i);
+      String[] keyAndValue = Util.splitAtFirst(vorbisComment, SEPARATOR);
+
+      if (keyAndValue.length != 2) {
+        Log.w(TAG, "Failed to parse Vorbis comment: " + vorbisComment);
+      } else {
```
icbaker: nit: consider turning this into a [guard clause](https://medium.com/@scadge/if-statements-design-guard-clauses-might-be-all-you-need-67219a1a981a) to reduce indentation:

```
if (keyAndValue.length != 2) {
  Log.w(TAG, "Failed to parse Vorbis comment: " + vorbisComment);
  continue;
}
if (keyAndValue[0].equals("METADATA_BLOCK_PICTURE")) {
  ...
```

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/FlacStreamMetadata.java
```diff
@@ -275,7 +269,7 @@ public FlacStreamMetadata copyWithVorbisComments(List<String> vorbisComments) {
     @Nullable
     Metadata appendedMetadata =
         getMetadataCopyWithAppendedEntriesFrom(
-            buildMetadata(vorbisComments, Collections.emptyList()));
+            VorbisUtil.buildMetadata(vorbisComments, Collections.emptyList()));
```
icbaker: I think you can use the 1-arg overload here?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
```diff
@@ -248,6 +263,89 @@ public static CommentHeader readVorbisCommentHeader(
...
+  }
+
+  /**
+   * Builds a metadata instance from Vorbis comments.
+   *
+   * METADATA_BLOCK_PICTURE comments will be transformed into picture frames.
+   * All others will be transformed into vorbis comments.
+   *
+   * @param vorbisComments The raw input of comments, as a key-value pair KEY=VAL.
+   * @param pictureFrames Any picture frames that were parsed beforehand.
```
icbaker: This is an overload of `buildMetadata(List<String>)` above, so I'm not sure this behaviour of 'oh and also append these pre-parsed things too' is super-clear (because it's slightly different to the behaviour of the function with the same name).

I appreciate you just copy-pasted this signature from `FlacStreamMetadata` - but now we're promoting this to a public method I think it's worth checking if we're completely happy with its design.

WDYT about removing this overload completely and just letting callers (actually I think it's only `FlacStreamMetadata`) manually append the entries?

Something like:

```java
Metadata metadata = VorbisUtil.buildMetadata(vorbisComments);
metadata = metadata.copyWithAppendedEntries(pictureComments);
```

The actual code would have to be more fiddly, because `metadata` might be `null` :(

Or maybe:
1. This method (with single `List<String>` parameter) should return `List<Entry>` instead of `Metadata`.
2. We could name it `parseVorbisComments` - since that's really the fiddly work it's doing, bulding the `Metadata` object at the end is relatively trivial.
3. It would never return `null`, just an empty-list if no valid comments are found (either because input is empty or all its contents are invalid).

Then the `FlacStreamMetadata` consuming could would be:

```java
Metadata metadata =
   new Metadata(
       Lists.concat(
           parseVorbisComments(vorbisComments),
           pictureComments));
```

EDIT: Just realised `FlacStreamMetadata` is nearly always passing the resulting `Metadata` into `getMetadataCopyWithAppendedEntriesFrom(Metadata)`. It would be pretty reasonable to add a `getMetadataCopyWithAppendedEntries(List<Entry>)` method - which combined with my suggestion above would clean things up a little.

docs/doc/reference/allclasses-index.html
```diff
@@ -5055,7 +5055,7 @@ <h1 title="All&amp;nbsp;Classes" class="title">All&nbsp;Classes</h1>
 </th>
 </tr>
 <tr id="i811" class="rowColor">
-<td class="colFirst"><a href="com/google/android/exoplayer2/metadata/flac/PictureFrame.html" title="class in com.google.android.exoplayer2.metadata.flac">PictureFrame</a></td>
+<td class="colFirst"><a href="com/google/android/exoplayer2/metadata/flac/PictureFrame.html" title="class in com.google.android.exoplayer2.metadata.vorbis">PictureFrame</a></td>
```
icbaker: Can you revert all the javadoc changes?

I appreciate this is strange, but we only update the javadoc each time we do a numbered release, rather than every time we change the code.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
```diff
@@ -69,13 +76,41 @@ protected boolean readHeaders(ParsableByteArray packet, long position, SetupData
...
       checkNotNull(setupData.format); // Has been set when the header was read.
-      boolean headerPacket = packet.readInt() == OPUS_CODE;
-      packet.setPosition(0);
-      return headerPacket;
+
+      // Handle the comments frame if we are working with one.
+      boolean isComment;
+
+      if (packet.bytesLeft() < OPUS_TAG_SIGNATURE.length) {
+        isComment = false;
```
icbaker: nit: might be clearer to just `return false` here? Then you can also turn this into a 'guard clause' too and remove the `else` below.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
```diff
@@ -69,13 +76,41 @@ protected boolean readHeaders(ParsableByteArray packet, long position, SetupData
...
+        VorbisUtil.CommentHeader commentHeader = VorbisUtil.readVorbisCommentHeader(
+            packet, false, false);
+
+        @Nullable Metadata vorbisMetadata = VorbisUtil.buildMetadata(
+            Arrays.asList(commentHeader.comments)
+        );
+
+        setupData.format = setupData.format
+            .buildUpon()
+            .setMetadata(vorbisMetadata)
```
icbaker: When I first saw this code I was suspicious we might be overwriting some existing `Metadata` here.

However it looks like the [spec says](https://datatracker.ietf.org/doc/html/rfc7845#section-5):
> An Ogg Opus logical stream contains exactly two mandatory header packets: an identification header and a comment header.

And the only other source of `setupData.format` is above, which doesn't set any `Metadata`.

So the code as-is looks safe - but I wonder if it's more  robust to just append here anyway, something like:
```java
if (vorbisMetadata != null) {
  setupData.format = setupData.format
      .buildUpon()
      .setMetadata(
          setupData.format.metadata != null 
              ? setupData.format.metadata.copyWithAppendedEntriesFrom(vorbisMetadata)
              : vorbisMetadata)
      .build()
```

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
```diff
@@ -69,13 +76,41 @@ protected boolean readHeaders(ParsableByteArray packet, long position, SetupData
...
+            .build();
+
+        return true;
+      } else {
+        packet.setPosition(0);
```
icbaker: The old code would call `setPosition(0)` both when it found a comment header and when it didn't. This new code only calls it when it doesn't find a comment header.

The documentation of `StreamReader#readHeaders` isn't very clear what's expected here :( Specifically I wonder:
1. What's the expected behaviour of `StreamReader` when it does find a header? By the name I'd assume that it should always consume it - and so the previous code setting `position(0)` when it finds a header is wrong.
2. (potentially an existing bug) Is it safe to assume that `packet` is passed in with `position=0` (so resetting position to zero is correct)? In the general case it seems we should be allowed to pass in a `ParsableByteArray` with non-zero position (expecting the function to start reading from this position), and if the function wants to pretend it didn't read any of it should reset the position to this starting position, not zero.

@kim-vde do you have thoughts on this?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
```diff
@@ -17,29 +17,35 @@
...
 import java.util.List;
 import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;
 
 /** {@link StreamReader} to extract Opus data out of Ogg byte stream. */
 /* package */ final class OpusReader extends StreamReader {
 
-  private static final int OPUS_CODE = 0x4f707573;
-  private static final byte[] OPUS_SIGNATURE = {'O', 'p', 'u', 's', 'H', 'e', 'a', 'd'};
+  private static final byte[] OPUS_ID_SIGNATURE = {'O', 'p', 'u', 's', 'H', 'e', 'a', 'd'};
+  private static final byte[] OPUS_TAG_SIGNATURE = {'O', 'p', 'u', 's', 'T', 'a', 'g', 's'};
```
icbaker: Can we align these field names as closely as possible with RFC7845? 

[It calls this a 'comment header'](https://datatracker.ietf.org/doc/html/rfc7845#section-5.2), so I maybe we can call this `OPUS_COMMENT_HEADER_SIGNATURE`?

And maybe the one above therefore `OPUS_ID_HEADER_SIGNATURE`

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/vorbis/PictureFrame.java
```diff
@@ -0,0 +1,150 @@
...
+import static com.google.android.exoplayer2.util.Util.castNonNull;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+import androidx.annotation.Nullable;
+import com.google.android.exoplayer2.MediaMetadata;
+import com.google.android.exoplayer2.metadata.Metadata;
+import java.util.Arrays;
+
+/** A picture parsed from a FLAC file. */
```
icbaker: I don't think this is correct any more?

It seems the [VorbisComment spec refers to a 'FLAC picture block'](https://wiki.xiph.org/VorbisComment#General_encoding.2Fdecoding), so maybe this javadoc should be updated to refer to that wording (rather than concretely stating it was parsed from  a FLAC file).

icbaker (changes requested):

icbaker:
Thanks for adding this! I took a look through and added some comments - let me know if it seems like i've suggested anything nonsensical :)

OxygenCobalt:
Thanks for the feedback. I'll get to changing those.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/flac/VorbisComment.java
OxygenCobalt: ~~Perhaps we can create a package-private supertype called `VorbisCommentImpl` or similar and have the new and deprecated types implement that. I think that's a better approach that preserves the `final` trait and ensures backwards compat.~~ Dumb idea, ignore this

icbaker: If i'm understanding your proposal correctly, I don't think it does ensure backwards compat for the use-case I described above.

I think you're proposing something like this:

```java
class VorbisCommentImpl {
  public final String key;
  public final String value;

  // equals(), hashcode(), toString() etc.
}

// Existing class
@Deprecated
final class flac.VorbisComment extends VorbisCommentImpl {}

// New class
final class vorbis.VorbisComment extends VorbisCommentImpl {}
```

And the producing could would create instances of `vorbis.VorbisComment`?

The problem is that if the consuming developer is checking something like `entry instanceof flac.VorbisComment`, then in order for that check to continue passing the run-time type of `entry` must be `flac.VorbisComment` or a subclass thereof. And in your proposal `flac.VorbisComment` remains `final`, and therefore `entry` must be **exactly** an instance of `flac.VorbisComment`. I think the only way to keep an `instanceof` check like I've shown passing is to make the new class a subclass of the old one.

OxygenCobalt: Yeah, I realized that pretty quickly, which is why I crossed out the comment.

OxygenCobalt: Pushed commit 0ea8567 to resolve this.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: Looking at the return statement in this method, you're right:

```
return metadataEntries.isEmpty() ? null : new Metadata(metadataEntries);
```

Basically, if there are no valid vorbis comments AND no picture comments were supplied, null will be returned.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: It can't be private as `FlacExtractor` needs to call it to parse it's picture blocks:

```java
ParsableByteArray pic_scratch = new ParsableByteArray(length);
input.readFully(pic_scratch.getData(), 0, length);
pic_scratch.skipBytes(FlacConstants.METADATA_BLOCK_HEADER_SIZE);
PictureFrame pictureFrame = VorbisUtil.buildPictureFrame(pic_scratch);
```

Perhaps we should make it package-private?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: Resolved with 9597ecb

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: Resolved with 9597ecb.

docs/doc/reference/allclasses-index.html
OxygenCobalt: Oh. Seems like android studio decided to rename the javadoc when I changed the module name. I'll try to roll it back to the dev branch.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
OxygenCobalt: I think it would be best to append. I'm considering adding my own extension [not to be merged] to my app's variant of exoplayer that would add the "base gain" header field to the Metadata instance, and I would rather not want a gotcha where the `Metadata` is overwritten like that.

docs/doc/reference/allclasses-index.html
OxygenCobalt: Resolved with 67e1261 and 84c9d29.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
OxygenCobalt: I think this is an accidental side-effect of me trying to shoehorn comment header parsing into this routine.

More or less, `OpusReader` would normally do the following:
- Does this header begin with "O p u s"?
    - If yes, then keep going, we have more headers to consume
    - If no, then stop here.

Since `OpusReader` does this indiscriminately to all frames, it has to rewind with this method or it would corrupt the packet. However, this is no longer needed as we now parse this frame.

According to RFC7845, the comment header is "required", so perhaps we should two one of the following:
- Throw an error if the second packet does not have the comment signature, as we expect it to [I'm in favor of this one]
- Just stop parsing if the packet is invalid.

OxygenCobalt: Actually, after making that change, it seems like changing it so that it does something more "sane" causes the extractor test to break. 

@icbaker It seems like the semantics of `StreamReader` are the following:
- If `readHeaders() == true` -> We just consumed a header, keep going
- If `readHeaders() == false` -> This is not a header, stop now

This would explain why it breaks if we get rid of that ID check. I'll keep the current process of checking the ID, but I've simplified it and eliminated the possible bug that might occur with setting the position to 0. Instead, I'll get the start position from the packet and then rewind the packet if it turns out that it's not a header.

kim-vde: > The old code would call setPosition(0) both when it found a comment header and when it didn't. This new code only calls it when it doesn't find a comment header.

This seems correct (see point 1).

> The documentation of StreamReader#readHeaders isn't very clear what's expected here :(

Yes, I agree the documentation is not very good :(

1. As long as `OpusReader.readHeaders` returns true (i.e. the packet contains headers), it doesn't matter where the position of `packet` is at the end of the method execution, because `oggPacket` will be re-populated with the next packet in `StreamReader.readHeaders`. But if `OpusReader.readHeaders` returns false, then the position of `packet` must be set to 0 because it will be re-read in `StreamReader.readPayload`. The method name `OpusReader.readHeaders` is confusing because in some cases where it finds a header, it doesn't read it entirely and resets the position to 0.

2.  I think it depends on how you see it. Arguably, the parameter name `packet` implies that the `ParsableByteArray` contains the packet and the packet only (and therefore that the packet starts at position 0). I don't think there is any bug here because, from what I see in the code, `packet` should always only contain the packet. SGTM to remove the assumption that the packet starts at 0 but the code as it is now also seems fine to me.

OxygenCobalt: I've tried to resolve this with b919161. To prevent any issues from occuring with `setPosition(0)`, I instead get the initial position beforehand with `getPosition` and rewind to that if the packet is not a comment header.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
OxygenCobalt: Resolved with b919161.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
OxygenCobalt: Resolved with b919161.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
OxygenCobalt: Resolved with b919161.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: Resolved this by making it package-private in 7a88829.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: My idea is this:

Convert `buildMetadata` to a new method called `parseVorbisComments` that takes only a list of strings and then returns a `Metadata` instance. I do not want to return a list of `Metadata.Entry` as that method makes things far messier, even if I graft on a new method onto `Metadata` that appends a list to it. 

Most of the code handles this pretty well, except for the `FlacStreamMetadata` constructor that takes a list of Strings and PictureFrames. For this, I just added a new static method that concatenates those two lists into a single Metadata instance, returning null if there are no metadata. This seems to work properly.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/vorbis/PictureFrame.java
OxygenCobalt: Resolved in fc08f86.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/VorbisUtil.java
OxygenCobalt: Resolved this with fc08f86.

extensions/flac/proguard-rules.txt
```diff
@@ -12,6 +12,6 @@
 -keep class com.google.android.exoplayer2.extractor.FlacStreamMetadata {
     *;
 }
--keep class com.google.android.exoplayer2.metadata.flac.PictureFrame {
+-keep class com.google.android.exoplayer2.metadata.vorbis.PictureFrame {
```
icbaker, line 15: The comment at the top of this section suggests you also need to change lines like this:
https://github.com/google/ExoPlayer/blob/b3981be8b91992976d78c8803349db9680b648a7/extensions/flac/src/main/jni/flac_jni.cc#L126

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/flac/VorbisComment.java
icbaker: I think this inheritance is the wrong way round to solve the backwards compat use-case I described above.

ExoPlayer will now be producing `vorbis.VorbisComment` instances, but a consuming app will still be checking `instanceof flac.VorbisComment`. Because `flac.VorbisComment` is a sub-class of `vorbis.VorbisComment`, this check will return false when passed a `vorbis.VorbisComment` instance.

I think you need to flip it so that `vorbis.VorbisComment` (the new one) extends `flac.VorbisComment` (the old one). And then app code that is checking `instanceof flac.VorbisComment` will continue to work when passed a new `vorbis.VorbisComment` instance.

icbaker (changes requested):

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/flac/VorbisComment.java
OxygenCobalt: Okay, I can just move the files.

OxygenCobalt: Resolved with f6bee30.

extensions/flac/proguard-rules.txt
OxygenCobalt: Resolved with c2d0649. I grepped for any instance of "(flac.PictureFrame)|(flac.VorbisComment)" and could only find documentation and the class definitions themselves.

icbaker:
Thanks for all the changes - I'll work on getting this merged.

icbaker (approved):

OxygenCobalt:
Any updates on when this will be merged?

Issue #9886, opened 2022-01-23
Support VTT in Matroska


moneytoo:
I'm looking into adding support for WebVTT embedded in Matroska (MKV). 

I managed to get basic support in this patch: https://github.com/moneytoo/ExoPlayer/commit/09b07da83fb9cf6a0a6e511e7c0f6282c963e1da and it seems to be working with the sample provided to me in here: https://github.com/moneytoo/Player/issues/249

The patch still lacks support for styles as described in [Matroska specs](https://www.matroska.org/technical/subtitles.html#webvtt). I'm planning on looking into the styles as well, but I'm pretty busy lately so I'm wondering if I should send you PR with this basic support (is it worth an attempt?) or better wait and finish it all?

icbaker:
I think sending a PR with what you've got so far makes sense.

**Heads up:** We have some work planned that will change how subtitle data flows from muxed formats like MKV to the `TextRenderer` - and that will likely affect this issue too.

**Some background:** Currently we put samples in the `SampleQueue` containing data in the 'original' subtitle format, and then parse/decode it inside the `TextRenderer`. We recently added support for parsing the subtitle data **before** we put samples into the `SampleQueue`, which brings various advantages and simplifications. This currently only works for sideloaded subtitle files (those specified in `MediaItem.localConfiguration.subtitleConfigurations`), and is opt-in (controlled by [`DefaultMediaSourceFactory#experimentalUseProgressiveMediaSourceForSubtitles`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/DefaultMediaSourceFactory.html#experimentalUseProgressiveMediaSourceForSubtitles(boolean))). We are planning to standardise all subtitle-processing to follow this model, which (among other changes) will involve changing `MatroskaExtractor` to parse the subtitle data it reads from each sample as it goes along.

If you get the PR in before we start that work in `MatroskaExtractor` then it will help to ensure we don't somehow make this improvement completely impossible in the new architecture :)

icbaker:
I'm going to mark this as low priority to indicate that we (exoplayer-team) have no immediate plans to directly work on implementing this - but we will of course prioritise reviewing and merging any PRs you send us!

marek22k:
I would also like this feature. yt-dlp, for example, downloads all subtitles in vtt format. I have the following options for watching the videos with subtitles:
1. convert subtitles to .srt with ffmpeg.
2. use VLC Media Player.

Both options are not really nice.

PR #9893, opened 2022-01-24
Support MKV embedded WebVTT captions

moneytoo:
This adds the most basic support for VTT to MatroskaExtractor. #9886


icbaker:
Are you able to provide a small test sample as part of this PR that we can add to the [`MatroskaExtractorTest`](https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractorTest.java) and `MkvPlaybackTest` cases?

If you can just add the file to the PR I can take care of adding it to the test cases and generating the golden output dump files.

It might be easiest to take an [existing MKV sample](https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/mkv) and mux some webvtt subtitles into it, but I'm not sure which of the media-prep tools will support this best.

moneytoo:
I've used mkvmerge to generate samples.

icbaker:
This was merged in https://github.com/google/ExoPlayer/commit/e58ffc06827073710586c14ef80403eb0a910ed4 but our tooling didn't correctly link it back to this PR, so I'm closing it manually.

PR #9898, opened 2022-01-25
Add adaptive icons for demo

moneytoo:
It always bothers me when testing something in the ExoPlayer demo app, that its icon is the only one non adaptive among all my installed apps. ;-)

I based it on the svg icon from exoplayer.dev (and did some fixes as it wasn't pixel perfect), applied padding based on current round icon and included wifi arcs for the cast demo.

Before & after:

Screenshot_20220125-162728_One UI Home


marcbaechinger:
Thank you! May I ask you to add the licence header in each of the XML files?

Here is an example in XML:
https://github.com/google/ExoPlayer/blob/release-v2/demos/main/src/main/AndroidManifest.xml

moneytoo:
Will do that (later today) as well as will add the icon to the transformer demo as well.

icbaker:

PR #9907, opened 2022-01-29
api 31

hannesa2:
To make clear to be able to run with api31.
Btw, I guess the most important step to make api31 work was this https://github.com/google/ExoPlayer/blob/dev-v2/extensions/cast/build.gradle#L17 5d7641c6f1f09
`api 'com.google.android.gms:play-services-cast-framework:20.1.0'`

### DIFF
--- a/constants.gradle
+++ b/constants.gradle
@@ -16,10 +16,10 @@ project.ext {
     releaseVersion = '2.16.1'
     releaseVersionCode = 2016001
     minSdkVersion = 16
-    appTargetSdkVersion = 29
+    appTargetSdkVersion = 31
     // Upgrading this requires [Internal ref: b/193254928] to be fixed, or some
     // additional robolectric config.
-    targetSdkVersion = 30
+    targetSdkVersion = 31
     compileSdkVersion = 31
     dexmakerVersion = '2.28.1'
     junitVersion = '4.13.2'
--- a/demos/surface/src/main/AndroidManifest.xml
+++ b/demos/surface/src/main/AndroidManifest.xml
@@ -27,9 +27,8 @@
       android:label="@string/application_name"
       android:exported="true">
 
-    <activity
-        android:name=".MainActivity"
-        android:exported="true">
+    <activity android:name=".MainActivity"
+      android:exported="true">
       <intent-filter>
         <action android:name="android.intent.action.MAIN"/>
         <category android:name="android.intent.category.LAUNCHER"/>


constants.gradle
```diff
@@ -16,10 +16,10 @@ project.ext {
     releaseVersion = '2.16.1'
     releaseVersionCode = 2016001
     minSdkVersion = 16
-    appTargetSdkVersion = 29
+    appTargetSdkVersion = 31
     // Upgrading this requires [Internal ref: b/193254928] to be fixed, or some
     // additional robolectric config.
-    targetSdkVersion = 30
+    targetSdkVersion = 31
```
ojw28, line 22: Looking at the internal referenced issue, it appears this will cause tests to fail. Updating to Robolectric 4.7.0 or later will resolve some of the issues (and is now possible). Looking at the notes in the issue, it looks like two problems will remain after doing that:

1. Taking the Robolectric update causes heap space exhaustion when running the tests. This is because adding an additional SDK version to Robolectric increases baseline memory usage, because it needs to load a whole new set of Android classes during runtime. This is relatively easily fixed with some changes to the `build.gradle` files to request larger-than-default heap space when running the tests.
1. ExoPlayer defaults to using `MediaCodec` in asynchronous mode on API 31 or later, but the Robolectric `MediaCodec` shadow doesn't implement this mode yet. We'll need to add some logic to make sure tests that fail as a result are not run on API 31+. Alternatively, we can configure the player to always use synchronous `MediaCodec` mode specifically for the tests, or (least preferred) wait for Robolectric support.

So there's a bit of work to be done here. Feel free to take a look on your side. We'll keep this pull request open in the meantime.

constants.gradle
```diff
@@ -16,10 +16,10 @@ project.ext {
     releaseVersion = '2.16.1'
     releaseVersionCode = 2016001
     minSdkVersion = 16
-    appTargetSdkVersion = 29
+    appTargetSdkVersion = 31
```
ojw28, line 19: This is currently set to 29 on purpose because at least one of the demo apps uses `requestLegacyExternalStorage` as an easy way to access side-loaded media files for testing purposes. So I think we need to update the demo apps to not need that before we can increase this.

ojw28:
We'll take a look at making the necessary robolectric changes to get the library `targetSdkVersion` bumped. I don't think we need to keep this pull request open as well, since we'll jump bump at the same time.

PR #9908, opened 2022-01-29
Bump gradle

hannesa2:
* Bump Gradle to have Log4j fix https://github.com/gradle/gradle/releases/tag/v7.3.3
* and verify Gradle with Github action


tonihei:
Thanks for sending the PR, but I think it's faster to update the Gradle version ourselves instead of merging the PR. I don't think we would want to include the additional GitHub action.

hannesa2:
Of course you trust yourself most.
But it's also a question of trust of external developers. We have to trust as well and for this the action is valid.

PR #9909, opened 2022-01-29
Build any pull request with Github action

hannesa2:
Currently only assemble works properly.
When this is merged, a greadle task `check` or `test` could be added as well. But it needs some fixes.
Best would be a call of `build` task

### DIFF
--- a/.github/workflows/Android-CI.yml
+++ b/.github/workflows/Android-CI.yml
@@ -1,0 +1,20 @@
+name: PullRequest
+
+on: [ pull_request ]
+
+jobs:
+  buildTest:
+    name: Build SDK for Android
+    runs-on: macos-10.15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.2
+        with:
+          fetch-depth: 0
+      - name: Install JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: 11
+      - name: Build ExoPlayer
+        run: ./gradlew assembleDebug


tonihei:
Thanks for the suggestion to add this, but it's unclear if it provides any benefit. As far as I understand it only helps us to ensure PRs build. However, we never had any pull request that failed to build, so it seems maintaining this workflow only causes extra maintenance effort without any obvious benefit.

hannesa2:
Tests are failing and you even don't know it.

hannesa2:
Please let us know what's the purpose of a test, when it never runs or will be checked

tonihei:
> when it never runs or will be checked

Just to clarify: all our tests are constantly run on every change internally (including all PRs that are imported). That's why maintaining this additional build/test infrastructure via GitHub workflows is likely to cause extra maintenance effort without a huge benefit.

ojw28:
I think there's some (understandable) confusion here because you don't have full visibility into how ExoPlayer is developed. To provide some additional context:

* The majority of ExoPlayer development takes place inside Google, and does not use GitHub / Gradle tooling. Changes made to the internal project are pushed to the externally visible GitHub project. Even when there's an external pull request, we actually merge it via the internal project.
* As @tonihei mentions, our tests run as part of presubmit for every change internally, so they're definitely useful. Changes that appear to be merged via GitHub pull requests are actually merged as internal changes first. The internal codebase is the source of truth.
* There are some subtle differences in the way the tests are run internally vs externally that can occasionally cause problems. In the specific case of the currently broken tests, our internal infrastructure doesn't run tests that are placed in inner classes, where-as these tests do get run using Gradle. This discrepancy allowed some failing tests to be accidentally committed. It's quite rare that we discover such differences. In addition, it's possible for discrepancies to arise due to some code transformations that occur when pushing changes to the external project.
* It's valid feedback that we should ensure the external project builds & the tests pass using Gradle, either during our internal presubmit or prior to pushing changes from our internal project to GitHub. This would prevent such problems from happening. Your pull request doesn't solve this problem though. It's something that needs to be looked at in the context of our internal tooling.

hannesa2:
I'm on latest untouched commit 0c2e77dacae6 in branch `dev-v2` and I see this 
image

Sorry, your CI or process don't found this

Issue #2092, opened 2016-11-22
Support AVI


bene25:
Is it possible to play .avi format on ExoPlayer 2.0.
Due to FrameworkSampleSource are removed, are there any other way to play it?

ojw28:
No, this is not supported in V2. Someone would have to implement an ExoPlayer AVI extractor. It's unclear there's really any significant demand for it, however.

lodup29:
I would also like to see an ExoPlayer avi extractor implemented.

nero12:
Same here, i would love to see .avi extractor, nothing out there ?

dirondir:
I would be happy to see .avi extractor

Aditya94A:
Any updates on this? This should be bumped up.

The fact that AVI format is horribly outdated and nobody should use it is not going to deter our simple minded users! ~7% of the files that my users try to play every day are AVI. With 20k DAU, this is insane. I get at least one 1-star review everyday specifically about this.

danyangwu:
I would also like to see an ExoPlayer avi extractor implemented.

Pecana:
Yes , please add support for this kind of format :-)

mytvpro:
i believe there is in fact significant demand for this

> No, this is not supported in V2. Someone would have to implement an ExoPlayer AVI extractor. It's unclear there's really any significant demand for it, however.

mytvpro:
many older video files still use .avi need to turn to other players to view them. .avi was once very widely used and still very much in circulation i'm not sure why it's not higher priority

PriyankaThakkar612:
Please add support for .avi extractor. We would love to see that.

AshishPsaini:
We are getting 1 star from users for these videos formats. Please provide the support for avi, flv, wmv and mpg formats.

mdtuyen:
@AshishPsaini Don't use exoplayer, there are too many errors in this library, even worse than the videoview

AshishPsaini:
> @AshishPsaini Don't use exoplayer, there are too many errors in this library, even worse than the videoview

Please help me which player I should use in my app.

mdtuyen:
@AdityaAnand1 ijkplayer is better

AshishPsaini:
> @AdityaAnand1 ijkplayer is better

Thanks. Please share your email id or connect with me on ashishpsaini@gmail.com on hangout.

HaiderSaleem:
Still neither avi nor wmv support is added.

[...]

Issue #1614, opened 2016-06-21
MJPEG support


ened:
For a current project I'm evaluating various MJPEG players.
It can be very complicated (ffmpeg, native code, VLC) or simple with Java-only implementations like https://bitbucket.org/neuralassembly/simplemjpegview .

However, there are multiple MJPEG implementations in varying quality around. 
Ideally we would pick the best Java implementation and bring it into ExoPlayer as an extension (like flac, vp9).

@ojw28 is this something the project would support? If so, I will try to start the extension.

ojw28:
You're more than welcome to create your own extension to do this, but we're unlikely to provide official support (i.e. we're unlikely to merge it into this repository or ensure that it continues to build against new versions of ExoPlayer). It's unclear that there's sufficient demand, and every edge case feature we decide to take slows down the speed at which we're able to iterate on the core use cases we should be prioritising. If sufficient demand is demonstrated, we could reconsider.

PR #9915, opened 2022-01-31
Add AVI extraction support

dburckh:
This PR contains support for the AVI container as defined here:

https://docs.microsoft.com/en-us/windows/win32/directshow/avi-riff-file-reference

As a bonus it also supports adds a Renderer with MJPEG support.  The Render is based on BitmapFactory, so it could be easily extended to MPNG or MHEIF, etc.

Video Codecs supported are MJPEG, AVC, MP4V, MP43.  MP42 is theoretically supported.
Audio Codecs supported are PCM, MP3, AAC, AC3 and DTS

The primary goal of this PR is support video from older and lower end video cameras.  Some examples of devices that use AVI are PowerShot cameras, low end drones and dash cameras.  The are predominately MJPEG/PCM.

Further, the MJPEG Renderer could be useful as most webcams support MJPEG.


tonihei:
Thanks for providing this PR! We'd really like to add AVI extractor support to fix #2092 and this is a helpful contribution. 

However, it's also a huge pull request :)  To actually merge this, it would be helpful if you can limit the scope of the PR to just the extractor changes. That is, remove the additional renderer changes and any other updates you made like the ones in the demo app. Once the PR just contains the new extractor, we can look at it in more detail to get it merged. Thanks!

AquilesCanta:
This is super valuable to us!  But +1 to what Toni said, please kindly adjust the PR to only contain AviExtractor and dependencies. Once you've done that, please post below and we'll take another look.

dburckh:
Thanks for your interest!  Here's where I'm at:

- It makes total sense to break the Extractor apart from the Render.  The vast majority of the changes are for the Extractor, so it probably won't be much smaller.  AVI doesn't store timings, so I have to generate them on the fly.  It makes things more complex, especially for AVC.  Sorry.  :(
- I would still like to add the MJPEG Renderer in another PR.  It's needed for my use case and there is another related ExoPlayer issue.  https://github.com/google/ExoPlayer/issues/1614 Fortunately, it's much smaller.  :)  
- I have a bug that came up with Pixels.  That may delay the split for a few days.
- I can revert the demo changes as well.  However, I would like to add AVI.  Unfortunately, I can't find a reliably hosted AVI.  Do you know of any hosting resources I could use?

dburckh:
Good morning!  

I have removed all the Demo and Renderer changes.  I believe everything left is just AviExtractor and it's dependencies.  I've also taken the liberty to add headers and beef up the class comments.

dburckh:
Doing some dog fooding and found out that returning RESULT_SEEK caused some very bad things to happen.  I've fixed it.

library/extractor/build.gradle
```diff
@@ -19,7 +19,11 @@ android {
             testCoverageEnabled = true
         }
     }
-
+    testOptions{
+        unitTests.all {
+            jvmArgs '-noverify'
```
AquilesCanta: Seems quite likely that this change does not belong in this patch. Can you clarify why this is being added here?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
```diff
@@ -0,0 +1,79 @@
...
+package com.google.android.exoplayer2.extractor.avi;
+
+import android.util.SparseArray;
+import com.google.android.exoplayer2.util.MimeTypes;
+import java.nio.ByteBuffer;
+
+/**
+ * Wrapper for the WAVEFORMATEX structure
+ */
+public class AudioFormat {
```
AquilesCanta, line 25: Nit: Final?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
```diff
@@ -0,0 +1,79 @@
...
+  public static final short WAVE_FORMAT_PCM = 1;
+  static final short WAVE_FORMAT_AAC = 0xff;
+  private static final short WAVE_FORMAT_MPEGLAYER3 = 0x55;
+  private static final short WAVE_FORMAT_DVM = 0x2000; //AC3
+  private static final short WAVE_FORMAT_DTS2 = 0x2001; //DTS
+  private static final SparseArray<String> FORMAT_MAP = new SparseArray<>();
+  static {
+    FORMAT_MAP.put(WAVE_FORMAT_PCM, MimeTypes.AUDIO_RAW);
```
AquilesCanta: It seems these constants are only used to populate this map. If so, we can remove the constants and just add comments. Like so:

```
static {
    FORMAT_MAP.put(1 , MimeTypes.AUDIO_RAW); // PCM.
    FORMAT_MAP.put(0x55, MimeTypes.AUDIO_MPEG); // MPEG Layer 3.
    FORMAT_MAP.put(0xff, MimeTypes.AUDIO_AAC); // AAC.
    // ... etc ...
  }
```

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.android.exoplayer2.extractor.SeekMap;
+import com.google.android.exoplayer2.extractor.SeekPoint;
+import java.util.Arrays;
+
+/**
+ * Seek map for AVI.
+ * Consists of Video chunk offsets and indexes for all streams
+ */
+public class AviSeekMap implements SeekMap {
+  final int videoId;
```
AquilesCanta: Nit: Please add the visibility to these. (private?)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
```diff
@@ -0,0 +1,121 @@
...
+      return new SeekPoints(getSeekPoint(index));
+    }
+    final int firstSeekIndex = getFirstSeekIndex(index);
+    if (firstSeekIndex + 1 < keyFrameOffsetsDiv2.length) {
+      return new SeekPoints(getSeekPoint(firstSeekIndex), getSeekPoint(firstSeekIndex+1));
+    } else {
+      return new SeekPoints(getSeekPoint(firstSeekIndex));
+    }
+
+    //Log.d(AviExtractor.TAG, "SeekPoint: us=" + outUs + " pos=" + position);
```
AquilesCanta: Please remove.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
AquilesCanta: Ditto for other classes.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/ChunkHandler.java
```diff
@@ -0,0 +1,195 @@
...
+ * This acts a bridge between AVI and ExoPlayer
+ */
+public class ChunkHandler {
+
+  /**
+   * Constant meaning all frames are considered key frames
+   */
+  public static final int[] ALL_KEY_FRAMES = new int[0];
+
+  public static int TYPE_VIDEO = ('d' << 16) | ('c' << 24);
```
AquilesCanta: Nit: final?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/ListBox.java
```diff
@@ -0,0 +1,97 @@
...
+  //Header List
+  public static final int TYPE_HDRL = 'h' | ('d' << 8) | ('r' << 16) | ('l' << 24);
+  //Stream List
+  public static final int TYPE_STRL = 's' | ('t' << 8) | ('r' << 16) | ('l' << 24);
+
+  private final int listType;
+
+  final List<Box> children;
+
+  ListBox(int size, int listType, List<Box> children) {
```
AquilesCanta: nit: public.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
```diff
@@ -0,0 +1,93 @@
...
+
+  static {
+    //Although other types are technically supported, AVI is almost exclusively MP4V and MJPEG
+    final String mimeType = MimeTypes.VIDEO_MP4V;
+    //final String mimeType = MimeTypes.VIDEO_H263;
+
+    //I've never seen an Android devices that actually supports MP42
+    STREAM_MAP.put('M' | ('P' << 8) | ('4' << 16) | ('2' << 24), MimeTypes.BASE_TYPE_VIDEO+"/mp42");
+    //Samsung seems to support the rare MP43.
+    STREAM_MAP.put('M' | ('P' << 8) | ('4' << 16) | ('3' << 24), MimeTypes.BASE_TYPE_VIDEO+"/mp43");
```
AquilesCanta: Nit: Please convert these to the fourCC values and add a comment explaining their value, like so:
```
STREAM_MAP.put(0x3334504d, MimeTypes.VIDEO_MP43); // Key: little-endian "MP43" fourCC.
```
Which is what we do for MP4's atoms fourCCs.

Also, please note the mime type should be added to MimeTypes.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/AviExtractorRoboTest.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.android.exoplayer2.util.MimeTypes;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(AndroidJUnit4.class)
+public class AviExtractorRoboTest {
```
AquilesCanta: Please remove "Robo".

dburckh:
@AquilesCanta Haven't heard anything from you guys in a while.  How are things on your end?  Let me know if you need anything from me.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
```diff
@@ -0,0 +1,199 @@
...
+import com.google.android.exoplayer2.testutil.FakeTrackOutput;
+import com.google.android.exoplayer2.testutil.TestUtil;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class DataHelper {
+  static final int FPS = 24;
```
AquilesCanta: nit: please add visibility to these.

AquilesCanta: Note: if the visibility is intended to be package, then the code should say /* package */.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
```diff
@@ -0,0 +1,199 @@
...
+
+  public static ChunkHandler getAudioChunkHandler(int sec) {
+    final FakeTrackOutput fakeTrackOutput = new FakeTrackOutput(false);
+    return new ChunkHandler(AUDIO_ID, ChunkHandler.TYPE_AUDIO, fakeTrackOutput,
+        new ChunkClock(sec * 1_000_000L, sec * FPS * AUDIO_PER_VIDEO));
+  }
+
+  public static AviSeekMap getAviSeekMap() {
+    final int[] keyFrameOffsetsDiv2= {4, 1024};
+    final UnboundedIntArray videoArray = new UnboundedIntArray();
```
AquilesCanta: Are these the only uses of UnboundedIntArray? If so, I would suggest using instead a regular array with the necessary manual handling. Some operations might be simplified by calling guava util methods.

AquilesCanta (changes requested):
Thank you for the high quality contribution! Sorry for the delay, failed to hit the send button last week.

library/extractor/build.gradle
dburckh: This allows you to run Robolectric tests with coverage from the Studio UI.  I find it very useful, but it is definitely optional.  FYI: You used to be able to specify -noverify in the UI, but that seems to be broken in the latest version of Studio.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
dburckh: I generally dislike final classes, especially in an open source projects.  I've had to use way too much reflection in the past to work-around bugs in final classes.  All that said, if it's your standard, I will follow it.

library/extractor/build.gradle
AquilesCanta: Thanks for the clarification. Please kindly fresh a new issue or pull request for this specifically. I would like to ask the Android studio team about this. If it's just an Android Studio bug, it might make sense not to add it in case it has other unintended effects.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
dburckh: Yeah, I wondered if you would call me on that.  :)  I left some of the key debug log points commented in the code because I found myself enabling them constantly.  Do you have any guidelines about debugging log statements?  This one isn't bad, but others can be quite verbose.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
dburckh: I would like to challenge that.  I think this syntax greatly improves the readability and understandability of the code.  Further, the compiler converts these to constants, so there is no runtime overhead.

I completely agree on the mime types constants.  My bad.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/AviExtractorRoboTest.java
dburckh: There are two sets of tests for this class one which requires Robolectric (Android) and the other that does not (POJO).  The POJO tests run much faster because they avoid the Robolectric overhead.  All that said, I would need to merge them or change the name to something else.  Let me know which direction I should takes.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: I've fine either way.  Do you have a peferrence?

library/extractor/build.gradle
dburckh: I've used Robolectric for a number of years and the verifier has always been an issue with IntelliJ code coverage.  Jacoco doesn't have an the same issue, but it has it's own set of problems with Robolectric.  We used IntelliJ in Studio and Jacoco for CI, so I have some battle scars!

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: You have a similar class in core called IntArrayQueue, but I didn't want to pull in the dependency.  The seek point array can get quite large, so I was trying to be as efficient as possible.  I can always do a List<Integer> and convert back to int[].  I'll look into the Guava stuff as well.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
dburckh: Started this change, but some of them are used in tests.  Would prefer not to copy/paste.  Thoughts?

library/extractor/build.gradle
dburckh: Done

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
dburckh: Yeah, this was me being lazy for testing purposes.  Fixed.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/ChunkHandler.java
dburckh: Good catch! Done

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/ListBox.java
dburckh: Done

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
dburckh: Fixed the MimeTypes, still would like to discuss the constants.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: I spent some time on this.  I didn't see anything in Guava that was a full drop-in replacement.  The array does grow dynamically and it feels clunky to move the logic from UnboundIntArray into AviSeekMap.  As a compromise, we could replace UnboundIntArray with ArrayList&lt;Integer&gt; and make a call to Ints.toArray(), that would be more overhead on the front end, but keep long term structures efficient.  Thoughts?

dburckh:
@AquilesCanta Made some of the changes.  There were a few bumps and one thing I wanted to discuss.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
AquilesCanta: Commented out code is almost never allowed in. 

Internal debug log lines are very rarely used these days, I think experience shows that they are useful during development stage only, and they become less useful as the library grows, which means they are unlikely to reach a release. I don't think you will easily find a single log.d line in extractor code. This is not true for warnings, but it's easier to determine when a warning is ok to add.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
AquilesCanta: I agree it is more readable. And I agree it's likely that the compiler will fold constants.

However, even more readable than tis was the approach used before https://github.com/google/ExoPlayer/commit/21be28431840f8adf73ea1cbed3e2c7fa7cf86c7, where you'd have `Util.getIntegerCodeForString("34pm")` (or similar). To allow inlining, hex constants where adopted instead. You can rightfully argue that operating over chars is more readable, but someone in the internal review will most likely argue that this goes against what we do everywhere else (see for instance AtomParsers.java, or Atom.java), and usually consistency takes precedence.

It's up to you, I can make this change when I merge this otherwise.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/AviExtractorRoboTest.java
AquilesCanta: I don't think we need non-robolectric tests in this Pull Request. Do we?

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
AquilesCanta: I'm not super clear on why a regular array would not work. Otherwise, you can also look into an IntBuffer.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
AquilesCanta: The rule is to always use the lowest possible visibility.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
AquilesCanta: It's fine to use the literal in the tests.

AquilesCanta:
I think the most important part right now is that we avoid the unbounded int array, which is adding a lot of code without lots of benefit. Otherwise, I think I'll be able to merge some time soon.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/AviExtractorRoboTest.java
dburckh: Maybe I didn't explain that well.  Allow me to clarify.  There is both an AviExtractorRoboTest.java and a AviExtractorTest.java, so I can't rename the former because it would clash with the later.   The difference is one requires Robolectric due testing Android SDK code, while use just standard Java code.  When running tests individually, plain Java tests run essentially instantly from the IDE, where Robolectric require more time.  All that said, I can merge them, just wanted to explain why they are currently separated.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: Let me give you all the detail.  The seek array is a subset of key frames.  For AVI, the key frame flag is stored in the index (idx1).  I don't know the number of key frames until after I have parsed the index.  As such, I need a dynamic array (like Java List).  There are two sets of arrays.  A single array that manages the seek position and a multidimensional array that manages the timestamp of the streams (usually 1 video and 1 audio).  Both int[] and IntBuffer are fixed length.  UnboundIntArray acts similar to List in that it manages the unbounded nature much as List does.

Part of the confusion is that I'm using parallel arrays, which I know is not a great design pattern.  It is very memory efficient.  A more OO implementation might be a List<> with an Object that has the attributes position: long and chunkIndex:int[n] where n is the number of streams in the file.

@AquilesCanta If you want to have a quick Duo call we can do it in the morning.  I'm generally available after 8:30am MST, except Weds

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
dburckh: > I think the most important part right now is that we avoid the unbounded int array, which is adding a lot of code without lots of benefit. Otherwise, I think I'll be able to merge some time soon.

I covered this in detail above, but conceptually, I think it's required.  It would be possible to inline the code into readIdx(), but I think that makes an already complex method more complex.  I prefer to write small reusable objects over monolithic methods.  It makes the testing much easier.  It also may be possible to replace it with a List<> of some variety.

* Not sure how this comment ended up in this conversion.

dburckh: The Util.getIntegerCodeForString() method did not allow the constants to be used in switch{} statements!  I AssUMe it was because they were getting resolved at run time.

Just because you have always done things one way, does not mean it's the best way.  :)  I would challenge you to bring this (or another) approach up to the team.  Writing char values as hex values in reverse order is clunky and highly error prone.

That said a non-optimal standard is better than no standard.  I'll make the update.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: Done

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/VideoFormat.java
dburckh: Done, but I cried a little.  :)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AviSeekMap.java
dburckh: Done + 3 more spots.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/avi/AudioFormat.java
dburckh: Done

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: FYI:  MP4/ISO and AVI are amazing similar, so I spent some time looking at the MP4 stuff to see how they got around this.  Their ChunkIndex is akin to my AviSeekMap (parallel arrays and all).  It appears they know the size of their index so that's how they get around this problem, so it's a different situation.  No help there.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/AviExtractorRoboTest.java
dburckh: Merged.

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/avi/DataHelper.java
dburckh: @AquilesCanta Alright, I think I have something that meets both of our goals.  I moved UnboundIntArray into AviExtractor.  This allowed me to simplify the interface and remove the guard code.  My intention was that it could be reused, but that might have been a bit YAGNI.  I think at this point it's as simple as it can get.  Hopefully we are good to merge now!

AquilesCanta:
I will look into merging this in roughly 3 weeks, as I'm starting my holidays. Thanks again for your contribution.

dburckh:
> I will look into merging this in roughly 3 weeks, as I'm starting my holidays. Thanks again for your contribution.

Thank you for your time and effort.  Enjoy your holiday!

dburckh:
@AquilesCanta How was holiday?  Are we ready to merge this?  

I've been building my app with Exo from source.  Studio does not like that.  :)

AquilesCanta:
I've started looking into this. Unfortunately, the internal infrastructure is having some inconveniences because of the AndroidX media3 work. I'll try to do this quickly, but we'll need to be patient. I'll post an update here once this is done, but please don't count on this being merged for a particular release, because we haven't even started the internal review process.

zhoushsong:
Thank you very much for sharing. I compiled and tested your code and found that two avi format videos could not be played. I hope you can fix this bug
http://www.melodvb.com/testVideo/1280.avi
http://www.melodvb.com/testVideo/67289.avi

dburckh:
> Thank you very much for sharing. I compiled and tested your code and found that two avi format videos could not be played. I hope you can fix this bug http://www.melodvb.com/testVideo/1280.avi http://www.melodvb.com/testVideo/67289.avi

For future reference, please open any issues in my fork.  https://github.com/dburckh/ExoPlayer/issues

The first link contained an MP42 aka MPEG-4 Visual (Microsoft) Video codec.  It's highly unlikely your device has a codec for it.  FYI:  I am aware VLC will play those, but VLC has it's own proprietary CODECs that are not generally available to ExoPlayer.  The second link was a zero length file.  If you will like to discuss this further please open an issue with the link above.

AquilesCanta:
Thank you all for your interest and patience. We are still working on this.

Just a quick heads up, I've already internally imported your PR some weeks ago, so please don't make any modifications to the PR because those won't be picked up. Please keep the fixes around so that we can incorporate them later, once the initial version has been pushed to Github.

AquilesCanta:
I'm still working on this pull request. The reason it's taking so long is that the implementation was not I/O error resilient, so I had to restructure it pretty much completely in order to make the extractor able to resume after an IOException or an interruption, which was not the case before. What would help at this stage is if folks could provide sample media using different container formats for me to test the new implementation. The only sample I have right now uses AVC + mpeg/audio. Thanks for your patience.

dburckh:
H264 is not a typical AVI.  I almost didn't support it.  There are a lot of problems with AVI because of out of sequence (B) frames.  Most streams are XVID/H263.  Let me know where you want the samples.  I have a ton of them.

My use for AVI is read from a local file system, so IOExceptions are extremely rare.  Curious to see what issues you had with interruption and what you had to do to fix them.  If you have any questions about the format feel free to ping me.

AquilesCanta:
What I need is as few samples as possible that cover as much as possible. So, at least one occurrence per compression format among all samples. A sample where the seek offsets start from the start of the file.

> Curious to see what issues you had with interruption

Actually, ExoPlayer uses interruptions to cancel loads, but I'm not 100% sure it can reuse extractors after a cancel so in practice I'm not sure interruptions are relevant. But IO Exceptions are a real problem. For instance, you can run out of network access at any point, which will be surfaced as an IO Exception. This is well covered by our ExtractorTest infrastructure, which simulates IO Exceptions for every byte in the stream.

The concrete consequences of this when writing extractor code is that you cannot have an ExtractorInput.readFully (or skip) followed by another, where the first is not protected by the state of the extractor. So for instance:

```
void readSample() {
  extractorInput.readFully(the header);
  extractorInput.readFully(the body);
}
```

Is problematic. If an exception occurs reading the body, a resumption of the extraction (which means extractor.read() will be called again) will make the extractor try to read the header again, but using the read position from before, that's pointing at the body. So we use states for that reason. If any read() or skip() fails, the extractor should be able to continue normally next time read() is called.

Please kindly share the samples, and thanks again for the contribution.

dburckh:
Sorry, was travelling last week.  Here are some links to bugs.  They have samples.

This one is probably the worst of them H264 with a hacky encoding to simulate variable frame rates.  It also has VBR MP3.
https://github.com/dburckh/ExoPlayer/issues/11

This one is XVID(H263) with MP3 and an incorrect encoding variant (I frames are flagged as uncompressed video).
https://github.com/dburckh/ExoPlayer/issues/5#issuecomment-1019809395

This one is MJPEG with PCM audio, unfortunately we pulled the software MJPEG codec out of this PR, so the only thing I've found that with a hardware mjpeg codec it is a Nvidia Shield TV.
[DV000101.zip](https://github.com/google/ExoPlayer/files/8779592/DV000101.zip)

AquilesCanta:
The PR has been merged internally, so it will be available soon. I've stripped many features to get it merged, in favor of letting the community contribute those features. Please kindly send future PRs to https://github.com/androidx/media, please keep PRs minimally scoped, and please include an AviExtractorTest test (using a minimal sample) that covers the feature that you are sending our way. Our testing infrastructure tests a few essential aspects of the solution for you, like IO Error resilience.

Once the internal change reaches this repository, this PR should be closed automatically. We are grateful for your contribution!

ghost (changes requested):
library/common/src/main/java/com/google/android/exoplayer2/util/FileTypes.java

dburckh:
I built what got merged (release-v2) and it's failing on most of my samples.  It appears most of the clock/AV sync code was removed.  This is what needs to get added back in (in order) to get a functional release:

1. The clock code.  Since AVI does not store PTS for each frame, it needs to be generated via some sort of clock generator.
2. The XXX ChunkHandler, this code feeds the clock and keeps it in sync with the stream.  Usually you will need one of these per codec.  Probably the most import are MP4V and MpegAudio.
3. The Audio Stream positioning in the SeekMap.  The audio stream is muxed in chunks (groups of frames), not frames.  The audio frame number is stored for each video frame that can be seeked to sync audio clock.  I'll admit that my impl. is hard to follow, but it's hard to store all that info efficiently.  Note:  Some streams (MJPEG) are 100% key frames, so just storing the key frames can be problematic.

It would be nice to have back the code that fixes up the aspect ratio based on the underlying codec data.  A lot of the AVI muxers don't pick up on the anamorphic flags in the video stream when muxing.  This results in the video being played back with the wrong aspect ratio.

AquilesCanta:
> This is what needs to get added back in (in order) to get a functional release

Right. We will need to rely on PRs to get that functionality back. Ideally sent to the androidx repo, being self contained (one key functionality per PR), including one test per feature. This will guarantee that future PRs conform to the Extractor's interface requirement around IOException resilience, and that reviews are done quickly.

dburckh:
Unfortunately, I won't have time to work on this until September(?) and that's assuming I don't get a real job.  :P  I'll leave my fork available for reference.  Maybe some other motivated individual can pick up some of the work?

AquilesCanta:
Hopefully. Thanks again for your contributions!

Issue #2, opened 2014-06-23
Support wider range of DASH manifests


ojw28:
Currently ExoPlayer only supports a narrow subset of possible DASH manifests. Support should be added for manifests containing segment lists and segment templates.

Note: As a workaround, application developers can already support arbitrary DASH manifests through implementing their own parser and ChunkSource (it's probably useful to use DashMp4ChunkSource as a starting point).

hafidben:
Do you have an example of an MPD file with segmentList ?

ojw28:
You could try: http://www-itec.uni-klu.ac.at/dash/ddash/mpdGenerator.php?segmentlength=6&type=full

PR #9934, opened 2022-02-04
Changing library loader, so we could use dynamic delivery.

dryganets:
**Background**
Dynamic delivery uses secondary class loader to load the classes.
According to JNI specification the native library should be downloaded by the same classloader that was used to load the classes with the native method.

The way android detects the class loader to use for the
System.loadLibrary call is a bit spooky:
It uses `Reflection.getCallerClass()` method which is basically takes the stack frame #2 from the call stack to find the caller class.

The caller class is always the LibraryLoader in this scenario. As this is the class loaded by the default system class loader the native library can't be used.

**Change**
- LibraryLoader is abstract class now
- There is an abstract method that should be overridden by every user of the class
- all call-sites updated

### DIFF
--- a/extensions/av1/src/main/java/com/google/android/exoplayer2/ext/av1/Gav1Library.java
+++ b/extensions/av1/src/main/java/com/google/android/exoplayer2/ext/av1/Gav1Library.java
@@ -25,7 +25,12 @@ public final class Gav1Library {
     ExoPlayerLibraryInfo.registerModule("goog.exo.gav1");
   }
 
-  private static final LibraryLoader LOADER = new LibraryLoader("gav1JNI");
+  private static final LibraryLoader LOADER = new LibraryLoader("gav1JNI") {
+    @Override
+    protected void loadLibrary(String name) {
+      System.loadLibrary(name);
+    }
+  };
 
   private Gav1Library() {}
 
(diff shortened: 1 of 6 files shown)


ojw28:
> The way android detects the class loader to use for the System.loadLibrary call is a bit spooky:
> It uses Reflection.getCallerClass() method which is basically takes the stack frame
> https://github.com/google/ExoPlayer/issues/2 from the call stack to find the caller class.

Please could you provide a reference / where you found this information? If you've found the code that actually does this, a direct reference to that would be great. Else, I assume you found this information somewhere? I think we should at least have something we can refer to as justification for why this change is necessary. That aside, the change looks reasonable to me. Thanks!

ojw28:
No particular need to change this PR, but for future reference please note that changes need to be made to the `dev-v2` branch, as per our [contributing guidelines](https://github.com/google/ExoPlayer/blob/release-v2/CONTRIBUTING.md).

dryganets:
Hi @ojw28, [this is](https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/lang/System.java;l=1654-1657?q=.loadLibrary0) the reference to android implementation.
```
  @CallerSensitive
    public static void loadLibrary(String libname) {
        Runtime.getRuntime().loadLibrary0(Reflection.getCallerClass(), libname);
    }
```

[getCallerClass implementation](https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/sun/reflect/Reflection.java;l=74-78;bpv=1;bpt=1)

```
    public static Class<?> getCallerClass() {
        // This method (getCallerClass()) constitutes another stack frame,
        // so we need to call getStackClass2() rather than getStackClass1().
        return VMStack.getStackClass2();
    }
```

ojw28:
We merged this manually into `dev-v2` (since this PR was sent to `release-v2`), in https://github.com/google/ExoPlayer/commit/63104064cac8606ec98fed8fa169985af3f76754. Hence we can close this. Thanks for the contribution!

PR #9935, opened 2022-02-04
[Dynamic Delivery] VP9 R8 bug workaround

dryganets:
**background**

This is an attempt to workaround the issues with the R8
We are using multiple class loaders for delivering this code. Due to the [bug in the R8](https://partnerissuetracker.corp.google.com/issues/181185126) the lambda ::releaseBuffers leaks into the main apk.

All our attempts to keep lambdas by proguard rules were futile.

In order to work around this we have to get rid of the extra class.

We tried other options to fix the issue:
- Internal owner class
- external owner class

in both cases we bumped into antother R8 bug - it doesn't pass the VpxDecoder reference to those classes.

As the last resort we decided to implement the interface as part of the VpxDecoder to fix the problem at it's source.
This solution is also generates less classes.

### DIFF
--- a/extensions/vp9/src/main/java/com/google/android/exoplayer2/ext/vp9/VpxDecoder.java
+++ b/extensions/vp9/src/main/java/com/google/android/exoplayer2/ext/vp9/VpxDecoder.java
@@ -34,7 +34,19 @@ import java.nio.ByteBuffer;
 /** Vpx decoder. */
 @VisibleForTesting(otherwise = PACKAGE_PRIVATE)
 public final class VpxDecoder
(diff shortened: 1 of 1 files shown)


ojw28:
The bug you're referring to looks like it was fixed 6 months ago. Are you sure you can't resolve your problem just by upgrading the Android Gradle Plugin to 7.0.0 or later?

dryganets:
Fair enough.
We didn't upgrade yet.
The change is simple enough, but I guess it is matter of taste.
Cons are - we expose something that is not needed outside of the class
Pros - no extra classes generated.

Will close this PR as eventually it wouldn't matter anyway.

Thank you for looking!

PR #9924, opened 2022-02-02
Cea708Decoder: Honor service block size

jruesga:
The current implementation of ExoPlayer Cea708 is processing the whole service block without taking into account the defined service block size, which could cause the execution of unwanted command.

The following set of Cea708 represents a real use case of the above.

``` hex
FC9420FD152FFF0929FE8CFCFE9818FEE332FE731FFE1044FE8B03FE8CFCFE0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC942CFD1520FF4649FE8CFEFE9918FEE332FE731FFE1000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC942FFD9470FF8924FE8B03FE8CFCFE4A92FE0300FE9005FE0091FE2A00FE0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC9420FD97A1FFC829FE8CFEFE9918FEE332FE731FFE1043FE9203FE0100FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC94D6FDCEEFFF082AFE9202FE0C90FE0500FE912AFE0000FE424EFE6F00FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FCC168FD20F4FF4422FE4168FE4220FE7400FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC2C20FDE520FF8422FE2C20FE4265FE2000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
FC64EFFD70EFFFC422FE646FFE4270FE6F00FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
[...]

PR #9967, opened 2022-02-10
Cea708Decoder: Handle multiple service blocks

jruesga:
When a asset contains multiple cc embedded tracks, current Cea708Decoder service block parsing logic is discarding all the data in that frame.

For example, a manifest with 2 embedded CEA-708 close captions tracks:

```xml
<Accessibility schemeIdUri="urn:scte:dash:cc:cea-708:2015" value="1=lang:eng;2=lang:spa"/>
```

when the spanish track is selected (service number 2), when processing the following CEA-708 frame:

```
FC9420FD2062FF0829FE8CFCFE9818FEE332FE731FFE1042FE2062FE0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000
```

the frame can be decomposed as:

```
FF0829 FE8CFC FE9818 FEE332 FE731F FE1042 FE2062 FE0000
  08 (00 001000) 0 (sequence) 8 (frame size)
  298CFC9818E332731F10 422062 0000
  =====
    29 (001 01001) 1 (service number) 9 (service block length)
    8CFC9818E332731F10

    42 (010 00010) 2 (service number) 2 (service block length)
    2062

    0000 Null block service
```

The current processCurrentPacket logic will discard the whole frame is discarded because the first service number found in the frame belongs to service number 1, which is not the one we are currently looking for.

This commit modifies the processCurrentPacket decoding logic, to take into account all service blocks available in the frame, by iteraring over the full frame data and skipping those service blocks we are not interesting in.

Related to https://github.com/google/ExoPlayer/pull/9924

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
@@ -143,7 +143,7 @@ public final class Cea708Decoder extends CeaDecoder {
   private static final int CHARACTER_UPPER_LEFT_BORDER = 0x7F;
 
   private final ParsableByteArray ccData;
-  private final ParsableBitArray serviceBlockPacket;
+  private final ParsableBitArray captionChannelPacketData;
   private int previousSequenceNumber;
   // TODO: Use isWideAspectRatio in decoding.
   @SuppressWarnings({"unused", "FieldCanBeLocal"})
@@ -161,7 +161,7 @@ public final class Cea708Decoder extends CeaDecoder {
 
   public Cea708Decoder(int accessibilityChannel, @Nullable List<byte[]> initializationData) {
     ccData = new ParsableByteArray();
-    serviceBlockPacket = new ParsableBitArray();
+    captionChannelPacketData = new ParsableBitArray();
     previousSequenceNumber = C.INDEX_UNSET;
     selectedServiceNumber = accessibilityChannel == Format.NO_VALUE ? 1 : accessibilityChannel;
     isWideAspectRatio =
@@ -303,7 +303,7 @@ public final class Cea708Decoder extends CeaDecoder {
     boolean cuesNeedUpdate = false;
 
     // Streams with multiple embedded cc tracks (different language tracks) can be delivered
-    // in the same frame packet, so current serviceBlockPacket buffer can contain information
+    // in the same frame packet, so current captionChannelPacketData buffer can contain information
     // for different service numbers.
     // xe: consider the following service block 298CFC9818E332731F104220620000. This service
     // block must be split into 3 different blocks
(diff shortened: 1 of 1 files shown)


library/extractor/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
```diff
@@ -297,71 +297,87 @@ private void processCurrentPacket() {
...
+    // block must be split into 3 different blocks
+    //   298CFC9818E332731F10: Information for service number 1
+    //   422062: Information for service number 2
+    //   0000: Null data
+    //
+    // So here we iterate over the full buffer until we found a null block service or untll
+    // buffer is emptied, and, on each iteration, process only the block related to that
+    // service block. if the specific service number is not the selected one, then just skip
+    // it and continue with next service block.
+    serviceBlockPacket.reset(currentDtvCcPacket.packetData, currentDtvCcPacket.currentIndex);
```
icbaker: do you think we should rename this from `serviceBlockPacket` to e.g. `captionChannelPacketData` (based on the names used in Figure 4 in Section 5 of CEA-708-B).

The current name suggests it contains a single service block, and the whole point of this PR is that assumption is incorrect.

icbaker:
Sorry for the delay in taking a look at this, thanks for sending these fixes over - it's really appreciated.

I wonder if you'd be able to produce a very short (<500kB) test MPEG-TS file containing two service block IDs in a single packet that we could submit as part of this PR (so it's covered by the CLA) and use to drive some automated testing? No need for you to actually write the test - if you can supply the file I can do that bit. 

You can see some existing TS files (and you may be able to modify one of them for your purposes) here:
https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/ts

library/extractor/src/main/java/com/google/android/exoplayer2/text/cea/Cea708Decoder.java
jruesga: Agree. I will publish a new commit with the suggested modification.

jruesga:
> Sorry for the delay in taking a look at this, thanks for sending these fixes over - it's really appreciated.
> 
> I wonder if you'd be able to produce a very short (<500kB) test MPEG-TS file containing two service block IDs in a single packet that we could submit as part of this PR (so it's covered by the CLA) and use to drive some automated testing? No need for you to actually write the test - if you can supply the file I can do that bit.
> 
> You can see some existing TS files (and you may be able to modify one of them for your purposes) here: https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/ts

I'm afraid I cannot provide a MPEG-TS sample file as all them are DRM protected. But, if that helps, I think I can provide sereral cea-708 frames that can be used to test the decoding logic.

icbaker:
Thanks for the replies - I'll merge this PR as-is, if you get hold of the CEA-708 frames then please open a new PR containing just them and I can have a go at feeding them into the decoder in a test.

jruesga:
@icbaker First at all, sorry for the delay to provide the requested information. Attached are some cea708 frames (like 1 minute) for a track with 2 embedded languages (1: English; 2: Spanish)
 
The attached document contains the `presentationTimeUs` and the `hex encoded cea708 frame`, some of them containing both services numbers in the same frame, like the below one:
 
> 1197280000: FC9420FDAE20FF4829FE8CFCFE9918FEE332FE731FFE1042FE2E20FE0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000FA0000

[cea708_frames_data.txt](https://github.com/google/ExoPlayer/files/8371212/cea708_frames_data.txt)

Please, let me know if you require more information.

PR #9990, opened 2022-02-19
Update the gradle wrapper scripts

TacoTheDank:
Currently they are over 6 years old. That is quite ancient. It would be good for them to be kept completely up-to-date. This may improve building times on local machines. This will also fix potential security vulnerabilities the current outdated JAR may have. See [here](https://docs.gradle.org/7.3.3/userguide/gradle_wrapper.html#:~:text=Using%20the%20wrapper%20task%20ensures%20that%20any%20optimizations%20made%20to%20the%20Wrapper%20shell%20script%20or%20batch%20file%20with%20that%20specific%20Gradle%20version%20are%20applied%20to%20the%20project.%20As%20usual%2C%20you%20should%20commit%20the%20changes%20to%20the%20Wrapper%20files%20to%20version%20control.) for why updating these is recommended.

The command I ran was `./gradlew wrapper --gradle-version 7.3.3 --distribution-type all` in Git Bash. You can independently verify the integrity of these files by running this command on your own machine and then using a file comparison tool (like WinMerge for example) to compare the results between this PR and your machine.

The updated JAR's integrity can be verified through this as well: https://gradle.org/release-checksums/


icbaker:
Thanks for the prompt to update these.

I generated the equivalent change internally and committed it as https://github.com/google/ExoPlayer/commit/8416f465ee7fa4b8d1e46201cdf777dc9d1dafc8 (we generally don't accept PRs directly editing the gradle JAR or wrapper script, even though we can verify the integrity as you say).

TacoTheDank:
@icbaker That is completely fine. Thanks for updating these. I appreciate it. 👍

Issue #8636, opened 2021-02-25
Provide an easy way to set MPEG-D DRC codec parameters


ojw28:
Android supports setting various MPEG-D DRC codec parameters to adjust the listening experience. These can be passed in the `MediaFormat` used to configure the codec, and updated dynamically via `MediaCodec.setParameters`. The available parameters are listed [here](https://developer.android.com/reference/android/media/MediaFormat#KEY_AAC_DRC_ALBUM_MODE).

It's currently quite awkward to set these parameters when using ExoPlayer. We should provide a mechanism that makes it relatively straightforward. Any solution should (a) permit dynamic adjustment of multiple parameters via a single `MediaCodec.setParameters` call, (b) be smart enough to retain the parameter values across any cases where the player needs to release the codec and instantiate a new one, and (c) ditto for the case where the parameters are passed when the player has not yet instantiated a codec.

### Proposed solution
One fairly generic solution might be to define a `MSG_SET_CODEC_PARAMETERS` for passing codec parameters to `MediaCodecRenderer` instances. The payload object would have to be capable of holding multiple parameters, and would have to support transferring its parameters into both `MediaFormat` (for initial configuration) and `Bundle` (for dynamic reconfiguration). We could optionally define a `SimpleExoPlayer.setCodecParameters(trackType, parameters)` convenience method to take care of constructing the message and delivering it to the target renderers. That's just one idea. There are probably quite a few ways of doing it!

ronak2121:
Hi @ojw28 are there any updates on this item?

PR #10134, opened 2022-03-30
Add Dialog Enhancement Control support for AC4 audio tracks

g-lass:
The AC4 audio decoder features a dialog enhancement mode
that allows the user to boost the dialog from the played
content.
The boost of the dialog can be tuned to the listener's
need.
This patch maps the control of the dialog boost as a
user controllable playback parameter (like playback speed or
pitch) and proposes a possible implementation in the  UI
in the form of "off, low, mid, high" presets.

- tested on Lenovo P11 tablet and samsung phones

Signed-off-by: glass <glass@dolby.com>

### DIFF
--- a/library/common/src/main/java/com/google/android/exoplayer2/MediaCodecParameters.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/MediaCodecParameters.java
@@ -1,0 +1,44 @@
+package com.google.android.exoplayer2;
+
+import android.os.Build;
+import androidx.annotation.Nullable;
+import java.util.HashMap;
+
+/**
+ * MediaCodec parameters configuring the {@link android.media.MediaCodec} instances.
+ *
+ * once instantiated, the application can add key-value pairs calling {@link MediaCodecParameters#set}
+ * and send a message of type {@code Renderer#MSG_SET_CODEC_PARAMETERS} to the renderers.
+ *
+ */
+public class MediaCodecParameters {
+
+  public static final String KEY_DIALOG_ENHANCEMENT = "dialog-enhancement-gain";
+  public static final String DIALOG_ENHANCEMENT_OFF = "Off";
+  public static final String DIALOG_ENHANCEMENT_LEVEL_LOW = "Low";
+  public static final String DIALOG_ENHANCEMENT_LEVEL_MID = "Mid";
+  public static final String DIALOG_ENHANCEMENT_LEVEL_HIGH = "High";
+
+  private HashMap<String, Object> mediaCodecParameters;
+
+  public MediaCodecParameters() {
+    mediaCodecParameters = new HashMap<String, Object>();
+  }
+
+  public void set(HashMap<String, Object> parameters) {
+    mediaCodecParameters.putAll(parameters);
+  }
+  public void set(String key, Object value) {
+    mediaCodecParameters.put(key, value);
+  }
+  public HashMap<String, Object> get() { return mediaCodecParameters; }
+
+  public Object getOrDefault(String key, Object def) {
+    if (Build.VERSION.SDK_INT >= 24) {
+      return mediaCodecParameters.getOrDefault(key, def);
+    } else {
+      @Nullable Object o = mediaCodecParameters.get(key);
+      return (o == null)  ? def : o;
+    }
+  }
+}
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
@@ -27,6 +27,8 @@ import android.media.AudioFormat;
 import android.media.MediaCodec;
 import android.media.MediaCrypto;
 import android.media.MediaFormat;
+import android.os.Build;
+import android.os.Bundle;
 import android.os.Handler;
 import androidx.annotation.CallSuper;
 import androidx.annotation.Nullable;
@@ -47,6 +49,7 @@ import com.google.android.exoplayer2.decoder.DecoderReuseEvaluation;
 import com.google.android.exoplayer2.decoder.DecoderReuseEvaluation.DecoderDiscardReasons;
 import com.google.android.exoplayer2.mediacodec.MediaCodecAdapter;
 import com.google.android.exoplayer2.mediacodec.MediaCodecInfo;
+import com.google.android.exoplayer2.MediaCodecParameters;
 import com.google.android.exoplayer2.mediacodec.MediaCodecRenderer;
 import com.google.android.exoplayer2.mediacodec.MediaCodecSelector;
 import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
@@ -58,6 +61,7 @@ import com.google.android.exoplayer2.util.MimeTypes;
 import com.google.android.exoplayer2.util.Util;
 import com.google.common.collect.ImmutableList;
 import java.nio.ByteBuffer;
+import java.util.HashMap;
 import java.util.List;
 
 /**
@@ -106,7 +110,6 @@ public class MediaCodecAudioRenderer extends MediaCodecRenderer implements Media
   private boolean audioSinkNeedsReset;
 
   private boolean experimentalKeepAudioTrackOnSeek;
-
   @Nullable private WakeupListener wakeupListener;
 
   /**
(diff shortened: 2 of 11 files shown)


ojw28:
I think this should be implemented in a format agnostic way, more along the lines of the approach suggested in https://github.com/google/ExoPlayer/issues/8636 where there's a generic `MSG_SET_CODEC_PARAMETERS` that can be used to send arbitrary parameters to the codecs within the renderers.

g-lass:
Hi, 
thanks for sharing the view and proposing to implement a codec/renderer specific messaging and a setCodecParameters() as opposed to reusing the exisiting setPlaybackParameters().

How different would the suggested setCodecParametes() be from the existing sendRendererMessage():
``` java
 private void sendVolumeToRenderers() {
    float scaledVolume = volume * audioFocusManager.getVolumeMultiplier();
    sendRendererMessage(TRACK_TYPE_AUDIO, MSG_SET_VOLUME, scaledVolume);
  }
```
I understand that the MSG_SET_CODEC_PARAMETERS  would have to bundle multiple controls , say multiple TLV fields , one of them being the AC4_DIALOG_ENHANCEMENT, others the AAC_DRC parameters and so on and that the message parser would have to dispatch the TLVs to the actual codec.setParameter function as in the ConfigureAC4() from my commit : 

```java
public void ConfigureAC4(int dialogEnhancementGain) {

      if (audioCodecInfo != null && audioCodecInfo.name.contains("ac4")) {
        Bundle bundle = new Bundle(1);
        bundle.putInt("vendor.dolby.dialog-enhancement-gain.value", dialogEnhancementGain);

        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
          super.getCodec().setParameters(bundle);
        }
      }
  }
```

Is my understanding correct ? or did you target a larger refactoring that would replace AAC_DRC_PARAMETERs with codec agnostic DRC parameters and my AC4_DIALOG_ENHANCEMENT gain with a codec agnostic dialog enhancement gain without knowing whether the underlying codec will actually apply it or ignore it ?

ojw28:
> How different would the suggested setCodecParametes() be from the existing sendRendererMessage()

I would have thought that you'd want to send them to all renderers, which can apply or ignore them as needed. In which case the implementation would look something like:

```
@Override
public void setCodecParameters(CodecParameters params) {
  verifyApplicationThread();
  for (Renderer renderer : renderers) {
    createMessageInternal(renderer).setType(MSG_SET_CODEC_PARAMETERS).setPayload(params).send();
  }
}
```

Alternatively, the method could also take a `trackType` argument and use `sendRendererMessage` to send the parameters to all renderers of that track type.

`CodecParameters` should be a fairly generic `(key,value)` object. It shouldn't contain anything AC4 specific, except perhaps some `public static final String KEY_DOLBY_XYZ` style constants. In other words, it should be equally possibly to use `CodecParameters` to deliver some custom video parameter to a video codec.

ojw28:
@g-lass - Just checking, are you planning to work on a generic version of this? Thanks!

g-lass:
Hi, 
I have reworked my commit so that it implements a generic MediaCodecParameter messaging between the application and the exoplayer renderers.
The new MediaCodecParameters class defines the DIALOG_ENHANCEMENT_LEVEL Key and should be extendable with any other codec parameters.
In this commit only the audioRenderer has been modified to manage the message.

g-lass:
Hi, 
Any comment / further guidance with respect to the review of my last commit ?

library/ui/src/main/java/com/google/android/exoplayer2/ui/StyledPlayerControlView.java
```diff
@@ -60,6 +60,7 @@
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.ExoPlayerLibraryInfo;
 import com.google.android.exoplayer2.ForwardingPlayer;
+import com.google.android.exoplayer2.MediaCodecParameters;
```
ojw28, line 63: No need to back these changes out of the commit, but I think we will do so when we get it merged.

`StyledPlayerControlView` isn't supposed to "do everything", because it doesn't really scale very well to all of the different options that everyone might like to see added to it. It's supposed to provide a quick, out-of-the-box starting point for simple use cases. We expect developers to customize or build their own controls for more advanced cases.

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
```diff
@@ -168,6 +171,32 @@ public MediaCodecAudioRenderer(
...
+
+  private void setDialogEnhancement() {
+      if (audioCodecInfo != null && audioCodecInfo.name.contains("ac4")) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+          String dialogEnhancementLevel = (String) mediaCodecParameters.getOrDefault(MediaCodecParameters.KEY_DIALOG_ENHANCEMENT, "off");
+          int    gain = AC4_DIALOG_ENHANCEMENT_LEVEL_MAP.getOrDefault(dialogEnhancementLevel,0);
+          Log.e(TAG, "setDialogEnhancement on AC4 Codec to " + gain + " dB (" + dialogEnhancementLevel + ")");
+          Bundle bundle = new Bundle(1);
+          bundle.putInt("vendor.dolby.dialog-enhancement-gain.value", gain);
+          super.getCodec().setParameters(bundle);
```
ojw28, line 191: Do you know what happens if you call `setParameters("key1", value1)` and then later call `setParameters("key2", value2)`? Is the end result that the codec has both parameters set, or does the second call clear out the value passed by the first?

I think for this to work, you need the end result to be that the codec has both parameters set. Else we'll have to find a slightly different way to do things.

library/common/src/main/java/com/google/android/exoplayer2/MediaCodecParameters.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * MediaCodec parameters configuring the {@link android.media.MediaCodec} instances.
+ *
+ * once instantiated, the application can add key-value pairs calling {@link MediaCodecParameters#set}
+ * and send a message of type {@code Renderer#MSG_SET_CODEC_PARAMETERS} to the renderers.
+ *
+ */
+public class MediaCodecParameters {
+
+  public static final String KEY_DIALOG_ENHANCEMENT = "dialog-enhancement-gain";
```
ojw28, line 16: How confident are we that this generalizes properly to other codecs? As soon as some other codec manufacturer wants 5 levels rather than 3, wouldn't that be a problem? I kinda feel that in absence of standardization at the platform level, we should require apps wanting to use codec specific features to use them without too much abstraction. E.g.,

`codecParameters.setInt(DOLBY_DIALOG_ENHANCEMENT_GAIN, 3);`

This also removes the need for Dolby specific code in `MediaCodecAudioRenderer`. Open to discussion on this point though!

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
```diff
@@ -737,6 +768,9 @@ public void handleMessage(@MessageType int messageType, @Nullable Object message
       case MSG_SET_VOLUME:
         audioSink.setVolume((Float) message);
         break;
+      case MSG_SET_CODEC_PARAMETERS:
```
ojw28, line 771: This probably needs handling in the base class (`MediaCodecRenderer`). Note also that the decoder is recreated in some cases, and also may not exist when this message is received. You'll need to correctly handle these cases. Probably by configuring the decoder correctly somewhere under `initCodec`.

ojw28 (commented):
This looks a lot closer to what I think we're aiming for. Thanks! Please take a look at some additional comments inline.

library/common/src/main/java/com/google/android/exoplayer2/MediaCodecParameters.java
g-lass: Hi, 

IMO, going for a generic "dialog-enhancement-gain" also mean a "common" range.

We have proposed here 3 "levels" (while AC4 actually built-in Dialog Enhancement gain ranges from 0 to 12 dB and the gain we could apply on (e)AC3 content using our audio-effect toolbox ranges from 0 to 9 )  primarily to let the listener clearly hear the difference between 2 steps....  
If we allow a 1:1 mapping between the codec actual range and the application, then we essentially delegate to the application developer the understanding that this dialog-enhancement can go from [0,12] for AC4 , [0,9] for AC3, eAC3, potentially [0,4] for some other parties codec ... 

For a "codec" specific range, my take would be that the app developer would use this setCodecParameter API with the codec specific key directly... then indeed we would have to add the business logic to bundle the unmanaged parameter(s) and pass them directly to the codec. With Android12 onwards, exoplayer can now check which parameter is supported by the codec and filter only these params that can be passed through to the codec directly....

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
g-lass: I have verified that I could set aac-drc-boost-level and aac-drc-cut-level  in 2 consecutive calls to codec.setParameters(bundle)  and that each were applied at some point  (the 2 set could be applied on different output buffers because the codec may be decoding a few buffers in between addressing the 2 setParameter calls. 


...
05-31 17:35:45.018 15168 15228 D CODEC.SETPARAMETERS: set cut-level to 16
05-31 17:35:45.023 15168 15228 D CODEC.SETPARAMETERS:  ... boost-level to 24
**05-31 17:35:45.038 15168 15246 D CCodecBuffers:   int32_t aac-drc-boost-level = 24
05-31 17:35:45.038 15168 15246 D CCodecBuffers:   int32_t aac-drc-cut-level = 16**
...
05-31 17:35:45.092 15168 15228 D CODEC.SETPARAMETERS: set cut-level to 4
05-31 17:35:45.095 15168 15228 D CODEC.SETPARAMETERS:  ... boost-level to 12
**05-31 17:35:45.193 15168 15246 D CCodecBuffers:   int32_t aac-drc-boost-level = 12
05-31 17:35:45.193 15168 15246 D CCodecBuffers:   int32_t aac-drc-cut-level = 4**
...

library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
g-lass: Or call setDialogEnhancement() from the MediaCodecAudioRenderer.onCodecInitialized() callback

Eventually the setDialogEnhancement could be called from a method that applies any "managed" params and that method could be called in updateMediaCodecParameters() as well as in onCodecInitialized()

icbaker:

Issue #8435, opened 2021-01-07
Support SSA/ASS styling


jakobkukla:
Add support for SSA/ASS styling. Currently there is only "minimal" support for ASS subtitles since styling was removed from the initial PR. (See https://github.com/google/ExoPlayer/pull/2281)

It would be great to have full support of the spec. Many subtitles use these styling features to work directly with the video (e.g. to substitute text in a video). SSA/ASS styling is also very common in certain genres of media like Cartoons, Anime or Karaoke.

andrewlewis:
@icbaker Tentatively marking as an enhancement, but please dupe or close as needed. Thanks!

icbaker:
Thanks for the request.

Unfortunately we don't currently have any capacity to improve ExoPlayer's SSA/ASS styling support. However we would welcome high quality PRs in this area.

I'm also happy to discuss various approaches with PR authors if a new feature doesn't fit easily into the existing code structure (either before the PR, or in the review, whichever is easier).

13steinj:
Could we determine a basic checklist of what isn't supported, to determine what features of the spec can be focused on in order of size and/or ease of implementation?

(I'm not familiar with what exactly even *is* supported currently).

icbaker:
> Could we determine a basic checklist of what isn't supported, to determine what features of the spec can be focused on in order of size and/or ease of implementation?

That sounds like a good idea - the best way to do that is probably to compare the [`SsaDecoder`](https://github.com/google/ExoPlayer/blob/dev-v2/library/core/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java) to a copy of the spec. I've found the SSA/ASS spec to be fairly loose, and I'm not sure there's a canonical definition, but I've generally referred to this when working with these files in the past: http://moodub.free.fr/video/ass-specs.doc (linked from [Matroska's page on SSA](https://www.matroska.org/technical/subtitles.html#ssaass-subtitles)).

In case it's helpful, we would probably prioritise things roughly in this order:
* Features that look like they're implemented but don't work properly (basically bugs).
* Features that are similar to existing ones, and so can be implemented with small incremental changes.
* Features that don't fit into the existing code architecture - so implementing them requires significant refactoring.

I don't know if SSA has anything in the final category - but if it looks like it does then please come talk to us before spending ages refactoring `SsaDecoder`, as we might have some ideas about the best way to approach the problem.

cosmo321:
[...]

PR #10150, opened 2022-04-03
Add support for SSA OutlineColour (only background)

egor-n:
https://github.com/google/ExoPlayer/issues/8435

`OutlineColour` should be treated as the background color if `BorderStyle=3`.

~Since currently `BorderStyle` is ignored, we can always treat `OutlineColor` as the background color.~ **EDIT:** with the suggestions from @icbaker, the code no longer ignores `BorderStyle`. Instead, the background color is applied _only_ if `BorderStyle=3`.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -320,7 +320,7 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
             /* end= */ spannableText.length(),
             SpannableString.SPAN_EXCLUSIVE_EXCLUSIVE);
       }
-      if (style.outlineColor != null) {
+      if (style.borderStyle == SsaStyle.SSA_BORDER_STYLE_BOX && style.outlineColor != null) {
         spannableText.setSpan(
             new BackgroundColorSpan(style.outlineColor),
             /* start= */ 0,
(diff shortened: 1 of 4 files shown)


icbaker:
> `OutlineColour` should be treated as the background color if `BorderStyle=3`.

Can you please link to a citation/spec for this? The specs I'm looking at suggest there's a `BackColour` field which I'd expect would be used for the background.

I'm looking at:
* https://fileformats.fandom.com/wiki/SubStation_Alpha
* The `.doc` file linked from https://www.matroska.org/technical/subtitles.html

> Since currently `BorderStyle` is ignored, we can always treat `OutlineColor` as the background color.

I'm not sure I agree with this reasoning. The current implementation ignores both - but if we start considering only one, when the other has such a huge impact on how it's interpreted, I think the result could be very surprising. People would expect to see an outline + dropshadow and instead they'd see a big solid-coloured box.

As an extreme equivalent: What if we handled `ScaleX` but not `ScaleY`? The result would be that people expecting to grow/shrink their subtitles equally in both directions would see them change width but not height - which is probably more surprising than them not changing size at all!

It's probably not that hard to extend this PR to read `BorderStyle` too, and check it's set to 3? You can continue to ignore it (and this field) if it's set to 1 (i.e. you don't need to also implement dropshadow and outline in this PR as well).

egor-n:
@icbaker thank you for your input, it makes perfect sense to check that `BorderStyle` is set to 3 before applying the background color. I will update the PR right away 👍 

> `OutlineColour` should be treated as the background color if `BorderStyle=3`

I can't find a good reference for this claim, but it's what I understood from looking at how libass behaves.
https://fileformats.fandom.com/wiki/SubStation_Alpha#Data_types only mentions that `BackColour` is actually the color of the shadow.

icbaker:
Thanks, looks good - I'll work on getting this merged (sorry for the delay at turning this round).

icbaker (approved):

PR #10169, opened 2022-04-09
Add support for SSA (v4+) MarginL, MarginR, MarginV style

szaboa:
This PR is about adding support for `MarginL`, `MarginR`, `MarginV` from both the `Style` and `Dialogue` lines. I've used VLC as reference.

- Non-zero margins defined in `Dialogue` lines takes priority over the margins in `Style` lines.
- In case we have a `{\pos}` override, then no margins will be applied neither from `Dialogue` or `Style` (same as in VLC).
- In case we have a `{\an}` override, then VLC ignores the `Dialogue` margin but applies the `Style` margin which makes no sense for me, so in our case I've just ignored both `Dialogue` and `Style` margins (same as `{\pos}` case).

_Suggestion_: maybe we could update the `https://storage.googleapis.com/exoplayer-test-media-1/ssa/test-subs-position.ass` subtitle file with these new lines, then no need for a new media in `media.exolist.json`.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -274,20 +274,20 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
             .replace("\\n", "\n")
             .replace("\\h", "\u00A0");
 
-    float marginLeft = format.marginLeftIndex != C.INDEX_UNSET
+    float dialogueMarginLeft = format.marginLeftIndex != C.INDEX_UNSET
         ? SsaStyle.parseMargin(lineValues[format.marginLeftIndex]) : 0f;
-    float marginRight = format.marginRightIndex != C.INDEX_UNSET
+    float dialogueMarginRight = format.marginRightIndex != C.INDEX_UNSET
         ? SsaStyle.parseMargin(lineValues[format.marginRightIndex]) : 0f;
-    float marginVertical = format.marginVerticalIndex != C.INDEX_UNSET
+    float dialogueMarginVertical = format.marginVerticalIndex != C.INDEX_UNSET
         ? SsaStyle.parseMargin(lineValues[format.marginVerticalIndex]) : 0f;
 
     Cue cue = createCue(
         text,
         style,
         styleOverrides,
-        marginLeft,
-        marginRight,
-        marginVertical,
+        dialogueMarginLeft,
+        dialogueMarginRight,
+        dialogueMarginVertical,
         screenWidth,
         screenHeight);
 
@@ -322,9 +322,9 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
       String text,
       @Nullable SsaStyle style,
       SsaStyle.Overrides styleOverrides,
-      float marginLeft,
-      float marginRight,
-      float marginVertical,
+      float dialogueMarginLeft,
+      float dialogueMarginRight,
+      float dialogueMarginVertical,
       float screenWidth,
       float screenHeight) {
     SpannableString spannableText = new SpannableString(text);
@@ -357,20 +357,35 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
         && styleOverrides.position == null
         && cue.getPosition() != Cue.DIMEN_UNSET
         && cue.getLine() != Cue.DIMEN_UNSET) {
+
       // Margin from Dialogue lines takes precedence over margin from Style line.
-      cue.setPosition(cue.getPosition() + (marginLeft != 0f
-          ? marginLeft / screenWidth
-          : style != null ? style.marginLeft / screenWidth : 0f));
-      cue.setPosition(cue.getPosition() - (marginRight != 0f
-          ? marginRight / screenWidth
-          : style != null ? style.marginRight / screenWidth : 0f));
-      // Ignore vertical margin if alignment is middle.
+      float marginLeft = dialogueMarginLeft != 0f
+          ? dialogueMarginLeft / screenWidth
+          : style != null ? style.marginLeft / screenWidth : 0f;
+      float marginRight = dialogueMarginRight != 0f
+          ? dialogueMarginRight / screenWidth
+          : style != null ? style.marginRight / screenWidth : 0f;
+
+      // Apply margin left, margin right.
+      if (SsaStyle.hasLeftAlignment(style)) {
+        cue.setPosition(cue.getPosition() + marginLeft);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else if (SsaStyle.hasRightAlignment(style)) {
+        cue.setPosition(cue.getPosition() - marginRight);
+        cue.setSize(1 - marginRight - marginLeft);
+      } else {
+        // Center alignment or unknown.
+        cue.setPosition(cue.getPosition() + (marginLeft - marginRight) / 2);
+        cue.setSize(1 - marginRight - marginLeft);
+      }
+
+      // Apply margin vertical, ignore it when alignment is middle.
       if (!SsaStyle.hasMiddleAlignment(style)) {
-        float verticalMargin = marginVertical != 0f ? marginVertical / screenHeight
+        float marginVertical = dialogueMarginVertical != 0f ? dialogueMarginVertical / screenHeight
             : style != null ? style.marginVertical / screenHeight : 0f;
-        // Apply margin from top if alignment is top.
-        cue.setLine(cue.getLine() - (SsaStyle.hasTopAlignment(style)
-            ? -verticalMargin : verticalMargin), LINE_TYPE_FRACTION);
+        cue.setLine(
+            cue.getLine() - (SsaStyle.hasTopAlignment(style) ? -marginVertical : marginVertical),
+            LINE_TYPE_FRACTION);
       }
     }
 
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
@@ -273,7 +273,19 @@ public final class SsaDecoder extends SimpleSubtitleDecoder {
             .replace("\\N", "\n")
             .replace("\\n", "\n")
             .replace("\\h", "\u00A0");
-    Cue cue = createCue(text, style, styleOverrides, screenWidth, screenHeight);
+    float marginLeft = SsaStyle.parseMargin(lineValues[format.marginLeftIndex]);
+    float marginRight = SsaStyle.parseMargin(lineValues[format.marginRightIndex]);
+    float marginVertical = SsaStyle.parseMargin(lineValues[format.marginVerticalIndex]);
+
+    Cue cue = createCue(
+        text,
+        style,
+        styleOverrides,
+        marginLeft,
+        marginRight,
+        marginVertical,
+        screenWidth,
+        screenHeight);
 
     int startTimeIndex = addCuePlacerholderByTime(startTimeUs, cueTimesUs, cues);
     int endTimeIndex = addCuePlacerholderByTime(endTimeUs, cueTimesUs, cues);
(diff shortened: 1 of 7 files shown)


library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -376,11 +348,77 @@ private static Cue createCue(
...
       cue.setLine(computeDefaultLineOrPosition(cue.getLineAnchor()), LINE_TYPE_FRACTION);
     }
 
+    // Apply margins if there are no overrides and we have valid positions.
+    if (styleOverrides.alignment == SsaStyle.SSA_ALIGNMENT_UNKNOWN
+        && styleOverrides.position == null
+        && cue.getPosition() != Cue.DIMEN_UNSET
+        && cue.getLine() != Cue.DIMEN_UNSET) {
+      // Margin from Dialogue lines takes precedence over margin from Style line.
+      cue.setPosition(cue.getPosition() + (marginLeft != 0f
```
icbaker: I could be wrong, but it looks like if a user sets the margin via a style then they can't override it to `0` in the dialogue line, because `0` is interpreted to mean `unset` and therefore ignored?

Possibly we need to represent this slightly differently: Store intermediate `unset` values, so we can correctly decide whether to take the value from the style or dialogue line - and if both are unset then only then default to `0`.

EDIT: Actually looking below it seems you're already using `Cue.DIMEN_UNSET`, so maybe this line should just be updated to check for `DIMEN_UNSET` rather than `0`?

icbaker: I think adding the left margin onto the cue position only makes sense if the cue is being 'measured' from the left edge.

i.e. `alignment` is one of `[1, 2, 4, 5, 7, 8]` (left or centre)

If the text is right-aligned, then the cue 'start' position is unaffected by a left margin. The only thing that should be affected is the cue width (`Cue.size`), which is currently unset in this class. I think this PR will need to set `Cue.size` in order to respect `MarginL` when alignment is right or centre, and `MarginR` when alignment is left or centre.

In the same way, `MarginV` needs to respect the vertical part of `Alignment`. The spec I'm reading says:

> Field 16:	MarginV. This defines the vertical Left Margin in pixels.
> For a subtitle, it is the distance from the bottom of the screen.
> For a toptitle, it is the distance from the top of the screen.
> For a midtitle, the value is ignored - the text will be vertically centred

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
```diff
@@ -227,6 +245,15 @@ public static Integer parseColor(String ssaColorExpression) {
     return Color.argb(a, r, g, b);
   }
 
+  public static float parseMargin(String floatValue) {
+    try {
+      return Float.parseFloat(floatValue);
+    } catch (NumberFormatException e) {
+      Log.w(TAG, "Failed to parse margin value: '" + floatValue + "'", e);
+      return 0f;
```
icbaker, line 253: I think `DIMEN_UNSET` would be more consistent here, no?

icbaker (changes requested):

icbaker:
> _Suggestion_: maybe we could update the `https://storage.googleapis.com/exoplayer-test-media-1/ssa/test-subs-position.ass` subtitle file with these new lines, then no need for a new media in `media.exolist.json`.

Yep, sounds good - I'll update the file as part of merging this.

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
szaboa: Overriding a margin to `0` with dialogue lines is not a thing - based on my tests with VLC and also based on the SSA/ASS specification: 

> MarginL   4-figure Left Margin override. The values are in pixels. All zeroes means the default margins defined by the style are used.

szaboa: > If the text is right-aligned, then the cue 'start' position is unaffected by a left margin

Yes, you are right.

> The only thing that should be affected is the cue width (Cue.size)

Can you please elaborate on this? Why should we change the cue width in this case instead of doing nothing? :)

> In the same way, MarginV needs to respect the vertical part of Alignment

Yes, I've missed this.

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
szaboa: Probably it would be more consistent, however in case of margins, the default value is `0` as I understand. Also setting `DIMEN_UNSET` would lead to a bunch of extra checks before applying the margins. In this way we can just add/subtract the margins without any checks, as `0` won't have an effect. But I can go with `DIMEN_UNSET` if you say so. :)

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
icbaker: > > The only thing that should be affected is the cue width (Cue.size)
> 
> Can you please elaborate on this? Why should we change the cue width in this case instead of doing nothing? :)

If you consider a subtitle that's left aligned but really long, then without any margin settings you'd expect it to start on the left hand side of the screen, continue to the right, and then (because it's **really** long) wrap onto a second line that starts back on the LHS of the screen again.

If the left margin is set, then updating the position does the right thing here - it will start from an indented position, continue all the way to RHS and wrap back to the same indent.

If the right margin is set, then things start to go wrong without setting `Cue.size`, because (I assume) the person setting the right margin will expect a long subtitle to wrap **at the margin** and not at the edge of the screen. This can be achieved by setting `Cue.size`, and essentially making the cue box shorter than the available screen width - which will cause the long line to wrap 'early' and respect the margin.

The same argument can be made in reverse for right-aligned subtitles. And center-alignment too really.

szaboa: It should be fixed now.

szaboa:
@icbaker Ready to be reviewed again. :)

Rootax:
> @icbaker Ready to be reviewed again. :)

Thx you, we've missed  .ass/ssa support for years... I hope this can be merged soon.

Rootax:
> @icbaker Ready to be reviewed again. :)

It's a pretty important merge, I hope this can be reviewed soon...

icbaker:
Apologies for the delay on this - I'm looking at it now.

I've patched it internally and I'm going to make some changes before sending it for internal review, so please don't upload any more commits here. One thing I'm going to change is reshuffle the logic in `SsaDecoder` to avoid setting the `cue.position` and `cue.line` values without considering the margins, and then changing them later. Instead I plan to just calculate them once, considering the margin values from the beginning (if set). I think this will be easier to reason about.

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
```diff
@@ -376,11 +348,97 @@ private static Cue createCue(
...
       cue.setLine(styleOverrides.position.y / screenHeight, LINE_TYPE_FRACTION);
     } else {
-      // TODO: Read the MarginL, MarginR and MarginV values from the Style & Dialogue lines.
       cue.setPosition(computeDefaultLineOrPosition(cue.getPositionAnchor()));
       cue.setLine(computeDefaultLineOrPosition(cue.getLineAnchor()), LINE_TYPE_FRACTION);
     }
 
+    // Apply margins if there are no overrides and we have valid positions.
+    if (styleOverrides.alignment == SsaStyle.SSA_ALIGNMENT_UNKNOWN
+        && styleOverrides.position == null
```
icbaker, line 357: Is this logic implying that the margin values are completely ignored if the cue contains `\pos` position overrides? Is that the expected behaviour?

Ditto for alignment overrides - why should they cause the margins to be ignored?

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaStyle.java
icbaker: Yep, I agree we never need to use `DIMEN_UNSET` because `0` implies 'fall back to the next level of defaults' - thanks for clarifying!

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
icbaker: Agreed, thanks - I think we can probably remove all references to `DIMEN_UNSET` margin values. I'll do that when merging internally.

library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java
szaboa: I am not sure if this is the expected behaviour but I took VLC for reference, where it is ignored.

icbaker:
Another question: We currently use a default margin value of 5%:
https://github.com/google/ExoPlayer/blob/c3866449e248bb7e7fce6c78b51f4960e14a0c91/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java#L59

https://github.com/google/ExoPlayer/blob/c3866449e248bb7e7fce6c78b51f4960e14a0c91/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java#L462-L474

This value is used if we don't know `PlayResX` or `PlayResX` **or** if the cue doesn't have a position override:
https://github.com/google/ExoPlayer/blob/c3866449e248bb7e7fce6c78b51f4960e14a0c91/library/extractor/src/main/java/com/google/android/exoplayer2/text/ssa/SsaDecoder.java#L381-L390

I believe your current implementation still uses this default value as a starting point, and adds any margin from the SSA file on top. This doesn't really seem right (shouldn't we just use the file-specified margin directly?). But equally, removing this default completely would result in a file that doesn't specify a margin at all changing from 'nicely' indented cues to cues that are hard against the edge of the screen. This kinda seems 'correct' too - what do you think?

The other option is that we use the default margin when the file doesn't specify a margin (in neither style nor dialogue line).

szaboa:
@icbaker 
> I believe your current implementation still uses this default value as a starting point, and adds any margin from the SSA file on top.

This is right.

If we want to keep strictly to the standard then removing the default margin is the way to go (VLC does the same, if no margin is specified then the subtitle is rendered on the edges). However probably a couple of devs/customers got used to the default margin, so I'd say we can go with this:

> The other option is that we use the default margin when the file doesn't specify a margin (in neither style nor dialogue line).

Note that if you decide to remove the default margin in the end, then the unit tests needs to be corrected. :)

Rootax:
Is this ready to be merged ?

sjiawjbssj:
@icbaker could you please take a look at this

szaboa:
@icbaker what's up with this? should I resolve the conflicts or this is too outdated already? :)

FongMi:
@szaboa How to fix the problem ?
My subtitleView is setApplyEmbeddedFontSizes(false);
So can adjust ass text size.
If size too large, still overlapping.

https://github.com/google/ExoPlayer/assets/3471963/ee3a4b52-7d6b-4177-999c-078ef0a2a811

GlassedSilver:
> @icbaker what's up with this? should I resolve the conflicts or this is too outdated already? :)

I assume it's worth trying to fix the conflicts since at the very least it'd show implementing this much needed feature that a lot of other apps would immediately profit from wouldn't fail to implement due to lack of developer initiative.

Google certainly feels the need to keep up with media playback needs, that's obvious otherwise they wouldn't have just announced to switch to VideoLAN's AV1 software decoder for devices dating back to Android 12 even.

(they are switching away from their own developed decoder since it didn't reach VideoLAN's efficiency)

sjiawjbssj:
@szaboa not sure if you've already read it but @icbaker said they are willing to take another look if this PR is re opened in the androidx media repository and the conflicts are resolved.
See: [https://github.com/androidx/media/issues/1482#issuecomment-2189019408](https://github.com/androidx/media/issues/1482#issuecomment-2189019408)

szaboa:
@sjiawjbssj I've missed this, i'll try to find some time to re-open the PR in androidx/media. But no promises when that's going to happen :)

szaboa:
@sjiawjbssj reopened here https://github.com/androidx/media/pull/2008.

Issue #9752, opened 2021-12-03
Support for custom logger


adroitandroid:
### Use case description
Have logs available on crashlytics or other such places for diagnostic purposes.

### Proposed solution
Have a `com.google.android.exoplayer2.util.Log.setCustomLogger()` that routes logs if a custom logger is provided

tonihei:
The PR #10185 is merged now, so this feature is supported.

PR #10181, opened 2022-04-13
Route the logs to custom logger

shashikantkonded-git:
Use Case:
In the release mode of the app we can capture the logs and send it across(crashlyltics).

Proposed solution:
Set custom logger in release mode so that the logs can be diverted from standard console logging.

Solution to feature: https://github.com/google/ExoPlayer/issues/9752


icbaker:
Please sign the CLA so that we can take a look at this, thanks!

Issue #10159, opened 2022-04-07
DASH DTS Digital Surround Passthrough Fails in recent Exoplayer Versions. 


cedricxperi:
### ExoPlayer Version

2.17.1

### Devices that reproduce the issue

Nvidia Shield TV Pro

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Yes

### Reproduction steps

Playback a DASH DTS Digital Surround audio stream on Nvidia Shield TV Pro (connected to an AVR via HDMI)  running v2.17.1 Exoplayer demo app. 
The DTS bitstream should be passthrough to the AVR for decoding. This works in earlier versions of Exoplayer, E.g. v2.11.8. In v2.17.1 the playback stops with a warning message,  "Media includes audio tracks, but none are playable by this device".

Solution: To fix this issue,  add the following to DashManifestParser.parseAudioChannelConfiguration()

      case "urn:dts:dash:audio_channel_configuration:2012":
      case "tag:dts.com,2014:dash:audio_channel_configuration:2012":
      case "tag:dts.com,2018:uhd:audio_channel_configuration":
        audioChannels = parseDtsChannelConfiguration(xpp);
        break;
  
These are defined in DASH-IF guidelines: https://dashif.org/docs/Audio%20Amendment%20to%20DASH%20IOP%204.3.pdf

### Expected result

The DTS bitstream should be passthrough to the AVR for decoding. 

### Actual result

 Playback stops with a warning message, "Media includes audio tracks, but none are playable by this device"

### Media

Test media and code for parseDtsChannelConfiguration(xpp) will be emailed to dev.exoplayer@gmail.com

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

cedricxperi:
The following case should be added as well for backward compatibility:
case "urn:dts:dash:audio_channel_configuration:2012":
        audioChannels = parseInt(xpp, "value", Format.NO_VALUE); 
Thanks.

ojw28:
Thanks for the detailed report and proposed fix!

cedricxperi:
No worries. Thanks for looking in to this!

ojw28:
Please see email. Thanks!

PR #10214, opened 2022-04-22
Issue #10159 : DASH DTS Digital Surround Passthrough Fails in recent …

cedricxperi:
@ojw28 Here is the pull request. Let me know if this is ok. Thanks.

### DIFF
--- a/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
+++ b/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
@@ -1463,6 +1463,13 @@ public class DashManifestParser extends DefaultHandler
       case "urn:mpeg:mpegB:cicp:ChannelConfiguration":
         audioChannels = parseMpegChannelConfiguration(xpp);
         break;
+      case "urn:dts:dash:audio_channel_configuration:2012":
+      case "tag:dts.com,2014:dash:audio_channel_configuration:2012":
+        audioChannels = parseDtsChannelConfiguration(xpp);
+        break;
+      case "tag:dts.com,2018:uhd:audio_channel_configuration":
+        audioChannels = parseDtsxChannelConfiguration(xpp);
+        break;
       case "tag:dolby.com,2014:dash:audio_channel_configuration:2011":
       case "urn:dolby:dash:audio_channel_configuration:2011":
         audioChannels = parseDolbyChannelConfiguration(xpp);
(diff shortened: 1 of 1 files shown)


library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1880,6 +1887,58 @@ protected static int parseMpegChannelConfiguration(XmlPullParser xpp) {
...
+
+  /**
+   * Parses the number of channels from the value attribute of an AudioElementConfiguration with
+   * schemeIdUri as defined by table B-5 in ETSI TS 103 491 v1.2.1.
+   *
+   * @param xpp The parser from which to read.
+   * @return The parsed number of channels, or {@link Format#NO_VALUE} if the channel count could
+   * not be parsed.
+   */
+  protected static int parseDtsxChannelConfiguration(XmlPullParser xpp) {
```
ojw28, line 1920: Not sure (so please check), but is it equivalent to do:

```
protected static int parseDtsxChannelConfiguration(XmlPullParser xpp) {
  @Nullable String value = xpp.getAttributeValue(null, "value");
  if (value == null) {
    return Format.NO_VALUE;
  }
  int channelCount = Integer.bitCount(Integer.parseInt(value, /* radix= */ 16));
  return channelCount == 0 ? Format.NO_VALUE : channelCount;
}
```

I guess the main change is to use `Integer.bitCount`, on the assumption that it does the same thing?

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
```diff
@@ -1880,6 +1887,58 @@ protected static int parseMpegChannelConfiguration(XmlPullParser xpp) {
...
+   * @param xpp The parser from which to read.
+   * @return The parsed number of channels, or {@link Format#NO_VALUE} if the channel count could
+   * not be parsed.
+   */
+  protected static int parseDtsChannelConfiguration(XmlPullParser xpp) {
+    @Nullable String value = xpp.getAttributeValue(null, "value");
+    if (value == null) {
+      return Format.NO_VALUE;
+    } else {
+      int channelCount = Integer.decode(value);
```
ojw28, line 1903: Is there a particular reason to use `Integer.decode` rather than `Integer.parseInt`, given the spec defines the value as a decimal integer? I think this method can probably be written more simply as:

```
protected static int parseDtsChannelConfiguration(XmlPullParser xpp) {
  int channelCount = parseInt(xpp, "value", Format.NO_VALUE);
  return 0 < channelCount && channelCount < 33 ? channelCount : Format.NO_VALUE;
}
```

ojw28 (commented):
I've put a couple of comments inline for you to take a look at.

If my suggestions look good then there's no need for you to update the pull request. Just reply to confirm that they're equivalent, and I'll get the change merged.

If my suggestions look functionally different, then please let me know what's different about them :). Thanks!

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
cedricxperi: I was just trying to be careful in case some non-standard packager put in the channel config "0x3f" instead of the channel count 6 . But it's ok, let's trust that the standards will be followed. Please replace decode with ParseInt.  Thanks.

library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java
cedricxperi: Yes. That's fine. It is equivalent.  Thanks.

Issue #8220, opened 2020-11-13
Negative timestamps in TS file causes playback to stay in buffering state forever


alexeyr15:
### [REQUIRED] Issue description
Video playback doesn't start for .ts file. It works in other players.

### [REQUIRED] Reproduction steps
1. Open the provided test stream in ExoPlayer Demo app
2. Video playback doesn't start

### [REQUIRED] Link to test content
Provided by email

### [REQUIRED] A full bug report captured from the device
N/A

### [REQUIRED] Version of ExoPlayer being used
2.12.1

### [REQUIRED] Device(s) and version(s) of Android being used
N/A

kim-vde:
The problem comes from the fact that the first audio timestamps are negative.

The content starts with video samples and the adjuster time offset is set to the DTS of the first sample. Audio samples are then read, but their timestamp is less than the video DTS and therefore less than the offset, leading to negative timestamps.

The corresponding code is given [here](https://github.com/google/ExoPlayer/blob/r2.12.1/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/PesReader.java#L243).

alexeyr15:
Thanks for investigation. Is there some way to work around this? After changing position via a seekbar, the video is playing normally.

kim-vde:
@christosts Does ExoPlayer support negative timestamps? And does it support the case where audio samples start with negative timestamps, and video samples start with positive timestamps?

What is happening here is that the MediaCodecVideoRenderer is never getting ready. Execution goes through the following steps:
- The MediaCodecVideoRenderer is enabled with offsetUs = 0 and MediaCodecVideoRenderer#outputStreamOffsetUs is therefore set to 0.
- The extractor extracts audio frames with negative timestamps.
- [This line](https://github.com/google/ExoPlayer/blob/r2.12.1/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java#L565) is executed with buffer.timeUs < 0 and ExoPlayerImplInternal#rendererPositionUs therefore goes from 0 to a negative value.
- As a result, the MediaCodecVideoRenderer is never ready because, in [here](https://github.com/google/ExoPlayer/blob/r2.12.1/library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java#L792), positionUs is negative and outputStreamOffsetUs is 0.

Note that it works sometimes, so there is probably a race condition somewhere.

[...]

Issue #10035, opened 2022-03-08
TS stream don't start playing automatically


skogl:
My issue is that I feel that some mpeg-ts streams don't start playing automatically. The stream source is terrestrial dvb-t2 so I imagine that there are some noise in the stream which makes it unable to start.

I have tested this in the demo player in version 2.16.1 and the issue can be resolved by just fast forward and then the stream starts playing again. I haven't found any similar issues to this so that's why I'm filing this issue.

Here's the log output:

```
2022-03-08 14:15:31.036 3822-3963/? I/MitvPolicyDatabaseManager: updatePackageUsage com.google.android.exoplayer2.demo/com.google.android.exoplayer2.demo.SampleChooserActivity (0,1,55)
2022-03-08 14:15:31.106 3822-3822/? D/MiTv-Service-TvService-Parent: current foreground packagename is com.google.android.exoplayer2.demo
2022-03-08 14:15:31.113 12016-12016/com.google.android.exoplayer2.demo W/ActivityThread: handleWindowVisibility: no activity for token android.os.BinderProxy@f80037
2022-03-08 14:15:31.800 12016-12016/com.google.android.exoplayer2.demo I/DefaultRenderersFactory: Loaded LibvpxVideoRenderer.
2022-03-08 14:15:31.801 12016-12016/com.google.android.exoplayer2.demo I/DefaultRenderersFactory: Loaded Libgav1VideoRenderer.
2022-03-08 14:15:31.815 12016-12016/com.google.android.exoplayer2.demo I/DefaultRenderersFactory: Loaded LibopusAudioRenderer.
2022-03-08 14:15:31.817 12016-12016/com.google.android.exoplayer2.demo I/DefaultRenderersFactory: Loaded LibflacAudioRenderer.
2022-03-08 14:15:31.817 12016-12016/com.google.android.exoplayer2.demo I/DefaultRenderersFactory: Loaded FfmpegAudioRenderer.
2022-03-08 14:15:31.822 12016-12016/com.google.android.exoplayer2.demo I/ExoPlayerImpl: Init 6606001 [ExoPlayerLib/2.16.1] [aquaman, MiTV-AESP0, Xiaomi, 28]
2022-03-08 14:15:31.910 12016-12016/com.google.android.exoplayer2.demo D/EventLogger: playWhenReady [eventTime=0.01, mediaPos=0.00, window=0, true, USER_REQUEST]
[...]

PR #10245, opened 2022-05-06
Fix infinite buffering on start when audio timestamps are behind start pos

afedorychak:
### Background

I recently encountered the same issue as the reporter of #8220 but on a different container type (mp4). Looking at the analysis in this [comment](https://github.com/google/ExoPlayer/issues/8220#issuecomment-729628462) it should stall on any format/container type as long as first audio samples are behind video renderer's output pos (which coincides with start pos for a non-started renderer).

In my case the audio streams had extra audio at the beginning of the stream that extended behind the start pos (however, audio and video were in sync). I played the same stream on roku, shaka player and iOS built-in player and they all handle such case just fine (it appears they just discard the part of the audio stream that doesn't have corresponding video stream). It's unfortunate that ExoPlayer can't do the same.

My understanding of exoplayer rendering/decoding pipeline is pretty sketchy, so please let me know if this approach causes more harm than good. But in my limited testing it worked just fine.

### Reproduction steps

I can't provide the streams I found the issue on, but you can reproduce it by doing this:

1. Pretend that extracted samples extends behind start pos by tweaking SampleQueue offset: 

<details>
  <summary>Example</summary>
  
```patch
Index: library/core/src/main/java/com/google/android/exoplayer2/source/chunk/ContainerMediaChunk.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/library/core/src/main/java/com/google/android/exoplayer2/source/chunk/ContainerMediaChunk.java b/library/core/src/main/java/com/google/android/exoplayer2/source/chunk/ContainerMediaChunk.java
--- a/library/core/src/main/java/com/google/android/exoplayer2/source/chunk/ContainerMediaChunk.java	(revision 7826678f9b1e9d272b3dd366fd5bc9347a797b8b)
+++ b/library/core/src/main/java/com/google/android/exoplayer2/source/chunk/ContainerMediaChunk.java	(date 1651866485271)
@@ -110,7 +110,9 @@
     if (nextLoadPosition == 0) {
       // Configure the output and set it as the target for the extractor wrapper.
       BaseMediaChunkOutput output = getOutput();
-      output.setSampleOffsetUs(sampleOffsetUs);
+      if (trackFormat.sampleMimeType.contains("audio")) {
+        output.setSampleOffsetUs(sampleOffsetUs - 2000000);
+      }
       chunkExtractor.init(
           getTrackOutputProvider(output),
           clippedStartTimeUs == C.TIME_UNSET ? C.TIME_UNSET : (clippedStartTimeUs - sampleOffsetUs),

```
</details>

2. Open demo app and play any clear dash stream.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
@@ -653,7 +653,11 @@ public class MediaCodecAudioRenderer extends MediaCodecRenderer implements Media
       // TODO: Remove this hack once we have a proper fix for [Internal: b/71876314].
       // Allow the position to jump if the first presentable input buffer has a timestamp that
       // differs significantly from what was expected.
-      if (Math.abs(buffer.timeUs - currentPositionUs) > 500000) {
+      if (Math.abs(buffer.timeUs - currentPositionUs) > 500000
+          && (getState() == STATE_STARTED || buffer.timeUs > currentPositionUs)) {
+        // if current pos was allowed to move backwards while this renderer hasn't yet started it
+        // would cause MediaCodecVideoRenderer to never start and lead to infinite buffering (see
+        // issue #8220 for more details)
         currentPositionUs = buffer.timeUs;
       }
       allowFirstBufferPositionDiscontinuity = false;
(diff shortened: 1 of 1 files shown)


tonihei:
Thanks for the PR!

It seems this is attempting to solve #10035 as well (at least the root cause of both problems are likely the same). Of the potential solutions in https://github.com/google/ExoPlayer/issues/10035#issuecomment-1066227078, this PR seems to do both (1) and (2) in some way. We need to check again how this logic is supposed to work to decide which solution is the best one. If we end up doing something similar to your PR, I'll merge it first.

icbaker:

Issue #10246, opened 2022-05-07
Player not parsing Clearkey license acquisition url.


sr1990:
### ExoPlayer Version

2.17.1

### Devices that reproduce the issue

Pixel 5

### Reproducible in the demo app?

Yes

### Reproduction steps

Added
```
      {
        "name": "CLEARKEY (cenc)",
        "uri": "https://media.axprod.net/TestVectors/v7-MultiDRM-SingleKey/Manifest_1080p_ClearKey.mpd",
        "drm_scheme": "clearkey"
      },
```
to media.exolist.json.

Works if I specify the drm license uri
```
      {
        "name": "CLEARKEY (cenc)",
        "uri": "https://media.axprod.net/TestVectors/v7-MultiDRM-SingleKey/Manifest_1080p_ClearKey.mpd",
        "drm_scheme": "clearkey",
        "drm_license_uri": "https://drm-clearkey-testvectors.axtest.net/AcquireLicense"
      },
```

### Expected result

The media plays successfully.

### Actual result

Playback crashes with the following error: 
E/EventLogger: playerFailed [eventTime=4.38, mediaPos=0.00, window=0, period=0, errorCode=ERROR_CODE_DRM_LICENSE_ACQUISITION_FAILED

One workaround is 
in DashManifestParser.java
1. in parseContentProtection(), 
- add the case to parse clearkey
```
case "urn:uuid:e2719d58-a985-b3c9-781a-b030af78d30e":
          uuid = C.CLEARKEY_UUID;
          schemeType = "clearkey";
          break;
```
    
-  Parse the clearkey url in the same method
```
if (XmlPullParserUtil.isStartTag(xpp, "clearkey:Laurl")) {
        if (xpp.next() == XmlPullParser.TEXT) {
          licenseServerUrl = xpp.getText();
        }
 } 
```
 
2. In parseAdaptationSet(), after line 416 while parsing the Content Protection, add

```
     } else if (XmlPullParserUtil.isStartTag(xpp, "ContentProtection")) {
        Pair<String, SchemeData> contentProtection = parseContentProtection(xpp);
        if (contentProtection.first != null) {
          drmSchemeType = contentProtection.first;
        }
        if (contentProtection.second != null) {
          if (drmSchemeType != null && drmSchemeType == "clearkey") {
            // update the license url of cenc
            for (int i = 0;i<drmSchemeDatas.size();i++) {
[...]

PR #10260, opened 2022-05-14
[DASH][ClearKey] Parse ClearKey license URL in MPD.

sr1990:
This PR is related to https://github.com/google/ExoPlayer/issues/10246.


tonihei:
I finally got around to import this PR. Made some changes to avoid removing the "final" keyword from the otherwise immutable class. This should be marked as merged as soon as it's uploaded.

PR #10386, opened 2022-06-28
CacheDataSource chunking fix

dryganets:
## Background
Caching data source logic has been changed recently.

ExoPlayer corrupts cache in case the upstream data source doesn't read the requested bytes.
That allows us to specify the chunk size.

We at Snapchat might decide to issue multiple requests.

New behavior changes the content length property in case upstream decides to split the player request into multiple pieces which causes cache corruption for us.

### DIFF
--- a/library/datasource/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSource.java
+++ b/library/datasource/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSource.java
@@ -630,8 +630,7 @@ public final class CacheDataSource implements DataSource {
           bytesRemaining -= bytesRead;
         }
       } else if (isReadingFromUpstream()
-          && (currentDataSpec.length == C.LENGTH_UNSET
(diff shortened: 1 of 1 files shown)


icbaker:
The current code and comment in `CacheDataSource` looks correct to me - I don't think this PR should be needed.

What `DataSource` implementation are you using with `CacheDataSource` - if it's a custom one, is it possible that it's not strictly adhering to the `DataSource` interface contract?

We have some contract tests for `DataSource` which you could try running on it, they might catch some contract breaches: https://medium.com/google-exoplayer/standardising-datasource-implementations-through-testing-b7a8ab04f8d1

dryganets:
Yep, you are right.

I found that we were returning wrong number of bytes remaining in this scenario (original instead of the chunkSize)
Seems like in case we return exact size it works fine.

Seems like it wasn't noticed as the contract was lax in the CacheDataSource of the previous version of the ExoPlayer.

PR #10423, opened 2022-07-09
Removed inheritance restrictions from ``DefaultHttpDataSource``

### LINKED ISSUE #9626: DefaultHttpDataSource: Support additional customization
This issue is for tracking what I think are the two concrete requests in https://github.com/google/ExoPlayer/issues/9607:

1. Ability to set a `Proxy` when opening the `HttpURLConnection`
2. Ability to set a `SSLSocketFactory` when opening `https://` URLs.

We could either add support directly into `DefaultHttpDataSource`, or make it so that application developers can easily extend `DefaultHttpDataSource` to support these use cases.

litetex:
Removed all not required ``final``s and changed ``private``-modifiers to ``protected``.

(partially) Fixes #9626

Further customization might be required to fully solve #9626  (currently there are very long and complex methods like ``makeConnection`` that are not easily override-able) but I think this PR is a good starting point.

### DIFF
--- a/library/datasource/src/main/java/com/google/android/exoplayer2/upstream/DefaultHttpDataSource.java
+++ b/library/datasource/src/main/java/com/google/android/exoplayer2/upstream/DefaultHttpDataSource.java
@@ -62,17 +62,17 @@ import java.util.zip.GZIPInputStream;
 public class DefaultHttpDataSource extends BaseDataSource implements HttpDataSource {
 
   /** {@link DataSource.Factory} for {@link DefaultHttpDataSource} instances. */
-  public static final class Factory implements HttpDataSource.Factory {
+  public static class Factory implements HttpDataSource.Factory {
 
-    private final RequestProperties defaultRequestProperties;
+    protected final RequestProperties defaultRequestProperties;
 
-    @Nullable private TransferListener transferListener;
-    @Nullable private Predicate<String> contentTypePredicate;
-    @Nullable private String userAgent;
-    private int connectTimeoutMs;
-    private int readTimeoutMs;
-    private boolean allowCrossProtocolRedirects;
-    private boolean keepPostFor302Redirects;
+    @Nullable protected TransferListener transferListener;
+    @Nullable protected Predicate<String> contentTypePredicate;
+    @Nullable protected String userAgent;
+    protected int connectTimeoutMs;
+    protected int readTimeoutMs;
+    protected boolean allowCrossProtocolRedirects;
+    protected boolean keepPostFor302Redirects;
 
     /** Creates an instance. */
     public Factory() {
@@ -82,7 +82,7 @@ public class DefaultHttpDataSource extends BaseDataSource implements HttpDataSou
     }
 
     @Override
-    public final Factory setDefaultRequestProperties(Map<String, String> defaultRequestProperties) {
+    public Factory setDefaultRequestProperties(Map<String, String> defaultRequestProperties) {
       this.defaultRequestProperties.clearAndSet(defaultRequestProperties);
       return this;
     }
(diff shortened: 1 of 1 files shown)


icbaker:
I think this PR exposes way too much of `DefaultHttpDataSource` to subclasses. I don't read #9626 as proposing to expose every possible piece of internal logic of `DefaultHttpDataSource` to subclasses, but instead to carefully expose specific pieces of functionality in order to allow subclasses to implement the specific tasks described in the issue.

Robustly designing inheritable types that can be safely evolved without breaking subclasses is very challenging, and the more of the API surface that is exposed to the subclass the more challenging this is. I'm afraid we're very unlikely to merge this PR in its current form.

If you'd like to solve #9626 I think it would be best to have a high level design discussion in the comments of that issue, before you start writing code. We can use that discussion to work out what the shape of the inheritable API of `DefaultHttpDataSource` might look like.

Item 19 of Effective Java 3rd Edition is relevant here imo:
> Design and document for inheritance or else prohibit it

litetex:
@icbaker 

Thank you for the feedback.

Hear my backstory/use-case first:
Currently you can't override anything in the ``DefaultHttpDataSource`` which is bad because I have an app that requires specific requests depending on the URL (due to some legacy remote services which I can't affect).
Therefore I had to override ``makeConnection`` - but surprise because everything is private and final I couldn't. So for now I just copy-pasted ``DefaultHttpDataSource`` code and extended it which is not exactly optimal because when there are any upstream improvement I don't get them.

As I don't have the deep insight to completely redesign the code of ``DefaultHttpDataSource`` so that everyone will be happy with it (which will likely occur never) I thought the best way is to simply make everything overridable as described in the issue (``or make it so that application developers can easily extend DefaultHttpDataSource``). That way if you need to do any changes to a method you can simply override it (at your own risk).

Also I don't know why some parts of a java-library are sometimes private. Sure there is some point in ensuring that everything works as intended but I had enough cases in the past where something was broken inside a library, the library author(s) were unable to fix it and you had to copy-paste their code, do crazy reflection stuff or fork the complete project to do a tiny change that would have been easily possible when someone would have allowed inheritance.

> Item 19 of Effective Java 3rd Edition...

I didn't read this book and don't exactly know what's written there but according to some quotes [I found on the fly](https://cs108.epfl.ch/archive/19/c/i/EffectiveJava_Item19.pdf) they mostly talk over classes that are expected to be overridden in the first place like ``AbstractColIection``.

Also 

> This is not just a theoretical problem. It is not uncommon to receive subclassing-related bug reports after modifying the internals of a non final concrete class that was not designed and documented for inheritance.

can't happen if you write "extend on own risk above" them.

Wrapping the class/``makeConnection``-method is also not an option because I have to - among other things - change the request method (``setRequestMethod``) on the fly and that's not possible from the outside.

icbaker:
> Wrapping the class/`makeConnection`-method is also not an option because I have to - among other things - change the request method (`setRequestMethod`) on the fly and that's not possible from the outside.

Can you use a [`ResolvingDataSource`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/upstream/ResolvingDataSource.html) to transform the `DataSpec` before it gets to the underlying HTTP `DataSource` implementation? That would allow you to modify the [URI](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/upstream/DataSpec.Builder.html#setUri(android.net.Uri)) and [HTTP method](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/upstream/DataSpec.Builder.html#setHttpMethod(@com.google.android.exoplayer2.upstream.DataSpec.HttpMethodint)) (as well as other properties like request headers).

icbaker:
Closing because we're not going to accept this in its current form.

Issue #7279, opened 2020-04-22
Provide ForwardingMediaSource to simplify custom implementations


stari4ek:
### [REQUIRED] Use case description
Current implementation of `HlsMediaSource` starts with default position:
> We attempt to set the default start position to be at least twice the target duration behind the live edge.

So, if HLS has lots of segments or they are long enough - default start position will be pretty far from it's 0. This is desired behavior when playing "live" media.
But, for example, "catchup" of IPTV services is implemented in the way, when url identifies specific time (wall clock), and when player starts playback - it should start from 0 instead of "live edge" to match target wall clock.

### Alternatives considered
I have tried:
* `MediaItem.ClippingProperties` - does not work (since it clips only)
* seeking before preparations (#2197) - does not work

### Proposed solution
I have tried:
* custom media source + custom timeline, which forces `Window.defaultPositionUs` (based on https://github.com/google/ExoPlayer/issues/2403#issuecomment-330955125) - works

I think that some functionality can be provided by library itself:
* fully functional `CustomStartPositionSource`
* or at least easy-to-extend delegating media source. Otherwise `CompositeMediaSource` should be used (the same as it's used by `ClippingMediaSource`), which looks a bit of overkill.

stari4ek:
Proof of concept implementation looks like:
```java

public class CustomStartPositionSource extends CompositeMediaSource<Void> {

    private final MediaSource mWrappedSource;
    private final long mStartPos;

    private Timeline mTimeline;

    public CustomStartPositionSource(MediaSource wrappedSource, long startPositionMs) {
        this.mWrappedSource = wrappedSource;
        this.mStartPos = C.msToUs(startPositionMs);
    }

    // MediaSource

    @Nullable
    @Override
    public Object getTag() {
        return mWrappedSource.getTag();
    }

    // CompositeMediaSource

    @Override
    protected void prepareSourceInternal(@Nullable TransferListener mediaTransferListener) {
        super.prepareSourceInternal(mediaTransferListener);
        prepareChildSource(/* id= */ null, mWrappedSource);
    }

    @NonNull
    @Override
    public MediaPeriod createPeriod(
            @NonNull MediaPeriodId id,
            @NonNull Allocator allocator,
            long startPositionUs) {

        return mWrappedSource.createPeriod(id, allocator, startPositionUs);
    }

    @Override
    public void releasePeriod(@NonNull MediaPeriod mediaPeriod) {
[...]

PR #10446, opened 2022-07-16
Add WrappingMediaSource

stoyicker:
This is a CompositeMediaSource<Void> with a slightly simpler API. I thought it was a bit more readable than the current usage of CompositeMediaSource<Void> to signify a single-source one.


library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
```
tonihei: This should be 2022 :)

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+package com.google.android.exoplayer2.source;
+
+import androidx.annotation.Nullable;
+import com.google.android.exoplayer2.Timeline;
+
+/**
+ * Single-child {@link CompositeMediaSource}.
+ */
+public abstract class SimpleMediaSource extends CompositeMediaSource<Void> {
```
tonihei: I think it might be caller to call this class `WrappingMediaSource` because "simple" doesn't really tell anyone what it's for.

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Called when the source info has been refreshed.
+   *
+   * @param mediaSource The child source whose source info has been refreshed.
+   * @param timeline The timeline of the source.
+   */
+  protected abstract void onChildSourceInfoRefreshed(
+      MediaSource mediaSource, Timeline timeline
```
tonihei: The implementation of this class knows the wrapped inner `mediaSource`, so no need to pass it back here I think.

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Releases the child source.
+   */
+  protected final void releaseChildSource() {
+    releaseChildSource(CHILD_SOURCE_ID);
+  }
+
+  /**
+   * @deprecated - Use {@link #getWindowIndexForChildWindowIndex(int)}.
```
tonihei: This method isn't really deprecated and you can just remove the comment entirely. Same for the other cases below.

tonihei:
Thanks for the contribution! 

For reference: I believe this is solving #7279. 

Is there any chance you can rebase this change on the [main branch](https://github.com/androidx/media/tree/main) of the Media3 repo? This would make the merging process a lot easier for us.

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
stoyicker: Not sure I agree 100% - it of course knew when it called `prepareChildSource`, but for it to know when this is called it'd have to have kept a reference to it somehow. And that is possible, but so is that a regular `CompositeMediaSource` inheritor keeps a mapping between ids and mediasources itself, so that it could retrieve the child just by using the id parameter on `CompositeMediaSource.onChildSourceInfoRefreshed`, and therefore applying this same argument then the `MediaSource` parameter wouldn't be needed there either. Without wanting to create a discussion on whether that argument should be removed in `CompositeMediaSource.onChildSourceInfoRefreshed`, I think for the sake of consistency with `CompositeMediaSource` the argument is better kept here too. Am I making sense?

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
stoyicker: Why not deprecate it? Would receivers of this class really want a method that takes an `id`?

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
stoyicker: fixed

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
stoyicker: fixed

stoyicker:
I've rebased on top of 9271572e950230a24da69a6ec5838dd838ea672d from media3, feel free to change the PR base for a simpler diff if you have a more fitting one

tonihei:
> I've rebased on top of https://github.com/google/ExoPlayer/commit/9271572e950230a24da69a6ec5838dd838ea672d from media3

Sorry, I think my question was a bit unclear. I actually meant to file a new PR against the main branch of the Media3 project . The files look exactly the same, but have different package names.

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
tonihei: It does make sense generally, but I think all implementations of this `WrappingMediaSource` will need to have a local reference to the wrapped child media source in any case. For example, the implementation couldn't create `MediaPeriod` from `createPeriod` if it doesn't have a reference to the child source. 

> I think for the sake of consistency with CompositeMediaSource

We are already changing the method signature by removing the useless argument, so there is no consistency to start with?

library/core/src/main/java/com/google/android/exoplayer2/source/SimpleMediaSource.java
tonihei: The difference is that our libraries never use @deprecated for cases where the method itself is redirected. There also shouldn't be any caller besides CompositeMediaSource itself, so it seems easier to the deprecated tag.

stoyicker:
https://github.com/androidx/media/pull/123

Issue #10445, opened 2022-07-15
Burst of invalid frame renders on seek with Android 10 AmLogic


stevemayhew:
### ExoPlayer Version

2.18.0

### Devices that reproduce the issue

Google ChormeCast With Google TV running Android 10, many other AmLogic based devices.

### Devices that do not reproduce the issue

AmLogic based devices running Android 9.

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Play the content (`adb shell am start -n com.google.android.exoplayer2.demo/.PlayerActivity -a com.google.android.exoplayer.demo.action.VIEW -d ...`)
2. Seek to a position mid-segment: (`adb shell am start -n com.google.android.exoplayer2.demo/.PlayerActivity -a com.google.android.exoplayer.action.SEEK_TO --ei position 279000`

You can add the SEEK_TO intent with this simple code in PlayerActivity

```java
  @Override
  public void onNewIntent(Intent intent) {
    super.onNewIntent(intent);
    String action = intent.getAction();
    if ("com.google.android.exoplayer.action.SEEK_TO".equals(action)) {
        long seekTo =  intent.getIntExtra("position", C.POSITION_UNSET);
        player.seekTo(seekTo);
    } else {
      releasePlayer();
      releaseClientSideAdsLoader();
      clearStartPosition();
      setIntent(intent);
    }
  }
```

Note, for hardware I am using:

1. Google Chromecast with Google TV (GZRNL)
2. An LG webPS TV IF7600 (any 60hz refresh 4k monitor seems to reproduce this)

### Expected result

Playback starts almost immediately after the seek with no frame glitches or drops.

### Actual result

Over 100 frame drops, 50+ consecutive.   These are the result of frames that should have been marked for discard (`decodeOnlyPresentationTimestamps`) are not discarded.

Our working assumption is that AmLogic has added some kind of frame-rate matching adjustment that causes double the frames on decode (e.g. 29.9 FPS conversion to 60hz display refresh) as was fixed for Nvidia in this commit https://github.com/google/ExoPlayer/commit/ee1ec8d3d6c89dfa11ca5504fd4eae919fcd83ac

The `auto-src` flag does not work however to turn this off.

### Media

Will send via email.

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

stevemayhew:
@tonihei  Here's some note and comments from our internal bug with Evolution for this...

[...]

Issue #6284, opened 2019-08-12
Leverage Android Q Thermal API to request low bit rate stream as needed


Ethan1983:
### [REQUIRED] Use case description
Android devices getting warm can throttle CPU/GPU and this could affect streaming performance.

### Proposed solution
Android Q provides thermal API for apps to be notified and fallback to low performance mode. These callbacks could be used by ExoPlayer to request low quality stream from multiple tracks (if available).

### Alternatives considered
Android Apps could do this logic explicitly but this is something all apps would need and an out of box solution from ExoPlayer would help avoid redoing logic across applications.

AquilesCanta:
@tonihei, mind having a look in the context of performance?

tonihei:
Thanks for the feature request. We can probably look into this, but it's slightly unclear what the logic should be exactly. I have the feeling different apps may want different behavior depending on the available formats and the various thermal severity levels. 

One very basic approach could be to listen to the thermal status in SimpleExoPlayer and force set the `forceLowestBitrate` flag once we reach THERMAL_STATUS_SEVERE. This is very conservative and I guess no one would object to that. 

Note that the API is currently only supported by Pixel devices running Q, so the practical effect is limited for now.

Ethan1983:
@tonihei A default handling that can be overridden or disabled by applications for custom logic (if needed) would help. Yes, at this moment it helps only Pixel devices but hopefully it should see wider adoption by other vendors.

ojw28:
This is interesting :). When considering this, we should think about the pros/cons of using the thermal status directly vs more user visible metrics like dropped frame count (and also the possibility of using a combination of the two).

> One very basic approach could be to listen to the thermal status in SimpleExoPlayer and force set the forceLowestBitrate flag once we reach THERMAL_STATUS_SEVERE. This is very conservative and I guess no one would object to that.

Could this approach end up repeatedly switching back and forth between the highest and lower resolutions (i.e. in the case where the highest resolution is sufficient to cause the device to heat up to THERMAL_STATUS_SEVERE, and the lowest resolution causes the device to cool down again).

tonihei:
> Could this approach end up repeatedly switching back and forth between the highest and lower resolutions (i.e. in the case where the highest resolution is sufficient to cause the device to heat up to THERMAL_STATUS_SEVERE, and the lowest resolution causes the device to cool down again).

[...]

PR #10468, opened 2022-07-25
Add skip surplus outBuffer for amlogic #10445

WontsonWong:


tonihei:
Thanks for the PR. In #10445 we said that we should follow up with Amlogic to see if the problem can be prevented entirely instead of just adding a workaround. So ideally, this change isn't needed at all. 

In addition, we have pending work to remove the `decodeOnlyPresentationTimestamps` list, which means this change is also not needed eventually. But depending on when we make this larger change, we may want to merge this type of workaround in the meantime. I'll leave it open until we hear more about the problem from Amlogic.

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -2478,6 +2486,13 @@ private static boolean codecNeedsMonoChannelCountWorkaround(String name, Format
         && "OMX.MTK.AUDIO.DECODER.MP3".equals(name);
   }
 
+  private static boolean codecNeedsSkipSurplusOutBufferWorkaround(String name, Format format) {
+    return Util.SDK_INT >= 29
+        && format.width >= 1920
+        && format.height >= 1080
+        && "OMX.amlogic.avc.decoder.awesome2".equals(name);
+  }
```
stevemayhew, line 2500: Is it all 1920x1080 content that frame rate conversion happens for?  What about other 30FPS content (4k?) on a 60HZ refresh TV?

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -1132,6 +1133,7 @@ private void initCodec(MediaCodecInfo codecInfo, MediaCrypto crypto) throws Exce
       this.codecNeedsAdaptationWorkaroundBuffer =
           codecAdaptationWorkaroundMode != ADAPTATION_WORKAROUND_MODE_NEVER;
     }
+    codecNeedsSkipSurplusOutBufferWorkaround = codecNeedsSkipSurplusOutBufferWorkaround(codecName, codecInputFormat);
     if ("c2.android.mp3.decoder".equals(codecInfo.name)) {
```
stevemayhew, line 1137: The format change to 1920x1080 occurs as an adaptive track selection.  Because the codec it self does not change.  So the logic will not trigger on a track change

library/core/src/main/java/com/google/android/exoplayer2/mediacodec/MediaCodecRenderer.java
```diff
@@ -2128,6 +2130,12 @@ private boolean isDecodeOnlyBuffer(long presentationTimeUs) {
         return true;
       }
     }
+    if (codecNeedsSkipSurplusOutBufferWorkaround && size > 0) {
+      if (Math.min(decodeOnlyPresentationTimestamps.get(0), decodeOnlyPresentationTimestamps.get(size - 1)) < presentationTimeUs
+          && Math.max(decodeOnlyPresentationTimestamps.get(0), decodeOnlyPresentationTimestamps.get(size - 1)) > presentationTimeUs) {
+        return true;
+      }
```
stevemayhew, line 2143: After modifying the code to:

1. add the SEEK_TO intent (sample code in the bug)
2. so the `codecNeedsSkipSurplusOutBufferWorkaround` is set when switching to 1080p

The return true on line 2136 is never executed.

Please follow the test steps in the bug exactly and add some logging to prove this code actually works, as @tonihei  noted this is a path that Google team is contemplating in the future, but for now it would be much preferred to disable the Auto-FRC if possible rather than complicating this code (which is in the critical render path).

Thanks!

stevemayhew (changes requested):

stevemayhew:
Thanks @WontsonWong the latest pull request looks better, it is missing the first frame render so the seek resolution time is perceived to be longer. 

One more thing to consider, the proposed solution (to turn off the auto frame rate conversion) is still better because up-converting the 30FPS to 60FPS adds to the CPU load and provides little or no visual benefit to the video quality.  As you can see from this bug https://github.com/google/ExoPlayer/issues/6284 we are seeing issues with some AmLogic SoC Android TV boxes with frame dropping because of thermal throttling of the CPU.

stevemayhew:
@WontsonWong  any movement on answering these two questions:

1. Is it possible to turn off the AmLogic auto frame rate up conversion with a property (like `auto-frc`), now or in some future code?
2. What encoding parameters in the source video trigger the feature? (as we don't see it in our 1080p30 streams but the test stream I've provided in the bug certainly does)?

If you read @icbaker 's comment here; https://github.com/google/ExoPlayer/issues/10445#issuecomment-1202926619  it is clear this feature violates the MediaCodec API.

Thanks!

WontsonWong:
@stevemayhew 
1.As far as I know, there is none
2.1080P && (25fps || 2997fps)

stevemayhew:
@tonihei  and @icbaker  See https://github.com/google/ExoPlayer/issues/10445#issuecomment-1208437362 on the bug this fixes.  Bottom line, this pull request is not needed.

I would encourage simplifying the logic to determine decode only to avoid depending on `MediaCodec`.  Happy to review and / or test any changes like this from Google's side.

stevemayhew:
@WontsonWong if you could share more on what this change was from AMLogic:

> SEI / AMLogic determined that the issue was specifically happening on during tuning when the 1080 P bitrate profile was invoked on tune, so AMLogic was able to turn off the "forced I-Frame" function in this case

I'm sure the teams involved here would be grateful.  Thanks!

.

icbaker:
I'm closing this because as noted above this doesn't seem like the best way to fix the issue.

PR #10477, opened 2022-07-27
SEEK_NEAREST does not adjust based on stale playlists

stevemayhew:
`getAdjustedSeekPosition()` fails to return a valid value if a track selection
updates the selected track followed by a seek operation, but before the
updated track has issued the first playlist refersh.

This occurs because the playlist for the new selected track has not yet been:

1. established as the primary playlist (`getNextChunk()` will do this)
2. Performed an initial load
3. Reported the `onPrimaryPlaylistRefreshed()` up to the main playback thread

The old code used `getPlaylistSnapshot()` can return a cached but potentially quite stale playlist
that is unsuitable for resolving a seek position.

The fix is to introduce a method, `HlsPlaylistTracker.getFreshPrimaryPlaylist()` that only returns
a playlist snapshot that is currently primary and recently loaded.

### DIFF
--- a/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java
+++ b/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java
@@ -196,6 +196,15 @@ public final class DefaultHlsPlaylistTracker
     return snapshot;
   }
 
+  @Nullable
+  @Override
+  public HlsMediaPlaylist getFreshPrimaryPlaylist(Uri url) {
+    MediaPlaylistBundle playlistBundle = playlistBundles.get(url);
+    boolean isPrimary = url.equals(primaryMediaPlaylistUrl);
+    boolean isFresh = playlistBundle.isSnapshotFresh();
+    return isPrimary && isFresh ? playlistBundle.getPlaylistSnapshot() : null;
+  }
+
   @Override
   public long getInitialStartTimeUs() {
     return initialStartTimeUs;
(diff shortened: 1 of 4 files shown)


library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java
```diff
@@ -196,6 +196,15 @@ public HlsMediaPlaylist getPlaylistSnapshot(Uri url, boolean isForPlayback) {
...
 
+  @Nullable
+  @Override
+  public HlsMediaPlaylist getFreshPrimaryPlaylist(Uri url) {
+    MediaPlaylistBundle playlistBundle = playlistBundles.get(url);
+    boolean isPrimary = url.equals(primaryMediaPlaylistUrl);
+    boolean isFresh = playlistBundle.isSnapshotFresh();
+    return isPrimary && isFresh ? playlistBundle.getPlaylistSnapshot() : null;
+  }
+
```
stevemayhew, line 207: Only the current primary HlsMediaPlaylist is up to date with the Period/Timeline used by ExoPlayerImplInternal, and only if it has been loaded (and thus reported the `MediaSource`.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java
```diff
@@ -550,6 +559,18 @@ public boolean isSnapshotValid() {
...
+      if (playlistSnapshot == null) {
+        return false;
+      }
+      long currentTimeMs = SystemClock.elapsedRealtime();
+      long updateDelta = currentTimeMs - lastSnapshotLoadMs;
+      return playlistSnapshot.hasEndTag
+          || playlistSnapshot.playlistType == HlsMediaPlaylist.PLAYLIST_TYPE_EVENT
+          || playlistSnapshot.playlistType == HlsMediaPlaylist.PLAYLIST_TYPE_VOD
+          || updateDelta <= C.usToMs(playlistSnapshot.targetDurationUs * 2);
+    }
```
stevemayhew, line 572: Possible the `TYPE_EVENT` should also be checked for freshness rather then assuming it is.

Note once it has updated once it may be overkill to check delta < target duration (as this is already covered by the stuck playlist check for primary), so just is primary and has updated (`lastSnapshotLoadMs != 0` or just `playlistSnapshot != null`).  If so, this method may not be needed.

tonihei:
See conversation in https://github.com/google/ExoPlayer/commit/530dd3f733a13190a8f75fd133bcb4aa24b06aaf# for more details. It may be we can solve the problem in a slightly simpler way than this approach.

stevemayhew:
Canceling this,  the simpler fix is in the works.  Thanks @tonihei for your help on this!

PR #10484, opened 2022-07-28
Seek nearest SYNC does not adjust stale playlists

stevemayhew:
`getPlaylistSnapshot()` can return a cached but potentially quite stale playlist
that may be unsuitable for resolving a seek position.

This change fixes a race conditon that can cause a failed attempt to resolve
a seek position with a stale playlist.

Pre-conditions for the race:

* A is the cached snapshot from previous playback of playlist A
* B is the selected playlist at time of the seek
* \* &mdash; is current playback position in period
* QQQ &mdash; positions in A cannot be seek targets from B, they have rolled out from `Timeline.Window`
* WWW &mdash; are positions in A that cannot be resolved in A, but would be in A`

```
+=================*==+
|       Period       |
+=================*==+

     +QQQ-------+
     |     A    |
     +QQQ-------+
         +-------WWW+
         |     B    |
         +-------WWW+

           seek and switch back to A occur here...

          +----------+
          |    A'    |
          +----------+
```

The seek request is issued with `Timeline.Window` from playlist B, yet the call to `getAdjustedSeekPositionUs()`
occurs:

1. After the current track selection swithces back to A
2. But before playlist snapshot A` is loaded, replacing the old A

The check is simple, does basic sanity check on `targetPositionInPlaylistUs` (that it
is < durationUs of the selected playlist).  This covers the positions WWW.


library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -270,29 +270,34 @@ public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParame
...
+    }
+
     // Segments start with sync samples (i.e., EXT-X-INDEPENDENT-SEGMENTS is set) and the playlist
-    // is non-empty, so we can use segment start times as sync points. Note that in the rare case
-    // that (a) an adaptive quality switch occurs between the adjustment and the seek being
+    // is non-empty, and the playlist is fresh enough that the period position falls in side of
+    // the playlist bound, so we can use segment start times as sync points.
+    //
+    // If (a) an adaptive quality switch occurs between the adjustment and the seek being
     // performed, and (b) segment start times are not aligned across variants, it's possible that
```
stevemayhew, line 286: Add comment about position exceeds duration check, note track switch on seek is not really a rare condition.  And the stale playlist issue happens for live if there are lots of ABR changes.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -270,28 +270,35 @@ public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParame
...
     // the adjusted position may not be at a sync point when it was intended to be. However, this is
     // very much an edge case, and getting it wrong is worth it for getting the vast majority of
     // cases right whilst keeping the implementation relatively simple.
-    long startOfPlaylistInPeriodUs =
-        mediaPlaylist.startTimeUs - playlistTracker.getInitialStartTimeUs();
-    long relativePositionUs = positionUs - startOfPlaylistInPeriodUs;
-    int segmentIndex =
```
stevemayhew, line 283: Moved this up to allow early return style.

library/hls/src/main/java/com/google/android/exoplayer2/source/hls/HlsChunkSource.java
```diff
@@ -270,28 +270,35 @@ public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParame
...
+    int segIndex =
         Util.binarySearchFloor(
             mediaPlaylist.segments,
-            relativePositionUs,
+            targetPositionInPlaylistUs,
```
stevemayhew, line 293: Mixing renaming, code changes and reformatting really makes it confusing for us to maintain.   Please let us know any formatting or other restructuring changes before the pull request is approved so the "merge" commit can be pure and free from other changes.

stevemayhew: P.S. This  change to move `startOfPlaylistInPeriodUs` add to convert playlist time to period time at the end is much more clear, thanks.

I'm happy to debate the name, but `targetPositionInPlaylistUs` seems more descriptive to me then `relativePositionUs`

library/hls/src/test/java/com/google/android/exoplayer2/source/hls/HlsChunkSourceTest.java
```diff
@@ -177,7 +194,7 @@ public void getAdjustedSeekPositionUs_emptyPlaylist() throws IOException {
 
     long adjustedPositionUs =
         testChunkSource.getAdjustedSeekPositionUs(
-            playlistTimeToPeriodTimeUs(100_000_000), SeekParameters.EXACT);
+            playlistTimeToPeriodTimeUs(100_000_000), SeekParameters.NEXT_SYNC);
 
     assertThat(periodTimeToPlaylistTimeUs(adjustedPositionUs)).isEqualTo(100_000_000);
```
stevemayhew, line 199: Expect it will not adjust, even though we ask it too (with non-EXACT).

icbaker:

PR #10514, opened 2022-08-10
Commits IDR NALU on end of stream

zacklukem:
Commits the active IDR NALU when the stream ends, without waiting for the next
packet (next iFrame-only segment) to load.

This is required for HLS I-Frame only segments that contain only a single PES
frame that simply ends with the last byte of the IDR NAL Unit.  This is only
for TS, where the size of the IDR NAL is unknown, with fMP4 it is indiated in
the sample metadata.

Without this change, a single iFrame load will not produce any commited sample data.

Also inluded is a sample iFrame, and test case for HLS TS parsing.


icbaker:

PR #10534, opened 2022-08-18
Refine codec selection based on MediaFormat

ybai001:
Submit this pull request for Dolby Vision team.

Since Android L, framework adds one [new API](https://developer.android.com/reference/android/media/MediaCodecList#findDecoderForFormat(android.media.MediaFormat)) to evaluate decoder with given MediaFormat. After that, android adds below format keys when evaluating decoder since Android N.
MediaFormat#KEY_PROFILE
MediaFormat#KEY_LEVEL

Currently, MediaCodecSelector uses MIME type to select decoder for Dolby Vision content. And this will return all decoders which are capable for both HEVC and AVC based Dolby Vision content. After that, trying to choose among these decoders that supports specific profile/level. 
We just wonder if we can simplify this procedure via [new API](https://developer.android.com/reference/android/media/MediaCodecList#findDecoderForFormat(android.media.MediaFormat)), which can set a good example of codec selection(both for decoder and encoder) for other applications.

Please help to check whether this update is suitable for contribution. Feel free to let us know if you have any question. Thanks for your time in advance.


ybai001:
Hi, @christosts,
Could you help to check whether this update is suitable for contribution? Feel free to let us know if you have any question. Thanks for your time in advance.

library/core/src/main/java/com/google/android/exoplayer2/video/MediaCodecVideoRenderer.java
```diff
@@ -462,9 +498,24 @@ private static List<MediaCodecInfo> getDecoderInfos(
...
+    if (Util.SDK_INT >= 24) {
+      /*
+        platform with API level higher than 24 will consider following keys:
+        MediaFormat#KEY_PROFILE,
+        MediaFormat#KEY_LEVEL,
+        MediaFormat#KEY_BIT_RATE,
+       */
+      decoderInfos.addAll(
+          tryEvaluatingCodecByFormat(format, requiresSecureDecoder, requiresTunnelingDecoder));
+    }
```
christosts, line 512: This change modifies how codecs are selected, and it would apply for every app. The change also of negates the reason why MediaCodecSelector exists, since an app might inject a custom MediaCodecSelector but the MediaCodecVideoRenderer would ignore it.

I think a more suitable solution would be to change the MediaCodecSelector to receive a Format instead of the MIME type as a String, and then it will be possible for apps who wish to use MediaCodecList.findDecoderForFormat() to inject their own MediaCodecSelector.

That said, changing the MediaCodecSelector API is a wider change and it might be more efficient for the ExoPlayer team to do it internally. Please see releated comment on the conversation section

christosts:
Thank you for sending the pull request and I apologize for the delayed response. I'll repeat a comment that I also left on the code. I think the suggested change does not fit well in the MediaCodecVideoRenderer API as it negates the existence of the MediaCodecSelector. I'm also not sure that changing the default codec selection to using MediaCodecList.findDecoderForFormat() does not risk any regressions so we try to be cautious when making such changes that apply to all apps and devices.

To support this use-case, we could change MediaCodecSelector to receive a Format argument instead of String. This way apps can inject a custom MediaCodecSelector that calls MediaCodecList.findDecoderForFormat(). That is a change that requires wider changes in ExoPlayer, and it might be more efficient if the ExoPlayer does this directly.

May I ask, what are the benefits of using MediaCodecList.findDecoderForFormat() compared to the existing implementation in ExoPlayer? To my knowledge, ExoPlayer's default logic is almost identical to what the platform does in MediaCodecList.findDecoderForFormat(). Is there something that ExoPlayer has missed?

yzhon:
Hi, Christosts, 
Really appreciate your kindly feedback.
I remember that you have already showed this proposal before in your email to my colleagues. It looks good to leave a chance for apps to inject a customer MediaCodecSelector. My colleagues will also keep in contact with your site about your proposal. 
Really thanks for your great effort.

ybai001:
Dolby will discuss with Google team about this topic based on Google's suggestion directly. Close this pull request.

Issue #6749, opened 2019-12-11
Support MediaCodec audio Float output


krocard:
## Use case description
Exoplayer (using MediaCodec) always play compressed audio with a 16bit encoding, when it should play float if available
This has the following drawbacks:
 - the audio framework native mode is float since P and has has been supporting float since API 21
 - quality is unnecessarily decreased (https://github.com/google/ExoPlayer/issues/6643#issuecomment-564281724)

And no advantages that I could identify.

## Proposed solution
Use Media Codec float output for audio if supported (API>24).
https://developer.android.com/reference/android/media/MediaCodec#raw-audio-buffers
To do that, the `MediaFormat#KEY_PCM_ENCODING` has to be set to `AudioFormat#ENCODING_PCM_FLOAT` during before calling `MediaCodec.configure`.

Also `DefaultAudioSink.enableConvertHighResIntPcmToFloat` should probably default to true.

## Alternatives considered

 - Use the ffmpeg extension to decode the stream to high res audio.
   This leads to bigger app size and does not use the platform codecs.
 - overridde `MediaCodecAudioRenderer.getMediaFormat` to configure the codec to output FLOAT.
   Drawback: Need to subclass MediaCodecAudioRenderer for one utility method. Might not be stable with respect to new releases.

ojw28:
> Also DefaultAudioSink.enableConvertHighResIntPcmToFloat should probably default to true.

This is now called `enableFloatOutput`, and is still disabled by default. It can't be enabled by default in its current form, since this causes audio processing (e.g., playback speed adjustment) to be disabled. I think we should work toward being able to remove the option, by making it so that `DefaultAudioSink` can automatically configure resampling on both sides of the other audio processors to do the best thing given the sink's configuration. Which probably means:

* On the input side, automatically enabling re-sampling to 16-bit integer PCM only if other audio processors in the chain actually need to do some processing.
* On the output side, automatically enabling re-sampling 24-bit and 32-bit integer PCM to float PCM if the output supports it, and automatically enabling re-sampling to 16-bit integer PCM otherwise.

krocard:
I think AudioProcessors should expose what input format they support. That would avoid having to re-sample to 16bit needlessly (which also decrease quality). Only if a processor does not support the inputFormat should the AudioSink automatically resampling to 16bit (as I guess all AudioProcessors are required to support it right now).
On the output side, I'm not sure to understand why resampling to Float from 24bit or 32bit PCM is required.
[...]

PR #10565, opened 2022-08-28
Add support for custom float audio processors

### LINKED ISSUE #10516: Improve support for custom float audio processors
###  Use case description
My app is a music player with many audiophile users. These users tend to use very high-quality lossless audio that can have bit rates in the megabytes and usually output large samples like 32-bit or Float PCM. However, ExoPlayer's functionality for rendering this high quality audio `enableFloatOutput` is severely lacking for two reasons:

1. I cannot use `AudioProcessor` instances with float output. This prevents features that are standard for music players, like ReplayGain. This is presumably due to `SonicAudioProcessor` and other internal audio processors expecting 16-bit PCM.
2. ExoPlayer always resamples audio to a single format. On low-power devices playing high-definition files, this can result in distorted playback as the resampling task is simply too expensive. This is also probably due to the internal audio processors.

### Proposed solution
Replace `enableFloatOutput` with a new `AudioSink` flag called `enableRawOutput` that signals ExoPlayer to drop all of it's internal audio processors and just output in the source audio quality. This output should still be routed to the user's audio processors, however, with some documentation explaining that user audio processors should expect all possible PCM formats if `enableRawOutput` is `true`.

### Alternatives considered
I could do some evil reflection hacks in `DefaultAudioSink` (with float output enabled) to replace the internal audio processors which resample with my own audio processors that don't resample. I do not like this at all, especially since I don't know if there are any other reasons for why float output is structured this way outside of the internal audio processors.

### Additional info
If there is any other good reason for why resampling is done like this, please let me know. I am also happy to work on a PR that implements this if it is a reasonable addition.

OxygenCobalt:
### The problem

My app leverages audio processing in multiple locations. However, it is also in a category of apps that benefit from having HD audio output. With the current float output implementation, I cannot implement both of these features simultaneously without one having to be dropped.

### What this PR does

This PR reworks `DefaultAudioSink` to use a single audio processor chain. This way, custom audio processors can receive float output if enabled.

### What this PR does not do

This PR does not change `SilenceSkippingAudioProcessor` or `SonicAudioProcessor` to handle float audio. This is primarily because I am not familiar with the functionality of either, and because there are seemingly API limitations stopping this according to https://github.com/google/ExoPlayer/issues/6749#issuecomment-603833018. Since `enableFloatOutput` is still not enabled by default, I feel like it would be best to leave this as a task for the maintainers that know more about how to do this.

Feel free to raise any concerns with my change.

Resolves #10516

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/DefaultAudioSink.java
@@ -315,8 +315,9 @@ public final class DefaultAudioSink implements AudioSink {
     /**
      * Sets whether to enable 32-bit float output or integer output. Where possible, 32-bit float
      * output will be used if the input is 32-bit float, and also if the input is high resolution
-     * (24-bit or 32-bit) integer PCM. Float output is supported from API level 21. Audio processing
-     * (for example, speed adjustment) will not be available when float output is in use.
+     * (24-bit or 32-bit) integer PCM. Float output is supported from API level 21. Playback
+     * parameter functionality (for example, speed adjustment) will not be available when float
+     * output is in use.
      *
      * <p>The default value is {@code false}.
      */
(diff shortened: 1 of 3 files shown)


OxygenCobalt:
Actually, I'm going to close this PR. It's nowhere near ready enough, and I'll try to create a better one that actually fully stabilizes this feature.

PR #10570, opened 2022-08-30
fix ProgressiveDownloader infinite loop

Artemych:
We could get an infinite loop inside the ProgressiveDownloader as downloadRunnable(RunnableFutureTask) could cache the 
PriorityTooLowException and then we would loop forever until the ProgressiveDownloader would be cancelled.

Detailed explanation:
Thread 1 - ProgressiveDownloader working thread
Thread 2 - ExoPlayerImpl working thread

1. ProgressiveDownloader adds C.PRIORITY_DOWNLOAD (Thread 1)
2. ProgressiveDownloader proceeds with C.PRIORITY_DOWNLOAD (Thread 1)
3. ExoPlayerImpl add C.PRIORITY_PLAYBACK (Thread 2)
4. ProgressiveDownloader using PriorityDataSource trying proceedOrThrow and receives an PriorityTooLowException
5. RunnableFutureTask caches the PriorityTooLowException
6. ExoPlayerImpl removes C.PRIORITY_PLAYBACK
7. infinite loop as RunnableFutureTask on get check the cached error state.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/offline/ProgressiveDownloader.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/offline/ProgressiveDownloader.java
@@ -88,19 +88,7 @@ public final class ProgressiveDownloader implements Downloader {
   public void download(@Nullable ProgressListener progressListener)
       throws IOException, InterruptedException {
     this.progressListener = progressListener;
-    downloadRunnable =
-        new RunnableFutureTask<Void, IOException>() {
-          @Override
-          protected Void doWork() throws IOException {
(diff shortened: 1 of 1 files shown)


ryihan (approved):

tonihei:
Thanks for reporting! We interestingly already fixed a similar issue in https://github.com/roblav96/ExoPlayer/commit/18f7d85c91996c0b6109b2bdec6f8e5ed7a50e44, but only for external retry loops that call the `download` method again. The internal loop after a `PriorityTooLowException` has the same problem of course and needs to be fixed as well.

PR #10588, opened 2022-09-06
Add support for TS Teletext subtitles

cdongieux:
This PR adds support for Teletext subtitles in TS streams.

Some readings:
- EN 300 472 - V1.4.1 - Digital Video Broadcasting (DVB); Specification for conveying ITU-R System B Teletext in DVB bitstreams
- EN 300 468 - V1.16.1 - Digital Video Broadcasting (DVB); Specification for Service Information (SI) in DVB systems
- ETS 300 706 - Enhanced Teletext specification (good luck with this one)

You can test this PR with this sample file, its a dump of a French television channel stream "NRJ12" with 2 teletext tracks : https://drive.google.com/file/d/19ZYsoeUfH85gEilQkaAdLbPhC4CxhDEh/view?usp=sharing
You can also compare the result in ExoPlayer with VLC which handles Teletext subs too.

Let me know if you need further info :)


icbaker:
Thanks for sending this - it's quite a large change, so I'm afraid you'll have to bear with me while I find the time to review it.

icbaker:
Do you have any test clips containing teletext subtitles that you could add to this PR? We could use them in:
* [`TsExtractorTest`](https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/extractor/ts/TsExtractorTest.java) to ensure the track is extracted
* Unit tests for the new `DvbTeletextDecoder`

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextReader.java
```diff
@@ -0,0 +1,100 @@
...
+      TrackOutput output = extractorOutput.track(idGenerator.getTrackId(), C.TRACK_TYPE_TEXT);
+      Format.Builder formatBuilder = new Format.Builder()
+          .setId(idGenerator.getFormatId())
+          .setSampleMimeType(MimeTypes.APPLICATION_TELETEXT)
+          .setInitializationData(Collections.singletonList(teletextInfo.initializationData))
+          .setLanguage(teletextInfo.language);
+      if (teletextInfo.type == 0x02) { // Teletext subtitle page
+        formatBuilder.setRoleFlags(C.ROLE_FLAG_SUBTITLE);
+      } else if (teletextInfo.type == 0x05) { // Teletext subtitle page for hearing impaired people
+        formatBuilder.setRoleFlags(C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND);
```
icbaker: I think in this case the flags should **also** contain ROLE_FLAG_SUBTITLE:

```java
formatBuilder.setRoleFlags(C.ROLE_FLAG_SUBTITLE | C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND);
```

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +753,24 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
         } else if (descriptorTag == TS_PMT_DESC_ISO639_LANG) {
           language = data.readString(3).trim();
           // Audio type is ignored.
+        } else if (descriptorTag == TS_PMT_DESC_TELETEXT) {
+          // ETSI EN 300 468 - V1.16.1 - 6.2.43 Teletext descriptor
+          streamType = TS_STREAM_TYPE_TELETEXT;
+          dvbTeletextInfos = new ArrayList<>();
+          while (data.getPosition() < positionOfNextDescriptor) {
+            String teletextLanguage = data.readString(3).trim();
```
icbaker: Can you use the `readString` overload that takes a `Charset` and pass [`Charsets.ISO_8859_1`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Charsets.html#ISO_8859_1)? The version you're using atm assumes UTF-8

From ETSI EN 300 468 - V1.16.1 - 6.2.43 (emphasis mine):
> **ISO_639_language_code:** This 24-bit field contains the 3 character ISO 639-2 [36] language code of the language of the teletext. Both ISO 639-2/B and ISO 639-2/T may be used. **Each character is coded into 8 bits according to ISO/IEC 8859-1 [17] and inserted in order into the 24-bit field.**

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +753,24 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
...
+            byte[] initializationData = new byte[2];
+            data.readBytes(initializationData, 0, 2);
+            byte typeAndMag = initializationData[0];
+            int type = typeAndMag >> 3;
```
icbaker: I think this is dodgy if `typeAndMag` has its MSB set to 1, i.e. it's negative in Java's signed view of the world. It looks like the Java sign-propagation and expansion to `int` causes problems here:

```java
byte typeAndMag = (byte) 0x80;  // Java thinks this is -128

// Java expands the -128 byte (0x80) to -128 int (0xFFFF_FFC0) and then shifts this by
// 3 to give type=0xFFFF_FFF8 (-8). I think the code below would expect type to be equal
// to 0x10 (16).
int type = typeAndMag >> 3;  
```

You can avoid gotchas like this by using [`ParsableBitArray`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/util/ParsableBitArray.html) and only reading the number of bits you care about. I'd suggest doing that, and reading the bytes directly from `ParsableByteArray` into it using the `readBytes(ParsableBitArray, int)` overload.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.java
```diff
@@ -111,6 +117,27 @@ public DvbSubtitleInfo(String language, int type, byte[] initializationData) {
     }
   }
 
+  /**
+   * Holds information about a DVB Teletext, as defined in ETSI EN 300 468 V1.16.1 section 6.2.43.
+   */
+  final class DvbTeletextInfo {
+
+    public final String language;
+    public final int type;
```
icbaker: This can be a byte, right? (it's max 5 bits afaict)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +753,24 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
...
+            if (type != 0x02 /* Teletext subtitle page */
+                && type != 0x05 /* Teletext subtitle page for hearing impaired people */) {
+              // other types are not subtitles, ignore them
+              continue;
+            }
+            dvbTeletextInfos.add(
+                new DvbTeletextInfo(teletextLanguage, type, initializationData));
```
icbaker: optional: i'd consider flipping this, to put the `.add()` call inside the `if()` block.

Also, nit: Can you move the comments out of the conditional statement and onto their own line inside the if-block?

```java
if (type == 0x02 || type == 0x05) {
  // Only 'Teletext subtitle page' (0x02) and 'Teletext subtitle page for hearing 
  // impaired people' (0x05) are supported.
  dvbTeletextInfos.add(
      new DvbTeletextInfo(teletextLanguage, type, initializationData));
}
```

(this may be irrelevant if you remove the conditional completely)

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextReader.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.android.exoplayer2.extractor.TrackOutput;
+import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.DvbTeletextInfo;
+import com.google.android.exoplayer2.extractor.ts.TsPayloadReader.TrackIdGenerator;
+import com.google.android.exoplayer2.util.MimeTypes;
+import com.google.android.exoplayer2.util.ParsableByteArray;
+import java.util.Collections;
+import java.util.List;
+
+/** Parses DVB Teletext subtitle data and extracts individual frames. */
+public final class DvbTeletextReader implements ElementaryStreamReader {
```
icbaker: The name of this class makes it sound like it handles all aspects DVB teletext data, but the javadoc only mentions subtitles.

Should it be renamed to make it clear it only supports subtitles? e.g. `DvbTeletextSubtitleReader`

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +753,24 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
...
+          dvbTeletextInfos = new ArrayList<>();
+          while (data.getPosition() < positionOfNextDescriptor) {
+            String teletextLanguage = data.readString(3).trim();
+            byte[] initializationData = new byte[2];
+            data.readBytes(initializationData, 0, 2);
+            byte typeAndMag = initializationData[0];
+            int type = typeAndMag >> 3;
+            if (type != 0x02 /* Teletext subtitle page */
+                && type != 0x05 /* Teletext subtitle page for hearing impaired people */) {
+              // other types are not subtitles, ignore them
```
icbaker: What do you think about handling all values of `teletext_type` here, and moving the 'is it 0x02 or 0x05?' logic into `DvbTeletext(Subtitle)Reader` (which actually already has conditional statements for these two types)

I don't think there's a problem with handling all teletext packets the same here - but I could be wrong.

It feels better to keep all the 'type checking/filtering' in one place if possible.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.java
```diff
@@ -111,6 +117,27 @@ public DvbSubtitleInfo(String language, int type, byte[] initializationData) {
...
+    public final byte[] initializationData;
+
+    /**
+     * @param language The ISO 639-2 three-letter language code.
+     * @param type The Teletext type.
+     * @param initializationData The magazine number and page number.
```
icbaker, line 132: I'm not sure the current implementation follows this javadoc? On L763 you read all 16 bits into the two bytes of this array, which covers 5 bits of type, 3 bits of mag number and 8 bits of page number. But AFAICT you never modify the first byte of this array to 'remove' the type bits.

This might also be clearer if you switch to `ParsableBitArray` as suggested above.

If you do this, I would consider having explicit fields in this type for magazine number and page number, and building the `initializationData` array later.

(oh, it looks like in `DvbTeletextDecoder` you've compensated for this by only reading the lowest 3 bits of the first byte: `int magazineNumber = data.readUnsignedByte() & 0x07;` - which does suggest this javadoc is being ignored?)

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextDecoder.java
```diff
@@ -0,0 +1,36 @@
...
+
+  private final DvbTeletextParser parser;
+
+  /**
+   * @param initializationData The initialization data for the decoder. The initialization data must
+   *     consist of a single byte array containing 2 bytes: magazine_number (1), page_number (2).
+   */
+  public DvbTeletextDecoder(List<byte[]> initializationData) {
+    super("DvbTeletextDecoder");
+    ParsableByteArray data = new ParsableByteArray(initializationData.get(0));
```
icbaker: nit: i think you can avoid this allocation by using guava's [`UnsignedInts.toInt`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedBytes.html#toInt(byte)) directly on the `initializationData` array.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
```diff
@@ -0,0 +1,513 @@
...
+import com.google.android.exoplayer2.util.Log;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/** Parses {@link Cue}s from a DVB Teletext subtitle bitstream. */
+/* package */ final class DvbTeletextParser {
```
icbaker: Does this need to be its own class? It kinda feels like it could be in-lined into `DvbTeletextDecoder`

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
```diff
@@ -0,0 +1,513 @@
...
+import com.google.android.exoplayer2.text.Cue;
+import com.google.android.exoplayer2.util.Log;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/** Parses {@link Cue}s from a DVB Teletext subtitle bitstream. */
```
icbaker: Please add a reference to the relevant spec here

I haven't fully reviewed the rest of this file yet since I'm not sure which spec I should be consulting

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
```diff
@@ -0,0 +1,513 @@
...
+/* package */ final class DvbTeletextParser {
+
+  private static final String TAG = "DvbTeletextParser";
+
+  private static final int TTX_HEIGHT = 576;
+  private static final int TTX_COLS = 40;
+  private static final int TTX_ROWS = 24;
+  private static final int CHAR_HEIGHT = 21;
+
+  private static final int[] TXT_COLORS = {
```
icbaker: Don't use abbreviations like TXT unless it's a literal reference to something in the spec

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
```diff
@@ -0,0 +1,513 @@
...
+   *
+   * @param data The subtitling packet data to decode.
+   * @param size The limit in {@code data} at which to stop decoding.
+   * @return The parsed {@link Cue}s.
+   */
+  public List<Cue> decode(byte[] data, int size) {
+    int offset = 0;
+    int len = size;
+
+    int dataIdentifier = data[offset++] & 0xFF;
```
icbaker: This might be better with a `ParsableByteArray`? 

If you want to avoid the allocation you could use a scratch instance as a field, and reset it with `data` each time.

icbaker (changes requested):

wazerstar:
> Do you have any test clips containing teletext subtitles that you could add to this PR? We could use them in:
> 
>     * [`TsExtractorTest`](https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/extractor/ts/TsExtractorTest.java) to ensure the track is extracted
> 
>     * Unit tests for the new `DvbTeletextDecoder`


3 samples, you click over to the right side of the page "Download"

https://cyberfile.is/Fbo

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextReader.java
cdongieux: Don't you think this would be better?
```
formatBuilder.setRoleFlags(C.ROLE_FLAG_CAPTION | C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND);
```
Because this type of subs adds information about music and sound, on top of dialogs.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
cdongieux: Well, I applied exactly the same logic as the existing `DvbParser`/`DvbDecoder` classes.

icbaker:
> 3 samples, you click over to the right side of the page "Download"
> 
> https://cyberfile.is/Fbo

This seems to just open repeated NSFW pop-ups, please upload these somewhere less dubious (I haven't been able to download them yet).

It also looks like these are going to be too large to be added to git and used for automated testing, but may be useful for manual testing. In order to for us to accept this PR we also need some very small sample files (<1MB each) (that can have any copyright assigned under the CLA you've signed) into a git commit and put directly in this PR (in this directory: https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/ts).

It might also be good to write a `DvbTeletextDecoderTest` - I recently started a `Cea608DecoderTest` that might be a useful starting point: https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/text/cea/Cea608DecoderTest.java

wazerstar:
> > 3 samples, you click over to the right side of the page "Download"
> > https://cyberfile.is/Fbo
> 
> This seems to just open repeated NSFW pop-ups, please upload these somewhere less dubious (I haven't been able to download them yet).
> 
> It also looks like these are going to be too large to be added to git and used for automated testing, but may be useful for manual testing. In order to for us to accept this PR we also need some very small sample files (<1MB each) (that can have any copyright assigned under the CLA you've signed) into a git commit and put directly in this PR (in this directory: https://github.com/google/ExoPlayer/tree/dev-v2/testdata/src/test/assets/media/ts).
> 
> It might also be good to write a `DvbTeletextDecoderTest` - I recently started a `Cea608DecoderTest` that might be a useful starting point: https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/text/cea/Cea608DecoderTest.java


Just as a note I'm trying to help with samples here, I'm not the owner of this topic.

With all respect for me I cannot see it's possible to have a raw sample under ~15-20mb with .ts format that includes teletext subtitles, I tried, unless somebody else knows a way and can contribute with ?

Here you go.

gdrive link.
https://drive.google.com/file/d/1MOeoHCdQ4M9HIuItL-_vkfaYwwcFMsrt/view?usp=sharing

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
cdongieux: Done in 0037e8917c8fc3f5510d3e56e520f0f9cb8ef6ff

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
cdongieux: Done in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.java
cdongieux: Done in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
cdongieux: Done in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextReader.java
cdongieux: Done in 4578299b5cfe65018085a24537c6e29e676a1cb2

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
cdongieux: Done in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.java
cdongieux: Fixed in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextDecoder.java
cdongieux: Done in cb97b5637c0906fb72a10a6ffc2e90f1db98da94

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
cdongieux: Done in 69f49d5388fbdc6972d7ca7517171fad3617d4de

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
cdongieux: Done in 69f49d5388fbdc6972d7ca7517171fad3617d4de

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
cdongieux: Done in 69f49d5388fbdc6972d7ca7517171fad3617d4de

cdongieux:
> Do you have any test clips containing teletext subtitles that you could add to this PR? We could use them in:
> 
> * [`TsExtractorTest`](https://github.com/google/ExoPlayer/blob/dev-v2/library/extractor/src/test/java/com/google/android/exoplayer2/extractor/ts/TsExtractorTest.java) to ensure the track is extracted
> * Unit tests for the new `DvbTeletextDecoder`

I have added a sample TS file containing Teletext subtitle only (no audio, no video), and I updated TsExtractorTest with a new test.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextSubtitleReader.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public void seek() {
+    writingSample = false;
+    sampleTimeUs = C.TIME_UNSET;
+  }
+
+  @Override
+  public void createTracks(ExtractorOutput extractorOutput, TrackIdGenerator idGenerator) {
+    for (int i = 0; i < outputs.length; i++) {
+      DvbTeletextInfo teletextInfo = teletextInfos.get(i);
```
icbaker, line 56: This looks dodgy, because above in the constructor you only add items to `outputs` for a subset of `teletextInfos`, so these indices might not match up.

What about using an `ArrayList` for `outputs` instead? So above in the constructor you just store `teletextInfos` into a field and instantiate an empty `outputs` list, then move all the filtering logic into this method and add items to `outputs` as you go.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/DvbTeletextSubtitleReader.java
```diff
@@ -0,0 +1,112 @@
...
+      TrackOutput output = extractorOutput.track(idGenerator.getTrackId(), C.TRACK_TYPE_TEXT);
+      Format.Builder formatBuilder = new Format.Builder()
+          .setId(idGenerator.getFormatId())
+          .setSampleMimeType(MimeTypes.APPLICATION_TELETEXT)
+          .setInitializationData(Collections.singletonList(teletextInfo.initializationData))
+          .setLanguage(teletextInfo.language);
+      if (type == SUBTITLE_PAGE) {
+        formatBuilder.setRoleFlags(C.ROLE_FLAG_SUBTITLE);
+      } else {
+        formatBuilder.setRoleFlags(C.ROLE_FLAG_DESCRIBES_MUSIC_AND_SOUND);
```
icbaker, line 68: As discussed in a [previous comment](https://github.com/google/ExoPlayer/pull/10588#discussion_r993382189), should this be OR'd with `C.ROLE_FLAG_CAPTION`?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +754,23 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
...
           language = data.readString(3).trim();
           // Audio type is ignored.
+        } else if (descriptorTag == TS_PMT_DESC_TELETEXT) {
+          // ETSI EN 300 468 - V1.16.1 - 6.2.43 Teletext descriptor
+          streamType = TS_STREAM_TYPE_TELETEXT;
+          dvbTeletextInfos = new ArrayList<>();
+          while (data.getPosition() < positionOfNextDescriptor) {
+            ParsableBitArray scratch = new ParsableBitArray(new byte[5]);
+            data.readBytes(scratch, 5);
+            String teletextLanguage = scratch.readBytesAsString(3, Charsets.ISO_8859_1).trim();
```
icbaker, line 764: nit: you can probably read the string from the `ParsableByteArray` and only use the `ParsableBitArray` for the next fiddly bit?

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsExtractor.java
```diff
@@ -749,6 +754,23 @@ private EsInfo readEsInfo(ParsableByteArray data, int length) {
...
+            int type = scratch.readBits(5);
+            int magazineNumber = scratch.readBits(3);
+            int pageNumber = scratch.readBits(8);
+            dvbTeletextInfos.add(
+                new DvbTeletextInfo(
+                    teletextLanguage, (byte) type,
+                    new byte[] { (byte) magazineNumber, (byte) pageNumber })
```
icbaker, line 771: Consider moving the byte casting up to where the locals are declared and assigned (i.e. declare them as type `byte`). And please use Guava's [`UnsignedBytes.checkedCast`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedBytes.html#checkedCast(long)).

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ts/TsPayloadReader.java
icbaker: What do you think about this suggestion?

> If you do this, I would consider having explicit fields in this type for magazine number and page number, and building the initializationData array later.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextParser.java
icbaker: Ack - let's leave as-is.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      0x8F, 0x42, 0x4F, 0x8F, 0x4A, 0x8F, 0x8F, 0x49,
+      0x08, 0x48, 0x48, 0x8F, 0x48, 0x8F, 0x8F, 0x49,
+      0x48, 0x8F, 0x8F, 0x4B, 0x8F, 0x4E, 0x43, 0x8F,
+      0x8F, 0x4C, 0x4F, 0x8F, 0x44, 0x8F, 0x8F, 0x49,
+      0x4F, 0x8F, 0x0F, 0x4F, 0x8F, 0x4E, 0x4F, 0x8F,
+      0x48, 0x8F, 0x8F, 0x45, 0x8F, 0x4E, 0x4D, 0x8F,
+      0x8F, 0x4E, 0x4F, 0x8F, 0x4E, 0x0E, 0x8F, 0x4E,
+  };
+
+  private static final int[] SPECIAL_CHARS_IDX;
```
icbaker, line 143: I don't see the word 'special' used in ETS 300 706 to describe characters or character sets.

What about sticking with the terminology in the spec and naming this something about "national option sub-set"?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    if ((a & 0x80) != 0 || (b & 0x80) != 0) {
+      Log.w(TAG, "uncorrectable hamming error");
+    }
+    return ((b & 0x0F) << 4) | (a & 0x0F);
+  }
+
+  private static void handleSpecialChars(byte[] data, int offset, int len, int lang) {
+    boolean mosaic = false;
+    for (int i = offset; i < offset + len; ++i) {
+      int c = data[i] & 0xFF;
```
icbaker, line 222: imo this is clearer if spelled [`UnsignedBytes.toInt(data[i])`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedBytes.html#toInt(byte))

Same elsewhere

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      0x4F, 0x8F, 0x0F, 0x4F, 0x8F, 0x4E, 0x4F, 0x8F,
+      0x48, 0x8F, 0x8F, 0x45, 0x8F, 0x4E, 0x4D, 0x8F,
+      0x8F, 0x4E, 0x4F, 0x8F, 0x4E, 0x0E, 0x8F, 0x4E,
+  };
+
+  private static final int[] SPECIAL_CHARS_IDX;
+  private static final char[][] SPECIAL_CHARS = {
+      /* for latin-1 font */
+      /* English (100%) */
+      {0, '£', '$', '@', '«', '½', '»', '¬', '#', '­', '¼', '¦', '¾', '÷'},
```
icbaker, line 147: it looks like this is derived from Table 36 in section 15.6.2 in ETS 300 706, but I'm not sure I understand why the leading `0` is needed (which afaict isn't present in the table).

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      0x08, 0x48, 0x48, 0x8F, 0x48, 0x8F, 0x8F, 0x49,
+      0x48, 0x8F, 0x8F, 0x4B, 0x8F, 0x4E, 0x43, 0x8F,
+      0x8F, 0x4C, 0x4F, 0x8F, 0x44, 0x8F, 0x8F, 0x49,
+      0x4F, 0x8F, 0x0F, 0x4F, 0x8F, 0x4E, 0x4F, 0x8F,
+      0x48, 0x8F, 0x8F, 0x45, 0x8F, 0x4E, 0x4D, 0x8F,
+      0x8F, 0x4E, 0x4F, 0x8F, 0x4E, 0x0E, 0x8F, 0x4E,
+  };
+
+  private static final int[] SPECIAL_CHARS_IDX;
+  private static final char[][] SPECIAL_CHARS = {
```
icbaker, line 144: This charset stuff is really hard to follow...

What about using a Guava ImmutableTable<Integer, Byte, Character> to directly replicate table 36 in the spec. That might remove lots of these arrays and make the usage a bit easier to reason about?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      0x1b, 0x9b, 0x5b, 0xdb, 0x3b, 0xbb, 0x7b, 0xfb,
+
+      0x07, 0x87, 0x47, 0xc7, 0x27, 0xa7, 0x67, 0xe7,
+      0x17, 0x97, 0x57, 0xd7, 0x37, 0xb7, 0x77, 0xf7,
+      0x0f, 0x8f, 0x4f, 0xcf, 0x2f, 0xaf, 0x6f, 0xef,
+      0x1f, 0x9f, 0x5f, 0xdf, 0x3f, 0xbf, 0x7f, 0xff
+  };
+
+  // 7bit with 1 bit Odd parity
+  private static final byte[] PARITY_TABLE = {
```
icbaker, line 84: I know this pattern already exists in `Cea608Decoder` - but I'm not sure this is better than just using [`Integer.bitCount`](https://developer.android.com/reference/java/lang/Integer#bitCount(int))? I was actually considering switching `Cea608Decoder` to use that - since imo it's more obviously correct to a future reader (analysing this array by hand for correctness is very tricky).

Something like:

```java
private static void checkParity(byte[] data, int offset, int len) {
  for (int i = 0; i < len; ++i) {
    int a = data[offset + i] & 0xFF;
    if (Integer.bitCount(UnsignedBytes.toInt(data[offset + i])) % 2 == 0) {
      Log.w(TAG, "parity error at " + (offset + i));
    }
    data[offset + i] &= 0x7F;
  }
}
```

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      {0, '#', '¢', 'Þ', 'Â', 'ª', 'Ã', 'Î', 'i', 'þ', 'â', 'º', 'ã', 'î'},
+  };
+
+  private static final int[] CHARSET_INDEXES = {0, 4, 2, 6, 1, 5, 3, 7};
+
+  private static void checkParity(byte[] data, int offset, int len) {
+    for (int i = 0; i < len; ++i) {
+      int a = data[offset + i] & 0xFF;
+      if ((PARITY_TABLE[a] & 32) == 0) {
+        Log.w(TAG, "parity error at " + (offset + i));
```
icbaker, line 188: shouldn't decoding fail here rather than just logging and continuing?

Same below for unrecoverable hamming codes

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      {0, '#', 'Ë', 'È', 'Æ', '®', 'Ð', '©', 'ë', 'è', 'æ', '®', 'ð', '¹'},
+      /* Czech/Slovak (100%) */
+      {0, '#', 'ù', 'è', '»', '¾', 'ý', 'í', 'ø', 'é', 'á', 'ì', 'ú', '¹'},
+      /* Rumanian (95%) */
+      {0, '#', '¢', 'Þ', 'Â', 'ª', 'Ã', 'Î', 'i', 'þ', 'â', 'º', 'ã', 'î'},
+  };
+
+  private static final int[] CHARSET_INDEXES = {0, 4, 2, 6, 1, 5, 3, 7};
+
+  private static void checkParity(byte[] data, int offset, int len) {
```
icbaker, line 184: This method also removes the parity bit, I think that should be made clear in the name - e.g. `checkAndRemoveParityBit`

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+          break;
+        case 0xFF:
+          // stuffing
+          break;
+        default:
+          // ignore unknown data unit id
+          break;
+      }
+    }
+    return new ArrayList<>(cues);
```
icbaker, line 324: Can you switch this to `ImmutableList.copyOf`?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    return cueBuilder
+        .setText(stringBuilder)
+        .setTextSize(size, Cue.TEXT_SIZE_TYPE_FRACTIONAL)
+        .setLine((float) line.row / (TELETEXT_ROWS + 1), Cue.LINE_TYPE_FRACTION)
+        .build();
+  }
+
+  @Nullable
+  TeletextLine processTeletextSubtitleField(ParsableByteArray data, int len) {
+    if (len != 0x2C) {
```
icbaker, line 383: Section 7.1 suggests a teletext packet is 45 bytes, but this is checking `len == 44` - why the discrepancy?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+
+    // ETSI EN 300 706 - 6.2 - Framing code
+    // The bit pattern in transmission order is: 1110 0100 (0xE4)
+    if (framingCode != 0xE4) {
+      // ignore invalid framing code
+      return null;
+    }
+
+    data.setPosition(data.getPosition() - 2);
+    revertBytes(data.getData(), data.getPosition(), len);
```
icbaker, line 399: Is this needed because of this bit of section 7.1?
> In each byte, the bits are numbered 1 to 8 (LSB to MSB), and are normally transmitted LSB first.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      int a = data[offset + i] & 0xFF;
+      if ((PARITY_TABLE[a] & 32) == 0) {
+        Log.w(TAG, "parity error at " + (offset + i));
+      }
+      data[offset + i] &= 0x7F;
+    }
+  }
+
+  // revert bits order from LSB first to MSB first
+  private static void revertBytes(byte[] data, int offset, int len) {
```
icbaker, line 195: I don't think 'revert' is the right word here

Maybe reverse? But that might suggest reversing `data` rather than reversing each byte

["Bit endian"](https://en.wikipedia.org/wiki/Endianness#Bit_endianness[citation_needed]) is probably a good phrase that's less ambiguous, so maybe `correctBitEndianness` and reference section 7.1 of the ETS 300 706 spec from the javadoc of this method.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+   * Decodes a subtitling packet, returning a list of parsed {@link Cue}s.
+   *
+   * @param data The subtitling packet data to decode.
+   * @param size The limit in {@code data} at which to stop decoding.
+   * @return The parsed {@link Cue}s.
+   */
+  public List<Cue> decode(byte[] data, int size) {
+    scratch.reset(data, size);
+
+    int dataIdentifier = scratch.readUnsignedByte();
```
icbaker, line 286: I couldn't work out where in ETS 300 706 this is defined - could you reference the relevant section from a code comment here?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+          break;
+        default:
+          // ignore unknown data unit id
+          break;
+      }
+    }
+    return new ArrayList<>(cues);
+  }
+
+  private Cue buildCue(TeletextLine line) {
```
icbaker, line 327: nit: please move this below `processTeletextSubtitleField` since that is called first in `decode` above.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+        Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+    return cueBuilder
+        .setText(stringBuilder)
+        .setTextSize(size, Cue.TEXT_SIZE_TYPE_FRACTIONAL)
+        .setLine((float) line.row / (TELETEXT_ROWS + 1), Cue.LINE_TYPE_FRACTION)
+        .build();
+  }
+
+  @Nullable
+  TeletextLine processTeletextSubtitleField(ParsableByteArray data, int len) {
```
icbaker, line 382: can this be private?

icbaker, line 382: The ETS 300 706 spec seems (I think?) to refer to this as a 'teletext packet' rather than a 'field' - would it be clearer to use that wording here too?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    if (len != 0x2C) {
+      // ignore invalid data unit length
+      return null;
+    }
+
+    data.setPosition(data.getPosition() + 1);
```
icbaker, line 388: this is probably more clearly spelled as `data.skipBytes(1)`

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+        data.getData()[i] = 0x20;
+      }
+    }
+
+    handleSpecialChars(data.getData(), data.getPosition(), len, page.charset);
+    page.updateLine(packetNumber, data.getData(), data.getPosition(), rowOffset, len);
+    return updatePage(page, packetNumber);
+  }
+
+  private TeletextLine updatePage(@NonNull TeletextPage page, int row) {
```
icbaker, line 471: This is a slightly confusing name for a method that mostly (?) creates a `TeletextLine` instance.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    int magazineAndFrameAddr = hamming84Decode(data.readUnsignedByte(), data.readUnsignedByte());
+    int magazine = magazineAndFrameAddr & 0x7;
+    if (magazine == 0) {
+      magazine = 8;
+    }
+    int packetNumber = (magazineAndFrameAddr >> 3) & 0x1F;
+
+    len -= 2;
+
+    if (magazine != magazineNumber) {
```
icbaker, line 413: please don't use textually distinct but semantically identical identifiers like this, it's very hard for a reader to follow what the difference between `magazine` (which is a number) and `magazineNumber` is.

In this case it's probably clearer to use `this` to disambiguate.

You could also rename the intance field to `selectedMagazine` or similar? Since that seems to be how it operates (we only keep packets with the same magazine number)

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      return null;
+    }
+
+    if (packetNumber > 25) {
+      // ETSI EN 300 706 - 7.1.4 - Packet types: non-displayable
+      return null;
+    }
+
+    int rowOffset = 0;
```
icbaker, line 422: looks like this is never updated from 0?

Maybe just in-line it below on L467 with a `/* parameterName= */` comment? Or remove the parameter completely since `updateLine` is only called from one place anyway

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      page.isSubtitle = (s4AndC5C6 & 0x8) != 0;
+      int c7C10 = hamming84Decode(data.readUnsignedByte());
+      page.suppressHeader = (c7C10 & 0x1) != 0;
+      page.inhibitDisplay = (c7C10 & 0x8) != 0;
+      int c11C14 = hamming84Decode(data.readUnsignedByte());
+      page.charset = CHARSET_INDEXES[c11C14 >> 1];
+      len -= 8;
+    }
+
+    if (page.pageNumber != pageNumber) {
```
icbaker, line 448: imo this would also be clearer if the field was called `selectedPageNumber` - does that work?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      // ETSI EN 300 706 - 7.1.4 - Packet types: non-displayable
+      return null;
+    }
+
+    int rowOffset = 0;
+
+    if (packetNumber == 0) {
+      // ETSI EN 300 706 - 7.1.4 - Packet types: Page header
+      int pageNumber = hamming84Decode(data.readUnsignedByte(), data.readUnsignedByte());
+      if (pageNumber == 0xFF) {
```
icbaker, line 427: I'm not sure I understand why this implies we need to return `null` here?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+        return null;
+      }
+      page.pageNumber = pageNumber;
```
icbaker, line 430: Not sure this if this is possible, just a thought:

It seems we don't really ever need/expect `this.page.pageNumber != this.pageNumber` - so what about checking `this.pageNumber == pageNumber` here, and returning early if they're different - maybe you could then completely remove the `Page.pageNumber` field? 

What about exiting early here so

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+
+      // ETSI EN 300 706 - 9.3.1 - Page header
+      data.skipBytes(1);
```
icbaker, line 433: nit: can you add a comment here that we're skipping 'Subcode S1' - just to help a reader reacquaint themselves.

```java
data.skipBytes(1);  // Subcode S1
```

Same below

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      int s2AndC4 = hamming84Decode(data.readUnsignedByte());
+      page.erase = (s2AndC4 & 0x08) != 0;
```
icbaker, line 435: Table 2 describes C4 as being in location 'Byte 9, bit 8'. `& 0x08` would seem to select either bit 3 or 4 (if counting from LSB, 0 or 1 indexed), or bit 4 or 5 (if counting from MSB, 0 or 1 indexed)

Have I misunderstood what's happening here?

EDIT: Ah, maybe the spec is referring to bit positions before the Hamming protection is removed?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      data.skipBytes(1);
+      int s4AndC5C6 = hamming84Decode(data.readUnsignedByte());
+      page.isNewsflash = (s4AndC5C6 & 0x4) != 0;
+      page.isSubtitle = (s4AndC5C6 & 0x8) != 0;
+      int c7C10 = hamming84Decode(data.readUnsignedByte());
+      page.suppressHeader = (c7C10 & 0x1) != 0;
+      page.inhibitDisplay = (c7C10 & 0x8) != 0;
+      int c11C14 = hamming84Decode(data.readUnsignedByte());
+      page.charset = CHARSET_INDEXES[c11C14 >> 1];
+      len -= 8;
```
icbaker, line 445: it looks like you're duplicating the position/limit tracking that `ParsableByteArray` already provides. What about just using the `position` and `bytesLeft` from there?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    if (page.pageNumber != pageNumber) {
+      return null;
+    }
+
+    if (packetNumber >= 1 && packetNumber <= 24 && page.inhibitDisplay) {
```
icbaker, line 452: nit: move the `page.inhibitDisplay` check to the front, i think it reads more naturally

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      page.erase = true;
```
icbaker, line 453: This doesn't quite match the definition of Inhibit Display in Table 2:
> Data addressed to rows 1 to 24 is not to be displayed.

It doesn't say anything about erasing existing data - I would assume we could just return here before decoding these packets instead?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      }
+    }
+
+    handleSpecialChars(data.getData(), data.getPosition(), len, page.charset);
+    page.updateLine(packetNumber, data.getData(), data.getPosition(), rowOffset, len);
+    return updatePage(page, packetNumber);
+  }
+
+  private TeletextLine updatePage(@NonNull TeletextPage page, int row) {
+    TeletextLine dr = new TeletextLine(row, page.erase, page.lines[row]);
```
icbaker, line 472: I think it's a little confusing to have mutable array references shared between `TeletextLine` and `TeletextPage` like this. Is it necessary?

I wonder if we need the `TeletextLine` type at all, and we could just encode everything in `TeletextPage`?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
icbaker, line 327: It's not obvious to me that we want one `Cue` per line

What about encoding multiple lines into a `Cue` using `\n` to separate the lines?

This is what we do in `Cea608Decoder`:
https://github.com/google/ExoPlayer/blob/r2.18.1/library/extractor/src/main/java/com/google/android/exoplayer2/text/cea/Cea608Decoder.java#L974

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+
+  private Cue buildCue(TeletextLine line) {
+    Cue.Builder cueBuilder = new Cue.Builder();
+    byte[] printable = new byte[TELETEXT_COLUMNS];
+    int color = TELETEXT_COLORS[7];
+    float size = (float) CHAR_HEIGHT / TELETEXT_HEIGHT;
+    boolean box = false;
+    boolean newBox = false;
+
+    for (int i = 0; i < line.line.length; ++i) {
```
icbaker, line 335: can we rename something to get rid of this `line.line`? It's a bit confusing to think of a line 'having' a line inside it.

(this may be moot if you get rid of `TeletextLine` as floated below)

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      int c = line.line[i] & 0xFF;
+      if (c < 0x20) {
```
icbaker, line 337: At this point I was lost in the spec again - can you add another signpost comment?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+        if (c > 0 && c < 8) {
+          color = TELETEXT_COLORS[c];
+        } else if (c == 0xa) {
+          newBox = false;
+        } else if (c == 0xb) {
+          newBox = true;
+        } else if (c == 0xd) {
+          size = (float) CHAR_HEIGHT / TELETEXT_HEIGHT * 2;
```
icbaker, line 345: I think size is either equal to `CHAR_HEIGHT / TELETEXT_HEIGHT` or half that (or is it double? not sure about operator precedence between `/` and `*`), and `CHAR_HEIGHT` and `TELETEXT_HEIGHT` are otherwise unused.

What about removing the `CHAR_HEIGHT` and `TELETEXT_HEIGHT` constants and just having a `float TEXT_SIZE = 21f / 576f` constant that you can divide/multiply by 2 if needed?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    String latin1Line;
+    try {
+      String charsetName;
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+        charsetName = StandardCharsets.ISO_8859_1.name();
+      } else {
+        charsetName = "ISO-8859-1";
+      }
```
icbaker, line 364: You can use Guava [`Charsets`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Charsets.html) on any build

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      latin1Line = new String(printable, 0, printable.length, charsetName);
+    } catch (UnsupportedEncodingException e) {
+      latin1Line = "";
+    }
+    SpannableStringBuilder stringBuilder = new SpannableStringBuilder(latin1Line);
```
icbaker, line 369: nit: no need to use `SpannableStringBuilder` (which lets you modify both the string content and the spans) - [`SpannableString`](https://developer.android.com/reference/android/text/SpannableString) should be fine.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+      }
+
+      if (!box) {
+        c = ' ';
+      }
+      printable[i] = (byte) c;
+      box = newBox;
+    }
+
+    String latin1Line;
```
icbaker, line 357: nit: no need to name this `latin1Line`, can't it just be `line` or `cueText`?

You could even in-line it into the `SpannableString(Builder)` constructor on L369 imo to avoid having to name it at all.

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+        charsetName = "ISO-8859-1";
+      }
+      latin1Line = new String(printable, 0, printable.length, charsetName);
+    } catch (UnsupportedEncodingException e) {
+      latin1Line = "";
+    }
+    SpannableStringBuilder stringBuilder = new SpannableStringBuilder(latin1Line);
+    stringBuilder.setSpan(new ForegroundColorSpan(color), 0, latin1Line.length(),
+        Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+    stringBuilder.setSpan(new BackgroundColorSpan(0x00000000), 0, latin1Line.length(),
```
icbaker, line 372: It looks like this adds a transparent background - does that have any effect?

library/extractor/src/main/java/com/google/android/exoplayer2/text/dvb/DvbTeletextSubtitleParser.java
```diff
@@ -0,0 +1,511 @@
...
+  private static final int[] TELETEXT_COLORS = {
+      0xff000000, // Alpha Black
+      0xffff0000, // Alpha Red
+      0xff00ff00, // Alpha Green
+      0xffffff00, // Alpha Yellow
+      0xff0000ff, // Alpha Blue
+      0xffff00ff, // Alpha Magenta
+      0xff00ffff, // Alpha Cyan
+      0xffffffff, // Alpha White
+  };
```
icbaker, line 39: For lots of these you can use [Color](https://developer.android.com/reference/android/graphics/Color) constants which will be self-documenting and you can remove the comments.

icbaker (changes requested):

zolsen23:
Hi @cdongieux, I'm really looking forward to this improvement :) I did some testing on my side with a MPEG-TS file and I can see that the subtitles appear for a few seconds and then disappear completely. I have attached a video of what I see. [https://www.dropbox.com/s/6rx12iifbms6kq7/device-2022-11-29-092900.mp4?dl=0](url)

This is the video file details:
`General
ID                                       : 1 (0x1)
Complete name                            : C:\Users\zolsen\Videos\239.1.1.16.ts
Format                                   : MPEG-TS
File size                                : 2.47 GiB
Duration                                 : 1 h 59 min
Overall bit rate mode                    : Variable
Overall bit rate                         : 2 951 kb/s

Video
ID                                       : 161 (0xA1)
Menu ID                                  : 785 (0x311)
Format                                   : MPEG Video
Format version                           : Version 2
Format profile                           : Main@Main
Format settings                          : CustomMatrix / BVOP
Format settings, BVOP                    : Yes
Format settings, Matrix                  : Custom
Format settings, GOP                     : Variable
Codec ID                                 : 2
Duration                                 : 1 h 59 min
Bit rate mode                            : Variable
Bit rate                                 : 2 288 kb/s
Maximum bit rate                         : 2 332 kb/s
Width                                    : 720 pixels
Height                                   : 576 pixels
Display aspect ratio                     : 16:9
Active Format Description                : Full frame 16:9 image
Frame rate                               : 25.000 FPS
Standard                                 : PAL
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Progressive
Scan order                               : Top Field First
Compression mode                         : Lossy
Bits/(Pixel*Frame)                       : 0.221
Time code of first frame                 : 12:55:39:22
Time code source                         : Group of pictures header
GOP, Open/Closed                         : Open
Stream size                              : 1.92 GiB (78%)

Audio #1
ID                                       : 81 (0x51)
Menu ID                                  : 785 (0x311)
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 2
Codec ID                                 : 3
Duration                                 : 1 h 59 min
Bit rate mode                            : Constant
Bit rate                                 : 256 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 48.0 kHz
Frame rate                               : 41.667 FPS (1152 SPF)
Compression mode                         : Lossy
Delay relative to video                  : -488 ms
Stream size                              : 220 MiB (9%)
Language                                 : English

Audio #2
ID                                       : 82 (0x52)
Menu ID                                  : 785 (0x311)
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 2
Codec ID                                 : 3
Duration                                 : 1 h 59 min
Bit rate mode                            : Constant
Bit rate                                 : 256 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 48.0 kHz
Frame rate                               : 41.667 FPS (1152 SPF)
Compression mode                         : Lossy
Delay relative to video                  : -488 ms
Stream size                              : 220 MiB (9%)
Language                                 : English
Language, more info                      : Visual impaired commentary
Editorial classification                 : Visual impaired commentary
Mix type                                 : Independent

Text
ID                                       : 41 (0x29)-801
Menu ID                                  : 785 (0x311)
Format                                   : Teletext Subtitle
Language                                 : English`

wazerstar:
@cdongieux You got an update on this code?

cdongieux:
@wazerstar sorry no, I didn't have time to check on this.

wazerstar:
@google-admin @icbaker @cdongieux can we get this ticket migrated to https://github.com/androidx/media

icbaker:

PR #10596, opened 2022-09-08
check all renderer mapped TrackGroups

yuzhidi:
as MappingTrackSelectorTest#selectTracks_multipleVideoAndAudioTracks_mappedToSameRenderer defined 4 renderers, 2 video renderers two audio renderers, for better, check all renderer mapped TrackGroups


icbaker:
Please propose this change against the `dev-v2` branch as described in https://github.com/google/ExoPlayer/blob/release-v2/CONTRIBUTING.md#pull-requests

Could you add a bit detail about why this needed? The test sets up four tracks, and it asserts that they're all mapped once - so assuming each track can only be mapped once, there's no need to check they're **not** mapped to the other two renderers. Or are you suggesting we should test that 'each track can only be mapped once' assumption?

Additionally: It's a bit confusing to read the method calls because you're passing an empty var-args, I wonder if explicitly passing an empty array would be more clear.

Issue #9553, opened 2021-10-10
Improve stuck buffering detection to explicitly detect that no buffer is available for rendering


vovan888:
When playing a specific HLS VOD stream the playback always stalls (video freezes, audio stops, CC stops) at the same time and at the same frame (around the 8 minutes from the beginning). Pausing/unpausing does not help. Seeking helps to resume playback.

I tried to debug the issue after playback stops: LoadControl reports that the media buffer is full (at 50seconds). But both Video and Audio renderers report that they are not ready in the following line:

ExoPlayerImplInternal.java:
```
        boolean allowsPlayback =
            isReadingAhead || isWaitingForNextStream || renderer.isReady() || renderer.isEnded();

```

VLC and ffplay play this stream without issues.

Unfortunately I cannot provide a link to the stream. But it looks like the following:
```
#EXTM3U
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-VERSION:3
#EXT-X-START:TIME-OFFSET=0
#EXT-X-TARGETDURATION:17
#EXT-X-MEDIA-SEQUENCE:11048
#EXT-X-KEY:METHOD=AES-128,URI="[edit]"
#EXTINF:8.341667,
11048.ts
.......
#EXTINF:8.341667,
11090.ts
#EXTINF:16.683333,
11091.ts
#EXTINF:16.683333,
11092.ts
#EXTINF:8.341667,
11093.ts

```
Logcat output:
```

2021-10-10 22:21:26.503 28898-28898/com.player D/EventLogger: downstreamFormat [eventTime=6.90, mediaPos=0.00, window=0, period=0, id=0, mimeType=null, bitrate=3500000, codecs=mp4a.40.2, avc1.4d401e, res=1280x720]
2021-10-10 22:21:26.913 28898-28898/com.player D/EventLogger: videoDecoderInitialized [eventTime=7.29, mediaPos=0.00, window=0, period=0, OMX.rk.video_decoder.hevc]
2021-10-10 22:21:26.934 28898-28898/com.player D/EventLogger: videoInputFormat [eventTime=7.33, mediaPos=0.00, window=0, period=0, id=0, mimeType=video/hevc, bitrate=3500000, codecs=hvc1.1.2.L93, res=1280x720]
2021-10-10 22:21:27.006 28898-28898/com.player D/EventLogger: audioDecoderInitialized [eventTime=7.41, mediaPos=0.00, window=0, period=0, OMX.google.aac.decoder]
[...]

PR #10613, opened 2022-09-13
Tunneling mode Audio stall detection logic

sneelavara:
This is the pull request for bug #10615
In tunneling mode, when audio stalls the playback is not failing. Video freezes or start playing with very slow rate. In this pull request we are attempting to fix that by tweaking isReady() in MediaCodecRenderer. The MediaCodecAudioRenderer reports not ready when the decoder FIFO is empty. This code is active only in Tunneling mode.

Test stream to reproduce was sent to exoplayer dev email with "**Missing Audio Segments**"
In this test stream audio is missing after 100th segment. Each segment is 6 second long. Without this fix, playback freezes after 10 minutes. With this fix, playback error is seen after 10 minutes.

To reproduce the issue, build the demo App with tunneling mode enabled. i.e. change PlayerActivity.java -
```
//trackSelectionParameters = new TrackSelectionParameters.Builder(/* context= */ this).build();
      trackSelectionParameters = new DefaultTrackSelector.ParametersBuilder(/* context= */ this).setTunnelingEnabled(true).build();
```

There is another stream we generated to evaluate video stall detection. Link to this test stream was sent to exoplayer dev email with "**Missing Video Segments**".  In this stream video is missing after 100th segment. Playback failure is reported after 10 minutes for this stream (expected result). So, Video stall detection seem to work fine.

In non-tunneling mode, audio and video stalls are detected and playback fails after 10 minutes for the above 2 test streams (expected result). 

Please suggest if there is a better way to handle this situation. Thank you.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java
@@ -109,6 +109,9 @@ public class MediaCodecAudioRenderer extends MediaCodecRenderer implements Media
 
   @Nullable private WakeupListener wakeupListener;
 
+  private long lastOutputTimeUs;
+  private boolean tunneling = false;
+
   /**
    * @param context A context.
    * @param mediaCodecSelector A decoder selector.
@@ -252,6 +255,7 @@ public class MediaCodecAudioRenderer extends MediaCodecRenderer implements Media
     context = context.getApplicationContext();
     this.context = context;
     this.audioSink = audioSink;
+    lastOutputTimeUs = C.TIME_UNSET;
     eventDispatcher = new EventDispatcher(eventHandler, eventListener);
     audioSink.setListener(new AudioSinkListener());
   }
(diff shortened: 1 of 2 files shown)


tonihei:
I'm trying to clarify my understanding of the issue:
  - Playback never fails because it's stuck in a "ready" state.
  - The audio renderer is ready because `audioSink.hasPendingData() || super.isReady()` returns true in the error state. Is it the `AudioSink` that has pending data, or is the `super.isReady()` call returning true? Since you removed `audioSink.hasPendingData()` I assume this method returned true? 
  - You replaced `hasOutputBuffer()` (which just checks if there is a pending fully processed output buffer) with a more elaborate check that basically tries to see if there are buffers "in" the codec (currently being processed). This is presumably meant to replace the removed `audioSink.hasPendingData()` check.

Could you verify my understanding is correct? I'm also not really sure whether this change makes sense as such, since we would still want to check if the `AudioSink` has pending buffers.

sneelavara:
Hi @tonihei, I could not get back earlier, since I was was on vacation. Thank you for taking a look at this. 

- Correct. Playback never fails. Player state remains "ready"
- Your assumption is correct. The `audioSink.hasPendingData` returns true. 
- Yes, in `hasOutputReady()` I am checking for decode FIFO level to workaround removed `audioSink.hasPendingData()`

This code is active only in tunneling mode.

tonihei:
Thanks for confirming. As I pointed out above though, I think the fix may not be appropriate since we still need to check if `audioSink.hasPendingData` in the general case. I don't have a good suggestion of how to fix this while maintaining the existing functionality.  Given the file is essentially broken, maybe it's not worth finding a fix for this? 

We have a more generic feature request to detect stuck rendering and this should probably cover this case as well (See #9553). 

I'll close this PR based on this reasoning. If you have a good idea of how this solve this without breaking existing logic for normal playback and without a huge amount of code, happy to continue the discussion though :)

PR #10618, opened 2022-09-14
[Issue #10685] Do not send content complete if any midroll is skipped

vishnuchilakala:
Let Google IMA SDK handle the event if the user seeks to the very end of the content and skips any midrolls.

This fix will ensure that Exoplayer will behave the same way as [the Google IMA sample application.](https://github.com/googleads/googleads-ima-android/) i.e. Play the last of the skipped midrolls before sending content complete / playing post roll.

### DIFF
--- a/extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/AdTagLoader.java
+++ b/extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/AdTagLoader.java
@@ -1122,12 +1122,17 @@ import java.util.Map;
   }
 
   private void ensureSentContentCompleteIfAtEndOfStream() {
-    if (!sentContentComplete
-        && contentDurationMs != C.TIME_UNSET
-        && pendingContentPositionMs == C.TIME_UNSET
-        && getContentPeriodPositionMs(checkNotNull(player), timeline, period)
-                + THRESHOLD_END_OF_CONTENT_MS
-            >= contentDurationMs) {
+
+    if (sentContentComplete
+        || contentDurationMs == C.TIME_UNSET
+        || pendingContentPositionMs != C.TIME_UNSET) {
+      return;
+    }
+
+    long contentPeriodPositionMs = getContentPeriodPositionMs(checkNotNull(player), timeline, period);
+    int pendingAdGroupIndex = adPlaybackState.getAdGroupIndexAfterPositionUs(contentPeriodPositionMs, contentDurationMs);
+    if (!adPlaybackState.getAdGroup(pendingAdGroupIndex).shouldPlayAdGroup()
+        && contentPeriodPositionMs + THRESHOLD_END_OF_CONTENT_MS >= contentDurationMs) {
       sendContentComplete();
     }
   }


vishnuchilakala:
@marcbaechinger Can you please take a look and let me know what you think? Thanks.

**To summarise the issue:**
Both the IMA Sample application and Exoplayer play the last of the skipped mid-rolls if the user seeks to any point in the duration of the playback. But if the user seeks directly to the very end of the duration of the playback, Exoplayer immediately is triggering the content complete event which can stop the playback of the last of the skipped mid-rolls if Google IMA wants to play it (Which is what is happening on the IMA Sample application).

Thus, checking if the user skipped over any mid-rolls will let IMA play the last mid-roll, post which the content complete signal will be sent.

tonihei:
Thanks for sending a fix for this problem. Your current fix only checks if a midroll exists though and will never allow to send the content complete flag. Could you change it to only prevent sending content complete if the last midroll group hasn't been fully played yet (`AdPlaybackState.shouldPlayAdGroup()`)?

vishnuchilakala:
Thanks for the review @tonihei ! Made the change you suggested, please take a look again now.

extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/AdTagLoader.java
```diff
@@ -1122,9 +1122,15 @@ private void handleAdPrepareError(int adGroupIndex, int adIndexInAdGroup, Except
   }
 
   private void ensureSentContentCompleteIfAtEndOfStream() {
+    if (player == null) {
+      return;
+    }
+
+    int pendingAdGroupIndex = adPlaybackState.getAdGroupIndexAfterPositionUs(player.getContentPosition(), player.getContentDuration());
```
tonihei: Instead of player.getContentPosition and player.getContentDuration, you need to use the values already used below (`getContentPeriodPositionMs(...)` and `contentDurationMs`).  

To make this easier, maybe reorder the method slightly:
 1. Check for the simple conditions first and return from this method if they don't apply. 
 2. Get the content position using the helper method `getContentPeriodPositionMs`.
 3. Check the more complicated conditions to see if we need to trigger `sendContentComplete`.

extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/AdTagLoader.java
```diff
@@ -1122,9 +1122,15 @@ private void handleAdPrepareError(int adGroupIndex, int adIndexInAdGroup, Except
   }
 
   private void ensureSentContentCompleteIfAtEndOfStream() {
+    if (player == null) {
```
tonihei: The `checkNotNull(player)` below suggests this condition isn't needed because the player should always be non-null here.

vishnuchilakala:
@tonihei Worked on the comments and updated the PR.

extensions/ima/src/main/java/com/google/android/exoplayer2/ext/ima/AdTagLoader.java
```diff
@@ -1122,12 +1122,17 @@ private void handleAdPrepareError(int adGroupIndex, int adIndexInAdGroup, Except
...
-            >= contentDurationMs) {
+
+    if (sentContentComplete
+        || contentDurationMs == C.TIME_UNSET
+        || pendingContentPositionMs != C.TIME_UNSET) {
+      return;
+    }
+
+    long contentPeriodPositionMs = getContentPeriodPositionMs(checkNotNull(player), timeline, period);
+    int pendingAdGroupIndex = adPlaybackState.getAdGroupIndexAfterPositionUs(contentPeriodPositionMs, contentDurationMs);
```
tonihei, line 1133: Should this be getAdGroupIndexForPositionUs? I assume there is no ad group following in the case where the user seeks to the end of the content. 

Otherwise, looks great! Thanks for making all these additional fixes :)

vishnuchilakala: I thought so too, but `getAdGroupIndexForPositionUs` returns a -1 for some reason.

I did test Midroll and Midroll + Postroll using this code and it works as expected.

tonihei: That's because the position you pass into the method is in milliseconds but the method expects microseconds. I'll make this update when merging the change internally. Thanks!

vishnuchilakala: Oooh, missed that. Thanks for pointing it out. Just tested it, that fixes the issue.
I can update the PR again using the `Util.msToUs` method, should I? Or would you suggest any other change to the `getAdGroupIndexForPositionUs` method itself?

tonihei: No worries about that. I already imported the PR to our internal code and will make the update there. You should see the merge here (including my additional changes) once we sync back to GitHub.

vishnuchilakala: Great. Happy to contribute :)

Issue #10684, opened 2022-10-11
Android 12L - AbstractMethodError calling onSpatializerAvailableChanged


asos-edgeorge:
### ExoPlayer Version

2.18.0

### Devices that reproduce the issue

All devices that have this issue are Samsung devices within the Galaxy Z Fold/Flip series, running Android 12 (presumably 12L)

For example, the three most popular devices experiencing this issue are:

* Galaxy Z Flip3 5G
* Galaxy Z Flip4
* Galaxy Z Fold3 5G

### Devices that do not reproduce the issue

Non-Galaxy Z Flip/Fold series devices

### Reproducible in the demo app?

Not tested

### Reproduction steps

We have seen this issue in the wild and unfortunately do not have access to the affected devices to find steps to reproduce this crash.

Based on the crash reports received, it seems 86% of crashes occur when the app is in the background. Due to the nature of foldable devices, this could be perhaps linked to a form config change (i.e. folding/unfolding) or something similar.

Should it help, our player is defined as follows (with some minor code omitted)

```kotlin
// Cache
val cache = SimpleCache(createVideoCacheDir(context),
                LeastRecentlyUsedCacheEvictor(VIDEO_CACHE_SIZE.toLong()),
                ExoDatabaseProvider(context))

// CacheDataSource
val upstreamFactory: DataSource.Factory = OkHttpDataSourceFactory({ request -> okHttp.newCall(request) }, USER_AGENT)
val cacheDataSource = CacheDataSource.Factory()
                .setCache(exoCache)
                .setUpstreamDataSourceFactory(upstreamFactory)
                .setFlags(CacheDataSource.FLAG_IGNORE_CACHE_ON_ERROR)

// ExoPlayer
ExoPlayer.Builder(context)
                .setLooper(Looper.getMainLooper())
                .setMediaSourceFactory(DefaultMediaSourceFactory(cacheDataSource))
                .setTrackSelector(DefaultTrackSelector(context, AdaptiveTrackSelection.Factory()))
                .build()
```

### Expected result

The media plays successfully and backgrounding the app works as expected

### Actual result

Exoplayer crashes with the following stacktrace 

```
Fatal Exception: java.lang.AbstractMethodError: abstract method "void android.media.Spatializer$OnSpatializerStateChangedListener.onSpatializerAvailableChanged(android.media.Spatializer, boolean)"
[...]

PR #10709, opened 2022-10-20
Bugfix

XiaoLongTX:
Hello, I found a playback crash caused by a compilation problem.
My app uses compileSdkVersion of 30, but Exoplayer uses 31. I compile the release version of the apk, when minifyEnabled is set to true, the app plays video on Android 12 mobile phones, and when using 5g, there will be java.lang.AbstractMethodError: abstract method "void android.telephony .TelephonyCallback$DisplayInfoListener.onDisplayInfoChanged(android.telephony.TelephonyDisplayInfo)"
crash.
The reason is that when compileSdkVersion 30 is used to compile the release version apk, the compilation environment cannot find the class android.telephony.TelephonyCallback$DisplayInfoListener, and there is no other code calling related logic, so it is deleted or inline optimized. So the logic of keep must be added here to ensure that the onDisplayInfoChanged method is not deleted. I tried a lot of methods, and finally found that -keepattributes LineNumberTable must be added at the same time to make the following keep logic take effect.
As for the keep of DefaultTrackSelector$SpatializerWrapperV32, it is the same reason.


icbaker:
The recommended fix for this problem is to increase your `compileSdkVersion` as described here: https://github.com/google/ExoPlayer/issues/10684#issuecomment-1286564259

I'm going to close this as we don't think playing proguard config whack-a-mole is the right way to fix this.

We're looking into changing the library metadata so that any app depending on it is forced to upgrade their `compileSdkVersion` (i.e. the error happens at compile time, rather than runtime).

Issue #10728, opened 2022-10-28
Network type in DefaultBandwidthMeter is always unknown on first playback


vishnuchilakala:
### ExoPlayer Version

2.18.1

### Devices that reproduce the issue

All devices

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Create a DefaultBandwidthMeter and override the initial bandwidth estimate for your network type. Since I am working on a WiFi network, I have overridden the estimate as below. 
2. Hook it to the player in `PlayerActivity.java` using Exoplayer.Builder.
```kotlin
DefaultBandwidthMeter defaultBandwidthMeter = new DefaultBandwidthMeter.Builder(getApplicationContext())
          .setInitialBitrateEstimate(C.NETWORK_TYPE_WIFI, 100L)
          .build();
// PlayerActivity.java - Line 273
ExoPlayer.Builder playerBuilder = new ExoPlayer.Builder(/* context= */ this)
              .setMediaSourceFactory(createMediaSourceFactory())
              .setBandwidthMeter(defaultBandwidthMeter);
```
3. Check the bitrate using `BandwidthMeter.getBitrateEstimate`.

### Expected result

The bitrate estimate should be overridden by the value I provided i.e. 100L.

### Actual result

The bitrate estimate always is 100000L i.e. the default value for unknown network type on the first playback attempt. On the subsequent playbacks, the bandwidth value is as expected.

### Media

Any media would work.

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

vishnuchilakala:
PR raised to fix this issue: https://github.com/google/ExoPlayer/pull/10729

tonihei:
> The bitrate estimate always is 100000L i.e. the default value for unknown network type on the first playback attempt. 

How did you test this? I just checked in our demo with the default config and with an explicit override for the initial estimate and it all worked as expected. 

> Check the bitrate using BandwidthMeter.getBitrateEstimate

Note that the network type is determined asynchronously (it requires a single message sent to the main handler to be precise) and if you immediately query the bitrate estimate in the method that creates the BandwidthMeter, it is still on the unknown network type. This doesn't effect playback though as the network init happens instantly within one message sent to the main thread and starting playback requires much more preparation than that. 


I'll close this issue assuming everything is working as intended.

PR #10729, opened 2022-10-28
[Issue #10728] Set network type initial value using context

vishnuchilakala:
The network type is initialised to `C.NETWORK_TYPE_UNKNOWN`, which causes any overrides in the `DefaultBandwidthMeter` to not work for the first playback. Since the network observer is a static entity, it gets reset to the actual value for the second playback and would not be a problem thereafter. However, the network type will always be `NETWORK_TYPE_UNKNOWN` for the first playback after creating the context.

This can be fixed by initialising the network type using the context and not default it to `C.NETWORK_TYPE_UNKNOWN`.

(Reproduction steps and code snippet provided in the issue)

### DIFF
--- a/library/common/src/main/java/com/google/android/exoplayer2/util/NetworkTypeObserver.java
+++ b/library/common/src/main/java/com/google/android/exoplayer2/util/NetworkTypeObserver.java
@@ -90,7 +90,7 @@ public final class NetworkTypeObserver {
     mainHandler = new Handler(Looper.getMainLooper());
     listeners = new CopyOnWriteArrayList<>();
     networkTypeLock = new Object();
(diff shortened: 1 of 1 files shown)


rohitjoins:
Thanks for sending a fix for this problem. Can you please change the PR to merge into `google:dev-v2` branch instead of `google:release-v2`? We only push changes to release branch when doing a release.

tonihei:
See discussion in #10728. I think this isn't fixing a real bug and we can't assign a potentially wrong network type initially as it would cause problems for tools (e.g. analytics systems) listening to network type changes if the network type is reported wrongly in the beginning.

Issue #1419, opened 2016-04-06
.srt encoding problem


AbdelrahmanSalem:
I am trying to render the following .srt file which is in turkish language 
https://drive.google.com/a/microdoers.com/file/d/0B4Rir8tNZWg3ZDlTN1FrT2Jfc1k/view?usp=sharing
 the srt file opens ok on any editor on my machine but the TextTrackRenderer that is setup by the following lines

````
        DataSource textDataSource = new DefaultUriDataSource(context, bandwidthMeter, userAgent);

        MediaFormat mediaFormat = MediaFormat.createTextFormat("0", MimeTypes.APPLICATION_SUBRIP, MediaFormat.NO_VALUE, C.MATCH_LONGEST_US, null);

        SingleSampleSource textSampleSource = new SingleSampleSource(Uri.parse(subUri), textDataSource, mediaFormat);

        TrackRenderer subTitlerenderer = new TextTrackRenderer(textSampleSource, player, player.getMainHandler().getLooper());

        renderers[EMExoPlayer.RENDER_TEXT] = subTitlerenderer;
````

does not display characters like ý instead displays �
I have tried to get bytes in UTF-8 and create a new string but I think data is already lost 
So I guess it is an encoding problem with the TextTrackRenderer.

AquilesCanta:
Please send the file to `dev.exoplayer@gmail.com`(access permission was not granted to the given resource) and post back so that we go look for it.

AquilesCanta:
Same as [#1362](https://github.com/google/ExoPlayer/issues/1362). The issue here is that exoplayer only supports UTF-8 right now. I think you have two possible courses of action:

1. Convert your file to UTF-8 and serve the resulting file.
2. Programmatically do the conversion before feeding the file to the exoplayer parser.

Issue #1362, opened 2016-03-11
Support UTF-16 srt files


anphunl:
I tried to add an external srt file to play.

My code : 

> MediaFormat mediaFormat = MediaFormat.createTextFormat("0", MimeTypes.APPLICATION_SUBRIP, MediaFormat.NO_VALUE, C.MATCH_LONGEST_US, null);
      Uri uri = Uri.parse("http://s.vn-hd.com:8080/store_06_2013/15062013/Rio_2011_1080p_2D_Bluray_DTS_x264_DON/Rio_2011_1080p_2D_Bluray_DTS_x264_DON_VIE.srt");
      DataSource textDataSource = new DefaultUriDataSource(context, bandwidthMeter, userAgent);
      SingleSampleSource textSampleSource = new SingleSampleSource(uri, textDataSource, mediaFormat);
      TrackRenderer textRenderer = new TextTrackRenderer(textSampleSource, player, player.getMainHandler().getLooper());
//...
renderers[DemoPlayer.TYPE_TEXT] = textRenderer;

It didn't show the subtitle, but returns errors in the console: Skipping invalid index: ��1��

ojw28:
The file looks like it's UTF-16 encoded to me. We only support UTF-8. For what it's worth, several text editors on my machine are also unable to open this file (I could open it in emacs though). I converted it to UTF-8 and it worked fine.

anphunl:
@ojw28 : Thank you for your note. How should we handle this situation? Do we have any way to detect its encoding?

ojw28:
Someone could add UTF-16 srt support to ExoPlayer. It's unclear whether it's really worthwhile though. It's not high priority for us, in any case.

anphunl:
@ojw28 : I use juniversalchardet to detect charset encoding of the stream, then use Scanner to parse SRT.

> compile 'com.googlecode.juniversalchardet:juniversalchardet:1.0.3'

Please check my file [https://github.com/anphunl/ExoPlayer/blob/master/library/src/main/java/com/google/android/exoplayer/text/subrip/SubripParser.java](Here).

I think we can merge it to the dev branch.

ojw28:
We cannot merge things that take dependencies such as `org.mozilla.universalchardet`

ojw28:
[...]

PR #10750, opened 2022-11-05
UTF-16 support for .srt files

Stronger197:
Issues:
#1362
#1419

The issue here is that exoplayer only supports UTF-8 for .srf files. 
I have added UTF-16 support for files containing BOM otherwise UTF-8 will be used.

More details:
I checking that file contains BOM to detect UTF-16 encoding. It can be little-endian (LE) or big-endian (BE)
If the file doesn't contain any BOM we can't determine whether UTF-16 or UTF-8 is used, so we use UTF-8 by default (no changes here)
For decoding, I use the same logic as in the public String readLine() method, but I take two bytes for each character.


Starmel (approved):

library/common/src/main/java/com/google/android/exoplayer2/util/Util.java
```diff
@@ -682,6 +682,30 @@ public static String fromUtf8Bytes(byte[] bytes, int offset, int length) {
...
 
+  /**
+   * Returns a new {@link String} constructed by decoding UTF-16 (LE) encoded bytes in a subarray.
+   *
+   * @param bytes  The UTF-16 encoded bytes to decode.
+   * @param offset The index of the first byte to decode.
+   * @param length The number of bytes to decode.
+   * @return The string.
+   */
+  public static String fromUtf16LEBytes(byte[] bytes, int offset, int length) {
```
icbaker: I would remove these methods from Util and just in-line them at the call-sites.

I think the UTF-8 equivalents above are already of borderline usefulness tbh (since the in-lined version is pretty clear and not much longer than the method call), and UTF-8 is **much** more commonly used throughout our library than UTF-16, so the argument for having UTF-16 variants is much weaker imo.

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -75,8 +77,25 @@ protected Subtitle decode(byte[] bytes, int length, boolean reset) {
     LongArray cueTimesUs = new LongArray();
     ParsableByteArray subripData = new ParsableByteArray(bytes, length);
 
+    @Nullable Charset utf16Charset;
+    if (bytes.length >= 2) {
```
icbaker: You should use `length` here, not `bytes.length`

From the documentation on `SimpleSubtitleDecoder.decode`:
> `size` - The size of the data to be decoded.

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -75,8 +77,25 @@ protected Subtitle decode(byte[] bytes, int length, boolean reset) {
...
+      utf16Charset = null;
+    }
+
     @Nullable String currentLine;
-    while ((currentLine = subripData.readLine()) != null) {
+    while (true) {
+      if (utf16Charset != null) {
+        currentLine = subripData.readLineUtf16(utf16Charset.equals(Charsets.UTF_16LE));
+      } else {
+        currentLine = subripData.readLine();
```
icbaker: What about creating `ParsableByteArray.readLine(Charset)`?

That would allow you to keep more of the existing structure of this `while` loop.

I think the current proposed code is a little strange that it does UTF-8 'by default' and has a nullable UTF-16 charset to trigger an alternative behaviour - why not just have a non-null `Charset` local?

It's also consistent with the existing `ParsableByteArray.readString(int)` (UTF-8) and `readString(int, Charset)` methods.

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
```diff
@@ -138,6 +169,21 @@ protected Subtitle decode(byte[] bytes, int length, boolean reset) {
     return new SubripSubtitle(cuesArray, cueTimesUsArray);
   }
 
+  @Nullable
+  private Charset getUtf16Charset(byte first, byte second) {
```
icbaker: Consider rejigging this into a 'detectUtfCharset' method which:
* Takes a `ParsableByteArray`
* Consumes the BOM if present, and uses it to distinguish between UTF-16 and UTF-8
* Returns UTF-8 if no BOM is detected (in which case it ensures the read pointer of the `ParsableByteArray` is restored)

icbaker:
Thanks for the contribution (and for including tests :))! The general idea looks good, I've added some suggestions on reshuffling the implementation a bit.

Stronger197:
@icbaker Hi. Do you have any updates?

icbaker (changes requested):
I'm so sorry - I wrote these comments ages ago but didn't hit 'send' on GitHub!

library/common/src/main/java/com/google/android/exoplayer2/util/Util.java
Stronger197: Ok, if you think this will be a better solution.

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
Stronger197: I don't think it's a good solution, and I did it on purpose.

If you have a method that contains Charset in signature. The caller side will assume that any Charset can be passed and it will work well. 
But in this case, we will only support 2 encodings from the entire set that contains Charsets class.
So, in my opinion, I don't think this is a good solution. @icbaker What about you?

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
Stronger197: Thanks, sounds much better.

Stronger197:
@icbaker ok, i'm done with it

Stronger197:
@icbaker ok, i'm done with it.

library/extractor/src/main/java/com/google/android/exoplayer2/text/subrip/SubripDecoder.java
Stronger197: Ok, resolved by creating `ParsableByteArray.readUtfLine(Charset)`

icbaker:
Looks good - thanks for all the changes. I'll work on getting this merged.

icbaker (approved):

Issue #10760, opened 2022-11-08
FFMpeg error with some flac files


Tolriq:
### ExoPlayer Version

2.18.1

### Devices that reproduce the issue

All

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Not tested

### Reproduction steps

Play the file send by mail with the ffmpeg extension enabled.
Skip or wait until near the end of the file and see an error.

### Expected result

No error (And no error with default Android mediacodec)

### Actual result

```
ExoPlayerImplInternal - Playback error
                                                                               com.google.android.exoplayer2.ExoPlaybackException: FfmpegAudioRenderer error, index=0, format=Format(null, null, null, audio/flac, null, -1, null, [-1, -1, -1.0], [2, 48000]), format_supported=YES
                                                                                   at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:566)
                                                                                   at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                                   at android.os.Looper.loopOnce(Looper.java:201)
                                                                                   at android.os.Looper.loop(Looper.java:288)
                                                                                   at android.os.HandlerThread.run(HandlerThread.java:67)
                                                                               Caused by: com.google.android.exoplayer2.ext.ffmpeg.FfmpegDecoderException: Unexpected decode error
                                                                                   at com.google.android.exoplayer2.ext.ffmpeg.FfmpegAudioDecoder.createUnexpectedDecodeException(FfmpegAudioDecoder.java:95)
                                                                                   at com.google.android.exoplayer2.ext.ffmpeg.FfmpegAudioDecoder.createUnexpectedDecodeException(FfmpegAudioDecoder.java:32)
                                                                                   at com.google.android.exoplayer2.decoder.SimpleDecoder.decode(SimpleDecoder.java:245)
                                                                                   at com.google.android.exoplayer2.decoder.SimpleDecoder.run(SimpleDecoder.java:202)
                                                                                   at com.google.android.exoplayer2.decoder.SimpleDecoder.access$000(SimpleDecoder.java:29)
                                                                                   at com.google.android.exoplayer2.decoder.SimpleDecoder$1.run(SimpleDecoder.java:73)
                                                                               Caused by: java.lang.IllegalArgumentException
                                                                                   at java.nio.Buffer.limit(Buffer.java:290)
                                                                                   at java.nio.ByteBuffer.limit(ByteBuffer.java:820)
[...]

PR #10762, opened 2022-11-08
Fix ffmpeg jni wrapper returning invalid result codes

Tolriq:
This ensure that ffmpeg error code are properly translated to values that the ExoPlayer decoder understand. 
The main gain is that it allows the decoder to properly ignore more cases of invalid data and recover.
The second gain is that the other errors are now proper ExoPlayer errors and no more obscure buffer ones.

Fixes: #10760

### DIFF
--- a/extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
+++ b/extensions/ffmpeg/src/main/jni/ffmpeg_jni.cc
@@ -283,7 +283,8 @@ int decodePacket(AVCodecContext *context, AVPacket *packet,
         break;
       }
       logError("avcodec_receive_frame", result);
-      return result;
+      return result == AVERROR_INVALIDDATA ? AUDIO_DECODER_ERROR_INVALID_DATA
+                                           : AUDIO_DECODER_ERROR_OTHER;
     }
 
     // Resample output.
@@ -330,7 +331,8 @@ int decodePacket(AVCodecContext *context, AVPacket *packet,
     av_frame_free(&frame);
     if (result < 0) {
       logError("swr_convert", result);
-      return result;
+      return result == AVERROR_INVALIDDATA ? AUDIO_DECODER_ERROR_INVALID_DATA
+                                           : AUDIO_DECODER_ERROR_OTHER;
     }
     int available = swr_get_out_samples(resampleContext, 0);
     if (available != 0) {


Tolriq:
@rohitjoins @microkatz The version you merged is not functionally equivalent.

https://github.com/google/ExoPlayer/commit/6d2e7a1b5760a05eade18b1a028b3d27f714715c#diff-95449691f939a34d0360a842f3a6770fb12f11063cc5f30d11c5004299a32198R318

Should be `return AUDIO_DECODER_ERROR_INVALID_DATA;` to be equivalent to current (But it might actually be another proper fix no idea)

https://github.com/google/ExoPlayer/commit/6d2e7a1b5760a05eade18b1a028b3d27f714715c#diff-95449691f939a34d0360a842f3a6770fb12f11063cc5f30d11c5004299a32198R337

Should be `return transformError(result);` this one is important to keep as is to avoid the decoder ignoring some case that can't be recovered and would trigger unknown errors later.

rohitjoins:
@Tolriq Based on the documentation for ffmpeg, [`swr_init`](https://ffmpeg.org/doxygen/3.2/group__lswr.html#gae173e8ed91717700471a1dcd06f00f67) returns `AVERROR`, hence it returns `transformError(result)`. For [`swr_convert`](https://ffmpeg.org/doxygen/3.2/group__lswr.html#gaa5bb6cab830146efa8c760fa66ee582a), it only mentions that it can return a negative value and because we are not sure if it can ever return `AVERROR_INVALIDDATA` we simply return  `AUDIO_DECODER_ERROR_INVALID_DATA`.

rohitjoins:
This was merged in [6d2e7a1](https://github.com/google/ExoPlayer/commit/6d2e7a1b5760a05eade18b1a028b3d27f714715c) but unfortunately it didn't mark this PR as closed - closing manually.

Please feel free to discuss any further confusions you have about the PR.

Tolriq:
@rohitjoins But are `swr_convert` errors actually recoverable? `AUDIO_DECODER_ERROR_INVALID_DATA` will make the decoder to just ignore and keep going so potentially just playing nothing for the whole file, that would be a change from current state that could hide issues no? Isn't it safer to return AUDIO_DECODER_ERROR_OTHER ?

In all cases for my bad files the issue was about "avcodec_receive_frame" and the PR properly handle the case, but just wondering.

rohitjoins:
Before this change, we just returned `result` which can be any negative value. Although we only recognise values of -1(AUDIO_DECODER_ERROR_INVALID_DATA) & -2(AUDIO_DECODER_ERROR_OTHER), any other negative value returned from `swr_convert` would have resulted in error because of us not handling it properly. Behaviour change would only occur if `swr_convert` returned -2 on error which I'm not sure about. 

For other function calls related to resampling, like `swr_get_out_samples`, we return `AUDIO_DECODER_ERROR_INVALID_DATA` and hence it made sense to return this same error in case of `swr_convert`. Does this make sense?

Tolriq:
@rohitjoins I'm not ffmpeg expert so can't answer about swr_convert vs swr_get_out_samples but previously any negative value would have stopped the decoder with an error https://github.com/google/ExoPlayer/blob/release-v2/extensions/ffmpeg/src/main/java/com/google/android/exoplayer2/ext/ffmpeg/FfmpegAudioDecoder.java#L139 (related to buffers as the limit can't be negative). A -2 value would have returned a better error but would have errored too. Only a previous -1 would have given the same result as the PR.

Now by returning AUDIO_DECODER_ERROR_INVALID_DATA the error is hidden. 

I'm not saying that it's bad or a wrong decision, just wondering if the change is fully validated. Seems that you are sure so nothing more to add and thanks for the merge.

PR #10776, opened 2022-11-14
Add a method to support clearing all download manager helpers on DownloadService class

dongvanhung:
This method is necessary in case your application supports downloading for multiple users on the same application and the same device.

For example, 
Login with account A, downloads content A, then logout. 
Login with account B,  downloads content A (same as content that downloaded on account A).
After switched to account B, we must to reset the `DownloadManager` to avoid multiple users use the same download folder.
```
DownloadManager?.release()
DownloadManager = null
```
In the case we call `DownloadManager?.release()`, if we do not clear all `downloadManagerHelpers`, `DownloadService` can not be started again.

But now `downloadManagerHelpers` is private and we cannot reset it in inherited subclasses. Therefore, it is necessary to have a method to reset it.

Please review and approve this PR.
Thank you.


dongvanhung:
@marcbaechinger 
Can you please review this ? Thank you !

donny9095:
@marcbaechinger @tonihei 
We want to this method too. Can you please add this method to the next release version ?

devno44:
No one wants to review this ?

icbaker:

PR #10799, opened 2022-11-21
Add support for ID3v2.4 multi-value tags

OxygenCobalt:
#### The problem

My app uses ExoPlayer's metadata parser to work around issues with `MediaStore`'s metadata handling. However, recent developments have required me to start handling multi-value tags, as many users want my app to support multiple artist/genre entries.

While the vorbis extractor supports multiple instances of the tag, the ID3v2 extractor does not, even though ID3v2.4 specifies that text information frames can have multiple values in them if delimited by a null terminator. This worsens the user experience with multi-value tags.

#### What this PR does

This PR deprecates the `value` field in `TextInformationFrame` and replaces it with a new array field called `values`. `Id3Decoder` populates this field with multiple values if it finds multiple tags delimited by null terminators. The now-deprecated `value` field is populated with the first item in `values`, to retain compatibility with the old implementation that would parse the first value only. 

This new algorithm should be backwards compatible with ID3v2.3 and ID3v2.2, as they only leverage null terminators as an informal shortcut to make serializing to C strings easier.

#### Further notes

I've modified other parts of the codebase to use the non-deprecated value to the most backwards-compatible extent possible. I have also added tests into `Id3Decoder` to handle the new functionality.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
@@ -29,10 +29,8 @@ import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
-import org.w3c.dom.Text;
 
 /** Decodes ID3 tags. */
 public final class Id3Decoder extends SimpleMetadataDecoder {
@@ -457,22 +455,22 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     byte[] data = new byte[frameSize - 1];
     id3Data.readBytes(data, 0, frameSize - 1);
 
-    int descriptionEndIndex = indexOfEos(data, 0, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, 0, encoding);
     String description = new String(data, 0, descriptionEndIndex, charset);
-    
+
     // In ID3v2.4, text information frames can contain multiple values delimited by a null
     // terminator. Thus, we after each "end of stream" marker we actually need to keep looking
     // for more data, at least until the index is equal to the data length.
     ArrayList<String> values = new ArrayList<>();
 
     int valueStartIndex = descriptionEndIndex + delimiterLength(encoding);
-    int valueEndIndex = indexOfEos(data, valueStartIndex, encoding);
+    int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     while (valueStartIndex < valueEndIndex) {
       String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
       values.add(value);
 
       valueStartIndex = valueEndIndex + delimiterLength(encoding);
-      valueEndIndex = indexOfEos(data, valueStartIndex, encoding);
+      valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     }
 
     return new TextInformationFrame("TXXX", description, values.toArray(new String[0]));
@@ -498,13 +496,13 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     ArrayList<String> values = new ArrayList<>();
 
     int valueStartIndex = 0;
-    int valueEndIndex = indexOfEos(data, valueStartIndex, encoding);
+    int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     while (valueStartIndex < valueEndIndex) {
       String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
       values.add(value);
 
       valueStartIndex = valueEndIndex + delimiterLength(encoding);
-      valueEndIndex = indexOfEos(data, valueStartIndex, encoding);
+      valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     }
 
     return new TextInformationFrame(id, null, values.toArray(new String[0]));
@@ -524,7 +522,7 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     byte[] data = new byte[frameSize - 1];
     id3Data.readBytes(data, 0, frameSize - 1);
 
-    int descriptionEndIndex = indexOfEos(data, 0, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, 0, encoding);
     String description = new String(data, 0, descriptionEndIndex, charset);
 
     int urlStartIndex = descriptionEndIndex + delimiterLength(encoding);
@@ -571,11 +569,11 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     String mimeType = new String(data, 0, mimeTypeEndIndex, "ISO-8859-1");
 
     int filenameStartIndex = mimeTypeEndIndex + 1;
-    int filenameEndIndex = indexOfEos(data, filenameStartIndex, encoding);
+    int filenameEndIndex = indexOfTerminator(data, filenameStartIndex, encoding);
     String filename = decodeStringIfValid(data, filenameStartIndex, filenameEndIndex, charset);
 
     int descriptionStartIndex = filenameEndIndex + delimiterLength(encoding);
-    int descriptionEndIndex = indexOfEos(data, descriptionStartIndex, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, descriptionStartIndex, encoding);
     String description =
         decodeStringIfValid(data, descriptionStartIndex, descriptionEndIndex, charset);
 
@@ -613,7 +611,7 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     int pictureType = data[mimeTypeEndIndex + 1] & 0xFF;
 
     int descriptionStartIndex = mimeTypeEndIndex + 2;
-    int descriptionEndIndex = indexOfEos(data, descriptionStartIndex, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, descriptionStartIndex, encoding);
     String description =
         new String(
             data, descriptionStartIndex, descriptionEndIndex - descriptionStartIndex, charset);
@@ -642,11 +640,11 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     data = new byte[frameSize - 4];
     id3Data.readBytes(data, 0, frameSize - 4);
 
-    int descriptionEndIndex = indexOfEos(data, 0, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, 0, encoding);
     String description = new String(data, 0, descriptionEndIndex, charset);
 
     int textStartIndex = descriptionEndIndex + delimiterLength(encoding);
-    int textEndIndex = indexOfEos(data, textStartIndex, encoding);
+    int textEndIndex = indexOfTerminator(data, textStartIndex, encoding);
     String text = decodeStringIfValid(data, textStartIndex, textEndIndex, charset);
 
     return new CommentFrame(language, description, text);
@@ -823,7 +821,7 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
         : String.format(Locale.US, "%c%c%c%c", frameId0, frameId1, frameId2, frameId3);
   }
 
-  private static int indexOfEos(byte[] data, int fromIndex, int encoding) {
+  private static int indexOfTerminator(byte[] data, int fromIndex, int encoding) {
     int terminationPos = indexOfZeroByte(data, fromIndex);
 
     // For single byte encoding charsets, we're done.
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
@@ -464,6 +464,10 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     ArrayList<String> values = new ArrayList<>();
 
     int valueStartIndex = descriptionEndIndex + delimiterLength(encoding);
+    if (valueStartIndex >= data.length) {
+      return new TextInformationFrame("TXXX", description, new String[0]);
+    }
+
     int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     while (valueStartIndex < valueEndIndex) {
       String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
@@ -496,6 +500,10 @@ public final class Id3Decoder extends SimpleMetadataDecoder {
     ArrayList<String> values = new ArrayList<>();
 
     int valueStartIndex = 0;
+    if (valueStartIndex >= data.length) {
+      return new TextInformationFrame(id, null, new String[0]);
+    }
+
     int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
     while (valueStartIndex < valueEndIndex) {
       String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
@@ -29,8 +29,10 @@ import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import org.w3c.dom.Text;
 
 /** Decodes ID3 tags. */
 public final class Id3Decoder extends SimpleMetadataDecoder {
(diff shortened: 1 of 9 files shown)


library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -455,14 +455,29 @@ private static TextInformationFrame decodeTxxxFrame(ParsableByteArray id3Data, i
     byte[] data = new byte[frameSize - 1];
     id3Data.readBytes(data, 0, frameSize - 1);
 
-    int descriptionEndIndex = indexOfEos(data, 0, encoding);
+    int descriptionEndIndex = indexOfTerminator(data, 0, encoding);
     String description = new String(data, 0, descriptionEndIndex, charset);
 
+    // In ID3v2.4, text information frames can contain multiple values delimited by a null
+    // terminator. Thus, we after each "end of stream" marker we actually need to keep looking
+    // for more data, at least until the index is equal to the data length.
```
icbaker: I think you can remove this comment

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -455,14 +455,29 @@ private static TextInformationFrame decodeTxxxFrame(ParsableByteArray id3Data, i
...
+    ArrayList<String> values = new ArrayList<>();
```
icbaker: nit: move this down a few lines closer to where it's first used (probably L470)

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
 
   @Nullable public final String description;
+
+  /** @deprecated Use {@code values} instead. */
+  @Deprecated
   public final String value;
 
-  public TextInformationFrame(String id, @Nullable String description, String value) {
+  @NonNull
+  public final String[] values;
```
icbaker: Let's use `ImmutableList<String>` here instead - since arrays are unfortunately mutable

https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableList.html

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -479,10 +494,26 @@ private static TextInformationFrame decodeTextInformationFrame(
...
+    }
+
+    int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
+    while (valueStartIndex < valueEndIndex) {
+      String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
+      values.add(value);
+
+      valueStartIndex = valueEndIndex + delimiterLength(encoding);
+      valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
+    }
```
icbaker, line 504: This looks very duplicated with the code above - would pulling this into a private method make sense?

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -455,14 +455,29 @@ private static TextInformationFrame decodeTxxxFrame(ParsableByteArray id3Data, i
...
     int valueStartIndex = descriptionEndIndex + delimiterLength(encoding);
-    int valueEndIndex = indexOfEos(data, valueStartIndex, encoding);
-    String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
+    if (valueStartIndex >= data.length) {
+      return new TextInformationFrame("TXXX", description, new String[0]);
+    }
+
+    int valueEndIndex = indexOfTerminator(data, valueStartIndex, encoding);
+    while (valueStartIndex < valueEndIndex) {
+      String value = decodeStringIfValid(data, valueStartIndex, valueEndIndex, charset);
```
icbaker: I don't think this will ever hit the 'not valid' branch of `decodeStringIfValid`, since you've asserted in the `while` that `valueStartIndex < valueEndIndex` and L467 and L477 ensure that `valueStartIndex < data.length` and `valueEndIndex <= data.length` respectively. So this can probably just be replaced with 

```
new String(data, valueStartIndex, valueEndIndex - valueStartIndex, charsetName)
```

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
+import java.util.Arrays;
 import java.util.List;
 
 /** Text information ID3 frame. */
 public final class TextInformationFrame extends Id3Frame {
+  private final static String MULTI_VALUE_DELIMITER = ", ";
 
   @Nullable public final String description;
+
+  /** @deprecated Use {@code values} instead. */
```
icbaker: Can you also mention that the first entry of `values` contains this value?

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
+  public TextInformationFrame(String id, @Nullable String description, @NonNull String[] values) {
     super(id);
     this.description = description;
-    this.value = value;
+    this.values = values;
+
+    if (values.length > 0) {
+      this.value = values[0];
+    } else {
+      this.value = null;
```
icbaker: I think this should `""` to be backwards compatible with the previous behaviour (please add a test for this)

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
 import androidx.annotation.Nullable;
 import com.google.android.exoplayer2.MediaMetadata;
 import com.google.android.exoplayer2.util.Util;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /** Text information ID3 frame. */
 public final class TextInformationFrame extends Id3Frame {
+  private final static String MULTI_VALUE_DELIMITER = ", ";
```
icbaker: I would just in-line this, i think it makes the usages clearer

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
   @Override
   public void populateMediaMetadata(MediaMetadata.Builder builder) {
+    // Depending on the context this frame is in, we either take the first value of a multi-value
+    // frame because multiple values make no sense, or we join the values together with a comma
+    // when multiple values do make sense.
     switch (id) {
       case "TT2":
       case "TIT2":
-        builder.setTitle(value);
+        builder.setTitle(values[0]);
```
icbaker: I think for this index-access to be safe you need to transform an empty array in the constructor to a single-entry array containing an empty string. Probably better to make this line (and similar lines below) robust in the case of an empty values list?

Please add a test for the case of passing in an empty values array.

icbaker (changes requested):
Thanks for the contribution! I added some initial thoughts

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: You're right about this. In fact, it's actually more consistent with the rest of the metadata module to make `values == [""]` rather than make it an empty list. I'll make that change.

OxygenCobalt: Resolved in 157f1f0.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 157f1f0.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 157f1f0.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 157f1f0.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 157f1f0.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
OxygenCobalt: Resolved in 0eb56f6.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
OxygenCobalt: Resolved in 0eb56f6.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
OxygenCobalt: Resolved in 0eb56f6.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
OxygenCobalt: Resolved in 0eb56f6.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
-
   @Nullable public final String description;
+
+  /** @deprecated Use {@code values[0]} instead. */
+  @Deprecated
   public final String value;
 
-  public TextInformationFrame(String id, @Nullable String description, String value) {
+  @NonNull
+  public final List<String> values;
```
icbaker, line 41: Can you add javadoc guaranteeing that this will always have at least one element?

Something like:

> The values of this ID3 frame. Will always have at least one element.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,77 @@
...
+
+  public TextInformationFrame(String id, @Nullable String description, @NonNull List<String> values) {
     super(id);
     this.description = description;
-    this.value = value;
+    
+    if( values.size() > 0 ) {
```
icbaker: Given that the spec says (emphasis mine):
> All ID3v2 frames consists of one frame header followed by **one or more** fields containing the actual information.

I think we should make this type match the spec, and reject an empty list with `IllegalArgumentException`

Then where we construct the list in `Id3Decoder` we can do the transform from 'no field' -> `[""]` before passing it in here, and it's part of the existing handling 'invalid' files logic (similar to how the current code uses `decodeStringIfValid` inside `Id3Decoder` instead of trying to do that transform in this constructor).

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
```diff
@@ -474,15 +475,38 @@ private static TextInformationFrame decodeTextInformationFrame(
...
+      byte[] data, final int encoding, final int index) throws UnsupportedEncodingException {
+    ArrayList<String> values = new ArrayList<>();
+
+    if (index >= data.length) {
+      return Collections.emptyList();
+    }
+
+    // In ID3v2.4, text information frames can contain multiple values delimited by a null
+    // terminator. Thus, we after each "end of stream" marker we actually need to keep looking
+    // for more data, at least until the index is equal to the data length.
```
icbaker: I think you can remove this comment too. IMO it only really makes sense as part of the 'diff' from the previous (current) behaviour, but that diff won't be visible to a reader in future (e.g. in 1 year). We should just assume that multiple values are 'natural' so no need to explicitly comment it.

icbaker (changes requested):

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/Id3Decoder.java
OxygenCobalt: Resolved in 85c4d98.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 85c4d98.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
OxygenCobalt: Resolved in 85c4d98.

icbaker:
Looks good, thanks for all the changes! I'll get this merged.

icbaker (approved):

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,76 @@
...
-    value = castNonNull(in.readString());
+    values = ImmutableList.copyOf(in.createStringArray());
+    value = values.get(0);
   }
 
   @Override
   public void populateMediaMetadata(MediaMetadata.Builder builder) {
+    // Depending on the context this frame is in, we either take the first value of a multi-value
+    // frame because multiple values make no sense, or we join the values together with a comma
+    // when multiple values do make sense.
```
icbaker, line 71: During internal review of this PR, @tonihei asked whether this split of "multiple values makes sense for field X but not field Y" is defined anywhere in the spec - or whether it's just your intuition?

I see that https://id3.org/id3v2.4.0-frames says:
> All text information frames supports multiple strings

Given that, we're wondering if it might be safer to just comma-separate every value when writing into a string field.

For `trackNumber` and other 'transform to a single int' fields we would probably still need to only use the first value, since there's no real way to join multiple values together.

What do you think?

(no need to make any additional changes to this PR, just reply to this comment thread and I'll make the changes to the internal version and they will get propagated back here after it's submitted internally)

OxygenCobalt: It's largely intuition from my personal experience @icbaker. Feel free to make it more generalized if you believe it aligns with the specification better.

library/extractor/src/main/java/com/google/android/exoplayer2/metadata/id3/TextInformationFrame.java
```diff
@@ -19,52 +19,76 @@
...
     description = in.readString();
-    value = castNonNull(in.readString());
+    values = ImmutableList.copyOf(in.createStringArray());
+    value = values.get(0);
   }
 
   @Override
   public void populateMediaMetadata(MediaMetadata.Builder builder) {
+    // Depending on the context this frame is in, we either take the first value of a multi-value
+    // frame because multiple values make no sense, or we join the values together with a comma
```
icbaker, line 70: It's been pointed out that these strings may end up directly in a UI (e.g. in a playback notification) and using a comma to separate a list of items is not appropriate for all locales.

We have an [`exo_item_list` resource definition](https://github.com/google/ExoPlayer/blob/r2.18.2/library/ui/src/main/res/values/strings.xml#L127) which gives a localised way to build up a list of items but it's currently defined in the UI module, so not available from here (the extractor module).

I understand you sent this PR because you need access to these multiple values. Do you need them to be automatically put in the `MediaMetadata`  object (i.e. handled in this method) or is it enough for them to be available from `TextInformationFrame` and your app can handle them from there? We're considering just keeping the existing behaviour of this method for now, due to the additional complexity of moving `exo_item_list` to e.g. the common module and also accessing it here (we'd need a `Context`). But if that makes this PR completely pointless for you then we should probably rethink that simplification :)

OxygenCobalt: It's enough to just be available in `TextInformationFrame` @icbaker. Feel free to revert it back to it's original behavior. What I might propose is a refactor of MediaMetadata itself to handle multiple values, but that's something that can be done on your end.

Issue #10502, opened 2022-08-03
Some vorbis files does not play in ogg container 


Tolriq:
### ExoPlayer Version

2.18.1

### Devices that reproduce the issue

All

### Devices that do not reproduce the issue

None

### Reproducible in the demo app?

Yes

### Reproduction steps

Play the attached file

### Expected result

The media should play.

From my understanding after some research, it seems that this file also embed a "fake" Theora track that stream the cover.
But the file have a valid Vorbis track that plays normally on other players.

The ogg parser should support those file and only return the Vorbis tracks, specially when providing a music only player.

### Actual result

2022-08-03 10:25:55.919 Error/ExoPlayer: playerFailed [eventTime=3707.62, mediaPos=0.00, window=3, period=3, errorCode=ERROR_CODE_PARSING_CONTAINER_UNSUPPORTED
  com.google.android.exoplayer2.ExoPlaybackException: Source error
      at com.google.android.exoplayer2.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:632)
      at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:602)
      at android.os.Handler.dispatchMessage(Handler.java:102)
      at android.os.Looper.loopOnce(Looper.java:201)
      at android.os.Looper.loop(Looper.java:288)
      at android.os.HandlerThread.run(HandlerThread.java:67)
  Caused by: com.google.android.exoplayer2.source.UnrecognizedInputFormatException: None of the available extractors (Mp3Extractor, FlacExtractor, OggExtractor, WavExtractor, MatroskaExtractor, Mp4Extractor, FragmentedMp4Extractor) could read the stream.
      at com.google.android.exoplayer2.source.BundledExtractorsAdapter.init(BundledExtractorsAdapter.java:92)
      at com.google.android.exoplayer2.source.ProgressiveMediaPeriod$ExtractingLoadable.load(ProgressiveMediaPeriod.java:1017)
      at com.google.android.exoplayer2.upstream.Loader$LoadTask.run(Loader.java:412)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
      at java.lang.Thread.run(Thread.java:920)

### Media

[...]

PR #10807, opened 2022-11-22
Ogg fixes

Tolriq:
Improve ogg support by supporting multistream container and selecting the first supported.

Properly support comment headers present in a future page.
Properly only decode a single stream when selected by the extractor.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/DefaultOggSeeker.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/DefaultOggSeeker.java
@@ -69,6 +69,7 @@ import java.io.IOException;
    */
   public DefaultOggSeeker(
       StreamReader streamReader,
+      long streamSerialNumber,
       long payloadStartPosition,
       long payloadEndPosition,
       long firstPayloadPageSize,
@@ -86,7 +87,7 @@ import java.io.IOException;
     } else {
       state = STATE_SEEK_TO_END;
     }
-    pageHeader = new OggPageHeader();
+    pageHeader = new OggPageHeader(streamSerialNumber);
   }
 
   @Override
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/FlacReader.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/FlacReader.java
@@ -42,6 +42,10 @@ import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;
   @Nullable private FlacStreamMetadata streamMetadata;
   @Nullable private FlacOggSeeker flacOggSeeker;
 
+  public FlacReader(long streamSerialNumber) {
+    super(streamSerialNumber);
+  }
+
   public static boolean verifyBitstreamType(ParsableByteArray data) {
     return data.bytesLeft() >= 5
         && data.readUnsignedByte() == 0x7F
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
@@ -90,7 +90,7 @@ public class OggExtractor implements Extractor {
 
   @EnsuresNonNullIf(expression = "streamReader", result = true)
   private boolean sniffInternal(ExtractorInput input) throws IOException {
-    OggPageHeader header = new OggPageHeader();
+    OggPageHeader header = new OggPageHeader(-1L);
 
     while (header.populate(input, true) && (header.type & 0x02) == 0x02) {
       int length = min(header.bodySize, MAX_VERIFICATION_BYTES);
@@ -98,13 +98,13 @@ public class OggExtractor implements Extractor {
       input.peekFully(scratch.getData(), 0, length);
 
       if (FlacReader.verifyBitstreamType(resetPosition(scratch))) {
-        streamReader = new FlacReader();
+        streamReader = new FlacReader(header.streamSerialNumber);
         return true;
       } else if (VorbisReader.verifyBitstreamType(resetPosition(scratch))) {
-        streamReader = new VorbisReader();
+        streamReader = new VorbisReader(header.streamSerialNumber);
         return true;
       } else if (OpusReader.verifyBitstreamType(resetPosition(scratch))) {
-        streamReader = new OpusReader();
+        streamReader = new OpusReader(header.streamSerialNumber);
         return true;
       }
       try {
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
@@ -91,24 +91,31 @@ public class OggExtractor implements Extractor {
   @EnsuresNonNullIf(expression = "streamReader", result = true)
   private boolean sniffInternal(ExtractorInput input) throws IOException {
     OggPageHeader header = new OggPageHeader();
-    if (!header.populate(input, true) || (header.type & 0x02) != 0x02) {
-      return false;
+
+    while (header.populate(input, true) && (header.type & 0x02) == 0x02) {
+      int length = min(header.bodySize, MAX_VERIFICATION_BYTES);
+      ParsableByteArray scratch = new ParsableByteArray(length);
+      input.peekFully(scratch.getData(), 0, length);
+
+      if (FlacReader.verifyBitstreamType(resetPosition(scratch))) {
+        streamReader = new FlacReader();
+        return true;
+      } else if (VorbisReader.verifyBitstreamType(resetPosition(scratch))) {
+        streamReader = new VorbisReader();
+        return true;
+      } else if (OpusReader.verifyBitstreamType(resetPosition(scratch))) {
+        streamReader = new OpusReader();
+        return true;
+      }
+      try {
+        input.skip((int) input.getPeekPosition());
+        header.skipToNextPage(input);
+      } catch (IOException e) {
+        return false;
+      }
     }
 
-    int length = min(header.bodySize, MAX_VERIFICATION_BYTES);
-    ParsableByteArray scratch = new ParsableByteArray(length);
-    input.peekFully(scratch.getData(), 0, length);
-
-    if (FlacReader.verifyBitstreamType(resetPosition(scratch))) {
-      streamReader = new FlacReader();
-    } else if (VorbisReader.verifyBitstreamType(resetPosition(scratch))) {
-      streamReader = new VorbisReader();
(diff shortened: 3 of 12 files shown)


rohitjoins:
Thank you for sending the PR. I'll take a look.

Tolriq:
@rohitjoins it lacks new tests and probably some more code refactoring to reduce duplicate code and position handling in the buffers. But wanted to discuss before. (And probably extracting a new C.XXX long value for the -1L as there was none)

For the tests I'll need more details about how to create the new files and the corresponding expectations if you expect me to add them.

Tolriq:
@rohitjoins Any news? If this kind of approach is not acceptable in it's current form, is there an easy way to handle this on my side in a module and not by maintaining a fork ?

rohitjoins:
@Tolriq I apologise for the delay. Will look into this next week. If you want to speed up the process, please consider raising the same PR in https://github.com/androidx/media. 

Please raise any future PRs in the media repository as well. Thanks.

Tolriq:
@rohitjoins No problem but as said this was first for discussion and tuning before anything else. I'm in prod with this in a fork, but it requires some improvements to really be merged I suppose.

About media 3, I do not have plans to migrate currently as everything it brings are not things I want nor want to use. I really hope exoplayer can still be used as a single library without forced dependencies on the rest. I have my fork of media router and many other custom things that does not fit the "media3" architecture.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/VorbisReader.java
```diff
@@ -140,7 +144,11 @@ protected boolean readHeaders(ParsableByteArray packet, long position, SetupData
     }
 
     if (commentHeader == null) {
-      commentHeader = VorbisUtil.readVorbisCommentHeader(scratch);
+      // The comment header can be in a future page
+      if (VorbisUtil.verifyVorbisHeaderCapturePattern(/* headerType= */ 0x03, scratch, /* quiet= */ true)) {
+        scratch.setPosition(scratch.getPosition() - 7);
+        commentHeader = VorbisUtil.readVorbisCommentHeader(scratch);
+      }
```
rohitjoins, line 151: Based on the [spec](https://www.xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-610004.2), the Ogg stream consists of three header packets which are in order. 

Do we need to do this also for `vorbisIdHeader`? Based on my understanding, I think we skip pages not belonging to vorbis stream in `OggPageHeader` and we don't need to do any changes here. WDYT?

Tolriq: Yes the page is skipped before if needed for the vorbisIdHeader so no need for change. If it's not present at that moment it's an error.

Previously it was assumed that the comment was in the same page and directly read, but the spec does not enforce that and I had multiple files during my tests that had the comment on the next page.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggPageHeader.java
```diff
@@ -59,6 +61,10 @@
 
   private final ParsableByteArray scratch = new ParsableByteArray(MAX_SEGMENT_COUNT);
 
+  public OggPageHeader(long streamSerialNumber) {
+    expectedStreamSerialNumber = streamSerialNumber;
+  }
```
rohitjoins, line 66: Should we create both parameterised and default constructors here? The default constructor can call the parameterised one with default value for streamSerialNumber. The default value can be C.INDEX_UNSET or create a new constant if wanted.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/DefaultOggSeeker.java
```diff
@@ -69,6 +69,7 @@
    */
   public DefaultOggSeeker(
       StreamReader streamReader,
+      long streamSerialNumber,
```
rohitjoins, line 72: We won't need to change the signature of this function if we can make streamSerialNumber as public member of StreamReader class.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/FlacReader.java
```diff
@@ -42,6 +42,10 @@
   @Nullable private FlacStreamMetadata streamMetadata;
   @Nullable private FlacOggSeeker flacOggSeeker;
 
+  public FlacReader(long streamSerialNumber) {
+    super(streamSerialNumber);
+  }
+
```
rohitjoins, line 48: Javadoc is missing for the constructor.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
```diff
@@ -90,25 +90,32 @@ public int read(ExtractorInput input, PositionHolder seekPosition) throws IOExce
...
+      } else if (VorbisReader.verifyBitstreamType(resetPosition(scratch))) {
+        streamReader = new VorbisReader(header.streamSerialNumber);
+        return true;
+      } else if (OpusReader.verifyBitstreamType(resetPosition(scratch))) {
+        streamReader = new OpusReader(header.streamSerialNumber);
+        return true;
+      }
+      try {
+        input.skip((int) input.getPeekPosition());
+        header.skipToNextPage(input);
```
rohitjoins, line 112: Based on the documentation of Extractor.sniff(ExtractorInput input), we can only modify the reading position when returning true from this method. If false is returned only peek position is modified.

Here we are changing the read position by calling skip on input and may still return false.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggPacket.java
```diff
@@ -29,7 +29,11 @@
 /** OGG packet class. */
 /* package */ final class OggPacket {
 
-  private final OggPageHeader pageHeader = new OggPageHeader();
+  private final OggPageHeader pageHeader;
+  public OggPacket(long streamSerialNumber) {
+    pageHeader = new OggPageHeader(streamSerialNumber);
+  }
```
rohitjoins, line 35: Javadoc is missing. Also this constructor should be moved after all variables are declared.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggPageHeader.java
rohitjoins: Javadoc is missing.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OpusReader.java
```diff
@@ -38,6 +38,10 @@
     'O', 'p', 'u', 's', 'T', 'a', 'g', 's'
   };
 
+  public OpusReader(long streamSerialNumber) {
+    super(streamSerialNumber);
+  }
+
```
rohitjoins, line 44: Javadoc is missing.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/VorbisReader.java
```diff
@@ -43,6 +43,10 @@
   @Nullable private VorbisUtil.VorbisIdHeader vorbisIdHeader;
   @Nullable private VorbisUtil.CommentHeader commentHeader;
 
+  public VorbisReader(long streamSerialNumber) {
+    super(streamSerialNumber);
+  }
+
```
rohitjoins, line 49: Same, Javadoc is missing here.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/VorbisReader.java
rohitjoins: We read one packet at a time when calling this method. And as these packets are in order, I expect this to work as intended. Can you please share some examples where the comment header is present in the next page and we are not able to read it based on the current implementation?

library/extractor/src/test/java/com/google/android/exoplayer2/extractor/ogg/OggPageHeaderTest.java
```diff
@@ -48,7 +48,7 @@ public void skipToNextPage_success() throws Exception {
                 new byte[] {'O', 'g', 'g', 'S'},
                 TestUtil.buildTestData(20, random)),
```
rohitjoins, line 49: This test fails based on the changes. As we set the limit now to EMPTY_PAGE_HEADER_SIZE which of size 27. So we should use a number greater than 27 here.

rohitjoins (changes requested):
Hi @Tolriq,

There are few minor issues and some major in this PR. I would have fixed the minor comments changes if everything else worked as expected. 

Also this PR requires extensive testing in my opinion. You can create some tests by extracting headers from the sample multi-stream media files and check if we can read the correct first supported header from it. One other test would be too see that we only read packets for the streamSerialNumber we derived and ignore the rest. And lastly, it would be nice to add some multi-stream ogg media files and check if can be played successfully.

As the issue to support multi stream in Ogg is a low priority for us now, we cannot afford to spend so much time in merging this PR at this moment in time. Please address all the comments and add tests for this to be merged.

Tolriq:
@rohitjoins For the tests I asked about more details about how to create them I do not know how the files are generated and how to reference them from the test data as it fails when just adding here on Windows.

As a reminder this is a PR for https://github.com/google/ExoPlayer/issues/10502 that have the file for test.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/VorbisReader.java
Tolriq: The sample file from the OG issue have the case. As explained you mix packet and page. The packets are put in pages. But the pages can be anysize and interleaved specially for those multi tracks.

So you have [Stream1-Page1][Header][Packet1][Stream2-Page1][Header][Packet1][Stream1-Page2][Header][Packet2][Stream2-Page2][Header][Packet2]

Packet 1 and packet 2 are in order and adjoined as the spec define. But the actual bytes are not at all, there's the other stream + the page header between them.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/ogg/OggExtractor.java
Tolriq: This one is the only one a little problematic :( 
While it's relatively easy to build a PeekExtractorInput wrapper, I could imagine an attack vector of a on purpose wrong ogg file with hundreds of fake streams to oom ExoPlayer.
It would probably need some limits to be defined to address that, either in the buffer size of the wrapper or in the number of attempted track tested. WTDY?

icbaker:

PR #10816, opened 2022-11-26
Enable CastPlayer to report live stream duration

ptsekov:


extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
```diff
@@ -1321,6 +1321,9 @@ private final class StatusListener extends RemoteMediaClient.Callback
     @Override
     public void onProgressUpdated(long progressMs, long unusedDurationMs) {
       lastReportedPositionMs = progressMs;
+      if (remoteMediaClient != null && remoteMediaClient.isLiveStream()) {
+        remoteMediaClient.requestStatus();
```
marcbaechinger: This will be called every second following the constant we use to request progress update requests. 

I'm not super familiar with the behaviour of the SDK with live streams.
So, aside this PR and from your experience with live streams and the Casts SDK, you can probably help me out a bit with my questions. :)

We here request the status to then read the `MediaLiveSeekableRange` to get the window. On the cast device, a change in the range is a result of reloading the manifest/timeline frequently. So, naively, can't we just expect the cast SDK to propagate that change to clients with a status update? This would then be in sync with the manifest updates, meaning, we don't pull too frequently and without the chance to get a no-op update.

Have you tested with your stream whether the status is updated without us requesting the update here?
I wonder further whether there is an API that gives us the manifest/playlist interval at which the Cast device fetches the manifest/playlist for a live stream being played?

marcbaechinger (commented):
Thanks for the CL! That looks good in general!

I moved things a bit and will merge this internally soon and then sync back to the dev-v2 branch here on GitHub. Many thanks!

extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
ptsekov: > This will be called every second following the constant we use to request progress update requests.

Yes.

> I'm not super familiar with the behaviour of the SDK with live streams. So, aside this PR and from your experience with live streams and the Casts SDK, you can probably help me out a bit with my questions. :)

I am not super familiar with the Cast SDK myself and the documentation is lacking.
 
> We here request the status to then read the `MediaLiveSeekableRange` to get the window. On the cast device, a change in the range is a result of reloading the manifest/timeline frequently. So, naively, can't we just expect the cast SDK to propagate that change to clients with a status update? This would then be in sync with the manifest updates, meaning, we don't pull too frequently and without the chance to get a no-op update.

This is how I thought it would work but testing revealed that status updates only happen when the user performs some action on the receiver device i.e. using any of the media keys would trigger a status update. I tested with a combination of three Android TV devices as receivers and three phones as senders:

- A Sony TV (Android 9)
- An operator STB (Android 10)
- A Chromecast for Google TV (Android 12)
- Xiaomi MI 5 (Android 8)
- Xiaomi MI Mix 3 (Android 10)
- Samsung Galaxy A41 (Android 12)

I have tested both the default Cast receiver and the styled Cast receiver.

> Have you tested with your stream whether the status is updated without us requesting the update here? I wonder further whether there is an API that gives us the manifest/playlist interval at which the Cast device fetches the manifest/playlist for a live stream being played?

I will check if there is such an API.

ptsekov: According to this link: https://developers.google.com/cast/docs/media/streaming_protocols#http_live_streaming_hls:

[...]
Every #EXT-X-TARGETDURATION seconds, we reload the playlist/manifest to get new segment lists and we update the new internal representation of all the segments to the new one. Any time a seek is requested, we only seek within the seekable range. For live, we only allow seeking from the beginning of the newest list until a three target duration from the end. So for example, if you have a 10 segment list, and you are on segment 6, you can only seek up to 7, but not 8.
[...]

I guess we could determine the refresh interval by subtracting the value of MediaLiveSeekableRange.getEndTime() between two status updates and update the timer which triggers requestStatus() accordingly?!

ptsekov:
@marcbaechinger, there is an issue with the PR - I will fix it and update the PR. Apparently RemoteMediaClient.requestStatus() triggers RemoteMediaClient.ProgressListener when the remote client is in the BUFFERING state. This could lead to infinite status polling if the client remains in the BUFFERING state.

ptsekov:
I've moved requestStatus() out of the progress listener.

extensions/cast/src/main/java/com/google/android/exoplayer2/ext/cast/CastPlayer.java
```diff
@@ -976,6 +997,15 @@ private boolean updateTimelineAndNotifyIfChanged() {
                     getCurrentMediaItem(), MEDIA_ITEM_TRANSITION_REASON_PLAYLIST_CHANGED));
       }
       updateAvailableCommandsAndNotifyIfChanged();
+
+      if (playingPeriodChanged) {
+        // If the new source is a live stream enable live stream status updates.
+        refreshLiveStreamStatusHandler.removeCallbacksAndMessages(null);
+        if (remoteMediaClient != null && remoteMediaClient.isLiveStream()) {
```
marcbaechinger, line 1004: I tested this with a couple of live streams but I never got `remoteMediaClient.isLiveStream()` return true. It is always false so it does not start polling. 

What live streams are you using when this returns true? HLS or DASH?

ptsekov: Ok. I was planning to create a question issue for this but here it comes - you have to remove the following line from 
from DefaultMediaItemConverter.toMediaQueueItem():
```
.setStreamType(MediaInfo.STREAM_TYPE_BUFFERED)
```

The Cast receiver behaves properly with both types of streams this way.

icbaker:

Issue #10865, opened 2022-12-13
Let AudioSink indicate if playback speed changes are supported to avoid unnecessary live speed adjustments


psharma676:
### ExoPlayer Version

2.15.0

### Devices that reproduce the issue

Tivo Streaming device running Android 10
Nvidia Sheild running Android 11

### Reproducible in the demo app?

Yes

### Reproduction steps

1) Apply attached patch "patch-r2.15.0.txt" to branch at tag r2.15.0. This will generate logs referenced in "Actual result" section below, or you can use your own logging.  
2) Play a working (AAC audio) stream and compare with Non-Working (AC3 stream).

### Expected result

Playback speed for AC3 streams should adjust to push liveOffsetUs up to idealTargetLiveOffsetUs. 

### Actual result

Note in the logs:
       For AAC: liveOffsetUs → currentTargetLiveOffsetUs and adjustedPlaybackSpeed → 1.0 eventually
       For AC3: liveOffsetUs does not move towards currentTargetLiveOffsetUs and adjustedPlaybackSpeed stays at 1.3

In both cases above currentTargetLiveOffsetUs →idealTargetLiveOffsetUs, which should be expected behavior. So AAC will eventually end up at the playback position of idealTargetLiveOffsetUs, but not AC3. 

The same behavior is seen in r.2.18.X as well as androidx/media (some adjustments are required to trigger/force Playback Speed Control because of [commit](https://github.com/google/ExoPlayer/commit/b09b8dc2aba5f4b54d76a0b958c62aeb5e5b8075), https://github.com/google/ExoPlayer/issues/9329)

### Media

[patch-r2.15.0.txt](https://github.com/google/ExoPlayer/files/10222594/patch-r2.15.0.txt)
Media streaming information will be sent separately via email. 

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

tonihei:
> liveOffsetUs does not move towards currentTargetLiveOffsetUs and adjustedPlaybackSpeed stays at 1.3

This sounds odd and logically impossible :)  Could you also share the logs you obtained with the attached patch? I'm curious to see how the values change (or not change) and I don't have an NVidea Shield available right now.

psharma676:
[Issue-10865-Dec-14-22-log.zip](https://github.com/google/ExoPlayer/files/10230555/Issue-10865-Dec-14-22-log.zip)
[...]

Issue #10882, opened 2022-12-22
User setPlaybackParameters() is overwritten by live adjusment


stevemayhew:
### ExoPlayer Version

2.18.2

### Devices that reproduce the issue

Evolution Digital eStream 4k (https://evolutiondigital.com/estream-4k/)  (probably many others)

Easiest to reproduce if device is running AC-3 audio with off-board decode.

### Reproducible in the demo app?

Yes

### Reproduction steps

Pretty difficult, working on a ExoPlayer unit test case (not sure it is even possible)

The steps are to repeatedly tune to a live channel, while speed adjustment is progress, then:
1. Disable the audio 
2. Set the speed to > 1.0. (eg 8x)

### Expected result

Media plays at 2.0x (or whatever the setting was).

### Actual result

Playback reverts immediately to 1.0 (with adjustment +/- 3% default).

The branch [x-live-speed-overwrites-user-logging](https://github.com/TiVo/ExoPlayer/tree/x-live-speed-overwrites-user-logging) has the logging shown in the description below.  Here's the sequence that causes the bug with my analysis:

1. `DO_SOME_WORK` &mdash; sees renderer clock is 1.0, does live speed adjustment, sets speed to 1.03

```
12-21 14:34:41.564  7303  7343 I ExoPlayerImplInternal: updatePlaybackPositions() changing speed - 
		old: PlaybackParameters(speed=1.00, pitch=1.00) 
		new speed: 1.03 
		MediaClock: DefaultMediaClock isUsingStandaloneClock: false standaloneClockIsStarted: true 
		      rendererClock: PlaybackParameters(speed=1.00, pitch=1.00) 
		      standaloneClock: PlaybackParameters(speed=1.00, pitch=1.00)

12-21 14:34:41.564  7303  7343 I ExoPlayerImplInternal: handlePlaybackParameters()  updatePlaybackInfo:false acknowledgeCommand:false 
        currentPlaybackSpeed: 1.03 currentParameters: PlaybackParameters(speed=1.00, pitch=1.00) playbackParameters: PlaybackParameters(speed=1.00, pitch=1.00)
```

2a. `DO_SOME_WORK` &mdash; Renderer reports speed 1.0 (ignore 1.03 setting), and this queues a `MSG_PLAYBACK_PARAMETERS_CHANGED_INTERNAL`, and resets standalone clock to 1.0

```
[...]

PR #10883, opened 2022-12-22
setPlaybackParameters() not overwriten by live adjusment

stevemayhew:
There are multiple sequences of handler messages which can cause a user request to update playback speed to be overwritten by the `LivePlaybackSpeedControl` value.

All the sequeunces share a common feature that:

1. a queued `MSG_PLAYBACK_PARAMETERS_CHANGED_INTERNAL` executes after the `MSG_SET_PLAYBACK_PARAMETERS`
2. The `PlaybackParameters` in this internal change are not the intended speed.

The easiest way to reproduce this is with AC-3 pass-thru audio, in this case the audio renderer does not properly support speed change, see bug:
  https://github.com/google/ExoPlayer/issues/10865

According to this commit,

 [Add back support for setting audio pitch](https://github.com/google/ExoPlayer/commit/74c493f51e) [andrewlewis]

This callback to report the internal speed change is needed for the renderers to run slow motion properly.

In any event, there is no reason to propegate this internal speed change to the user speed setting (`ExoPlayerImplInternal.playbackParameters`). The change simply fixes `MSG_PLAYBACK_PARAMETERS_CHANGED_INTERNAL` not to update the user setting.

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -527,7 +527,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
           reselectTracksInternal();
           break;
         case MSG_PLAYBACK_PARAMETERS_CHANGED_INTERNAL:
-          handlePlaybackParameters((PlaybackParameters) msg.obj, /* acknowledgeCommand= */ false);
+          PlaybackParameters updatedParameters = (PlaybackParameters) msg.obj;
+          reportPlaybackParametersUpdateInternally(updatedParameters, updatedParameters.speed);
           break;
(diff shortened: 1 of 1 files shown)


stevemayhew:
This is the fix for #10882

stevemayhew:
I'll open the same pull request on AndroidX.media if this one is acceptable.

tonihei:
I'd suggest we continue the discussion on #10882. This PR doesn't fully work anyway I believe because it cuts out all potential updates to `PlaybackInfo` when a speed change is triggered from `MSG_PLAYBACK_PARAMETERS_CHANGED_INTERNAL`. This may be correct for the speed changes related to automatic live stream adjustments, but we definitely want to see the `PlaybackInfo` update for cases where a manually set speed can't be applied or has to be changed.

Issue #10996, opened 2023-02-17
AV Sync issue in DASH Stream


mayurk2:
### ExoPlayer Version

2.18.3

### Devices that reproduce the issue

All devices.
Tested on Redmi Note 9 Pro, Samsung M32.

### Reproducible in the demo app?

Yes

### Reproduction steps

1. I saved the attached stream in some folder on Phone.
2. Hard coded the path in "media.exolist.json".
3. Played the media and observed AV sync.

### Expected result

Media should play with correct AV sync.

### Actual result

AV sync is not present.

### Media

https://drive.google.com/file/d/156XrRCy0Nxk6_u1fgQa9JVEExy3JeLRk/view?usp=sharing 

### Bug Report

- [ ] You will email the zip file produced by `adb bugreport` to dev.exoplayer@gmail.com after filing this issue.

tonihei:
As far as I can see, your source media is simply not aligned. You can check yourself by listening to the audio-only file, check the time of the initial conversation ("You are a jerk Tom..."), it starts at 00:23 and ends at 00:27, and then compare the the video-only file where it still shows the previous scene between 00:23 and 00:27.

mayurk2:
@tonihei Event I thought that is the issue first.
I saw that there is a difference in first PTS of audio and video when I printed the timestamp value here - https://github.com/google/ExoPlayer/blob/f72ca8338399c9e3c19e73fc0779f24baff14d35/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java#L1057

But I later saw that VLC and Shaka player are able to play this content correctly and AV sync issue was not observed.

Is it possible to provide more info on why is this issue happens only with Exoplayer?

mayurk2:
@tonihei From the VLC source code here - https://code.videolan.org/videolan/vlc/-/blob/master/modules/demux/mp4/mp4.c#L4749
It looks like VLC is using separate PTS and DTS values. The i_composition_time_offset (sampleCompositionTimeOffset in Exoplayer) is added only to PTS.
But Exoplayer does not have separate PTS and DTS values. Maybe that is resulting in AV sync issue?

mayurk2:
@tonihei another observation:
I played the original video on YouTube - https://youtu.be/41hv2tW5Lc4
The dialogue starts before the the people show-up on the video - i.e. at ~0:24s.


I played audio and video separately on "QuickTime Player". 
[...]

PR #11010, opened 2023-02-23
Use edts media time offset if edits list duration is entire media duration.

mayurk2:
It will fix the AV sync issue mentioned here - https://github.com/google/ExoPlayer/issues/10996 

(Test media file: https://drive.google.com/file/d/156XrRCy0Nxk6_u1fgQa9JVEExy3JeLRk/view?usp=sharing)

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
@@ -129,8 +129,6 @@ public class FragmentedMp4Extractor implements Extractor {
   private static final int STATE_READING_SAMPLE_START = 3;
   private static final int STATE_READING_SAMPLE_CONTINUE = 4;
 
-  private static final long MICROS_PER_MILLI = 1000L;
-
   // Workarounds.
   private final @Flags int flags;
   @Nullable private final Track sideloadedTrack;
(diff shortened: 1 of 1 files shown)


tonihei:
Thanks for trying to fix this! Could you open the PR against the [main branch of the AndroidX media3 library](https://github.com/androidx/media/tree/main)? This makes it a lot easier for us to merge the change. I'll leave some comment in this PR still for your review.

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -968,6 +970,27 @@ private static long parseTfdt(ParsableByteArray tfdt) {
     return version == 1 ? tfdt.readUnsignedLongToLong() : tfdt.readUnsignedInt();
   }
 
+  /**
+   * Retunrs if edts list duration is for the entire media timeline.
```
tonihei: There is no need to document a private method with Javadoc as long as the method name is self-explanatory (which it is).

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
```diff
@@ -968,6 +970,27 @@ private static long parseTfdt(ParsableByteArray tfdt) {
...
+    // Currently we only support a single edit that moves the entire media timeline (indicated by
+    // duration == 0 or duration == track duration).
+    // Other uses of edit lists are uncommon and unsupported.
+    if (track.editListDurations != null && track.editListDurations.length == 1) {
+      long edtsDurationUs = Util.scaleLargeTimestamp(track.editListDurations[0],
+          C.MICROS_PER_SECOND, track.movieTimescale);
+      return edtsDurationUs == 0
+          || (track.durationUs > MICROS_PER_MILLI
+          && edtsDurationUs >= track.durationUs - MICROS_PER_MILLI
+          && edtsDurationUs <= track.durationUs + MICROS_PER_MILLI);
```
tonihei: The `edtsListDuration` is the duration of the edit itself. So the end time of the edit is `editListMediaTime + editListDuration`. And I think if we want to check if the edit list covers the entire segment, we can accept all edits that exceed the track duration. So the condition can just be:
`scaleToUs(editListMediaTime + editListDuration) >= track.durationUs`

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
mayurk2: Updated @tonihei .

library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java
mayurk2: Agreed. Updated the code @tonihei .

tonihei:
Thanks for the updates - did you see my comment above about opening this PR against the main branch of media3?

mayurk2:
> Thanks for trying to fix this! Could you open the PR against the [main branch of the AndroidX media3 library](https://github.com/androidx/media/tree/main)? This makes it a lot easier for us to merge the change. I'll leave some comment in this PR still for your review.

@tonihei Opened a PR - https://github.com/androidx/media/pull/255

mayurk2:
@tonihei should we merge this PR too? Or there are no more releases (minor/major) for Exoplayer?

mayurk2:
@tonihei should I raise PR to dev-v2 branch? Please let me know.
Opened it anyway :)
https://github.com/google/ExoPlayer/pull/11017

tonihei:
The media3 one is enough, thanks. We have a single codebase (in media3 package names) internally and the code gets pushed to both repositories automatically.

Issue #11050, opened 2023-03-09
Seek to default live does not restore idealTargetLiveOffset


stevemayhew:
### ExoPlayer Version

2.18.4

### Devices that reproduce the issue

* SEI500 SEI Robotics (AmLogic SoC STB) running Android 29
* Probably everything else with AAC audio

### Devices that do not reproduce the issue

None I'm aware of.

### Reproducible in the demo app?

Yes

### Reproduction steps

1. Play live stream with non-default `LiveConfiguration`
2. Seek back (jump back 5 button) a few times, notice `setTargetLiveOffsetOverrideUs()` is called to move the target (not sure why this is a feature, IMO live adjustment should shut off here until seek back to live edge)
3. call `seekTo(C.TIME_UNSET)` (from the UI play next button with no next item will do this.

### Expected result

Live adjustment resumes (continues) using the original target set on playback startup (`mediaConfigurationTargetLiveOffsetUs`)

### Actual result

Playback seeks to the live edge, but Live adjustment desperately tries to move **back** to the last non default seek.

Logging shows clearly what is happening:

Step 1 - normal startup and live adjustment

````
03-09 14:47:30.744 20204 20204 I ExoPlayerImpl: Init c9a8d43 [ExoPlayerLib/2.18.3] [SEI500RCN, Pyxis-RCN, SEI Robotics, 29]
03-09 14:47:31.665 20204 21891 D DefaultLivePlaybackSpeedControl: setLiveConfiguration(LiveConfiguration - min/max speed: 0.9/1.1 targetOffset: 33284) - updated
03-09 14:47:33.414 20204 21891 D DefaultLivePlaybackSpeedControl: adjusting offset, speed: 1.1 liveOffsetUs: 38107723 currentTargetLiveOffsetUs: 33284000 idealTargetLiveOffsetUs: 33284000 smoothedMinPossibleLiveOffsetUs: 32112240
````

Step 2 - three back button presses jump back 15 seconds total, override is called (`setTargetLiveOffsetOverrideUs()`)

````
03-09 14:48:02.377 20204 21891 D ExoPlayerImplInternal: seekToInternal() - timeline empty: false  seekPosition.windowPositionUs: 3575642000
03-09 14:48:02.378 20204 21891 D ExoPlayerImplInternal: resolvedSeekPosition periodUid: Pair{java.lang.Object@9be3c3e java.lang.Object@c575286} resolvedContentPositionUs: 3599666000
[...]

PR #11051, opened 2023-03-10
Removes liveOffsetTarget override on seek to live edge

stevemayhew:
Any calls that issue a `seekTo()` where the position is `TIME_UNSET` will now restore the live offset target to the MediaItem set default.

The code handles an intra-period seek in the same timeline by simply updating or removing the `setTargetLiveOffsetOverrideUs()`.  Other cases call the original code.

This fixes issue #11050

### DIFF
--- a/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
+++ b/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
@@ -1248,7 +1248,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
             /* newPeriodId= */ periodId,
             /* oldTimeline= */ playbackInfo.timeline,
             /* oldPeriodId= */ playbackInfo.periodId,
-            /* positionForTargetOffsetOverrideUs= */ requestedContentPositionUs);
+            /* positionForTargetOffsetOverrideUs= */ requestedContentPositionUs,
+            /* positionRequestedBySeek */true);
+
       }
     } finally {
       playbackInfo =
@@ -1897,7 +1899,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
           /* oldPeriodId= */ playbackInfo.periodId,
           /* positionForTargetOffsetOverrideUs */ positionUpdate.setTargetLiveOffset
               ? newPositionUs
-              : C.TIME_UNSET);
+              : C.TIME_UNSET,
+          /* positionRequestedBySeek */false);
       if (periodPositionChanged
           || newRequestedContentPositionUs != playbackInfo.requestedContentPositionUs) {
         Object oldPeriodUid = playbackInfo.periodId.periodUid;
@@ -1935,7 +1938,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
       MediaPeriodId newPeriodId,
       Timeline oldTimeline,
       MediaPeriodId oldPeriodId,
-      long positionForTargetOffsetOverrideUs)
+      long positionForTargetOffsetOverrideUs,
+      boolean positionRequestedBySeek)
       throws ExoPlaybackException {
     if (!shouldUseLivePlaybackSpeedControl(newTimeline, newPeriodId)) {
       // Live playback speed control is unused for the current period, reset speed to user-defined
@@ -1965,8 +1969,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
         int oldWindowIndex = oldTimeline.getPeriodByUid(oldPeriodId.periodUid, period).windowIndex;
         oldWindowUid = oldTimeline.getWindow(oldWindowIndex, window).uid;
       }
-      if (!Util.areEqual(oldWindowUid, windowUid)) {
-        // Reset overridden target live offset to media values if window changes.
+      if (!Util.areEqual(oldWindowUid, windowUid) || positionRequestedBySeek) {
+        // Reset overridden target live offset to media values if window changes or if seekTo
+        // default live position.
         livePlaybackSpeedControl.setTargetLiveOffsetOverrideUs(C.TIME_UNSET);
       }
     }
@@ -2089,7 +2094,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
         /* newPeriodId= */ readingPeriodHolder.info.id,
         /* oldTimeline= */ playbackInfo.timeline,
         /* oldPeriodId= */ oldReadingPeriodHolder.info.id,
-        /* positionForTargetOffsetOverrideUs= */ C.TIME_UNSET);
+        /* positionForTargetOffsetOverrideUs= */ C.TIME_UNSET,
+        /* positionRequestedBySeek */false);
 
     if (readingPeriodHolder.prepared
         && readingPeriodHolder.mediaPeriod.readDiscontinuity() != C.TIME_UNSET) {


library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -1237,12 +1237,29 @@ private void seekToInternal(SeekPosition seekPosition) throws ExoPlaybackExcepti
...
+            if (requestedContentPositionUs != C.TIME_UNSET) {
+              livePlaybackSpeedControl.setTargetLiveOffsetOverrideUs(
+                  getLiveOffsetUs(playbackInfo.timeline, periodId.periodUid, requestedContentPositionUs));
+            } else {
+              if (!playbackInfo.timeline.isEmpty()) {
+                // Seek resolved to window.getDefaultPositionMs(), so just remove the override.
+                livePlaybackSpeedControl.setTargetLiveOffsetOverrideUs(C.TIME_UNSET);
+              }
+            }
+          }
```
stevemayhew: basically this is an 'inline' of the logic from `updatePlaybackSpeedSettingsForNewPeriod` when the periodId does not change.

I'm curious for the source update call to `updatePlaybackSpeedSettingsForNewPeriod`, I'm assuming this is the possible in a DASH playlist to have multiple live periods?

marcbaechinger: Thanks for the contribution!

`updatePlaybackSpeedSettingsForNewPeriod` is called each time when the playing period potentially changed. This is not necessarily a multi-period stream. Currently this is called from `seekToInternal` (after a seek), `handleMediaSourceListInfoRefreshed` (after the timeline has changed), and `maybeUpdateReadingPeriod` when the reading period advances to the next period. 

From this I rather think we should move this logic into `updatePlaybackSpeedSettingsForNewPeriod`. So we can call it in any case because we want to potentially reset the playback parameters in the `!shouldUseLivePlaybackSpeedControl(...)` case.

The only issue with this is that `updatePlaybackSpeedSettingsForNewPeriod` is potentially called from the other call sites with `C.TIME_UNSET`. When `C.TIME_UNSET` is passed `updatePlaybackSpeedSettingsForNewPeriod` currently only remove the override when the old period is different to the new period (for instance when moving to the next reading period or when the timeline change has removed the playing period).

Concretely, I think your change does actually what `updatePlaybackSpeedSettingsForNewPeriod` does except that L1945 [1] does only remove the override when the periods are not the same, what is not the case in the `seekTo` case.

Allow me to suggest the following so you can test whether this is equivalent for you:

- add a parameter `boolean positionRequestedBySeek` to the parameter list of `updatePlaybackSpeedSettingsForNewPeriod`
- All call sites call the method then with `false` except the call from `seekToInternal`.
- Revert your change in `seekToInternal` and instead call `updatePlaybackSpeedSettingsForNewPeriod` with `positionRequestedBySeek=true`
- change ExoPlayerImplInternal.java#L1945 to `if (positionRequestedBySeek || !Util.areEqual(oldWindowUid, windowUid))`

I think that should boil down to the same, avoid some duplicated code and makes sure that the `!shouldUseLivePlaybackSpeedControl(..)` case is covered as well.

Can you apply this and probably even more important, test this with your app/stream to see if it is equivalent to your change? 

[1] https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java#L1945

marcbaechinger (commented):
Thanks Steve! Please see my suggestion in the comment!

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
stevemayhew: > Thanks for the contribution!
You're welcome

> `updatePlaybackSpeedSettingsForNewPeriod` is called each time when the playing period potentially changed.

Ah, ok as long as this alway the case then yes, it would duplicate less code to keep it all in that method.  Guess we need a name like `updatePlaybackSpeedSettingsForNewPeriodOrSeek` (:-)) 

> 
> From this I rather think we should move this logic into `updatePlaybackSpeedSettingsForNewPeriod`. So we can call it in any case because we want to potentially reset the playback parameters in the `!shouldUseLivePlaybackSpeedControl(...)` case.

That sounds good, I'll rework the code to follow this pattern and submit a second commit so we can review the diffs.  When it is merge ready I'll squash to a single commit.

> 
> The only issue with this is that `updatePlaybackSpeedSettingsForNewPeriod` is potentially called from the other call sites with `C.TIME_UNSET`. When `C.TIME_UNSET` is passed `updatePlaybackSpeedSettingsForNewPeriod` currently only remove the override when the old period is different to the new period (for instance when moving to the next reading period or when the timeline change has removed the playing period).
> 
> Concretely, I think your change does actually what `updatePlaybackSpeedSettingsForNewPeriod` does except that L1945 [1] does only remove the override when the periods are not the same, what is not the case in the `seekTo` case.

Correct, the main reason I pull the code out was to make it clear what is going on, and to avoid adding a parameter to 
`updatePlaybackSpeedSettingsForNewPeriod`.  If we are good for adding the parameter it will remove the duplicated logic.  
> Allow me to suggest the following so you can test whether this is equivalent for you:
> ...
> I think that should boil down to the same, avoid some duplicated code and makes sure that the `!shouldUseLivePlaybackSpeedControl(..)` case is covered as well.
> 
> Can you apply this and probably even more important, test this with your app/stream to see if it is equivalent to your change?

Sure, happy to do that. Also, you want me to cherry-pick the commit to `androidx.media` or does that just make managing the two repositories more of a headache for your team?

marcbaechinger: SG, thanks! It's fine to keep it here for this PR that is a single file. But you are right, for future PRs we'd prefer `androidx.media` if that's possible.

stevemayhew:
Thanks @marcbaechinger. I tested the change and forced pushed the single commit I rebased before updating,  so I picked up @tonihei 's fix for #10882.  Diff's to HEAD of dev-v2 are strait forward now.

It works just the same (logic is same), but the code is cleaner all going through the same method.

I will back port his changes and this code to our 2.15.1 based release so we are running with the same code.

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -1897,7 +1899,8 @@ timeline, rendererPositionUs, getMaxRendererReadPositionUs())) {
           /* oldPeriodId= */ playbackInfo.periodId,
           /* positionForTargetOffsetOverrideUs */ positionUpdate.setTargetLiveOffset
               ? newPositionUs
-              : C.TIME_UNSET);
+              : C.TIME_UNSET,
+          /* positionRequestedBySeek */false);
       if (periodPositionChanged
```
stevemayhew, line 1904: false every place but the` seekToInternal()`

library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImplInternal.java
```diff
@@ -1965,8 +1969,9 @@ private void updatePlaybackSpeedSettingsForNewPeriod(
         int oldWindowIndex = oldTimeline.getPeriodByUid(oldPeriodId.periodUid, period).windowIndex;
         oldWindowUid = oldTimeline.getWindow(oldWindowIndex, window).uid;
       }
-      if (!Util.areEqual(oldWindowUid, windowUid)) {
-        // Reset overridden target live offset to media values if window changes.
+      if (!Util.areEqual(oldWindowUid, windowUid) || positionRequestedBySeek) {
+        // Reset overridden target live offset to media values if window changes or if seekTo
+        // default live position.
         livePlaybackSpeedControl.setTargetLiveOffsetOverrideUs(C.TIME_UNSET);
```
stevemayhew, line 1975: on the else side of `if (positionForTargetOffsetOverrideUs != C.TIME_UNSET)` so we can unconditionally remove the liveOffsetOverride if this is from a seek.

PR #11127, opened 2023-04-18
Dts direct passthrough contri

cedricxperi:
This adds DTS Audio Support for Direct Passthrough playback. This is required for Tunnel mode playback on Android TVs that supports tunnel mode playback (e.g. MTK and Realtek based TVs). Test streams will be provided during the review if required.
Thanks.


cedricxperi:
Moving over to androidx/media.

PR #11181, opened 2023-05-25
Save state throwsWhenUsingWrongThread when ListenerSet::copy()

changxiangzhong:
`ListenerSet::copy()` does not save the state `throwsWhenUsingWrongThread`. This PR is to address that issue.


tonihei:
Can you describe the problem you are solving with this? I agree that the value is missing in the copy and we should probably add it for completeness, but as far as I see, there is also no way for your app to depend on this behavior at the moment.  

Note: We are also no longer accepting PRs in the ExoPlayer repository. So you can either open the same change in the Media3 repository or I can make the equivalent change internally. (Depends on whether you want to be contributed as the author in the git history).

changxiangzhong:
Thank you for your feedback, I've created a new [PR](https://github.com/androidx/media/pull/420), with more explanation. Please close this issue then. We can carry the discussion in the new PR.

PR #11270, opened 2023-08-21
PgsParser: reuse ArrayList

xufuji456:
Hello, guys:
We could reuse ArrayList with clearing the collection, instead of creating a new instance.

### DIFF
--- a/library/extractor/src/main/java/com/google/android/exoplayer2/text/pgs/PgsParser.java
+++ b/library/extractor/src/main/java/com/google/android/exoplayer2/text/pgs/PgsParser.java
@@ -52,11 +52,13 @@ public final class PgsParser implements SubtitleParser {
   private final ParsableByteArray inflatedBuffer;
   private final CueBuilder cueBuilder;
   @Nullable private Inflater inflater;
+  private final ArrayList<Cue> cues;
 
   public PgsParser() {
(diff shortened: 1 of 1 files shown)


icbaker:
This violates (in spirit) [4.8.2.2 of the Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.8.2-variable-declarations), by promoting a local variable to a field with implicitly extended scope and lifetime: 
> 4.8.2.2 Declared when needed
> Local variables are not habitually declared at the start of their containing block or block-like construct. Instead, local variables are declared close to the point they are first used (within reason), to minimize their scope. Local variable declarations typically have initializers, or are initialized immediately after declaration.

We do sometimes do this throughout the library, sometimes using the name `scratch` to indicate this - but generally only in places that are very sensitive to allocations (e.g. running very frequently). We don't routinely use fields over local variables. In this case, there are already other allocations happening every time `PgsParser.parse` is invoked, I don't think that also allocating an empty list at the same time is likely to cause a problem (and corollary: if there's a problem here with too many allocations, this change won't fix it).

Therefore I'm inclined to leave this code as-is.

icbaker:
I'm going to close this - if you want to continue the discussion please re-open it in the AndroidX media3 repo: https://github.com/androidx/media/pulls

PR #11285, opened 2023-09-07
Subsequent `selectTracks() after prepare does not override position

stevemayhew:
The MaskingMediaPeriod.onChildSourceInfoRefreshed() already has the position correct from the initial timeline update, since the `preparePositionUs` is 0 the second condition in the if:

```
    if (preparePositionOverrideUs != C.TIME_UNSET && positionUs == preparePositionUs) {
	...
```
is false, so the `preparePositionUs` is not reset.  This causes it to be used incorrectly on subsequent `selectTracks()` calls where `positionUs` does match `preparePositionUs`

The fix simply resets the `preparePositionOverrideUs` unconditionally, as only the initial `selectTracks()` after the prepared transition should use the override


icbaker:
Hi Steve - can you please make this PR against the https://github.com/androidx/media project instead? Unfortunately we're unable to merge PRs in this deprecated project. After the change is merged in that project it will get pushed back to the dev-v2 branch here.

PR #11304, opened 2023-09-27
Add AV1 color space parsing in MP4 atom parser

haixia-meta:
This adds av1C parsing based on the AV1 bitstream specification: https://aomediacodec.github.io/av1-spec/av1-spec.pdf

This fix is needed to have correct color space parsing in av1C atom when using hardware AV1 decoder.


haixia-meta:
Looks like this repository is deprecated.

Will close and make this PR against the https://github.com/androidx/media project instead.

PR #11375, opened 2024-02-06
Fixing Period Switch Ref Update Issue in Manifest Parsing

Lilsax:
This merge request addresses a critical issue in the manifest parsing functionality, specifically related to period transitions. The problem arises from the failure to update references correctly when encountering a period switch during parsing.

Previously, the parser would generate objects for each period with specific references. However, upon transitioning to a new period, these references were not appropriately updated since it creates a new tracks object with a new reference . Consequently, subsequent attempts to utilize the `getStreamIndexToTrackGroupIndex` function to retrieve track group information would employ outdated references. As a result, the parser failed to locate the necessary data, leading to application crashes.

To rectify this issue, this merge request introduces a fix that ensures the accurate updating of references during period switches. The fix involves identifying the group index using the track ID, thereby ensuring the retrieval of correct track group information across period transitions.

This enhancement aims to enhance the stability and reliability of the manifest parsing process, preventing crashes and ensuring seamless functionality, particularly during period switches.


icbaker:
The `com.google.android.exoplayer2` project is now deprecated. We are not accepting any new PRs in this project. Please file your PR in the `androidx.media3` GitHub project instead: https://github.com/androidx/media

PR #11377, opened 2024-02-26
Package mappings match documentation

stevemayhew:
Update the PACKAGE_MAPPINGS to match the documentation on https://developer.android.com/media/media3/exoplayer/mappings


marcbaechinger:
Thanks Steve for the nudge. There where various entries wrong in the existing file not only in the package mappings. I updated the script in this branch and in the release branch to make it sync with the version we are having in the internal source tree that looked correct.

https://github.com/google/ExoPlayer/blob/release-v2/media3-migration.sh
https://github.com/google/ExoPlayer/blob/dev-v2/media3-migration.sh

I figure you wouldn't specifically need/want to add this PR to your CV ;), so it's fine to close this given the file is updated in release and dev-v2 branch?

many thanks!

stevemayhew:
Whatever is easiest for you to apply the change.  I'm looking at making an option to migrate patches for us to use to move some pull requests over to AndroidX I'll share that to if it proves useful.  

BTW what's the script you guys use to move patches between git repos (used to copy to/from one repo). Have you open sourced that?

We have an entire ExoPlayer tree we want to automate some of moving to AndroidX

marcbaechinger:
Ok, cool. The script is updated in the release and the dev-v2 branch to match again to what we have internally.

We are using [copybara](https://github.com/google/copybara) internally to transform the code from the internal source code to external media3 and exoplayers packages. We export from the internal piper to a Media3 git project. Earlier this has then also been transformed to the exoplayer2 packages but we stopped this. 

That's an internal configuration of copybara that also includes to remove close-source stuff and things like this. I don't think it is of general value in a sense that we open-source that. 

The migration script here includes the mappings from the copybara configuration we use. That's a semi automated process to get that config into the bash script. You've witnessed that that can go wrong when I do it.

icbaker:
Closing this as the requested changes have been merged.

